PR #3, opened 2023-12-31
Cloud server open source

afsalthaj:


afsalthaj:
@adamgfraser  @vigoo  @justcoon  I am merging this change as they are pretty straight forward. Now we have cloud-servers-base and cloud-servers-oss similar to the worker executor naming pattern

PR #19, opened 2024-01-11
Try to optimise CI

afsalthaj:


golem-cloud-server-oss/src/api/mod.rs
```diff
@@ -28,7 +28,7 @@ pub fn combined_routes(services: &Services) -> Route {
         .nest("/docs", ui)
         .nest("/specs", spec)
         .at(
-            "/v2/templates/:template_id/workers/:worker_name/connect",
+            "/templates/:template_id/workers/:worker_name/connect",
```
justcoon, line 31: is not better to have here `/v1`?

vigoo:
why are we doing this?

PR #37, opened 2024-01-19
Use target release directory as build was changed but not docker images

afsalthaj:


afsalthaj:
Perks of not including docker image build in PRs!!

PR #58, opened 2024-01-24
Return error when there is decode error for record

afsalthaj:
If we Invoke function with partial information in the function parameter which is of `Record` type, then golem-cli doesn't respond with any error and looks to be hanging forever.

This is a minor fix.

More details here: https://www.notion.so/golemcloud/Golem-OSS-CLI-Testing-893dbb95d2054044afdedc5a49906b9d

image

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -42,8 +42,8 @@ services:
       - GOLEM__GRPC_PORT=9090
       - GOLEM__HTTP_PORT=8080
     volumes:
-      - ./template_store:/template_store
-      - ./golem_db:/app/golem_db
+      - template_store:/template_store
+      - golem_db:/app/golem_db
     ports:
       - "8080:8080"
       - "9090:9090"
@@ -84,4 +84,8 @@ services:
 
 volumes:
   redis_data:
+    driver: local
   template_store:
+    driver: local
+  golem_db:
+    driver: local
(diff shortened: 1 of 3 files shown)


afsalthaj:
Testing in progress

adamgfraser (approved):

afsalthaj:
```scala
 golem-cli --golem-url http://localhost:8080 worker invoke-and-await  --template-id e604bc58-b3dd-4864-808f-b549ee5ffa90 --worker-name myworker --function golem:it/api/add-item --parameters '[{"product-id" : "hmm", "name" : "hmm" , "price" : 10}]'
Error: Response error: InternalError: GolemErrorBody { golem_error: RuntimeError(GolemErrorRuntimeError { details: "Value mismatch: Invalid record data in function parameters expected 4 value(s); got 3" }) }
```

vigoo:
Here is an old test case that tests that missing field is accepted as `None` in case of option fields:
https://github.com/zivergetech/golem/blob/059082a44a727782ae1d2ad5f32c361a208ad310/worker-executor-tests/src/it/scala/com/golem/integtests/ApiTests.scala#L344-L429

As this could not be migrated to a new worker executor test (because that calls the executor directly), I converted it to these tests:

https://github.com/golemcloud/golem-services/blob/main/golem-cloud-server-base/src/typechecker.rs#L1858-L1937

I think this property should be kept, while we definitely should fail in the type checker if there is any **non-optional** field missing.

afsalthaj:
Those tests are passing now

PR #59, opened 2024-01-24
Docker amd64 and arm64 build

justcoon:
added support for arm64 builds 
- Dockerfile-s with amd64 and arm64 platform support
- CI docker publish for amd64 and arm64 - publish to dockerhub needs to be tested, but build of docker images in scope of CI was successful, see: https://github.com/golemcloud/golem-services/actions/runs/7671053105/job/20908600837
- updated docker-compose.yml - tested locally on Intel Mac and ARM Mac
- updated contributing.md with instructions for ARM Mac build

NOTE: amd64 specific Dockerfile-s were not removed yet


golem-shard-manager/docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM debian:bookworm-slim
+
+ARG TARGETPLATFORM
```
afsalthaj, line 3: I strongly recommmend not to abstract things here

afsalthaj: Because a change for ARM will result in having to test the build for amd64.
I understand DRY principle, however, sometimes its ok to have simply two folders and have the most simplest Dockerfile. DevOps can quickly get convoluted and nasty.

afsalthaj: There might be requirements of additional installations for a target platform. I am not 100 percent sure. Unless we are so sure this will never change, then its ok. Good that base image is apparently same.

afsalthaj: This is also going with the assumption, our current docker build strategies is going to be permanent (i.e, building binaries outside targetting a platform).

Cross compilation in Rust is not that straight forward, and may be in a few months, we will have to revisit that as well, breaking this Dockerfile anyway

vigoo: I would rather have a single dockerfile than more than one. If we have more than one, with almost the same content, we will always forget to update one of them and it's going to be annoyoing.

justcoon: everything what i seen till now is using that approach (one common dockerfile), 
for multi platform build , if there are needs for platform customisation, it should be possible to do stages based on given platform https://github.com/docker/buildx/issues/805#issuecomment-946478949

but sure we can try to find if it is possible to do it in different way ...

.github/workflows/ci.yaml
```diff
@@ -29,10 +29,18 @@ jobs:
         with:
           toolchain: stable
           override: true
+      - name: Install Rust Targets
```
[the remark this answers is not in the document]
justcoon: i wanted to have some working solution as this seems to be priority in my opinion now,
sure we can split build (then platform specific part, will be build in parallel), so it should be also faster, but also more complicated ...
but i prefer to merge this one first, and do improvements in next iteration/PR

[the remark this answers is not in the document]
vigoo: Why is it more complicated? Isn’t it simpler, just a duplication of the job instead of doing multiple architectures in a single one?
I don’t think there is anything that is not platform specific.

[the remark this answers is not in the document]
justcoon: ok, i will take o look on that

.dockerignore
```diff
@@ -7,4 +7,10 @@ README.md
...
 !target/x86_64-unknown-linux-gnu/release/worker-executor-oss
-!target/x86_64-unknown-linux-gnu/release/cloud-server-oss
\ No newline at end of file
+!target/x86_64-unknown-linux-gnu/release/cloud-server-oss
+!target/aarch64-unknown-linux-gnu/debug/golem-shard-manager
+!target/aarch64-unknown-linux-gnu/debug/worker-executor-oss
+!target/aarch64-unknown-linux-gnu/debug/cloud-server-oss
+!target/aarch64-unknown-linux-gnu/release/golem-shard-manager
+!target/aarch64-unknown-linux-gnu/release/worker-executor-oss
+!target/aarch64-unknown-linux-gnu/release/cloud-server-oss
```
afsalthaj, line 16: Nice

.github/workflows/ci.yaml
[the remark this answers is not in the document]
afsalthaj: I agree as well. Two different parallel tasks. Duplication here is more important than reuse

.github/workflows/ci.yaml
```diff
@@ -56,60 +70,55 @@ jobs:
             COMMIT_SHORT_HASH=$(git rev-parse --short=7 HEAD)
             echo "DOCKER_VERSION=${COMMIT_SHORT_HASH}" >> $GITHUB_ENV
           fi
-      - uses: satackey/action-docker-layer-caching@v0.0.11
+      - uses: jpribyl/action-docker-layer-caching@v0.1.1
```
afsalthaj, line 73: I am sure this doesn't contribute much to build

CONTRIBUTING.md
```diff
@@ -66,4 +87,4 @@ cargo build --target x86_64-unknown-linux-gnu
 ### WINDOWS
 TBD
 
-We will be trying cargo chef and offload cargo build to docker context without impacting the build time.
+We will be trying cargo chef and offload cargo build to docker context without impacting the build time.
```
afsalthaj, line 90: Nice, again contributing to doc :)

golem-cloud-server-oss/docker/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+COPY /golem-cloud-server-oss/config/cloud-server.toml ./config/cloud-server.toml
+
+RUN apt-get update && apt-get install -y libssl-dev
+RUN apt-get update && apt-get install -y ca-certificates
+RUN update-ca-certificates
+
+EXPOSE 8080
+EXPOSE 9090
+
+ENTRYPOINT ["./cloud-server-oss"]
```
afsalthaj, line 26: I can't quite understand this dockerFile

justcoon: TARGETARCH is provided by buildkit see: https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

platforms: linux/amd64,linux/arm64
archs: amd64, arm64

based on `platform-${TARGETARCH}` then there will be `RUST_TARGET=x86_64-unknown-linux-gnu` or `RUST_TARGET=aarch64-unknown-linux-gnu`

which will be used in `COPY /target/$RUST_TARGET/release/cloud-server-oss ./`

golem-shard-manager/docker/Dockerfile
afsalthaj: Ok we are using standard solutions here, and I can now see why this approach you took. 👍

.github/workflows/ci.yaml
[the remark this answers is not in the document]
afsalthaj: I think, I can reason about this PR. After reading about buildx
https://github.com/golemcloud/golem-services/pull/59#discussion_r1468238014

afsalthaj (approved):
Have a discussion with @vigoo . I can approve it from my side

.github/workflows/ci.yaml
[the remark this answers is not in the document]
afsalthaj: See if you can paralllelise it for two in separate jobs making it faster as well. But not super concerned as I was before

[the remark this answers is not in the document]
vigoo: So, I remembered wrong as I thought there is a separate job that builds the release executables, and another that takes them and builds the container(s). I guess that's in the `golem` repo not here. Confusing :) So now I understand that this was really the simplest way to add this.

I still think it would be better to not build two different archs in the same job, as all the rust compilation will not be able to share anything between them so it will be sequential in time, and take double disk space, etc. 

So if it's possible (I feel it's easy but maybe I'm overlooking something) I would do it the same way as in the other repo - build the two different release artifacts in parallel and then have a third github actions job that takes the result of both and builds the containers. It would also make the two CI pipelines more similar, which reduces confusion. 

I'm removing the "request for changes" so I don't block you but please consider doing it :)

vigoo:
Approved but please read my latest comment above first

.github/workflows/ci.yaml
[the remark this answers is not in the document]
justcoon: sure, thanks for comment, in next PR, i will try to do cargo build per each arch in parallel ... and then publish

PR #63, opened 2024-01-26
GOL-23 Tests and fixes for filesystem durability

vigoo:
Fixes WASI filesystem durability issues regarding timestamps.

Also fixes the _recovery process_ to fail the recovery in case an invocation fails. This makes divergence detection work.


golem-worker-executor-base/src/durable_host/filesystem/types.rs
```diff
@@ -245,14 +292,52 @@ impl<Ctx: WorkerCtx> HostDescriptor for DurableWorkerCtx<Ctx> {
...
+            None => source_dir,
+        };
+
+        let target_path: PathBuf = new_path.into();
+        let target_parent = match target_path.parent() {
+            Some(p) => Arc::new(target_dir.open_dir(p)?),
+            None => target_dir,
+        };
+
+        self.durable_file_times(
```
vigoo, line 329: Here (and in other similar functions) we basically create two separate op-log entries, each of them restoring the access/modification times on a single file system entry. The entries are having the `ImportedFunctionInvoked` type though which feels a bit hacky. 

Probably we should always create a single entry, with storing a map of file times instead of a single one.

@adamgfraser what do you think?

vigoo: Or even create a dedicated "filesystem access log entry"

vigoo: The alternative appoach of only persisting `stat` and `stat-at` resolves this question

adamgfraser (approved):

PR #65, opened 2024-01-27
CI docker build

justcoon:
parallel build of targets for docker
successful run (without publish) https://github.com/golemcloud/golem-services/actions/runs/7681364498/job/20934477766

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,9 +15,20 @@ env:
   BUILD_TARGET: "x86_64-unknown-linux-gnu"
 
 jobs:
-  docker:
+  docker-targets-build: # WIP
     runs-on: ubuntu-latest
     if: github.event_name == 'push' && github.ref_type == 'tag'
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - platform: linux/amd64
+            target: x86_64-unknown-linux-gnu
+            build-config: ""
+          - platform: linux/arm64
+            target: aarch64-unknown-linux-gnu
+            build-config: "--config \"target.aarch64-unknown-linux-gnu.linker = 'aarch64-linux-gnu-gcc'\""
+    name: docker-targets-build (${{ matrix.platform.platform }})
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -15,29 +15,44 @@ env:
...
-          rustup target add x86_64-unknown-linux-gnu
-          rustup target add aarch64-unknown-linux-gnu
+      - name: Install Rust Target
+        run: rustup target add ${{ matrix.platform.target }}
       - name: Install Protoc
         uses: arduino/setup-protoc@v1
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Install gcc-aarch64-linux-gnu
+        if: matrix.platform.platform == 'linux/arm64'
```
afsalthaj, line 55: This indicates the problem of reuse.

.github/workflows/ci.yaml
```diff
@@ -50,12 +65,44 @@ jobs:
             ~/.cargo/git/db/
             target/
           key: ${{ runner.os }}-cargo-release-${{ hashFiles('**/Cargo.lock') }}
-      - name: Build amd64 Executables
+      - name: Build Executables
+        run: cargo build --release --target ${{ matrix.platform.target }} ${{ matrix.platform.build-config }}
```
afsalthaj, line 69: Now that has moved away what we usually do in local which is release. Some users refer to CI to see how they build docker. Anyway it's an opinionated approach if I say repeat is better than matrix. So feel free to discard my opinion :)

.github/workflows/ci.yaml
justcoon: well, it is like that as it is build is made under adm64 linux, i can remove condition, it will be installed always

.github/workflows/ci.yaml
justcoon: then it does not make sense to do it in matrix way, and also in parallel, because this in such way it will be even harder to extend it

justcoon: and sure, it can be done, even better, but we should try to do it step by step, when next platform will be needed, or something else change, which will require noticeable changes in CI

afsalthaj: Sure. I understand your concern. I wouldn't do it this way personally, and it is much easier for me to copy paste than understanding the nuances of when to use `if statements` for which platform and having to figure out the commonalities of these builds.

justcoon: currently it is 1 conditional step from 10, 
if it was 5 from 10, it could make sense to do copy/paste style, 
but with 1, it does not make sense to me to do it like that

afsalthaj (approved):

PR #68, opened 2024-01-29
Test docker compose

afsalthaj:


afsalthaj:
With arm and amd changes, I am not able to figure out the right way to test docker compose in CI.
Also I can see the build has become slow.

PR #72, opened 2024-01-31
Improve docker and make getting started easier

afsalthaj:
Avoid explicit URL in golem-cli

```scala
golem-cli --golem-url http://localhost:8080/ worker connect --template-id 3558cc5d-d08d-4c87-950f-defa42ef4be2 --worker-name myworker
```
can be just

```scala

golem-cli worker connect --template-id 3558cc5d-d08d-4c87-950f-defa42ef4be2 --worker-name myworker
```

Make use `.env` reusing certain ports across multiple docker compose files.


docker-compose-postgres.yaml
```diff
@@ -1,12 +1,12 @@
-version: '3'
+version: '3.8'
```
afsalthaj, line 1: 3.8 version https://docs.docker.com/compose/compose-file/compose-file-v3/#variable-substitution did't work for me. Looks like it is applicable to only variable block where the name of a key is same across the services, which is not in our case. Therefore falling back to the primitive use of .env file, where we substitute just the value of whatever key the service is expecting

afsalthaj:
Merging this, as it is definitely better than what is existing now.

PR #73, opened 2024-01-31
DON'T MERGE: Add docker-compose tests in CI without impacting build times

afsalthaj:
https://www.notion.so/DON-T-MERGE-Add-docker-compose-tests-in-CI-without-impacting-build-times-361dfc76ea2e45968482b34e71826fec

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1323304762) by [Unito](https://www.unito.io)


afsalthaj:
The CI is crucial to the maintenance and reliability of our  Golem OSS artifacts, and they are primarily docker images. User can use these docker images and go  ahead and deploy it with kubernetes or even directly refer/use one of our docker compose files with [ecs-cli ](https://github.com/aws/amazon-ecs-cli) to deploy straight to their AWS ECS.  So its important to test docker-compose in CI, and avoid regression and prove that golem services successfully work with docker-compose.
We don't need to test all examples though, because all we need is to prove this aspect, while finding a sweet spot in build timing and keep up developer productivity. Also the test is done with only one platform.

Another aspect is the complexity of build of artifacts for multiple platforms using `matrix` that exists in main branch. Matrix is great approach if workflow is one to one with only platform attribute changing. But that's not the case as far as I see. There are steps that are exclusive to certain platform and therefore it is suboptimal to use matrix approach. Note to mention the complexity with `tar` and `untar` that exists and the deviation of what's done by a developer in local, vs what exists in CI (main). So I chose to avoid matrix and repeat the steps and keeping it dead simple. We are still reusing OSS actions for CI such as docker-action, and I don't think we need to over-do it. By repeating the steps instead of using a matrix, we maintain clarity in the workflow definition for each artifact, and easier to debug.

.github/workflows/ci.yaml
```diff
@@ -283,3 +132,266 @@ jobs:
...
+
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-release-${{ hashFiles('**/Cargo.lock') }}
```
vigoo, line 311: Should have a different cache key than the amd64 build. Although the sources are the same the generated binaries are not so I think caching with the same key in both parallel jobs just invalidates te whole thing for each other.

vigoo (changes requested):
One simple request otherwise looks good to me

.github/workflows/ci.yaml
afsalthaj: Let me take a look. All I focussed was splitting what existed. May be I missed something

.github/workflows/ci.yaml
```diff
@@ -283,3 +132,266 @@ jobs:
...
+          images: golemservices/golem-worker-executor
+
+      # Build and push worker executor image
+      - name: Build and push worker executor image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: ./golem-worker-executor-oss/docker/Dockerfile
+          push: ${{ steps.push_decision.outputs.push }}
+          platforms: linux/arm64
```
justcoon, line 358: i am not sure if docker images can be published in such way, that there will be one version of docker  image for multiple platforms

.github/workflows/ci.yaml
vigoo: Just add an `-arm64` segment to this cache key in the arm64 job.

afsalthaj:
This PR is intentioanlly delayed to avoid last minute issues. There can be a chance there is a manifest merge required if we are pushing docker images separately for different architecture

.github/workflows/ci.yaml
afsalthaj: It's a good question, and this is what I mentioned in the post comment, and delayed this PR

vigoo:
@afsalthaj do we still need this?

afsalthaj:
Need to do the same with new test framework

PR #75, opened 2024-01-31
Add caching

afsalthaj:
https://www.notion.so/golemcloud/Worker-Executor-gRPC-client-caching-3b1ec1f782e24b0a9401b180130f763c?pvs=4


afsalthaj:
Initially implemented using dashmap out of interest (wasn't bad), but trying to change to what we usually do

golem-service-base/src/worker_executor_clients.rs
```diff
@@ -1,24 +1,45 @@
...
+
+        if let Some(client) = cache.get(pod) {
+            return Ok(client.clone());
+        }
+
         let client = WorkerExecutorClient::connect(pod.uri())
             .await
             .map_err(|e| e.to_string())?;
+
+        cache.insert(pod.clone(), client.clone());
```
justcoon, line 41: i think we may need some cache invalidation in future ...

golem-service-base/src/worker_executor_clients.rs
```diff
@@ -1,24 +1,45 @@
+use crate::model::Pod;
 use async_trait::async_trait;
 use golem_api_grpc::proto::golem::workerexecutor::worker_executor_client::WorkerExecutorClient;
+use std::collections::HashMap;
+use std::sync::Arc;
+use tokio::sync::Mutex;
 use tonic::transport::Channel;
 
-use crate::model::Pod;
+type WorkerExecutorCache = Arc<Mutex<HashMap<Pod, WorkerExecutorClient<Channel>>>>;
```
vigoo, line 9: You should use my cache implementation - it's async, supports different invalidation techniques and we use it for all caching things in the worker executor and the shard manager.

https://github.com/golemcloud/golem-services/blob/main/golem-common/src/cache.rs

vigoo (changes requested):

golem-service/config/cloud-server.toml
```diff
@@ -18,6 +18,10 @@ database = "golem"
 [templates.store]
 object_prefix = ""
 
+[worker_connect_cache]
+max_capacity = 1000
```
afsalthaj, line 22: what should be this? Pretty sure people will try to connect to million workers.

vigoo: You cannot open a million tcp connections to a single cloud server :) Leave it as is, we cannot figure out these numbers without the benchmarking framework

golem-service-base/src/worker_executor_clients.rs
```diff
@@ -1,25 +1,50 @@
...
+    pub fn new(max_capacity: usize, time_to_idle: Duration) -> WorkerExecutorClientsDefault {
+        WorkerExecutorClientsDefault {
+            cache: Cache::new(
+                Some(max_capacity),
+                FullCacheEvictionMode::LeastRecentlyUsed(1),
+                BackgroundEvictionMode::OlderThan {
+                    ttl: time_to_idle,
+                    period: Duration::from_secs(60),
+                },
+                "worker_connect_client",
```
vigoo, line 27: This name is a bit misleading, these clients are not only used for the `worker connect` feature, but for all gRPC calls to the worker executors

golem-service/src/config.rs
```diff
@@ -15,6 +16,13 @@ pub struct DbSqliteConfig {
     pub max_connections: u32,
 }
 
+#[derive(Clone, Debug, Deserialize)]
+pub struct WorkerConnectCacheConig {
```
vigoo, line 20: Let's call it WorkerExecutorClientCache or something, it's not for the "connect" thing only

golem-service/src/service/worker.rs
```diff
@@ -226,7 +227,7 @@ impl WorkerServiceDefault {
                                 routing_table.number_of_shards.value,
                             ),
                             worker_id,
-                            err
+                            err.to_json_string()
```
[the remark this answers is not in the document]
vigoo: By the way `lookup` returning a `GolemError` does not make any sense either - it is always `GolemError::Unknown`, containing a stringified gRPC client error.

justcoon (approved):

golem-service-base/src/worker_executor_clients.rs
afsalthaj: True, worker_grpc_client may be?

golem-service/src/config.rs
afsalthaj: Sure yea!

golem-service/src/service/worker.rs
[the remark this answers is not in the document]
afsalthaj: Yea, your second comment fixes the first one :D

vigoo (approved):

PR #85, opened 2024-02-03
docker compose suffix fix

justcoon:
fix suffix to be same like in other cases


adamgfraser (approved):

afsalthaj (changes requested):
Did you update documentations? And this was changed from yaml to yml

afsalthaj:
Ah got you

afsalthaj (approved):

PR #97, opened 2024-02-08
Add worker bridge

afsalthaj:


afsalthaj:
After a bit of rethinking, I will include more documentations and examples, working with Tyk, and then raise the PR again.
Please note that there are changes/renamings/optimisations/module-structure changes compared to what's in cloud.

PR #102, opened 2024-02-08
docker-examples - default values

justcoon:
default values for env variables  - to not require .env file


docker-examples/docker-compose-postgres.yaml
```diff
@@ -88,7 +90,7 @@ services:
       - GOLEM__COMPILED_TEMPLATE_SERVICE__TYPE="Local"
       - GOLEM__SHARD_MANAGER_SERVICE__TYPE="Grpc"
     ports:
-      - "${WORKER_EXECUTOR_HTTP_PORT}:${WORKER_EXECUTOR_HTTP_PORT}"
+      - "${WORKER_EXECUTOR_HTTP_PORT:-8082}:${WORKER_EXECUTOR_HTTP_PORT:-8082}"
```
afsalthaj, line 93: What is the difference of this with having the dafult values as .env variable?

afsalthaj: I am not into this style, of default values. Default values already exist in .env file.

Pros: They can download docker-compose file
cons: Prone to more user errors, because variable values are repeated. The sole reason for why I avoided repetitions, was, when I was writing compose file, I made errors with respect to port numbers

docker-examples/docker-compose-postgres.yaml
```diff
@@ -28,9 +28,9 @@ services:
       - RUST_BACKTRACE=1
       - RUST_LOG=info,h2=warn,hyper=warn,tower=warn
       - GOLEM__REDIS__HOST=redis
-      - GOLEM__REDIS__PORT=${REDIS_PORT}
-      - GOLEM__HTTP_PORT=${SHARD_MANAGER_HTTP_PORT}
-      - GOLEM_SHARD_MANAGER_PORT=${SHARD_MANAGER_GRPC_PORT}
+      - GOLEM__REDIS__PORT=6379
```
afsalthaj, line 31: So now it closes the opportunity to use a single .env file that has all configurations in 1 place. Hardcoded redis port!

docker-examples/docker-compose-postgres.yaml
```diff
@@ -6,7 +6,7 @@ services:
     volumes:
       - redis_data:/data
     ports:
-      - ${REDIS_PORT}:6379
+      - 6379:${REDIS_PORT-6379}
```
afsalthaj, line 9: So why ${REDIS_PORT} here and not in the lie number 31?

afsalthaj: These many repetitions
```scala
      9: 6379:${REDIS_PORT-6379}
      31: GOLEM__REDIS__PORT=6379
     75: GOLEM__REDIS__PORT=6379
```

afsalthaj: 9881 is repeated in 4 places

afsalthaj: 8081 is repeated in 2 places, and this will repeat with worker bridge

afsalthaj:
You can zip the docker-compose with a .env file and expose it elsewhere, if all you need is to download them as 1 single thing

docker-examples/docker-compose-postgres.yaml
justcoon: as i mentioned in different comment, with current redis setup, internal redis port will be 6379, REDIS_PORT will change just public value

afsalthaj:
@adamgfraser @justcoon 

`curl -O https://raw.githubusercontent.com/golemcloud/golem-services/main/docker-examples/docker-compose.yaml -O https://raw.githubusercontent.com/golemcloud/golem-services/main/docker-examples/.env`

This command downloads all required files in 1 single go, and that will be it

PR #106, opened 2024-02-08
Add worker bridge

afsalthaj:
* The parts gateway code in cloud changed for the following before moving to OSS, as the new worker-bridge
   * optimisations in parsing, that we really parse request and response mapping to actual types only if there are comparison nodes in the expression, making the code simpler elsewhere
   * Renamings, and package namings, and cleanups
   * Removal of ApiDeployment, Certificate, DomainRegistry etc
   * Removal of Project and Account
 
 * Documentation on integration with Tyk including instructiions. More cleanups here to come
 * Docker image, and docker compose and CI
 
 
 More details to follow: Tyk's middleware isn't working for me, and this will complete a proper example. Currently I am able to integrate and make requests through Tyk already reusing all its features
 
 
 Please note that the code is not one to one, and there are many changes between cloud and the code here. 
 A few tasks that's in my radar is porting API request validating, metrics etc.
 
 
 
## Things in progress

- [x]  Proper error types (to a great extent)

- [x]  Fix 9 failed tests related to expr that came in as part of the test case (2 more)
          test result: FAILED. 100 passed; 2 failed; 1 ignored; 0 measured; 0 filtered out; finished in 30.96s 

- [x]  Make sure we have an example with Tyk's middleware working and documented

- [x]  Make sure we reuse a feature in Tyk - probably caching, and ensure we are not hitting worker instance when hitting cache

- [x] Further clean ups if required

- [x] Keep adding more documentations and ideas.

- [x] Docker compose examples for API Gateway integration
- [x] Allow special characters in path(found a bug that endpoint path cannot have hiphens)
- [x] Choose OAS Spec for Tyk
- [x] Fix a bug that allowed "/" to match any path spec and add tests

### Follow up PR:

* Remove Box<dyn Error> when executing worker
* Local Environment Run without Redis
* Validation of API
* Better APiRegistrationError with version as part of ApiDefinition key
* metrics


afsalthaj:
As of now, I managed to integrate with Tyk

```scala
url -X GET http://localhost:8080/hmmm -H "X-Tyk-Authorization: foo" 

[[{"name":"hmm","price":10.0,"product-id":"hmm","quantity":2}]]%
```

.github/workflows/ci.yaml
```diff
@@ -166,6 +167,20 @@ jobs:
           platforms: ${{ env.PLATFORMS }}
           tags: ${{ steps.meta-golem-service.outputs.tags }}
           labels: ${{ steps.meta-golem-service.outputs.labels }}
+      - name: Extract metadata (tags, labels) for golem worker bridge
+        id: meta-golem-worker-bridge
+        uses: docker/metadata-action@v5
+        with:
+          images: golemservices/golem-worker-bridge
+      - name: Build and push cloud server image
```
justcoon, line 175: cloud server image -> worker bridge image

golem-worker-bridge/src/register.rs
```diff
@@ -0,0 +1,309 @@
...
+    pub async fn new(config: &RedisConfig) -> Result<RedisApiRegistry, Box<dyn Error>> {
+        let pool = golem_common::redis::RedisPool::configured(config).await?;
+        Ok(Self { pool })
+    }
+}
+
+#[async_trait]
+impl RegisterApiDefinition for RedisApiRegistry {
+    async fn register(&self, definition: &ApiDefinition) -> Result<(), Box<dyn Error>> {
+        debug!("Register account: id: {}", definition.id);
```
justcoon, line 89: `account`  should be removed from log messages, same in other cases

justcoon:
@afsalthaj binding vs template validation is missing, are you planning to add it?

afsalthaj:
Template validation will be added in next PR. I am hoping I can port your code as is.

I cherrypicked code because for technical reasons. Now that I don't think we can copy paste things as it is due to lot of renamings and implementation differences

Please give me some time next week, and I will re-add some validations

adamgfraser (approved):

PR #111, opened 2024-02-10
kube helm chart for golem-services

justcoon:
- template storage - persistence volume claim (in next version, support for different storage will be added: S3)
- golem-service - disabled health check - latest released version do not have health check  (changes are already in main, needs to be released)
- redis auth  - disabled  - latest released version do not have support for username and password (changes are already in main, needs to be released)


afsalthaj (approved):
LGTM

PR #112, opened 2024-02-11
WIP: Open api to worker bridge

afsalthaj:
Now we can upload the same open-api doc to both worker-bridge and external API gateways


afsalthaj:
Too early, let me make an example that work with shopping-cart

PR #113, opened 2024-02-12
Open api to worker bridge

afsalthaj:
Please wait before review/merge, as this PR is really meant to showcase a functionality which I would like to discuss within the team.  This is kept open to see what's going on.

This is more of an inspiration from Tyk.

Tyk is a popular API gateway. We can configure Tyk is it's own configuration language called Classic API definition.
But Tyk also document the fact, users already have an API documentation as an Open API spec for their backend service. So Tyk exposes a functionality where it accepts the OpenAPI format configuration language, such that users have the option to upload the same open api spec into Tyk, and they can configure the rest in UI, or as patch request.

This is exactly what I did here with worker bridge too.
Worker bridge can accept worker-bridge specific configuration language. If users are using this configuration, then they need to write a separate configuration mandatory for Tyk. With this PR, worker-bridge can also accept Open API spec format to define worker bridge functionality as well as all endpoints. By doing this, users have the option to upload the same API definition doc in Tyk as well as worker bridge. Note that, this is not limiting the language power of workerbridge, because we are using extension format of OpenAPI where you can simply put a worker bridge specific configuration.


golem-worker-bridge/src/oas_worker_bridge.rs
```diff
@@ -0,0 +1,125 @@
...
+
+pub fn get_api_definition(open_api: &str) -> Result<ApiDefinition, String> {
+    dbg!(open_api);
+    let openapi: OpenAPI = serde_json::from_str(open_api).map_err(|e| e.to_string())?;
+    let version = Version(openapi.info.version);
+    let id = ApiDefinitionId(
+        openapi
+            .extensions
+            .iter()
+            .find(|(key, _)| key.to_lowercase() == "x-api-definition-id")
```
justcoon, line 19: i would do `x-api-definition-id` -> `x-golem-api-definition-id` - if it is golem specific

afsalthaj: This is very rough PR, I can take reviews soon

afsalthaj: Will fix in next PR. Right now I got working examples, and didn't want to go through testing it again

justcoon: maybe is good to have constants

PR #115, opened 2024-02-14
Disallow updating the same version of endpoint definitions

afsalthaj:
While we think about versioning of worker bridge definitions (api-definition version), we should disallow mutating a definition for the same version.

cc @adamgfraser , something that we discussed.

### DIFF
--- a/golem-worker-bridge/src/api/custom_request_endpoint.rs
+++ b/golem-worker-bridge/src/api/custom_request_endpoint.rs
@@ -78,13 +78,7 @@ impl CustomRequestEndpoint {
         {
             Ok(api_definition_id) => ApiDefinitionId(api_definition_id.to_string()),
             Err(err) => {
-                error!(
-                    "{}",
-                    format!(
-                        "Invalid {} header: {}",
-                        GOLEM_API_DEFINITION_ID_EXTENSION, err
-                    )
-                );
+                error!(err);
                 return Response::builder()
                     .status(StatusCode::BAD_REQUEST)
                     .body(Body::from_string(err.to_string()));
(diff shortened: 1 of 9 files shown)


afsalthaj:
@nicoburniske  If possible you could add a follow up PR with more tests related to this change, and implement get_all functionality (get all versions of a particular api-id)

golem-worker-bridge/src/api/custom_request_endpoint.rs
```diff
@@ -72,46 +74,55 @@ impl CustomRequestEndpoint {
...
-
-        let api_definition_id = match api_definition_id_string {
-            Ok(api_definition_id) => ApiDefinitionId(api_definition_id.to_string()),
+        let version = match get_header_value(&headers, GOLEM_API_DEFINITION_VERSION) {
+            Ok(version) => Version(version),
             Err(err) => {
-                error!("Invalid X-API-Definition-Id header: {}", err);
+                error!(
+                    "{}",
+                    format!("Invalid {} header: {}", GOLEM_API_DEFINITION_VERSION, err)
```
justcoon, line 99: error in `get_header_value` 
is already formatted like `format!("Invalid value for the header {} error: {}", header_name, e)`

golem-worker-bridge/src/api/custom_request_endpoint.rs
```diff
@@ -72,46 +74,55 @@ impl CustomRequestEndpoint {
...
+        let api_definition_id = match get_header_value(&headers, GOLEM_API_DEFINITION_ID_EXTENSION)
+        {
+            Ok(api_definition_id) => ApiDefinitionId(api_definition_id.to_string()),
             Err(err) => {
+                error!(
+                    "{}",
+                    format!(
+                        "Invalid {} header: {}",
+                        GOLEM_API_DEFINITION_ID_EXTENSION, err
+                    )
```
justcoon, line 86: error in `get_header_value` 
is already formatted like `format!("Invalid value for the header {} error: {}", header_name, e)`

nicoburniske:
> @nicoburniske If possible you could add a follow up PR with more tests related to this change, and implement get_all functionality (get all versions of a particular api-id)

https://github.com/golemcloud/golem-services/pull/117/files

justcoon (approved):

PR #120, opened 2024-02-15
Add get_all Redis impl + RegisterService get_all_versions 

nicoburniske:
Introduces 2 new sets to aid in API Definition lookup
   - One for all API Definitions
   - One for all versions for a specific API Definition 

Introduces Transaction helper to have transactional guarantees in redis 

This can be further optimized using MGET to retrieve all API Definitions or using LUA scripting for a transaction.

### DIFF
--- a/golem-worker-bridge/src/register.rs
+++ b/golem-worker-bridge/src/register.rs
@@ -315,10 +315,7 @@ impl RegisterApiDefinition for RedisApiRegistry {
             futures::future::try_join_all(futures).await?
         };
 
-        Ok(all_definitions
-            .into_iter()
-            .filter_map(std::convert::identity)
-            .collect::<Vec<_>>())
+        Ok(all_definitions.into_iter().flatten().collect::<Vec<_>>())
     }
 
     async fn get_all_versions(
@@ -349,10 +346,7 @@ impl RegisterApiDefinition for RedisApiRegistry {
             .await
             .map_err(|e| ApiRegistrationError::InternalError(e.to_string()))?;
 
-        Ok(api_definitions
-            .into_iter()
-            .filter_map(std::convert::identity)
-            .collect())
+        Ok(api_definitions.into_iter().flatten().collect())
     }
 }
 
--- a/golem-worker-bridge/src/register.rs
+++ b/golem-worker-bridge/src/register.rs
@@ -300,22 +300,19 @@ impl RegisterApiDefinition for RedisApiRegistry {
             futures::future::try_join_all(futures).await?
         };
 
-        let all_definitions: Vec<Option<ApiDefinition>> = {
-            let futures = all_versions
-                .into_iter()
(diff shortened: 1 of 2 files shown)


golem-worker-bridge/src/register.rs
```diff
@@ -194,30 +232,145 @@ impl RegisterApiDefinition for RedisApiRegistry {
...
+        let all_definitions: Vec<Option<ApiDefinition>> = {
+            let futures = all_versions
+                .into_iter()
+                .flat_map(|(api_id, versions)| {
+                    versions.into_iter().map(move |version| ApiDefinitionKey {
+                        id: api_id.clone(),
+                        version: Version(version),
+                    })
+                })
+                .map(|api_id| async move { self.get(&api_id).await })
```
justcoon, line 312: calling `GET` for each api definition in iteration, is not very good idea 
i think that redis https://redis.io/commands/mget/  should be able to fetch multiple keys with one call

nicoburniske: yeah I mentioned that in PR description. I will implement

justcoon: nice, sorry, i missed description , but i think it will be good to implement it here or in following PR ...

golem-worker-bridge/src/register.rs
```diff
@@ -140,29 +160,47 @@ impl RegisterApiDefinition for RedisApiRegistry {
...
+                    move |transaction| async move {
+                        let definition_key = get_api_definition_redis_key(&key);
+                        let version_set_key = get_api_definition_redis_key_version_set(&api_id);
+
+                        transaction
+                            .set(definition_key, definition_value, None, None, false)
+                            .await?;
+                        transaction.sadd(version_set_key, vec![api_version]).await?;
+                        transaction
+                            .sadd(get_all_apis_set_redis_key(), vec![api_id.0])
```
justcoon, line 194: it may be hard to say, how many api definitions ids we will have and how many versions for specific api definitions

if `get_all_apis_set_redis_key()` is mainly used for get_all which suppose to return all api definitions with all versions, probably it could be better if there will be `id:version` stored in this set

afsalthaj: The redis key for api definitions is actually `id:version` already in main branch

afsalthaj: I thought we will be able to fetch it with key prefix.

nicoburniske: it's possible to do with [SCAN](https://redis.io/commands/scan/), but it's not performant. I think that the two sets is overkill and it's possible to do with a single set `id:version` as peter suggested. I will fix this tomorrow

justcoon (approved):

nicoburniske:
Closed in favor of https://github.com/golemcloud/golem-services/pull/132

PR #122, opened 2024-02-20
Re architecture

afsalthaj:
On a very high level

- [x] golem-service is replaced by golem-template-service

- [x] golem-worker-bridge is replaced by golem-worker-service

- [x] In golem-worker-service, we have a narrow-down service called TemplateService whose implementation is grpc that connects to template service, which is used by the `WorkerService`

- [x] Removal of confusing structures such as GolemWorkerExecutor in worker-bridge, and a slight redesign of the gateway service (or former worker bridge service). This is replaced by a service which takes a worker-request (that was resolved from input-http-request) and respond with `http::Response`. This is done to reduce confusions between worker related code and gateway related code in worker-service. There may be more clean ups required, but major ones are done 

- [x] Making sure all docker related services are running using curl requests

- [x] Currently worker-service and template-servicce are having different URLs. So the current golem-cli will run only commands that are template related.

# TODO

- [x]  Re-route to multiple services solution, and update docker-compose, and test with golem-cli

- [x]  Making sure default URLs in golem-cli works as it is

- [x]  Testing worker-connect with gateway:  websocat ws://localhost:9881/v2/templates/df599634-9b8e-433d-9bf3-eec04b84e68c/workers/worker-adam/connect

- [x]  Testing worker-connect with gateway with golem-cli: 

- [x] Fix client generation and related tests 

- [x] Update documentation on worker-bridge

image

image

image


docker-compose-sqlite.yaml
```diff
@@ -1,6 +1,26 @@
...
+    # as it is currently set to match the default port in golem-cli:
+    # Golem base url. Default: GOLEM_BASE_URL environment variable or http://localhost:9881
+    ports:
+      - "${GOLEM_GATEWAY_PORT}:80"
+    environment:
+      - GOLEM_WORKER_SERVICE_HOST=golem-worker-service
+      - GOLEM_WORKER_SERVICE_PORT=${WORKER_SERVICE_HTTP_PORT}
+      - GOLEM_TEMPLATE_SERVICE_HOST=golem-template-service
+      - GOLEM_TEMPLATE_SERVICE_PORT=${TEMPLATE_SERVICE_HTTP_PORT}
+    command: /bin/bash -c "envsubst < /etc/nginx/nginx.conf.template > /etc/nginx/nginx.conf && nginx -g 'daemon off;'"
```
justcoon, line 18: if we have custom docker image i think we should add this to startup script, so just configuration will be here

vigoo:
You will also need to update the integration tests to run the new service too

golem-api-grpc/proto/golem/template/template_service.proto
```diff
@@ -15,6 +15,7 @@ service TemplateService {
   rpc GetTemplate (GetTemplateRequest) returns (GetTemplateResponse);
   rpc GetLatestTemplateVersion (GetLatestTemplateVersionRequest) returns (GetLatestTemplateVersionResponse);
   rpc UpdateTemplate (stream UpdateTemplateRequest) returns (UpdateTemplateResponse);
+  rpc GetVersionedTemplate(GetVersionedTemplateRequest) returns (GetVersionedTemplateResponse);
```
vigoo, line 18: I think it would be better with the following name changes:
- `GetVersionedTemplate` => `GetTemplateMetadata`
- `GetTemplate` => `GetTemplateMetadataAllVersions`

Reason: these are only returning template metadata, not the template itself. Also all templates are "versioned" so it's not a good way to distinguish between the two

golem-template-service/src/grpcapi/template.rs
```diff
@@ -90,6 +92,29 @@ impl TemplateGrpcApi {
         Ok(result.into_iter().map(|p| p.into()).collect())
     }
 
+    async fn get_template_info(
```
vigoo, line 95: I'd call it `get_template_metadata`

golem-api-grpc/proto/golem/template/template_service.proto
vigoo: Another thing: until now the gRPC API and the REST API were the same - just two versions of the same API. I would like to keep that, so let's add this new endpoint to the REST API too.

golem-worker-service/src/worker_request.rs
```diff
@@ -1,23 +1,24 @@
...
     pub template: TemplateId,
     pub worker_id: String,
     pub function: String,
     pub function_params: Value,
-    pub response_mapping: Option<ResponseMapping>,
 }
 
-impl GolemWorkerRequest {
+impl ResolvedRouteAsWorkerRequest {
+    // A worker-request can bef formed a route with resolved variables (through incoming http request)
```
vigoo, line 18: typo

golem-api-grpc/proto/golem/template/template_service.proto
afsalthaj: Sure

docker-compose-sqlite.yaml
afsalthaj: The env variables are valid only at docker-compose stage. I don't think command for env substitution be used unless we set the env within Dockerfile which is a bad practice, especially it defaults to empty strings

adamgfraser (approved):

.github/workflows/ci.yaml
```diff
@@ -278,7 +292,8 @@ jobs:
           name: golem-binaries
           retention-days: 1
           path: |
-            target/release/golem-service
+            target/release/golem-template-service
+            target/release/golem-worker-service
             target/release/golem-service-yaml
```
justcoon, line 297: there should be:
 
target/release/golem-template-service-yaml
target/release/golem-worker-service-yaml

.github/workflows/ci.yaml
```diff
@@ -318,7 +333,7 @@ jobs:
       - name: Restore executable flag
         run: chmod a+x ./target/debug/*
       - name: Generate Golem OpenAPI yaml
-        working-directory: golem-service
+        working-directory: golem-public-api
         run: ../target/debug/golem-service-yaml > ../target/golem-service.yaml
```
justcoon, line 337: this needs to be fixes
there will be two

```
run: ../target/debug/golem-template-service-yaml > ../target/golem-template-service.yaml
```


```
run: ../target/debug/golem-worker-service-yaml > ../target/golem-worker-service.yaml
```

.env
```diff
@@ -1,10 +1,12 @@
 REDIS_PORT=6380
 POSTGRES_PORT=5432
-GOLEM_SERVICE_HTTP_PORT=9881
-GOLEM_SERVICE_GRPC_PORT=9090
+GOLEM_GATEWAY_PORT=9881
```
justcoon, line 3: since docker image and service name is `golem-router`, i think we should rename variable to `GOLEM_ROUTER_HTTP_PORT`

.github/workflows/ci.yaml
afsalthaj: No need of it for now

afsalthaj: It works now because it’s merged to be golem-service-yaml. Intentional it works in CI. Not accidental.

.github/workflows/ci.yaml
afsalthaj: No need of this now. There is an automated merge now

afsalthaj: Merge of yanls

.github/workflows/ci.yaml
justcoon: ok

.github/workflows/ci.yaml
justcoon: ok

golem-router/golem-services.conf.template
```diff
@@ -0,0 +1,36 @@
...
+            proxy_pass http://$GOLEM_WORKER_SERVICE_HOST:$GOLEM_WORKER_SERVICE_PORT;
+        }
+
+        location /v2/templates {
+            proxy_pass http://$GOLEM_TEMPLATE_SERVICE_HOST:$GOLEM_TEMPLATE_SERVICE_PORT;
+        }
+
+        location / {
+            proxy_pass http://$GOLEM_TEMPLATE_SERVICE_HOST:$GOLEM_TEMPLATE_SERVICE_PORT;
+        }
```
justcoon, line 34: i think it should be possible to add and serve openapi yaml by nginx

afsalthaj: We can do it later. We have the yaml generated in the yaml directory anyway as before

justcoon (approved):

PR #139, opened 2024-03-02
Increase client_max_body_size for golem-router

afsalthaj:
To avoid errors during upload templates of larger size (debug mode especially).
A shopping-cart debug mode creates 2+ megabytes. To be safer side I increased to 4m


afsalthaj:
@adamgfraser @justcoon @vigoo @senia-psm Merging this as this is a 1 liner

PR #142, opened 2024-03-04
Enhance expr

afsalthaj:
* Introduced concept `variables` into expr. This is primary requirement to achieving the rest.
* Make expr more intelligent by parsing it to numbers, variables and booleans (and avoid evaluator to guess). Variables that take the form of numbers and booleans will only get evaluated to numbers and booleans. Example this type of a logic: `${match worker.response { some(1) -> 2, none  -> 3 }`. 1 is not considered as a variable here.
* Support `match expressions` to easily manipulate the data that are `Result`, `Option` etc. This became almost an obvious need especially when working with golem-timeline where everything that returns is either an Option or Result.

```scala

${match worker.response { ok(value) => value.id, err(msg) => 0}}}


${if request.path.id == 1 then 2 else { match worker.response { ok(value) => value.id, err(msg) => 3 }} }

// there is infact no need of `{` before else

${if request.path.id == 1 then 2 else match worker.response { ok(value) => value.id, err(msg) => 3 }} 

// And users can try simple select field symbols to manipulate even further, such as handling arrays, map

${if request.path.id == 1 then 2 else match worker.response { ok(value) => value.ids[0], err(msg) => 3 }} 
```

* Along with pattern matching, a simple constructor is regarded as an expr, so that it automatically works with if conditions too. `if some(variable) = worker.response then variable.id else 1`. This pattern is not tested yet, but it's in my radar.

* Custom patterns are currently not supported yet but the design done in this PR already paves the path and easier to achieve. Just that I ran out of time, and would like follow up PR.  There are a few details to consider for random matches. Example: `match person(age, name) ` sort of assumes, the result has exactly the field `age`, `name` etc. I didn't want to bringin in such level of assumptions in this PR. Therefore, reverted some of these evaluations, and kept only the data structure `ConstructorPattern` that support it in future.

* The support of variable and distinction of a variable from a raw string is important here so as to look up the value for a variable in a particular context. Example: `${ match worker.response { some(person) => person.age, non => 0 } }`. Previously `person.age` was never supported, and now it does.
* Introduced the concept of `quotes` (`'`) to avoid the confusion between Strings and code within an interpolated aka code context. Example: Things like the following become possible and is more intuitive. There was a false positive tests (status 401 instead of 200, in evaluator.rs) which got fixed with this approach

```scala
 // If there were no quotes, evaluator tries to find the value of the variable `person` (which exists in the response throuh some(person)) and `bar` (which doesn't exist and will fail)

${ match worker.response { 
      some(person) => 'person', 
      none => 'bar' 
}}

// Within quotes, users will have the ability to interpolate. This way, they have maximum flexibility that within a code context, they can write literals and append it with proper variable result

  ${ match worker.response  { 
        some(person) => 'age-${person.age}', 
        none => 'age-0' 
   }} 

```
* More tests are added, while re-writing a dozen of existing ones (due to wrong logic, and harder maintainability)
As part of these changes, I had to revisit certain logic and design such that it's a scalable parsing logic. There are tests in evaluator.rs that automatically just worked as you can see. 

There will be follow up PRs in coming days, to make this as best as possible. Using gateway functionality in `golem-timeline` (so as to find the status of each timeline worker, and to retrieve results as an API call) will really help fine tuning this worker-bridge functionalities.


golem-worker-service/src/evaluator.rs
```diff
@@ -282,13 +282,140 @@ impl Evaluator<Value> for Expr {
...
+                    }
+
+                    resolved_result
+                        .map(|value| ValueTyped::from_json(&value))
+                        .ok_or(EvaluationError::Message(
+                            "Pattern matching failed".to_string(),
+                        ))
+                }
+                Expr::Constructor0(_) => Err(EvaluationError::Message(
+                    "Constructor0 is not supported yet".to_string(),
```
afsalthaj, line 417: Will clean this big logic in a follow up PR.

afsalthaj: Rust is not super friendly writing higher order functions in my view :P

vigoo (approved):

afsalthaj:
This issue closes #211

PR #150, opened 2024-03-06
GOL-188 Compilation Server

nicoburniske:
Configurable ahead of time template compilation service. If enabled, every time a template is created or updated, then we will call the template-compilation-service to compile the template and put result into `CompiledTemplateService`.

### TODO
- [ ]  Fix Docker Examples docker-compose files to include template-compilation-service
- [ ] Update Kubernetes helm charts

https://www.notion.so/GOL-188-Compilation-Server-61682806d9e14ba38ca2241d3a21bb81

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1323304668) by [Unito](https://www.unito.io)


golem-compilation-server/src/service/upload_worker.rs
```diff
@@ -0,0 +1,72 @@
...
+        compiled_template_service: Arc<dyn CompiledTemplateService + Send + Sync>,
+        recv: mpsc::Receiver<CompiledTemplate>,
+        cancellation: CancellationToken,
+    ) {
+        let worker = Self {
+            retry_config,
+            compiled_template_service,
+        };
+
+        let mut stream = ReceiverStream::new(recv)
```
nicoburniske, line 32: Multiple Upload "workers" uploading compiled files with buffered parallelism

nicoburniske: @vigoo

vigoo: I don't think it matters much to parallelize these uploads - I expect the compilation itself takes orders of magnitude more time, so this will likely never really parallelize anything. We can keep it simple and just do it one by one

vigoo:
Looks good, some remaining things to do:
- we should have queue size metrics otherwise we won't have any idea what's happening 
- add to CI, including docker container building
- add to integration tests (probably with the ability to turn it on and off - so we can verify that the it's not an essential component)
- add it to the docker-compose files

golem-compilation-server/src/service/upload_worker.rs
nicoburniske: yeah you're completely right. I didn't account for compilation happening slower, so there's no need for parallel uploads. great catch.

golem-template-compilation/src/config.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub host: String,
+    pub port: u16,
+    pub access_token: String,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct TemplateServiceConfig {
+    pub host: String,
+    pub port: u16,
+    pub access_token: Uuid,
```
afsalthaj, line 38: Probably this isn't required.

golem-template-compilation/src/grpc.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+#[async_trait]
+impl GrpcCompilationServer for CompileGrpcService {
+    async fn enqueue_compilation(
+        &self,
+        request: Request<TemplateCompilationRequest>,
+    ) -> Result<tonic::Response<TemplateCompilationResponse>, Status> {
+        let response = match self.enqueue_compilation_impl(request.into_inner()).await {
+            Ok(_) => template_compilation_response::Result::Success(Empty {}),
```
afsalthaj, line 34: I would rather inline this instead of separating it under another impl ....

afsalthaj (changes requested):

afsalthaj:
Some very minor stuffs.  LGTM

justcoon (approved):

PR #153, opened 2024-03-08
GOL-220 Shard manager improvements part 1

vigoo:
- Rebalancing extracted to separate module
- Tests for rebalancing
- Improved rebalancing logic based on the tests, support up/downscaling and registering more than one pod at once
- Routing table is managed in an "actor like" way, with registering and healthcheck only indicating changes in the available pods in a non-blocking way. A single thread applies the changes to the routing table and sends out the updates to the workers and manages the persistency.
- The register gRPC call no longer returns immediately the associated shards, so it can use the above mentioned non-blocking registration. The registered workers gets assign messages the next time the routing table gets updated (which is immediate if there was no change in process already)
- Some small refactorings to reduce the number of clones
- Switched to `BTreeMap` for reproducible results

The new/improved rebalacing algorithm:

- First we take a note if there are any registered pods that are completely empty. 
- Then we go through each _unassigned shard_ and associate them with pods. First we fill up the originally empty (new) pods, up to the optimal shard count which is (shard_count/pod_count). Once we reached that the rest of the unassigned shards are distributed evenly among all pods.
- At this point it is not guaranteed that the shard assignment is balanced, especially if there was a change in the number of available pods.
- So we improve the balancing in iterations, each iteration we pick a target pod which has less shards than the optimal (with some threshold - the threshold reduces the number of revokes/assigns by allowing some level of unbalancedness). 
- We move shards from other pods - except the ones that were initially empty as they got filled up with the unassigned pods and we want to keep those together to reduce movements - to the picked target pod
- We repeat this until we no longer have any pods below the threshold

### DIFF
--- a/api-gateway-examples/README.md
+++ b/api-gateway-examples/README.md
@@ -12,7 +12,7 @@ The API definition essentially comprises a set of endpoints, each associated wit
 
 The registration process for this endpoint definition is straightforward. Further configuration details can be discussed later as needed.
 
-## An example, including integration with an externa API Gateway
+## An example, including integration with an external API Gateway
 
 After deploying our Golem service, integration with external API Gateways, like Tyk, becomes straightforward. 
 This setup enables us to utilize the robust infrastructure and security features provided by the API Gateway while leveraging Golem for request processing.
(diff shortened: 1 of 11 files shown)


afsalthaj:
That was quick, I would say. Wow!

PR #156, opened 2024-03-13
sharding test

senia-psm:


scripts/it.sh
```diff
@@ -7,4 +7,4 @@ cd "${script_full_path}"/.. || exit
 
 ./scripts/build-all.sh
 cd golem-cli
-RUST_LOG=info cargo test
+RUST_LOG=info cargo test --test integration
```
vigoo, line 10: maybe run sharding tests here too?

senia-psm: It's quite unstable, but let's try any way.
Done.

vigoo (approved):

PR #250, opened 2024-03-18
Get/set oplog index host functions

### LINKED ISSUE #160: Implement get/set oplog index

vigoo:
- Resolves #160 
- Fixes an issue where previous instance of a retried worker was not dropped
- Unifies the oplog index type to be `u64` everywhere
- Adds version (`0.2.0`) to the Golem host WIT interface
- Required `golem-wit` change: https://github.com/golemcloud/golem-wit/pull/9


vigoo:
I have just realized this is not fault tolerant yet - as it ignores `set-oplog-index` in replay mode assuming that when it reaches it it already "performed the jump". But that is only true if it is replaying because previously it reached that point in live mode. If it did the jump once already and moved forward in the execution and got recovered for any other reason - failure, or just moving shards or recovering a suspended instance - then it won't work correctly.

In an earlier version of the PR i had a solution for it I think but got removed. It would be something like this:
- when reaching `set-oplog-index` in replay mode, if it is not in the active jumps table, then we append a new jump entry to the oplog and restart the recoevry
- it requires some attention to properly skip these extra jump entries when needed

vigoo:
Or an alternative is to not just look at the last oplog entry on startup but scan the whole oplog in advance and collect all jump entries. Then we always know about the jumps during recovery and we can do it in one run. This is probably more efficient than restarting the recoveries anyway.
And if we do this we don't need to accumulate the jumps in the jump entries like in the current version of the PR because we can accumulate them while traversing the oplog.

vigoo:
And this scanning approach can be made optimal similar to how the worker status is handled - store the jumps in a separate redis key with the oplog index it was last updated at, and then we only need to scan the newer oplog entries.

vigoo:
So to sum it up, the PR will be extended with the following:

- `OplogEntry::Jump` only stores a single jump not a list of jumps
- There is a separate redis key for storing the last known set of jumps
- On worker startup we don't just check the last oplog entry but load this cached value, (and use empty list @ idx = 0 as default) and scan the oplog entries after the last update index to look for additional jump entries

These changes will make sure the jumps are repeated correctly in any case of failure.

golem-worker-executor-base/src/grpc.rs
```diff
@@ -852,6 +852,7 @@ impl<Ctx: WorkerCtx, Svcs: HasAll<Ctx> + UsesAllDeps<Ctx = Ctx> + Send + Sync +
                 while let Ok(item) = receiver.recv().await {
                     match item {
                         worker_event::WorkerEvent::Close => {
+                            debug!("WORKER_EVENT_STREAM CLOSED");
```
justcoon, line 855: we should add worker id to log

vigoo: Ah, I did not want to leave this log in it it was just for debugging

golem-worker-executor-base/src/durable_host/golem/mod.rs
```diff
@@ -82,6 +84,76 @@ impl<Ctx: WorkerCtx> golem::api::host::Host for DurableWorkerCtx<Ctx> {
...
+                    self.private_state.active_jumps.add_jump(jump);
+                    self.set_oplog_entry(OplogEntry::jump(
+                        Timestamp::now_utc(),
+                        self.private_state.active_jumps.jumps().clone(),
+                    ))
+                    .await;
+                    self.commit_oplog().await;
+
+                    debug!(
+                    "Interrupting live execution for jumping from {jump_source} to {jump_target}"
```
justcoon, line 146: is good to add workerId to logs

golem-common/src/model.rs
```diff
@@ -732,45 +732,64 @@ impl Display for AccountId {
...
+    /// Worker failed
+    Error { timestamp: Timestamp },
+    /// Debug log entry
     Debug {
         timestamp: Timestamp,
         message: String,
     },
+    /// Marker entry added when get-oplog-index is called from the worker, to make the jumping behavior
+    /// more predictable.
+    Marker { timestamp: Timestamp },
```
jdegoes, line 786:
```suggestion
    NoOp{ timestamp: Timestamp },
```

Or `Noop`.

jdegoes (changes requested):
Looks great! A few suggested revisions:

1. Rename `Marker` to `NoOp`, or possibly remove it, if we are writing oplog entries for all invocations of Golem runtime API (in which case, `get_oplog_index` would always return the oplog index of itself).
2. Rename `ActiveJumps` to `DeletedRegions`, and update naming of all other relevant parts of the code, to reflect the conceptual model we discussed.
3. Make it an error to jump to an index inside a (currently) deleted region. This way, we can later, during compaction, safely delete without changing semantics.
4. Implement the optimizations you mentioned (caching deleted regions so we don't have to replay them during recovery, if already available in the instance cache).
5. Anything else we discussed.

golem-common/src/model/jumps.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+impl Display for DeletedRegions {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "[{}]",
+            self.jumps
+                .iter()
+                .map(|jump| format!("<{} => {}>", jump.source_oplog_idx, jump.target_oplog_idx))
```
justcoon, line 94: it could be good to add `Display` for `Jump`

golem-common/src/model/regions.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    /// Adds a new region to the list of deleted regions
+    pub fn add(&mut self, region: OplogRegion) {
+        self.regions.push(region);
+    }
+
+    /// Checks whether there is an deleted region starting at the given oplog index.
+    /// If there is, returns the oplog index where the execution should continue, otherwise returns None.
+    pub fn is_deleted_region_start(&self, oplog_index: u64) -> Option<u64> {
+        // TODO: optimize this by introducing a map during construction
```
jdegoes, line 77: I don't think this `TODO` can wait because this function is called on every single oplog entry during recovery.

Rather than using a map, it would be even more efficient to not check on a per oplog-entry-basis, and just grab the next offset during replay (i.e. boundary-aware replay logic).

vigoo: I agree. Did not want to leave it there just forgot about it! Will fix

jdegoes (approved):

golem-common/src/model/regions.rs
```diff
@@ -26,6 +32,28 @@ impl OplogRegion {
     pub fn contains(&self, target: u64) -> bool {
         target > self.end && target <= self.start
     }
+
+    pub fn union(&self, other: &OplogRegion) -> Option<OplogRegion> {
```
jdegoes, line 36: Nice!

jdegoes (approved):

PR #269, opened 2024-03-19
Worker Service Base: Rework Error Handling 

nicoburniske:


golem-worker-service-base/src/api_definition_repo.rs
```diff
@@ -41,30 +40,17 @@ pub trait ApiDefinitionRepo<Namespace: ApiNamespace> {
...
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, thiserror::Error)]
 pub enum ApiRegistrationRepoError {
+    #[error("AlreadyExists: ApiDefinition with id: {} and version: {} already exists in the namespace {}", .0.id, .0.version, .0.namespace)]
     AlreadyExists(ApiDefinitionKey<String>),
+    #[error("InternalError: {0}")]
     InternalError(String),
 }
 
-impl From<Box<dyn Error>> for ApiRegistrationRepoError {
```
afsalthaj, line 50: 👍

golem-worker-service-base/src/service/api_definition_service.rs
```diff
@@ -225,24 +187,22 @@ impl<Namespace: ApiNamespace, AuthCtx: Send + Sync> ApiDefinitionService<Namespa
...
+        Ok(result)
     }
 
     async fn delete(
         &self,
         api_definition_id: &ApiDefinitionId,
         version: &Version,
         auth_ctx: AuthCtx,
-    ) -> Result<ApiDefinitionIdAnnotated<Namespace>, ApiRegistrationError> {
+    ) -> Result<bool, ApiRegistrationError> {
```
afsalthaj, line 205: Shift this to Error, if `bool` is false. Clients don't care about this bool

afsalthaj: It is ok to propogate the info that "hey this api definition is deleted from this namespace", rather than just a `bool`. In other words, both ways are opinionated and can be justified, so we pick the one that will cause less  pain in the usage sites. Now if you change this to false, the cloud code base will need to change

afsalthaj (approved):

PR #272, opened 2024-03-19
stable sarding test

senia-psm:


vigoo:
Integrated into https://github.com/golemcloud/golem/pull/352

Issue #289, opened 2024-03-21
Make Failed and Exited status computable from oplog


vigoo:
Currently we don't store enough information in the oplog to be able to restore the `Failed` (after N retries) and `Exited` (when the error was the special exit signal) statuses.

The `Error` oplog entries are not holding any information and can only be used to decide if there were N retries at the end of the oplog.

Some changes should be made:
- Enough information in the error entries to be able to create a `RecoveryDecision` just from oplog entries
- Maybe make retry count a cached value in `WorkerStatusRecord`
-

PR #290, opened 2024-03-21
Interrupt oplog entry

### LINKED ISSUE #239: Add interruption mark to the oplog
We want worker status to be fully recalculateable from the oplog, even though we cache it.

vigoo:
Resolves #239 

Created follow-up task #289


vigoo:
`wasi::directories_replay` test broken, to be checked

jdegoes (approved):

PR #307, opened 2024-03-22
Write exit and error details to the oplog

### LINKED ISSUE #289: Make Failed and Exited status computable from oplog

vigoo:
Resolves #289


golem-client/build.rs
```diff
@@ -44,3 +46,20 @@ fn generate(yaml_path: PathBuf, out_dir: OsString) {
...
+        let b = std::fs::read(&dst)?;
+        if a != b {
+            std::fs::copy(src, dst)?;
+        }
+        Ok(())
+    } else {
+        std::fs::copy(src, dst)?;
+        Ok(())
+    }
+}
```
afsalthaj, line 65: 👍

golem-common/src/model/oplog.rs
```diff
@@ -218,6 +227,22 @@ pub enum WrappedFunctionType {
     WriteRemote,
 }
 
+/// Describes the error that occurred in the worker
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Encode, Decode)]
+pub enum WorkerError {
+    Unknown(String),
+    StackOverflow,
```
afsalthaj, line 234: Interesting

golem-worker-executor-base/src/durable_host/mod.rs
```diff
@@ -731,73 +733,66 @@ impl<Ctx: WorkerCtx> InvocationHooks for DurableWorkerCtx<Ctx> {
...
+                    true
+                }
+                TrapType::Exit => {
+                    self.set_oplog_entry(OplogEntry::Exited {
+                        timestamp: Timestamp::now_utc(),
+                    })
+                    .await;
+                    true
+                }
+                _ => false,
```
afsalthaj, line 763: So for rest of the types , we don't need to set the oplog entry

golem-common/src/model/oplog.rs
vigoo: Currently we only distinguish stack overflow because that is never retried.
I expect this will be extended in the future, for example with http-specific errors for which we need custom retry policies, etc.

jdegoes (approved):

PR #311, opened 2024-03-23
Worker Base Service: Reuse Refactoring

nicoburniske:
- Add `<Namespace, AuthContext>` to TemplateService 
- `<AuthContext>` to ApiDefinitionValidationService
- Template Service Errors are now agnostic to GPRC. 
- Template Service doesn't return optional template anymore. Already has NotFound error
- Added `NotFound` error case to `AuthError`. Useful when authorizing access to resources, which you are not sure if they exist yet. 
- Rely on Error Display implementations instead of inner string values. 

### Further Considerations
- Sanitizing error messages (especially Internal error)
- Use of anyhow::Error instead of String for `InternalError`. Useful to keep track of cause / backtrace. Can also use downcasting to see the inner error type. Useful for hiding internal implementation details. See `is_retriable` in TemplateService


golem-worker-service-base/src/service/api_definition.rs
```diff
@@ -112,14 +112,14 @@ pub enum ApiRegistrationError {
 pub struct RegisterApiDefinitionDefault<Namespace, AuthCtx> {
     pub auth_service: Arc<dyn AuthService<AuthCtx, Namespace> + Sync + Send>,
     pub register_repo: Arc<dyn ApiDefinitionRepo<Namespace> + Sync + Send>,
-    pub api_definition_validator: Arc<dyn ApiDefinitionValidatorService + Sync + Send>,
+    pub api_definition_validator: Arc<dyn ApiDefinitionValidatorService<AuthCtx> + Sync + Send>,
```
afsalthaj, line 115: @nicoburniske The type annotation conceptually says, validation requires some operation that requires authorisation. This might be due to the fact we authorise getting the template.

L:et me know if I am wrong here. I think, separation of concerns is the problem

I think Validation can be just a mere function, and let the top level worker-server service layer be responsible for fetching the template metadata, and then call this reusable validation function. This way validation can exist as a pure function testable without worrying about permissions.

 Or if we still think validation itself should be trait interface, still, worker-service (the caller) should be responsible for fetching the template metadata from an outside service, such that validation can stay as pure and testable as possible

golem-worker-service-base/src/service/api_definition_validator.rs
```diff
@@ -11,8 +12,12 @@ use golem_service_base::model::{Export, ExportFunction, ExportInstance, Template
 use serde::{Deserialize, Serialize};
 
 #[async_trait]
-pub trait ApiDefinitionValidatorService {
-    async fn validate(&self, api: &ApiDefinition) -> Result<(), ValidationError>;
+pub trait ApiDefinitionValidatorService<AuthCtx> {
+    async fn validate(
+        &self,
+        api: &ApiDefinition,
+        auth_ctx: &AuthCtx,
```
afsalthaj, line 19: Instead of auth_ctx, you could take template-metadata. I believe, conceptually, that's a much reusable way of doing it.  If not now, later we would need it.

golem-worker-service-base/src/service/api_definition_validator.rs
```diff
@@ -42,19 +47,30 @@ impl RouteValidationError {
...
-    template_service: Arc<dyn TemplateService + Send + Sync>,
+pub struct ApiDefinitionValidatorDefault<Namespace, AuthCtx> {
+    template_service: Arc<dyn TemplateService<Namespace, AuthCtx> + Send + Sync>,
 }
 
-impl ApiDefinitionValidatorDefault {
-    pub fn new(template_service: Arc<dyn TemplateService + Send + Sync>) -> Self {
+impl<Namespace, AuthCtx> ApiDefinitionValidatorDefault<Namespace, AuthCtx> {
+    pub fn new(
+        template_service: Arc<dyn TemplateService<Namespace, AuthCtx> + Send + Sync>,
```
afsalthaj, line 56: As mentioned before, Validator could forget about any other interface such as template-service here. Being able to fetch the output with defined=-input than in-directions generally pay off better in my experience. Example: When writing a test Generating invalid template-metadata using `arbitrary` for a given `ApiDefinition` is easier with that approach. Otherwise you have to define a type, which takes ApiMetadata, and then write a template-service for it that always emits conflicting template metadata, and then having to figure out another type (a new type) for which it always succeeds etc and that seems to be an overkill.  Let me know what you think.

golem-worker-service-base/src/service/error.rs
afsalthaj, line 1: Nice. I was thinking what to do with this orphan errors (struggled with what was already in there). Thanks for this one!

golem-worker-service-base/src/service/mod.rs
```diff
@@ -1,6 +1,23 @@
...
-pub mod error;
 pub mod http_request_definition_lookup;
 pub mod template;
 pub mod worker;
+
+#[derive(Clone, Debug)]
+pub struct TemplatePermission {
+    pub template: TemplateId,
+    pub permission: Permission,
+}
```
afsalthaj, line 14: Fine grained approach -that we aleays discussed. Its nice, but keep a general principle that a fine grained approach often result in too-many possibilities, of which some of them could be illegal. Example, what does it mean to have a product of `Permission::Create` for a `TemplateId` that already exist

afsalthaj: So if keeping fine-grained approach, the `Permission` should be different type too, to disallow bumping into this property. For an existing `TemplateId` , only `View` , `Update` and `Delete` is possible

golem-worker-service-base/src/service/template/error.rs
```diff
@@ -0,0 +1,120 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum TemplateServiceError {
+    #[error(transparent)]
+    Auth(#[from] AuthError),
+    #[error("Bad Request: {0:?}")]
+    BadRequest(Vec<String>),
+    #[error("Unauthorized: {0}")]
+    Unauthorized(String),
+    #[error("Limit Exceeded: {0}")]
+    LimitExceeded(String),
```
afsalthaj, line 18: Are we bumping into this error in OSS?

afsalthaj:
As always the changes looks like bringing cleanliness into the code. Nice and important refactoring!

I have commented on a few things. If any of them don't make sense, happy to pair up. Or feel free to take this up as a follow up PR if you want to progress quickly. In both ways,  approved already :)

afsalthaj (approved):

golem-worker-service-base/src/service/api_definition_validator.rs
justcoon: i think it is about, that it require to call TemplateService, which require authorisation

but sure other solution, could be that as namespace is result of authorisation, namespace could also contain authorisation data, which can be used in that invocation

golem-worker-service-base/src/service/template/error.rs
justcoon: i think it should not be in OSS, but not sure, if with this approach it possible to do it better, 
maybe here should be some Custom error, which will be specially handled in cloud version

golem-worker-service-base/src/service/api_definition.rs
nicoburniske: This makes perfect sense. Much better approach

golem-service-base/src/service/auth.rs
```diff
@@ -16,6 +21,8 @@ pub enum AuthError {
     Unauthorized(String),
     #[error("Forbidden: {0}")]
     Forbidden(String),
+    #[error("Not found: {0}")]
+    NotFound(String),
```
justcoon, line 25: it is little bit worrying  that Auth Service can return error like NotFound

scenario like: it will return error like Account not found which with some details, which in my opinion is security issue

golem-worker-service-base/src/service/api_definition.rs
```diff
@@ -141,12 +147,54 @@ impl<Namespace: ApiNamespace, AuthCtx> RegisterApiDefinitionDefault<Namespace, A
...
+        auth_ctx: &AuthCtx,
+    ) -> Result<Vec<Template>, ApiRegistrationError> {
+        let get_templates = definition
+            .routes
+            .iter()
+            .cloned()
+            .map(|route| (route.binding.template.clone(), route))
+            .collect::<HashMap<_, _>>()
+            .into_values()
+            .map(|route| async move {
```
justcoon, line 162: i would prefer more precise errors:

in this case: if there are 3 routes with same template id
it will report error like Error getting latest template for  route 3.
in my opinion proper error reporting should say:
- Error getting latest template for  route 1.
- Error getting latest template for  route 2.
- Error getting latest template for  route 3.

nicoburniske: This is how it currently works

golem-service-base/src/service/auth.rs
nicoburniske: What is the alternative? We already have similar errors everywhere in the codebase, especially in grpc types. What if you are trying to authenticate a given template id, and then that template id doesn't exist? What should AuthService return? 

The only alternative that I can think of is returning an anyhow::Error and downcasting it? Or including a new type parameter for the error type in AuthService? 

This is also an internal service, and the error messages don't necessarily need to be propagated to API level.

golem-worker-service-base/src/service/api_definition_validator.rs
nicoburniske: This was implemented. Removed auth in favor of explicitly passing all templates to validator

golem-worker-service-base/src/service/template/error.rs
nicoburniske: These errors are the same as the grpc errors

https://github.com/golemcloud/golem/blob/main/golem-api-grpc/proto/golem/template/template_error.proto

Issue #321, opened 2024-03-25
Reduce redundancy in DurableWorkerCtx


vigoo:
There is an unnecessary level of indirection between DurableWorkerCtx and its `private_state` member. Code should be simplified and cleaned up, avoiding host function implementations to call a mix of the outer and the inner layer.

PR #328, opened 2024-03-27
Implementation of atomic operation regions

### LINKED ISSUE #162: Implement begin/end markers

vigoo:
Resolves #162 

Contains some refactorings related to #321 (it was confusing what goes into the `DurableWorkerCtx` and what goes into the `.private_state` level so removed some unnecessary duplications, and added a macro for looking up the next oplog entry instead of duplicating the same loop for each type)


golem-common/src/model/oplog.rs
```diff
@@ -72,6 +72,16 @@ pub enum OplogEntry {
...
     },
+    /// Begins an atomic region. All oplog entries after `BeginAtomicRegion` are to be ignored during
+    /// recovery except if there is a corresponding `EndAtomicRegion` entry.
+    BeginAtomicRegion { timestamp: Timestamp },
+    /// Ends an atomic region. All oplog entries between the corresponding `BeginAtomicRegion` and this
+    /// entry are to be considered during recovery, and the begin/end markers can be removed during oplog
+    /// compaction.
+    EndAtomicRegion {
+        timestamp: Timestamp,
+        begin_index: u64,
```
afsalthaj, line 83: 👍

golem-worker-executor-base/src/durable_host/golem/mod.rs
```diff
@@ -173,35 +170,94 @@ impl<Ctx: WorkerCtx> golem::api::host::Host for DurableWorkerCtx<Ctx> {
...
+
+                    // We need to jump to the end of the oplog
+                    self.state.oplog_idx = self.state.oplog_size;
+
+                    // But this is not enough, because if the retried transactional block succeeds,
+                    // and later we replay it, we need to skip the first attempt and only replay the second.
+                    // Se we add a Jump entry to the oplog that registers a deleted region.
+                    let deleted_region = OplogRegion {
+                        start: begin_index + 1,         // need to keep the BeginAtomicRegion entry
+                        end: self.state.oplog_size + 1, // skipping the Jump entry too
```
afsalthaj, line 208: 👁️

golem-worker-executor-base/src/durable_host/golem/mod.rs
```diff
@@ -173,35 +170,94 @@ impl<Ctx: WorkerCtx> golem::api::host::Host for DurableWorkerCtx<Ctx> {
...
+                }
+                None => {
+                    debug!("Worker {}'s atomic operation starting at {} is not committed, ignoring persisted entries", self.worker_id, begin_index);
+
+                    // We need to jump to the end of the oplog
+                    self.state.oplog_idx = self.state.oplog_size;
+
+                    // But this is not enough, because if the retried transactional block succeeds,
+                    // and later we replay it, we need to skip the first attempt and only replay the second.
+                    // Se we add a Jump entry to the oplog that registers a deleted region.
```
afsalthaj, line 205: Could you explain this part? Not getting it properly

test-templates/runtime-service/src/lib.rs
```diff
@@ -74,6 +74,38 @@ impl Guest for Component {
...
         println!("Finished commit");
     }
+
+    fn atomic_region() {
+        let now = std::time::SystemTime::now();
+        println!("Starting atomic region at {now:?}");
+
+        let begin = mark_begin_operation();
+
+        remote_side_effect("1"); // repeated 3x
```
afsalthaj, line 84: repeated 3x here means?

golem-worker-executor-base/src/durable_host/golem/mod.rs
vigoo: The oplog is append only. So let's imagine that on first run, we fail in the middle of an atomic operation:

```
x = arbitrary oplog entry
B = atomic region start
F = failure
E = atomic region end

x x x B x x F
```

now we restart the worker, and replay the oplog. when reaching B, we jump to the end and start writing new entries (now `y`) but fail again:

```
x x x B x x F y y F
```

we restart it again, and when we replay we do the same, writing `z`s but now it does not fail within the block, but fails later:

```
x x x B x x F y y F z z z z E z z F
```

Then we restart it again and when we reach `B`, we find the corresponding `E` so we know the region was committed and we don't need to rerun it. 
This means that we need to act like only `z z z z` is between `B` and `E`, otherwise we go out of sync (because how it works is that the WASM code always runs thet same, and when it tries doing some side effect, we read the next oplog entry.

So we have to skip the entries between the first `x` and the last `F` in the region and we can use the same mechanism as we use for the "time travel jumps" here.

test-templates/runtime-service/src/lib.rs
vigoo: there is a failure between side effect "2" and "3" which is controlled from the outside in a way that first two times it fails, third time it is not. Because this side effect is between begin/end operation markers, it is going to be performed three times, but not more, because after the third the region is committed.

afsalthaj (approved):

PR #332, opened 2024-03-27
Remove AuthService and Namespace

nicoburniske:
AuthService and Namespace are removed. Auth will now have to be done by wrapping OSS with another service in cloud, that performs authorization. this pattern was extremely leaky and led to to many generics

- There are two TemplateService traits, one in `golem-worker-service-base` to lookup templates, and the full one in `golem-template-service`
- The RemoteTemplateService impl has a type constraint on the AuthCtx generic parameter to allow for adding arbitrary metadata entries such as authorization to grpc request

```rust
impl<AuthCtx> TemplateService<AuthCtx> for RemoteTemplateService
where
    AuthCtx: IntoIterator<Item = (String, String)> + Clone + Send + Sync {
//...
}
```

First part of https://github.com/golemcloud/golem/issues/335


golem-worker-service-base/src/service/template/default.rs
```diff
@@ -10,86 +11,65 @@ use golem_common::config::RetryConfig;
...
 }
 
 #[async_trait]
-impl<AuthCtx, Namespace> TemplateService<AuthCtx, Namespace>
-    for TemplateServiceDefault<AuthCtx, Namespace>
+impl<AuthCtx> TemplateService<AuthCtx> for RemoteTemplateService
 where
-    Namespace: Send + Sync,
-    AuthCtx: Clone + Send + Sync,
+    AuthCtx: IntoIterator<Item = (String, String)> + Clone + Send + Sync,
```
nicoburniske, line 50: For RemoteTemplateService, which will use grpc, I added a constraint to create a Metadata map, which will add entries to the grpc request metadata using `with_metadata`

In cloud we can put headers and such here.

golem-worker-service-base/src/auth.rs
```diff
@@ -16,6 +11,15 @@ impl Display for EmptyAuthCtx {
     }
 }
 
+impl IntoIterator for EmptyAuthCtx {
```
nicoburniske, line 14: EmptyAuth will not have any metadata entries

golem-worker-service-base/src/service/api_definition.rs
```diff
@@ -104,53 +110,34 @@ impl<Namespace: Display> ApiDefinitionKey<Namespace> {
...
 where
-    AuthCtx: Send + Sync,
-    ApiDefNamespace: ApiNamespace,
-    TemplateNamespace: std::fmt::Debug + Send + Sync,
+    Namespace: ApiNamespace + Send + Sync,
 {
     pub fn new(
-        template_service: Arc<dyn TemplateService<AuthCtx, TemplateNamespace> + Send + Sync>,
-        auth_service: Arc<dyn AuthService<AuthCtx, ApiDefNamespace> + Sync + Send>,
-        register_repo: Arc<dyn ApiDefinitionRepo<ApiDefNamespace> + Sync + Send>,
```
afsalthaj, line 132: Ok, this says its a good change.

golem-worker-service-base/src/service/api_definition.rs
```diff
@@ -191,33 +177,27 @@ where
...
-    ApiDefNamespace: ApiNamespace,
-    TemplateNamespace: std::fmt::Debug + Send + Sync,
+    Namespace: ApiNamespace + Send + Sync,
 {
     async fn register(
         &self,
         definition: &ApiDefinition,
-        auth_ctx: AuthCtx,
-    ) -> ApiResult<ApiDefinitionId, ApiDefNamespace> {
-        let namespace = self.is_authorized(Permission::Create, &auth_ctx).await?;
```
afsalthaj, line 218: Why is this deleted from ApiDefinitionService?

nicoburniske: the diff looks confusing. AuthCtx is still there

golem-worker-service-base/src/service/api_definition.rs
```diff
@@ -273,88 +248,83 @@ where
...
-        Ok(WithNamespace { value, namespace })
+        Ok(value)
     }
 
     async fn get_all_versions(
         &self,
         api_id: &ApiDefinitionId,
-        auth_ctx: AuthCtx,
-    ) -> ApiResult<Vec<ApiDefinition>, ApiDefNamespace> {
-        let namespace = self.is_authorized(Permission::View, &auth_ctx).await?;
```
afsalthaj, line 290: Why is this deleted? from worker-bridge part?

Issue #295, opened 2024-03-22
Enumeration in worker executor service


vigoo:
Create a new service trait and implementation in `worker-executor-service-base` with two public operations (in pseudo code):

- `get-running-workers(template_id, filter) -> vec<metadata>`
- `scan-workers(template_id, filter, precise, cursor) -> option<(cursor, vec<metadata>)>`

This service should have a default “live implementation”, which:

- uses the `ActiveWorkers`  to enumerate all workers in memory
- uses a combination of `Redis` and the `WorkerService` to perform the `SCAN` to lazily enumerate all workers

Both operations should use a common implementation of applying the filter to the chunk of worker metadata.

We should have a `precise` bool flag (or some mode enum with two modes for now) with the following meaning:

- When it is `true`, `scan-workers` must check every worker metadata’s cached part (`WorkerStatusRecord` in the code) to see if it’s `oplog_index` is the same as the oplog’s size (requires a `XLEN` redis operation - to be done through `OplogService`). If not, an up-to-date `WorkerStatusRecord` needs to be generated `calculate_last_known_status`. This needs to be done *before* applying the filters (at least that part of the filter which is related to status)

Issue #296, opened 2024-03-22
Expose the worker enumeration interface via gRPC


vigoo:
Expose the two methods defined in https://github.com/golemcloud/golem/issues/295 via gRPC

PR #337, opened 2024-03-28
Worker enumeration

justcoon:
fix: #295, #296


golem-common/src/model/mod.rs
```diff
@@ -777,12 +793,416 @@ pub fn parse_function_name(name: &str) -> ParsedFunctionName {
...
+//     pub comparator: FilterComparator,
+//     pub value: i32,
+// }
+//
+// #[derive(Clone, Debug, Serialize, Deserialize, Encode, Decode)]
+// pub struct WorkerVersionEnvFilter {
+//     pub name: String,
+//     pub comparator: StringFilterComparator,
+//     pub value: String
+// }
```
justcoon, line 819: will remove after we clarify Filters

golem-worker-executor-base/src/grpc.rs
```diff
@@ -668,6 +674,62 @@ impl<Ctx: WorkerCtx, Svcs: HasAll<Ctx> + UsesAllDeps<Ctx = Ctx> + Send + Sync +
...
+            _ => None,
+        };
+
+        let (new_cursor, workers) = self
+            .worker_enumeration_service()
+            .get(
+                &template_id,
+                filter,
+                request.cursor as usize,
+                request.count as usize,
```
justcoon, line 723: should we use `usize` or `u64`?

golem-worker-executor-base/src/services/mod.rs
```diff
@@ -158,6 +175,9 @@ pub struct All<Ctx: WorkerCtx> {
     template_service: Arc<dyn template::TemplateService + Send + Sync>,
     shard_manager_service: Arc<dyn shard_manager::ShardManagerService + Send + Sync>,
     worker_service: Arc<dyn worker::WorkerService + Send + Sync>,
+    worker_enumeration_service: Arc<dyn worker_enumeration::WorkerEnumerationService + Send + Sync>,
+    running_worker_enumeration_service:
+        Arc<dyn worker_enumeration::RunningWorkerEnumerationService + Send + Sync>,
```
justcoon, line 180: probably we should find out better naming - shorter one

golem-worker-executor-base/src/grpc.rs
vigoo: we cannot have `usize` on the gRPC api so it's probably simpler to just use `u64` if possible. If the redis client requires `usize` then we have to have a conversion somewhere. 

`usize` is `u64` on 64bit architectures so it does not really matter.

golem-common/src/model/mod.rs
```diff
@@ -520,6 +521,21 @@ impl WorkerMetadata {
...
+    fn from(value: WorkerMetadata) -> Self {
+        golem_api_grpc::proto::golem::worker::WorkerMetadata {
+            worker_id: Some(value.worker_id.worker_id.into_proto()),
+            account_id: Some(value.account_id.into()),
+            args: value.args,
+            env: HashMap::from_iter(value.env.iter().cloned()),
+            template_version: value.worker_id.template_version,
+            status: Into::<golem::worker::WorkerStatus>::into(value.last_known_status.status)
+                .into(),
+            retry_count: 0, // FIXME
```
vigoo, line 534: Maybe we should not have this here because we will forget that it cannot set the retry_count and misuse it 

Right now the get-metadata gRPC endpoint in worker executor is like this: https://github.com/golemcloud/golem/blob/main/golem-worker-executor-base/src/grpc.rs#L651-L669

But that call to calculate the retry count requires reading from the oplog, so I think it is good in general if worker enumeration returns 0, except if the `precise` flag is set.

I think in longer term the retry count should also be cached in the `last_known_status` but that change should not be part of this PR.

So for now I would remove this from here and have two more custom conversion functions somewhere in `worker-executor-base` that is used both by the worker enumeration, and the get-metadata endpoint - one that uses the last known status and sets retry count to 0 (for now), and one that calls the necessary functions to calculate the latest information ("precise mode").

golem-common/src/model/mod.rs
```diff
@@ -777,12 +793,416 @@ pub fn parse_function_name(name: &str) -> ParsedFunctionName {
...
+        value: String,
+    },
+    Status {
+        value: WorkerStatus,
+    },
+    Version {
+        comparator: FilterComparator,
+        value: i32,
+    },
+    // CreatedAt {
```
vigoo, line 834: Once https://github.com/golemcloud/golem/issues/238 is done we will have creation data (because each oplog entry has a timestamp) so I think we can "prepare" for it already

golem-common/src/model/mod.rs
```diff
@@ -777,12 +793,416 @@ pub fn parse_function_name(name: &str) -> ParsedFunctionName {
...
+                comparator.matches(&metadata.worker_id.template_version, &value)
+            }
+            WorkerFilter::Env {
+                name,
+                comparator,
+                value,
+            } => {
+                let mut result = false;
+                for env_value in metadata.env.clone() {
+                    if env_value.0 == name {
```
vigoo, line 882: minor thing, maybe this could be case-insensitive to make it more user friendly

golem-worker-executor-base/src/services/mod.rs
vigoo: we can keep it for now :)

golem-worker-executor-base/src/services/worker_enumeration.rs
```diff
@@ -0,0 +1,363 @@
...
+            filter.is_some()
+        );
+
+        let active_workers = self.active_workers.enum_workers();
+
+        let mut template_workers: Vec<WorkerMetadata> = vec![];
+        for worker in active_workers {
+            if worker.0.template_id == *template_id
+                && (worker.1.metadata.last_known_status.status == WorkerStatus::Running
+                    || worker.1.metadata.last_known_status.status == WorkerStatus::Idle)
```
vigoo, line 50: I think it should only be `Running`. For idle ones we cannot use this method because it does not return all the ones that are idle - just the ones that are in memory. On the other hand there cannot be anything that is running and is not in memory.

vigoo:
Very nice! Added a few small comments, other than that it's very close to be merged

afsalthaj (approved):

PR #340, opened 2024-03-29
Check cli vs server versions

### LINKED ISSUE #197: CLI self update
We should have a way to:
- notify the users if `golem-cli` needs to be updated
- possibly automate the update process

The check should not just look for the latest version, it should check the version of the Golem servers it is connecting to, and ensure we have the same version of the CLI tooling.

jupposessho:
Checks the server versions against the cli version and show a warning when the server is newer.

Resolves #197

### DIFF
--- a/golem-cli/src/clients/health_check.rs
+++ b/golem-cli/src/clients/health_check.rs
@@ -1,0 +1,40 @@
+// Copyright 2024 Golem Cloud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use async_trait::async_trait;
+use golem_client::model::VersionInfo;
+use tracing::debug;
+
+use crate::model::GolemError;
+
+#[async_trait]
+pub trait HealthCheckClient {
+    async fn version(&self) -> Result<VersionInfo, GolemError>;
+}
+
+#[derive(Clone)]
+pub struct HealthCheckClientLive<C: golem_client::api::HealthCheckClient + Sync + Send> {
+    pub client: C,
+}
+
+#[async_trait]
+impl<C: golem_client::api::HealthCheckClient + Sync + Send> HealthCheckClient
+    for HealthCheckClientLive<C>
+{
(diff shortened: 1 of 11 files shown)


golem-cli/src/main.rs
```diff
@@ -187,6 +189,32 @@ async fn async_main(cmd: GolemCommand) -> Result<(), Box<dyn std::error::Error>>
...
+        worker_client: health_check_client_for_worker,
+    };
+
+    let yellow = "\x1b[33m";
+    let reset_color = "\x1b[0m";
+
+    let version_check = update_srv.check().await;
+
+    if let Err(err) = version_check {
+        println!("{}{}{}", yellow, err.0, reset_color)
```
vigoo, line 215:
```suggestion
        eprintln!("{}{}{}", yellow, err.0, reset_color)
```

should use stderr, because stdout can be used programmatically

golem-cli/src/clients/health_check.rs
```diff
@@ -0,0 +1,40 @@
...
+pub struct HealthCheckClientLive<C: golem_client::api::HealthCheckClient + Sync + Send> {
+    pub client: C,
+}
+
+#[async_trait]
+impl<C: golem_client::api::HealthCheckClient + Sync + Send> HealthCheckClient
+    for HealthCheckClientLive<C>
+{
+    async fn version(&self) -> Result<VersionInfo, GolemError> {
+        info!("Getting version from health check");
```
vigoo, line 36:
```suggestion
        debug!("Getting server version");
```

vigoo (changes requested):
Two small requests otherwise good

vigoo:
Using the stderr will also fix the failing integration tests

vigoo:
And before we merge we should see if the "released" versions 
- built by CI for a tag and published to docker
- output of `cargo make build-release` if the current git head is pointing to a tag

are returning the proper version from the health endpoint and not `0.0.0`. I think this is not the case yet, but probably only requires something in the `Makefile.toml`.

golem-cli/src/main.rs
jupposessho: done, thanks!

golem-cli/src/clients/health_check.rs
jupposessho: done, thanks!

golem-cli/src/version.rs
```diff
@@ -0,0 +1,156 @@
...
+
+const VERSION: &str = env!("CARGO_PKG_VERSION");
+
+#[async_trait]
+pub trait VersionHandler {
+    async fn check(&self) -> Result<GolemResult, GolemError>;
+}
+
+pub struct VersionHandlerLive<
+    T: HealthCheckClient + Send + Sync + Send + Sync,
```
afsalthaj, line 29: Two `Send`s

jupposessho: fixed, thanks!

vigoo (approved):

Issue #298, opened 2024-03-22
Add worker enumeration to worker service


vigoo:
Worker service should only expose a single endpoint (both REST and gRPC) for worker enumeration, requiring:

- a template-id
- filter
- precise flag
- an optional cursor (missing cursor means starting a new query)

More information is in the [Worker Enumeration](https://www.notion.so/Worker-Enumeration-b269f13737f2483ab418fcd159ff558b?pvs=21) document. 

The implementation needs to split into two paths based on whether the filter parameter is **filtering only for running workers** or not.

**Filtering only for running workers**

- First it needs to get a snapshot of all the known worker executors. The necessary information should be already available through `RoutingTableService` but it may needs some work to expose it
- It should randomize the order of calls to them, and can apply some parallelization too
- Then it should call `get-running-workers` **on each worker service** and combine the results

**Returning arbitrary workers**

- Pick a random worker executor. Scanning on the worker executor side is stateless (stateful only in Redis itself) so we can choose a random one for each call and even retry in case of connection failures.
- Call `scan-workers` via gRPC and return the results

PR #343, opened 2024-03-31
Workers enumeration APIs

justcoon:
fixes: #298 

added REST and GRPC APIs

### DIFF
--- a/golem-service-base/src/model.rs
+++ b/golem-service-base/src/model.rs
@@ -2630,6 +2630,12 @@ pub struct InterruptResponse {}
 pub struct ResumeResponse {}
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Object)]
+pub struct WorkerMetadatasResponse {
+    pub workers: Vec<WorkerMetadata>,
+    pub cursor: Option<u64>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Object)]
 #[serde(rename_all = "camelCase")]
 #[oai(rename_all = "camelCase")]
 pub struct WorkerMetadata {
--- a/golem-service-base/src/model.rs
+++ b/golem-service-base/src/model.rs
@@ -19,6 +19,7 @@ use golem_common::model::{parse_function_name, ShardId, TemplateId, WorkerFilter
 use golem_wasm_ast::analysis::{AnalysedResourceId, AnalysedResourceMode};
 use http::Uri;
 use poem_openapi::{Enum, NewType, Object, Union};
+use rand::seq::IteratorRandom;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::{collections::HashMap, fmt::Display, fmt::Formatter};
 
@@ -3100,7 +3101,6 @@ impl RoutingTable {
     }
 
     pub fn random(&self) -> Option<&Pod> {
-        use rand::seq::IteratorRandom;
         self.shard_assignments
             .values()
             .choose(&mut rand::thread_rng())
--- a/golem-service-base/src/model.rs
+++ b/golem-service-base/src/model.rs
@@ -21,6 +21,7 @@ use http::Uri;
 use poem_openapi::{Enum, NewType, Object, Union};
 use rand::seq::IteratorRandom;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::collections::HashSet;
 use std::{collections::HashMap, fmt::Display, fmt::Formatter};
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Object)]
@@ -2631,7 +2632,7 @@ pub struct InterruptResponse {}
 pub struct ResumeResponse {}
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Object)]
-pub struct WorkerMetadatasRequest {
+pub struct WorkersMetadataRequest {
     pub filter: Option<WorkerFilter>,
     pub cursor: Option<u64>,
     pub count: Option<u64>,
@@ -2639,7 +2640,7 @@ pub struct WorkerMetadatasRequest {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Object)]
-pub struct WorkerMetadatasResponse {
+pub struct WorkersMetadataResponse {
     pub workers: Vec<WorkerMetadata>,
     pub cursor: Option<u64>,
 }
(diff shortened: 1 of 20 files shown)


vigoo:
Regarding the running-workers optimization:
- we can't have pagination for that so it should only work if the cursor is not specified (or 0 if it's not optional)
- then if the filter is only allowing status==running then we can use the running worker API and we need to ask _all_ executors and combine the results into one
- this leaves the question of what happens with the `count` parameter - if we limit the result by it, we can't get the rest of the items because we don't have support for the cursor. What I would do is to just ignore both `cursor` and `count` in the case the filter is for running workers, and state this explicitly in the documentation. The number of actually running workers any given time cannot be that high that this creates some unhandleable response size.

golem-service-base/src/model.rs
```diff
@@ -3084,6 +3098,17 @@ impl RoutingTable {
             self.number_of_shards.value,
         ))
     }
+
+    pub fn random(&self) -> Option<&Pod> {
+        use rand::seq::IteratorRandom;
```
afsalthaj, line 3103: Should this be within this function?

golem-worker-service-base/src/service/worker/default.rs
```diff
@@ -127,6 +127,16 @@ pub trait WorkerService<AuthCtx> {
...
 
+    async fn get_metadatas(
+        &self,
+        template_id: &TemplateId,
+        filter: Option<WorkerFilter>,
+        cursor: u64,
+        count: u64,
+        precise: bool,
+        auth_ctx: &AuthCtx,
+    ) -> WorkerResult<(Option<u64>, Vec<WorkerMetadata>)>;
```
afsalthaj, line 138: What is u64 representing here as metadata?. 
I would  newtype it

afsalthaj: Also not sure what would it mean to `None` -> `WorkerMetadata`

golem-worker-service-base/src/service/worker/default.rs
```diff
@@ -1053,6 +1210,18 @@ where
         })
     }
 
+    async fn get_metadatas(
```
afsalthaj, line 1213: Minor. Name is weird. `metadata` is already plural, isn't it :D

golem-worker-service/src/grpcapi/worker.rs
```diff
@@ -116,6 +119,25 @@ impl GrpcWorkerService for WorkerGrpcApi {
...
 
+    async fn get_worker_metadatas(
+        &self,
+        request: Request<GetWorkerMetadatasRequest>,
+    ) -> Result<Response<GetWorkerMetadatasResponse>, Status> {
+        let response = match self.get_worker_metadatas(request.into_inner()).await {
+            Ok((cursor, workers)) => {
+                get_worker_metadatas_response::Result::Success(GetWorkerMetadatasSuccessResponse {
+                    workers,
+                    cursor: cursor.unwrap_or(0), // FIXME can we do option in proto?
```
afsalthaj, line 130: I guess we can do option in proto?

golem-worker-service/src/grpcapi/worker.rs
```diff
@@ -331,6 +353,42 @@ impl WorkerGrpcApi {
...
+    ) -> Result<(Option<u64>, Vec<WorkerMetadata>), GrpcWorkerError> {
+        let template_id: golem_common::model::TemplateId = request
+            .template_id
+            .ok_or_else(|| bad_request_error("Missing template id"))?
+            .try_into()
+            .map_err(|_| bad_request_error("Invalid template id"))?;
+
+        let filter: Option<WorkerFilter> =
+            match request.filter {
+                Some(f) => Some(f.try_into().map_err(|error| {
```
afsalthaj, line 368: Same here. I strongly feel its a overuse of `TryInto`, `TryFrom`

golem-worker-service/src/grpcapi/worker.rs
```diff
@@ -331,6 +353,42 @@ impl WorkerGrpcApi {
...
+                filter,
+                request.cursor,
+                request.count,
+                request.precise,
+                &EmptyAuthCtx {},
+            )
+            .await?;
+
+        let result: Vec<WorkerMetadata> =
+            workers.iter().map(|worker| worker.clone().into()).collect();
```
afsalthaj, line 387: `workers.into_iter()` ?

golem-worker-service/src/grpcapi/worker.rs
justcoon: not sure what do you mean by that

golem-service-base/src/model.rs
justcoon: i can move it to common imports, but is it problem when it is here?

golem-worker-service/src/grpcapi/worker.rs
justcoon: in proto3, everything is optional, 

just i think tokio-rs/prost, for types like u64 (which have default value in rust), do not making them optional 

but custom type/message 

```proto
message Cursor {
   uint64 value = 1;
}
```
can solve that

golem-worker-service-base/src/service/worker/default.rs
jdegoes: `Cursor`

golem-worker-service-base/src/service/worker/default.rs
jdegoes: `get_metadata` / `get_all_metadata`

golem-worker-service/src/grpcapi/worker.rs
vigoo: I think you can write `optional` in proto and prost will handle that. (I was surprised by this once already because I also thought everything is optional in proto3)

golem-worker-service-base/src/service/worker/default.rs
```diff
@@ -735,6 +745,63 @@ where
...
+                        })
+                    })?;
+                    match response.into_inner() {
+                        workerexecutor::GetWorkerMetadatasResponse {
+                            result:
+                            Some(workerexecutor::get_worker_metadatas_response::Result::Success(workerexecutor::GetWorkerMetadatasSuccessResponse {
+                                workers, cursor })),
+                        } => {
+                            let cursor: Option<u64> = if cursor == 0 { None } else { Some(cursor) };
+                            let workers = workers.iter().map(| w| w.clone().try_into()).collect::<Result<Vec<_>, _>>().map_err(|_| GolemError::Unknown(GolemErrorUnknown {
```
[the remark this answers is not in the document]
vigoo: Using the `Into`/`From`/`TryInto`/`TryFrom` instances instead of manual conversion functions makes a lot of things cleaner and automatic so I'm fully supporting it, and already got rid of most of the manual conversion codes in our codebase.

If readability suffers that can easily be fixed by some more explicit type annotations.

(Actually I have similar feelings that you have, but for type inference in general :)

vigoo (approved):

PR #345, opened 2024-04-01
Reuse Template Service

nicoburniske:
- Reuse TemplateCompilationService
- Reuse Template WASM AST processing/transformation

Spent too much time trying to create an abstraction for TemplateRepo, and ultimately decided it's not worth it. We are supporting Postgres and SQLite in OSS and it got too complicated


golem-template-service/src/config.rs
```diff
@@ -114,7 +99,6 @@ impl Default for TemplateServiceConfig {
             db: DbConfig::default(),
             template_store: TemplateStoreConfig::default(),
             compilation: TemplateCompilationConfig::default(),
-            worker_executor_client_cache: WorkerExecutorClientCacheConfig::default(),
```
nicoburniske, line 117: I am not sure why we had this. it's not used anywhere

justcoon (approved):

golem-template-service/src/config.rs
afsalthaj: GRPC client cache for worker comms from template.  Its confusing we are not using this, something to consider

golem-template-service-base/src/service/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod template_compilation;
+pub mod template_processor;
```
afsalthaj, line 2: Interesting that reusable components here is far less than I expected.

golem-template-service/src/config.rs
vigoo: The template service is not calling worker executor at all, right?

afsalthaj: Oh okay, Ok it got not used after we have a separate worker-service that talks to worker-executor.

afsalthaj (approved):

afsalthaj:
Its unfortunate the data layer isn't reusable. It's unrelated to the PR and it was a good decision to solve it in this PR. Inmy view, its because some of suboptimal storage we did in the beginning. Things such as `Project` `account` etc overly tied to the core entities.

PR #349, opened 2024-04-02
Generic Response mapping using Expr

### LINKED ISSUE #316: Replace ResponseMapping with Expr that expects a record consisting between status, headers and body#1343

afsalthaj:
Fixes #316 

* Flexible mapping representation using Expr
* Although the mapping corresponding to a protocol is parsed during evaluation, a validation is done when we submit the API definition using open API spec as well.
* There isn't a performance implication for trying to parse it during evaluation, because Expr is already parsed, and all that's done pattern matching on whether it is Expr::Record, and fetch the status, body and header out of it which is a constant big O.


justcoon (approved):

afsalthaj:
Not waiting for the worker-executor tests, as the change don't impact this at all.

PR #352, opened 2024-04-03
Golem test framework

### LINKED ISSUE #234: Move the existing IT setup code into a shared library in the golem repo

### LINKED ISSUE #235: Add higher level DSL functions for writing integration tests

### LINKED ISSUE #236: Rebase the existing integration tests on top of the new library

### LINKED ISSUE #237: Have integration tests in the golem repo separate from the CLI specific tests

### LINKED ISSUE #327: Move the existing benchmark draft PR to the refactored test framework

vigoo:
Major refactoring involving the worker executor tests, integration and sharding tests, and the (not merged) benchmarking suite.

Introduces a base library for all tests and benchmarks, and rebases the existing test code to use it.

- Resolves #234
- Resolves #235 
- Resolves #236 
- Resolves #237 
- Resolves #327 

The primary motivation is to be able to use the same set of helper functions (test and benchmark DSL) for all layers of testing:
- worker executor tests (only running worker executor itself as a "libarary")
- integration tests (spawning multiple services, talking to them with gRPC)
- golem-cli tests 
- benchmarks (with the ability to set up and talk to remote infrastructure)

This is achieved by restructuring and unifying the existing code, having:

**test dependencies**

These are all the services (golem or third party) that are required for running the tests. Each has a couple of implementation, each implementation responsible for starting and stopping the dependency as well as being able to set up a communication channel to it, and to describe for other services how to access it (private / public host and IPs - private to be used "between dependencies", public to be used from the test runner)

**configurations**

There are three implementations of `TestDependencies`, each setting up them in different ways.

- `WorkerExecutorTestDependencies` only used for worker executor tests, and sets up the "fake", partial test dependency set that is enough to use the same DSL functions for worker executor tests as for the real integration tests
- `EnvBasedTestDependencies` makes decisions based on environment variables and should work the same as the previous test runners worked
- `CliTestDependencies` parses command line arguments and allows to configure all implementations of test dependencies (provided (already running), spawned (as native processes), docker, minikube and k8s-on-aws). This is intended to be used for benchmarks.

**DSL**
The current set of test functions work on any `TestDependencies`, and it is mostly copied from what prevously was the worker executor tests.
There is also a separate set of types and functions targeting benchmarks, with an initial capability of recording key-value pairs where value is either a duration or a count, calculating some statistics, etc.

**Key differences from the previous code:**
- The worker invocation test fuction now takes `wasm_rpc::Value`
- The K8s resource dropping is now reseted to the "drop each resource" implementation - this probably has to be changed back to drop a full namespace but I wanted to see if it works.
- I have dropped all `Result` propagation in the test code because I found that just panicking as early as possible provides a better test-debugging experience, and slightly simplifies the code
- The test dependency implementations are more separated than before - using `dyn` traits
- Avoided passing around the docker client lifetime (and had to because of the dyn traits) by assuming that for the tests, one global docker client is good enough.
- Had to change the minikube port forwarding implementation to work on non-linux too. Tried to keep the linux implementation as well but did not test it.

**Stability:**

- The sharding tests now contain the new "scripted" version as well as the previous random one. The random one is still flaky, probably because of the known sharding-manager issues. If it is annoyingly flaky we can ignore it but I don't want to compeltely delete it
- The benchmarks are now independent of the infrastructure setup in a way that they can be tested with fully local ("spawned" or "docker" modes). This works. However the minikube version fails for me for a weird reason (not being able to create the environment the _second_ time even though everything seems to be deleted).
- Did not test on real AWS infra, but the code is the same as in the previous benchmark PR so if that worked it should still work.


integration-tests/tests/sharding.rs
```diff
@@ -0,0 +1,434 @@
...
+
+#[ctor]
+pub static TRACING: Tracing = Tracing::init();
+
+#[tokio::test]
+#[tracing::instrument]
+async fn service_is_responsive_to_shard_changes() {
+    let (stop_tx, stop_rx) = std::sync::mpsc::channel();
+    let chaos = std::thread::spawn(|| {
+        unstable_environment(stop_rx);
```
senia-psm, line 67: Any reason for it to be random?

golem-cli/tests/main.rs
```diff
@@ -1,35 +1,35 @@
...
     libtest_mimic::run(&args, tests)
 }
 
 fn main() -> Result<(), Failed> {
     env_logger::init();
 
-    let docker = clients::Cli::default();
-    let context = Context::start(&docker, EnvConfig::from_env())?;
+    let deps: Arc<dyn TestDependencies + Send + Sync + 'static> =
+        Arc::new(EnvBasedTestDependencies::new(3));
```
afsalthaj, line 25: What is this 3?

vigoo:
Note that I tried to add the "CI test report" changes as well but reverted it because it was compiling things twice because it relies on unstable rust features.

golem-test-framework/src/components/k8s.rs
```diff
@@ -0,0 +1,441 @@
...
+    async fn async_drop(&mut self) {
+        info!("Stopping pod {}", self.name);
+
+        let name = self.name.clone();
+
+        let pods: Api<Pod> =
+            Api::namespaced(Client::try_default().await.unwrap(), &self.namespace.0);
+
+        let dp = DeleteParams {
+            grace_period_seconds: Some(0),
```
senia-psm, line 72: :+1: 

This might be the reason of my issues with `Drop` (why I commented out this code).

golem-test-framework/src/components/k8s.rs
```diff
@@ -0,0 +1,441 @@
...
+
+            let mut stderr_string = String::new();
+            stderr
+                .read_to_string(&mut stderr_string)
+                .await
+                .expect("Failed to read stderr");
+
+            debug!("minikube service stderr: {}", stderr_string);
+
+            if stderr_string.contains("SVC_UNREACHABLE") && attempts < 5 {
```
senia-psm, line 391: The loop is infinite if `attempts >= 5`.

golem-test-framework/src/components/k8s.rs
```diff
@@ -0,0 +1,441 @@
...
+                attempts += 1;
+                tokio::time::sleep(Duration::from_secs(1)).await;
+                continue;
```
senia-psm, line 394: This `continue` it the last line in `loop`. It does nothing.

senia-psm (changes requested):

golem-test-framework/src/components/rdb/k8s_postgres.rs
```diff
@@ -0,0 +1,164 @@
...
+                    "port": 5432,
+                    "protocol": "TCP"
+                }],
+                "selector": { "app": "golem-postgres" },
+                "type": "LoadBalancer"
+            }
+        }))
+        .expect("Failed to deserialize service description");
+
+        let _res_srv = services
```
senia-psm, line 110: There is no need for explicit name here. There is no `Drop`. Explicit name might be misleading - it keeps variable alive till the end of the method (i.e. prevents immediate `Drop` call).

jdegoes (approved):

golem-test-framework/src/components/k8s.rs
vigoo: Nice catch! Don't know how that happened :)

golem-test-framework/src/config/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn rdb(&self) -> Arc<dyn Rdb + Send + Sync + 'static>;
+    fn redis(&self) -> Arc<dyn Redis + Send + Sync + 'static>;
+    fn redis_monitor(&self) -> Arc<dyn RedisMonitor + Send + Sync + 'static>;
+    fn shard_manager(&self) -> Arc<dyn ShardManager + Send + Sync + 'static>;
+    fn template_directory(&self) -> PathBuf;
+    fn template_service(&self) -> Arc<dyn TemplateService + Send + Sync + 'static>;
+    fn worker_service(&self) -> Arc<dyn WorkerService + Send + Sync + 'static>;
+    fn worker_executor_cluster(&self) -> Arc<dyn WorkerExecutorCluster + Send + Sync + 'static>;
+
+    fn kill_all(&self) {
```
senia-psm, line 42: Methods here return `Arc` for real services. And I believe this is the reason why we have to use custom `kill/kill_all` methods instead of `Drop`. This is the reason why I created `*Info` structs. Most of the tests need only `*Info` which implements `Clon` and has no `Drop`. The only tests that needs real services is `sharding`, but in `sharding` we don't need `Arc` either because there is a single thread that can own real services with `Drop`.

The approach with `kill`/`kill_all` as a bit risky because someone can add a field with `Drop` in some of the structs and incorrectly assume it should work.

golem-test-framework/src/components/redis/k8s.rs
```diff
@@ -0,0 +1,182 @@
...
+                    "port": 6379,
+                    "protocol": "TCP"
+                }],
+                "selector": { "app": "golem-redis" },
+                "type": "LoadBalancer"
+            }
+        }))
+        .expect("Failed to deserialize service definition");
+
+        let _res_srv = services
```
senia-psm, line 112: Unused name

golem-test-framework/src/components/shard_manager/k8s.rs
```diff
@@ -0,0 +1,207 @@
...
+                    "image": format!("golemservices/golem-shard-manager:latest"),
+                    "env": env_vars
+                }]
+            }
+        }))
+        .expect("Failed to deserialize Pod definition");
+
+        let pp = PostParams::default();
+
+        let _res_pod = pods.create(&pp, &pod).await.expect("Failed to create pod");
```
senia-psm, line 105: Unused name

golem-test-framework/src/components/shard_manager/k8s.rs
```diff
@@ -0,0 +1,207 @@
...
+                        "protocol": "TCP"
+                    }
+                ],
+                "selector": { "app": Self::NAME },
+                "type": "NodePort"
+            }
+        }))
+        .expect("Failed to deserialize service definition");
+
+        let _res_srv = services
```
senia-psm, line 137: Unused name

golem-test-framework/src/components/template_service/k8s.rs
```diff
@@ -0,0 +1,201 @@
...
+                        "protocol": "TCP"
+                    }
+                ],
+                "selector": { "app": Self::NAME },
+                "type": "LoadBalancer"
+            }
+        }))
+        .expect("Failed to deserialize service definition");
+
+        let _res_srv = services
```
senia-psm, line 136: Not important: Unused name.

jdegoes (approved):

senia-psm (approved):

afsalthaj:
Approved

golem-test-framework/src/components/k8s.rs
senia-psm: Copied from the next `loop`. The next `loop` has `break` after `continue`

golem-test-framework/src/config/mod.rs
vigoo: I was thinking of keeping your approach with the `*Info` descriptors but in the end felt like this one is safer/easier to understand for me. The reason is that those are essentially weak references to the running services (not real references but host/port information) completely decoupled from the lifetime of the started services, while with my approach they are real shared reference counted references to the thing that controls the running services. This way it is guaranteed that if you have a reference to one of these dependencies, it still exists and that anyone who has a reference for it can "control it" (for example restart). On the other hand it is true that with an explicit `kill` the guarantee that "it still runs" is not really true anymore. 

In general this type of shared services of type `Arc<dyn Svc + Send + Sync + 'static>` worked well in other places of our code base (even if it is not the most idiomatic Rust way to do things) and I think it is easy to use.

PR #358, opened 2024-04-04
Expose api-definition endpoints to golem-cli

### LINKED ISSUE #242: Support for worker-bridge in Golem-CLI
Based on discussion with @jdegoes

jupposessho:
Resolves #242

### DIFF
--- a/golem-cli/src/api_definition.rs
+++ b/golem-cli/src/api_definition.rs
@@ -1,0 +1,215 @@
+// Copyright 2024 Golem Cloud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


afsalthaj (approved):

justcoon (approved):

golem-cli/src/api_definition.rs
```diff
@@ -0,0 +1,290 @@
...
+        where
+            'life0: 'async_trait,
+            Self: 'async_trait,
+        {
+            todo!()
+        }
+
+        #[must_use]
+        #[allow(clippy::type_complexity, clippy::type_repetition_in_bounds)]
+        fn get<'life0, 'life1, 'life2, 'async_trait>(
```
vigoo, line 139: You don't need these horrible function signatures - this is the desugared version of how async functions in a trait look like. Please use the `async_trait` annotation (as you did in other places) and just write async functions instead - this is very hard to read

vigoo (changes requested):

golem-cli/src/api_definition.rs
jupposessho: indeed, thanks! I was wondering what is missing to have the simplified signature 🤦

vigoo (approved):

Issue #297, opened 2024-03-22
Expose worker enumeration via host functions


vigoo:
The`scan-workers` function implemented in https://github.com/golemcloud/golem/issues/295 need to be exposed through the Golem Runtime API.

It should be a WIT _resource_ getting the template ID and the filter parameters as its constructor, and a method that returns the next chunk of metadata.

Later with WASI Preview 3 it can be a stream.

This involves:
- modifying the host WIT in `golem-wit` and publishing it
- making the new `EnumerationService` available from host functions by passing it to the `DurableCtx`
- connecting the dots in the host function implementation - it requires creating a new type that can be registered into wasmtime’s resource `Table`.

vigoo:
Detailed instructions:

First this needs to be added to the `golem-wit` repo, to https://github.com/golemcloud/golem-wit/blob/main/wit/deps/golem/golem-host.wit

I think we should create a WIT `resource`, each instance representing one search. This way we can hide the passing of the cursor from the user, in a language-independent way. So the constructor of the resource would take the template id and the filters, and it would have one method that returns an option<vec<metadata>> or something like that.

This of course requires defining the WIT data types for defining the filters too.

Once it is done, `golem` needs to use the latest `golem-wit`. It is now connected via a cargo dependency, so during the development you can use path dependency but then we should publish a new `golem-wit` version before merging the `golem` PR.

`golem-worker-executor-base` may need to be `clean`d before building to catch the new WITs. (Could not properly set up invalidation in `build.rs` yet)

Once it has the new WITs, you should get all kinds of errors because you are not implementing the new resource. 

It should be placed in https://github.com/golemcloud/golem/blob/main/golem-worker-executor-base/src/durable_host/golem/mod.rs

But not in the 
```rust
impl<Ctx: WorkerCtx> golem::api::host::Host for DurableWorkerCtx<Ctx> 
```
because if it is a resource, the binding generator generates a separate host trait for it.

So it would be something like

```rust
impl<Ctx: WorkerCtx> golem::api::host::HostEnumerateWorkers for DurableWorkerCtx<Ctx> 
```

In this you have to implement the constuctor and methods and a drop function.

You can see other examples of how resources are implemented, for example https://github.com/golemcloud/golem/blob/main/golem-worker-executor-base/src/durable_host/blobstore/types.rs#L35

The idea is that there is a `Table` (defined in wasmtime) for each worker, where you can register instances of resources.
[...]

PR #359, opened 2024-04-05
Expose worker enumeration via host functions

justcoon:
fixes: #297 

related golem-wit PR: https://github.com/golemcloud/golem-wit/pull/17

### DIFF
--- a/golem-worker-executor-base/Cargo.toml
+++ b/golem-worker-executor-base/Cargo.toml
@@ -44,7 +44,7 @@ futures = { workspace = true }
 futures-util = { workspace = true }
 gethostname = "0.4.3"
 #golem-wit = { version = "0.2.3" }
-golem-wit = { git = "https://github.com/golemcloud/golem-wit.git", rev = "6d537e1d" }
+golem-wit = { git = "https://github.com/golemcloud/golem-wit.git", rev = "04f6d125" }
 http = { workspace = true }
 http_02 = { workspace = true }
 http-body = "1.0.0" # keep in sync with wasmtime
--- a/golem-worker-executor-base/Cargo.toml
+++ b/golem-worker-executor-base/Cargo.toml
@@ -44,7 +44,7 @@ futures = { workspace = true }
 futures-util = { workspace = true }
 gethostname = "0.4.3"
 #golem-wit = { version = "0.2.3" }
-golem-wit = { git = "https://github.com/golemcloud/golem-wit.git", rev = "04f6d125" }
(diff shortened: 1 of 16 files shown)


golem-worker-executor-base/Cargo.toml
```diff
@@ -43,7 +43,8 @@ fs-set-times = "0.20.1"
 futures = { workspace = true }
 futures-util = { workspace = true }
 gethostname = "0.4.3"
-golem-wit = { version = "0.2.3" }
+#golem-wit = { version = "0.2.3" }
+golem-wit = { git = "https://github.com/golemcloud/golem-wit.git", rev = "04f6d125" }
```
justcoon, line 47: needs to be updated after golem-wit will be released

golem-worker-executor-base/Cargo.toml
```diff
@@ -43,7 +43,8 @@ fs-set-times = "0.20.1"
 futures = { workspace = true }
 futures-util = { workspace = true }
 gethostname = "0.4.3"
-golem-wit = { version = "0.2.3" }
+#golem-wit = { version = "0.2.3" }
+golem-wit = { git = "https://github.com/golemcloud/golem-wit.git", rev = "5079afa1" }
```
vigoo, line 47: Just adding a review comment here to make sure it's not fogotten :)

golem-worker-executor-base/src/durable_host/golem/mod.rs
```diff
@@ -27,12 +30,96 @@ use crate::get_oplog_entry;
...
+        Ok(())
+    }
+}
+
+pub struct GetWorkersEntry {
+    template_id: golem_common::model::TemplateId,
+    filter: Option<golem_common::model::WorkerFilter>,
+    precise: bool,
+    count: u64,
+    next_cursor: Arc<RwLock<Option<u64>>>,
```
vigoo, line 104: You don't have to protect it because the host functions never run in parallel. In fact the whole context (wasmtime's, which hold `DurableCtx` as its user-defined part) is protected by a big mutex.
So just use `Option<u64>` here, and use `get_mut` when getting it from the table instead of `get`.

vigoo (changes requested):
Very nice, there is one simplification I ask for, and to not forget to change the `golem-wit` dependency to a released version before merge.

golem-worker-executor-base/Cargo.toml
justcoon: updated

vigoo (approved):

Issue #300, opened 2024-03-22
Add worker enumeration to the CLI


vigoo:
A new subcommand for template command should be added with some syntax / parameters to express all the filtering possibilities.

PR #363, opened 2024-04-08
CLI - worker list

justcoon:
fixes: #300


golem-cli/src/clients/worker.rs
```diff
@@ -418,3 +462,91 @@ struct Log {
...
+                comparator: filter_comparator_from_str(comparator)?,
+                value: worker_status_from_str(value)?,
+            })),
+            "created_at" | "createdAt" => Ok(WorkerFilter::CreatedAt(WorkerCreatedAtFilter {
+                comparator: filter_comparator_from_str(comparator)?,
+                value: DateTime::parse_from_str(value, "%Y-%m-%dT%H:%M:%SZ")
+                    .map(|ts| ts.into())
+                    .map_err(|e| format!("Invalid filter value: {}", e))?,
+            })),
+            _ if arg.starts_with("env.") => {
```
afsalthaj, line 529: Looks a bit off. Could you explain what's going on here

golem-cli/src/clients/worker.rs
```diff
@@ -80,6 +83,14 @@ pub trait WorkerClient {
         name: WorkerName,
         template_id: RawTemplateId,
     ) -> Result<WorkerMetadata, GolemError>;
+    async fn find_metadata(
+        &self,
+        template_id: RawTemplateId,
+        filter: Option<Vec<String>>,
```
afsalthaj, line 89: I would rather call this as Option<workrfilter>. I don't think the call site need to be overly simplified to just pass vec<string>

afsalthaj (changes requested):

golem-cli/src/clients/worker.rs
justcoon: it is using this API https://github.com/golemcloud/golem/blob/main/openapi/golem-service.yaml#L63

and it is list of filters, in form `{property} {operator} {value}`, for example `name = worker-name`, we need to do some documentation ....

golem-cli/src/clients/worker.rs
```diff
@@ -227,6 +243,59 @@ impl<C: golem_client::api::WorkerClient + Sync + Send> WorkerClient for WorkerCl
...
+        template_id: RawTemplateId,
+        filter: Option<WorkerFilter>,
+        cursor: Option<u64>,
+        count: Option<u64>,
+        precise: Option<bool>,
+    ) -> Result<WorkersMetadataResponse, GolemError> {
+        info!(
+            "Getting workers metadata for template: {}, filter: {}",
+            template_id.0,
+            filter.is_some()
```
nicoburniske, line 257: Just curious, does filter not have a debug impl?

nicoburniske (approved):
LGTM

golem-cli/src/clients/worker.rs
justcoon: yes, it have

golem-cli/src/worker.rs
```diff
@@ -168,6 +168,29 @@ pub enum WorkerSubcommand {
         #[arg(short, long)]
         worker_name: WorkerName,
     },
+    /// Retrieves metadata about an existing workers in a template
+    #[command()]
+    List {
+        /// The Golem template the workers to be retrieved belongs to
```
senia-psm, line 174: Doc comment does not work on `#[command(flatten)]`.

golem-cli/src/worker.rs
```diff
@@ -168,6 +168,29 @@ pub enum WorkerSubcommand {
...
+        #[command(flatten)]
+        template_id_or_name: TemplateIdOrName,
+
+        // Filter
```
senia-psm, line 178: It would be nice to have some explanation or example here.

golem-cli/src/worker.rs
```diff
@@ -168,6 +168,29 @@ pub enum WorkerSubcommand {
...
+        #[arg(short, long)]
+        filter: Option<Vec<String>>,
+
+        // Cursor
```
senia-psm, line 182: Please add some details here (i.e. "skip this amount of records in result" or something)

golem-cli/src/worker.rs
```diff
@@ -168,6 +168,29 @@ pub enum WorkerSubcommand {
...
+        #[arg(short, long)]
+        cursor: Option<u64>,
+
+        // Count
```
senia-psm, line 186: Please add some details here.

golem-cli/src/worker.rs
```diff
@@ -168,6 +168,29 @@ pub enum WorkerSubcommand {
...
+        #[arg(short = 's', long)]
+        count: Option<u64>,
+
+        // Precise
```
senia-psm, line 190: Please add some details here.

PR #368, opened 2024-04-10
Support all constructors

afsalthaj:
### Features and fixes

* Support for pattern matching on variants

```scala
// given worker.response TypeAnnotatedValue::Variant { Foo ..}

match worker.response {
   Foo(value) => value.id
   Bar(value) => value.id
}

```

* Fixed nested pattern matching - recursively identify the patterns
```scala
// given worker.response TypeAnnotatedValue::Variant { Foo ..{ Option { Result }

match worker.response {
   Foo(some(ok(value)) => value.id
   Bar(none) => value.id
}
```

* Support on wild card

```scala
match worker.response {
   some(_) => worker.response.id
   none => 'not found'
}
```

* Support on alias

```scala
match worker.response { 
  a @ ok(b @ _) => b 
  c @ err(d @ ok(e)) => {p : c, q: d, r: e.id} 
}

```
* Introduction of TypeMismatch

```scala
// given worker.response is TypeAnnotatedValue::U64(1)

// Following will result in type-mismatch error for the first arm itself - where we tried to match 1 with `ok`
match worker.response {
   ok(_) => "success"
   err(_) => "failure"
}
```

* Stricter pattern matching. Previously any value from worker.response could be considered as `Option::Some`. Now that we moved to `TypeAnnotatedValue` which clearly has `Some` and `None` types for optional values. So we can pattern match strictly, otherwise we get a `TypeMismatch` error.

* Disallowed random constructions. Removed `Expr::Constructor0` to being explicit. That while we can pattern match on any variant, users cannot construct random constructors. 

```scala
// Disallowed
match worker.response {
   some(_) => Foo(1)
   none => 'not found'
}
```

`Expr` cannot evaluate `Foo(1)` because it simply doesn't know what `Foo` is.

* Expr::Result and Expr::Option (and removed Expr::Constructor0)

* Tests in `text` module (for read and write) and `evaluator` module. The ones PR reviewers might be interested in is evaluator, to see the outcome of this PR

* Cleanups that are as follows:
   * Removed higher order functions to evaluate if-else expressions
   * Separate module for if-else, to be consistent with other patterns
   * Removed the complex structure InternalExprResult (which was holding a function insider, on what to do in the next loop). Works well only in Scala!
   * Renaming. A pattern match should consist of `match_expression` and a collection of `arms` and LHS of `arm` is `arm_pattern` and rhs of arm is `arm_body`, following consistent naming around pattern matching expressions in the wild.
   
   
   
There can be many things where we could optimize, I do have an eye on these things, but I think time to take a step back and progress with what we have now, and optimize it with specific tickets in place.


golem-worker-service-base/src/parser/expr/expr_parser.rs
```diff
@@ -1,495 +1,299 @@
...
-                                    let first_result: Rc<Expr> = Rc::clone(&first_result);
-                                    InternalExprResult::incomplete(
-                                        InCompleteExpressionContext::Else,
-                                        move |else_result| {
-                                            let first_result: Expr =
-                                                (*Rc::clone(&first_result)).clone();
-                                            InternalExprResult::complete(Expr::Cond(
-                                                Box::new(first_result),
-                                                Box::new(second_result.clone()),
-                                                Box::new(else_result),
```
afsalthaj, line 294: Forming higher order functions was logiccally ok, but it just giving me lot of pain with Rust

golem-worker-service-base/src/api/register_api_definition_api.rs
```diff
@@ -282,10 +282,7 @@ impl TryFrom<crate::worker_binding::GolemWorkerBinding> for grpc_apidefinition::
 
     fn try_from(value: crate::worker_binding::GolemWorkerBinding) -> Result<Self, Self::Error> {
         let response: Option<String> = match value.response {
-            Some(v) => {
-                let r = v.0.to_string().map_err(|e| e.to_string())?;
-                Some(r)
-            }
+            Some(v) => Some(v.0.to_string()),
```
justcoon, line 285: maybe you could use  `Option.map`

afsalthaj: That was me doing a fianl compilation error that happened "elsewhere"

jdegoes (approved):
Looks like a big improvement! I would create a Github issue that comprehensively documents the limitations of the current expression language--including, for example, that variant constructors may not be used. We will want to address these expressive issues, as well as performance issues (that may come from an interpreter) in Post 1.0.

afsalthaj:
@justcoon the display shouldn't fail as such for logical reasons. The std.io fails are mainly panic. I should be using only display instance everywhere. Merging this now.

PR #369, opened 2024-04-11
Persistent invocation queue

### LINKED ISSUE #178: Persistent invocation queue
Currently we don’t have a persistent invocation queue / mailbox for workers - the incoming invocations are implicitly enqueued in the worker executor’s memory but in case it gets restarted, they get lost. 

This is not a critical problem if the clients are using the “invoke-and-await” with a timeout, because they can retry the same invocation with the same invocation key to eventually get a result.

However, when using the “invoke” API, the caller has no (easy) way to do a retry in such scenarios. The expectation is that once we trigger an invocation, Golem will perform that no matter what.

vigoo:
Requires https://github.com/golemcloud/wasm-rpc/pull/31

Resolves #178


golem-worker-executor-base/src/durable_host/mod.rs
```diff
@@ -299,24 +301,20 @@ impl<Ctx: WorkerCtx> DurableWorkerCtx<Ctx> {
 
     pub async fn store_worker_status(&self, status: WorkerStatus) {
         let oplog_idx = self.state.oplog_idx;
+        let status_record = WorkerStatusRecord {
```
jdegoes, line 304: The time spent building this (worker status record as a fold over oplog) is proving quite worthwhile, and the architecture for doing new things like durable invocation queue is becoming clearer.

vigoo: Yes I'm very happy about it :)

golem-worker-executor-base/src/durable_host/durability.rs
```diff
@@ -268,13 +268,13 @@ impl<Ctx: WorkerCtx> DurableWorkerCtx<Ctx> {
                     serializable_result
                 )
             });
-            self.state.set_oplog_entry(oplog_entry).await;
+            self.state.oplog.add(oplog_entry).await;
```
afsalthaj, line 271: Is this a logical change required for the persistent queue, or just refactor?

golem-worker-executor-base/src/durable_host/golem/mod.rs
```diff
@@ -132,7 +132,10 @@ impl<Ctx: WorkerCtx> golem::api::host::Host for DurableWorkerCtx<Ctx> {
         Ok(self
             .public_state
             .promise_service
-            .create(&self.worker_id.worker_id, self.state.oplog_idx)
+            .create(
+                &self.worker_id.worker_id,
+                self.state.current_oplog_index().await,
```
afsalthaj, line 137: Is this a logical change or refactor? :)  - trying to understand what's being done

afsalthaj (approved):

golem-worker-executor-base/src/durable_host/durability.rs
vigoo: Refactoring how the oplog is accessed was necessary, because now we need to write to it outside of the worker's context.

Issue #245, opened 2024-03-18
Use wasm-wave to represent types and values on CLI


vigoo:
Original idea was: Use the new `text` module of wasm-rpc for printing values and types in CLI

It is easier to work directly with wasm-wave, than with wasm-rpc, because of incompatible data types, unsupported function type and handle type.

PR #371, opened 2024-04-12
245 wasm wave representation

senia-psm:
Fix: #245 

`--human-readable` flag in `invoke-and-await` is temporary. I'll replace it with a new global format later.

### DIFF
--- a/golem-cli/src/model/wave.rs
+++ b/golem-cli/src/model/wave.rs
@@ -1,10 +1,12 @@
-use golem_client::model::{ExportFunction, ResourceMode, Type, TypeHandle, TypeRecord, TypeResult, TypeTuple, TypeVariant};
+use golem_client::model::{
+    ExportFunction, ResourceMode, Type, TypeHandle, TypeRecord, TypeResult, TypeTuple, TypeVariant,
+};
+use golem_wasm_ast::analysis::{AnalysedResourceId, AnalysedResourceMode, AnalysedType};
 use std::borrow::Cow;
 use std::fmt::Display;
-use golem_wasm_ast::analysis::{AnalysedResourceId, AnalysedResourceMode, AnalysedType};
 use wasm_wave::wasm::{DisplayType, WasmFunc};
 
-fn type_wave_compatible(typ: &Type) -> bool {
+pub fn type_wave_compatible(typ: &Type) -> bool {
     fn variant_wave_compatible(tv: &TypeVariant) -> bool {
         tv.cases
             .iter()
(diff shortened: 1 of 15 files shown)


golem-cli/src/worker.rs
```diff
@@ -79,12 +90,29 @@ pub enum WorkerSubcommand {
         function: String,
 
         /// JSON array representing the parameters to be passed to the function
-        #[arg(short = 'j', long, value_name = "json", value_parser = ValueParser::new(JsonValueParser))]
-        parameters: serde_json::value::Value,
+        #[arg(short = 'j', long, value_name = "json", value_parser = ValueParser::new(JsonValueParser), conflicts_with = "wave")]
+        parameters: Option<serde_json::value::Value>,
+
+        /// Function parameter in wasi-wave format
```
vigoo, line 96:
```suggestion
        /// Function parameter in WAVE format
```

Just a minor thing :) the library we use is called `wasm-wave` (not related to wasi) and actually WAVE means "Web Assembly Value Encoding" so i think that's what we should refer to as a format

golem-cli/src/model/wave.rs
```diff
@@ -0,0 +1,282 @@
...
+
+fn wrap_analysed_type(typ: AnalysedType) -> golem_wasm_rpc::AnalysedType {
+    golem_wasm_rpc::AnalysedType(typ)
+}
+
+pub fn wrap_type(typ: &Type) -> golem_wasm_rpc::AnalysedType {
+    wrap_analysed_type(type_to_analysed(typ))
+}
+
+impl WasmFunc for WrapExportFunction {
```
vigoo, line 73: Ah I did not know there is a function type class in wasm-wave too! That's nice. 
And we should clean this and wasm-rpc up by implementing these typeclasses directly in `golem-wasm-ast` (under a `wave` feature flag). But that can be a separate task of course.

senia-psm: I'll move this into `golem-wasm-rpc` after merging this one.
Please note that `DisplayWasmFunc` in `wasm-wave` is not usable for us - it renders `func` as function name (hardcoded).

vigoo: I think the correct place for the instances is https://github.com/golemcloud/golem-wasm-ast/blob/main/src/analysis/mod.rs where `AnalysedType` and `AnalysedFunction` is defined.

This means removing the wrapper and instance from `wasm-rpc` too.

vigoo (approved):

afsalthaj (approved):

PR #417, opened 2024-04-17
Update related oplog entries and removing VersionedWorkerId

### LINKED ISSUE #387: Introduce component version override, in-progress and failed updates to the oplog
- Define two or three new oplog entries and the associated cached status fields. 
- Use the latest overridden component version during worker instantiation

vigoo:
Resolves #387 

Also removes the concept of a `VersionedWorkerId` as it is confusing (there cannot be different versions of the same worker identified by a `WorkerId`) and used inconsistently.


golem-api-grpc/proto/golem/worker/worker_service.proto
```diff
@@ -47,7 +46,7 @@ message LaunchNewWorkerRequest {
 
 message LaunchNewWorkerResponse {
   oneof result {
-    golem.worker.VersionedWorkerId success = 1;
+    golem.worker.WorkerId success = 1;
```
justcoon, line 49: actually i think it is quiet useful, if this returning version of created worker, maybe instead there should be 

```proto
message LaunchNewWorkerSuccess {
    golem.worker.WorkerId worker_id = 1;
    uint64 template_version = 2;
    ...
}

message LaunchNewWorkerResponse {
  oneof result {
    LaunchNewWorkerSuccess success = 1;
      ...
   }
}
```
also in such way it is more prepared for future extensions of response

golem-worker-service/src/api/worker.rs
```diff
@@ -47,7 +30,7 @@ impl WorkerApi {
         &self,
         template_id: Path<TemplateId>,
         request: Json<WorkerCreationRequest>,
-    ) -> Result<Json<VersionedWorkerId>> {
+    ) -> Result<Json<WorkerId>> {
```
justcoon, line 33: similar like i mentioned in case of gRpc api 
instead of just `WorkerId`, have some payload 
```rust
struct WorkerCreationResponse {
    worker_id: WorkerId,
    template_version: u64
}

```

golem-api-grpc/proto/golem/worker/worker_service.proto
vigoo: Good idea, thanks

justcoon (approved):

PR #424, opened 2024-04-18
Add benchmarks with cold start

### LINKED ISSUE #379: Cold start benchmark
Write a new benchmark that measures cold start of workers.

- start a cluster of `size` 
- no warmup
- invoke `length` number of distinct workers once. each invocation will create a new worker.
- measure the time of the invocation

Once a single version of a benchmark like that exists, create more variants:
- have two or three versions of this benchmark, only difference is using a different component of different size
 - the easiest way is to write simple workers, but use different programming languages: C or stripped Rust can be the small, JS the middle, Python the big. 
 - You can create new test components for this

This task does not involve any benchmark framework / infrastructure work, those are managed as separate issues.

jupposessho:
Resolves #379


integration-tests/src/benchmarks/cold_start_large.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    async fn create(config: CliParams) -> Self {
+        Self { config }
+    }
+
+    async fn setup_iteration(&self) -> Self::IterationContext {
+        setup(self.config.clone(), "python-1").await
+    }
+
+    async fn warmup(&self, _: &Self::IterationContext) {}
```
senia-psm, line 41: Should we call some different workers (not the same as in `run`) to warmup worker executors?

integration-tests/src/lib.rs
```diff
@@ -11,3 +11,79 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
```
vigoo, line 14: Please move these helpers into the `benchmarks` module (create a `benchmarks/mod.rs` and just define it here as `pub mod benchmarks`.

Because this is the root module for integration-tests and benchmarks together.

integration-tests/src/benchmarks/cold_start_medium.rs
```diff
@@ -0,0 +1,55 @@
...
+    fn name() -> &'static str {
+        "cold-start-medium"
+    }
+
+    async fn create(config: CliParams) -> Self {
+        Self { config }
+    }
+
+    async fn setup_iteration(&self) -> Self::IterationContext {
+        setup(self.config.clone(), "js-1").await
```
vigoo, line 38: The `js-1` test component does not export the `golem:it/api/echo` that your `run_echo` function is calling. Let's create new dedicated test components for this (`rust-echo`, `js-echo`, `py-echo`) with the same WIT interface. *(It can be even simpler than the `option-service` component - no need for `option` in the API just a string -> string function)

vigoo (changes requested):
This will not work with the components referenced, need to write new ones.

integration-tests/src/benchmarks/cold_start_medium.rs
jupposessho: Added the echo component in 3 different languages. The bigger wasm files (> 4Mb) still cannot be loaded(with the changes from main branch).

vigoo (approved):

Issue #411, opened 2024-04-16
CLI: Add Api Definition Creation Command 


nicoburniske:
Follow up of https://github.com/golemcloud/golem/pull/405 

Add support to cli for creating an existing version of an API using new endpoints

`/v1/api/definitions`
- POST is for creating API
- PUT is for updating existing API. 

Before we had PUT to create and update, but the update didn't work, so I separated the functionality up into two endpoints

PR #425, opened 2024-04-18
411 api definition add/update/import and tests

senia-psm:
Fix: #411

### DIFF
--- a/golem-cli/src/clients/api_definition.rs
+++ b/golem-cli/src/clients/api_definition.rs
@@ -12,10 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::fmt::Display;
+
 use std::io::Read;
 
 use async_trait::async_trait;
+
 use golem_client::model::HttpApiDefinition;
+
 use tokio::fs::read_to_string;
 use tracing::info;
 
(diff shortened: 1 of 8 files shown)


golem-cli/src/clients/api_definition.rs
```diff
@@ -42,6 +48,70 @@ pub struct ApiDefinitionClientLive<C: golem_client::api::ApiDefinitionClient + S
...
+    Update,
+    Import,
+}
+
+impl Display for Action {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let str = match self {
+            Action::Create => "Creating",
+            Action::Update => "Updating",
+            Action::Import => "Creating or updating",
```
afsalthaj, line 63: That's a bit, I don't know. May be its ok, as there is no other way to detect

golem-cli/tests/api_definition.rs
```diff
@@ -0,0 +1,368 @@
...
+                    "type": "number"
+                  }
+                }
+              }
+            }
+        }
+        }
+    );
+
+    make_file(id, &open_api_json)
```
afsalthaj, line 196: 👍

afsalthaj (approved):

golem-cli/src/clients/api_definition.rs
senia-psm: It should be "Importing". I'll fix it. And it is just for logs.

Issue #414, opened 2024-04-17
Make sure component upload and download is streaming on all levels


vigoo:
Components can be big (hundreds of megabytes) and we need to handle them in a streaming way on all levels:
- S3
- template (component) service
- worker executor
- on both HTTP and gRPC APIs.

An example where it is incorrectly implemented:
https://github.com/golemcloud/golem/blob/main/golem-template-service/src/grpcapi/template.rs#L240-L250

Most likely it is already properly implemented in many places but make sure to check all and have end-to-end streaming in both upload and download directions.

PR #433, opened 2024-04-18
Download template as stream

justcoon:
related to: #414

### DIFF
--- a/golem-service-base/src/service/template_object_store.rs
+++ b/golem-service-base/src/service/template_object_store.rs
@@ -15,16 +15,30 @@
 use std::error::Error;
 use std::fs;
 use std::path::PathBuf;
+use std::pin::Pin;
+use std::task::{Context, Poll};
 
 use crate::config::{TemplateStoreLocalConfig, TemplateStoreS3Config};
 use async_trait::async_trait;
 use aws_config::BehaviorVersion;
 use aws_sdk_s3::primitives::ByteStream;
+use hyper::body::Bytes;
 use tracing::{debug, info};
+
+pub struct GetTemplateStream(ByteStream);
+
+impl futures::stream::Stream for GetTemplateStream {
+    type Item = Result<Bytes, Box<dyn Error>>;
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        Pin::new(&mut self.0).poll_next(cx).map_err(|e| e.into())
+    }
+}
 
 #[async_trait]
 pub trait TemplateObjectStore {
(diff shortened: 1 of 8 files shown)


nicoburniske (approved):
Looks much better than before

golem-service-base/src/service/template_object_store.rs
```diff
@@ -78,18 +100,34 @@ impl TemplateObjectStore for AwsS3TemplateObjectStore {
...
+
+        match self
+            .client
+            .get_object()
+            .bucket(&self.bucket_name)
+            .key(key)
+            .send()
+            .await
+        {
+            Ok(response) => response.body.into(),
```
afsalthaj, line 116: How is the whole impl different to using `into_async_read` https://dtantsur.github.io/rust-openstack/tokio/io/trait.AsyncRead.html

vigoo (approved):

golem-service-base/src/service/template_object_store.rs
justcoon: in case of gRpc Api returning stream, 
then we had to transform AsyncRead back to stream

afsalthaj: AsyncRead can be Stream. But just looking through the APIs. take it with a grain salt as I have never done this before and may be reading these APIs ina. wrong way. But ofcourse you can see if the following works, without much wrappers of our own, if you are interested (you already got approvals :) )

```
    async fn get(&self, object_key: &str) -> Result<impl AsyncRead, Box<dyn Error>>;



```

```
 let response = self
            .client
            .get_object()
            .bucket(&self.bucket_name)
            .key(key)
            .send()
            .await?;

        let body = response.body.into_async_read();
        Ok(body)
```

```
        let bytes: impl AsyncRead+Sized = self
            .template_service
            .download(&template_id.0, version.0)
            .await?;

        Ok(Response::builder()
            .header("Content-Type", "application/octet-stream")
            .body(Body::from_bytes_stream(bytes))
            .expect("failed to construct response"))
```

justcoon: but this is more like case of rest api
where this can be done 

```rust
 async fn download_template(
        &self,
        template_id: Path<TemplateId>,
        version: Query<Option<u64>>,
    ) -> Result<Binary<Body>> {
        let bytes = self
            .template_service
            .download_stream(&template_id.0, version.0)
            .await?;
        Ok(Binary(Body::from_bytes_stream(bytes.map_err(|e| {
            std::io::Error::new(std::io::ErrorKind::Other, e.to_string())
        }))))
    }
```

afsalthaj (approved):

PR #444, opened 2024-04-19
Automatic worker update

### LINKED ISSUE #390: Implement automatic update during recovery
Implement the automatic update handling during recovery, mark the update’s result when it diverges or reaches live. At this point we have the worker-executor level of the automatic updates done.

### LINKED ISSUE #391: Define worker-executor tests for testing the automatic update feature

### LINKED ISSUE #396: Expose the hot update feature in worker-service

vigoo:
Resolves #390 
Resolves #391 
Resolves #396


golem-worker-service/src/api/worker.rs
```diff
@@ -314,6 +314,31 @@ impl WorkerApi {
 
         Ok(Json(ResumeResponse {}))
     }
+
+    #[oai(
+        path = "/:component_id/workers/:worker_name/update",
+        method = "post",
```
justcoon, line 320: should not it be `PUT` with  path `/:component_id/workers/:worker_name` ?

golem-service-base/src/model.rs
```diff
@@ -2620,6 +2623,41 @@ pub struct InterruptResponse {}
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd, Serialize, Deserialize, Object)]
 pub struct ResumeResponse {}
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd, Serialize, Deserialize, Object)]
+pub struct UpdateResponse {}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Enum)]
+pub enum UpdateMode {
```
justcoon, line 2630: as this is in common services namespace i would consider to rename it to WorkerUpdateMode

golem-service-base/src/model.rs
```diff
@@ -2620,6 +2623,41 @@ pub struct InterruptResponse {}
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd, Serialize, Deserialize, Object)]
 pub struct ResumeResponse {}
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd, Serialize, Deserialize, Object)]
+pub struct UpdateResponse {}
```
justcoon, line 2627: as this is in common services namespace i would consider to rename it to WorkerUpdateResponse

vigoo: Hm makes sense, but currently it follows the pattern of the other ones. (See `ResumeResponse` for example above)

golem-worker-service/src/api/worker.rs
vigoo: I'm not sure, because it is not really an update in itself - it just registers an update attempt that happens asynchronously and may fail (and this failure is not reported by this request's response). So for me it is more like the pause, resume, etc actions on a worker.

golem-service-base/src/model.rs
justcoon: yes i noticed this part, maybe then we should do also `ResumeResponse` -> `ResumeWorkerResponse` , ....

justcoon (approved):

golem-worker-service/src/api/worker.rs
jdegoes: `POST` = create (id not specified in URL)
`PUT` = update (id specified in URL)

PR #445, opened 2024-04-20
get-workers test template

justcoon:
simple test template using get-workers golem-wit resource


vigoo:
What's the purpsoe of this? We already had tests for this in the `runtime-service` test component, which we used to have all the test functions for different Golem runtime APIs

vigoo:
There was intentionally not a separate test-component per tested function - it is just a lot of overhead (in terms of files checked in etc - especially that we now need to copy all the WIT files to each).

justcoon:
hi @vigoo 
for runtime tests, i added get workers function to `runtime-service`
then for documentation, i wanted to write some simple component, to show how can it be used, and once i wrote it, i was thinking to add it to golem test templates, if I already spent time to work on that, but we can decline PR, 
as you mentioned, probably in current state it will require additional maintainnance

justcoon:
ok, i will decline this PR, and will add it to https://github.com/golemcloud/golem-examples

PR #456, opened 2024-04-23
Use Golem WIT from JS

nicoburniske:


vigoo (approved):
You could have called something that is easier to check if it actually worked - for example the wasi keyvalue interface :) but good enough for now

PR #461, opened 2024-04-25
Worker update in CLI

### LINKED ISSUE #397: Add the hot update feature to the CLI

jupposessho:
Resolves #397

### DIFF
--- a/golem-cli/src/worker.rs
+++ b/golem-cli/src/worker.rs
@@ -33,7 +33,7 @@ use crate::model::text::WorkerAddView;
 use crate::model::wave::type_to_analysed;
 use crate::model::{
     ComponentId, ComponentIdOrName, Format, GolemError, GolemResult, InvocationKey,
-    JsonValueParser, WorkerName,
+    JsonValueParser, WorkerName, WorkerUpdateMode,
 };
 use crate::parse_key_val;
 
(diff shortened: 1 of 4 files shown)


nicoburniske (approved):

golem-cli/src/worker.rs
```diff
@@ -231,6 +231,25 @@ pub enum WorkerSubcommand {
         #[arg(short, long)]
         precise: Option<bool>,
     },
+    /// Updates a worker
+    #[command()]
+    Update {
+        /// The Golem component of the worker, identified by either its name or its component ID
```
senia-psm, line 237: Doc comment on `#[command(flatten)]` is not used in `--help`.

golem-cli/tests/worker.rs
```diff
@@ -724,3 +725,38 @@ fn worker_list(
...
+        &cfg.arg('w', "worker-name"),
+        &worker_name,
+        &cfg.arg('C', "component-id"),
+        &component_id,
+        &cfg.arg('m', "mode"),
+        "auto",
+        &cfg.arg('t', "target-version"),
+        "1",
+    ])?;
+
```
senia-psm, line 760: Please check the new version after update via `worker list` command.

senia-psm (approved):

golem-cli/tests/worker.rs
jupposessho: check added, thanks.

golem-cli/tests/worker.rs
```diff
@@ -724,3 +725,59 @@ fn worker_list(
...
+        &worker_name,
+        &cfg.arg('C', "component-id"),
+        &component_id,
+        &cfg.arg('m', "mode"),
+        "auto",
+        &cfg.arg('t', "target-version"),
+        "1",
+    ])?;
+    let worker_updates_after_update = workers_list()?.workers[0].updates[0].clone();
+    let target_version = if let UpdateRecord::PendingUpdate(pu) = worker_updates_after_update {
```
senia-psm, line 774: Could you please match `SuccessfulUpdate` as well? In case of slow test runner the server might be able to update worker faster than the test asks for worker list.

senia-psm: Or even better you can put `workers_list()?.workers[0].component_version` into `else` branch.

golem-cli/tests/worker.rs
vigoo: There is a race condition here because the pending update will become a failed update (as you are trying to update to a non-existing version). Maybe the test will be flaky because of this.

PR #468, opened 2024-04-29
429 immutable api definition

### LINKED ISSUE #429: Immutable API definition versions
Clean separation between draft and immutable api definition version.
2 Options:

- `draft` field in api definition version, that allows updates and prevents deployment
- prevent updates of deployed versions (in this case we need a virtual `deployed` field in version)

senia-psm:
Resolves #429

### DIFF
--- a/golem-worker-service-base/src/repo/api_definition_repo.rs
+++ b/golem-worker-service-base/src/repo/api_definition_repo.rs
@@ -10,7 +10,7 @@ use tracing::{debug, info};
 use golem_common::config::RedisConfig;
 use golem_common::redis::{RedisError, RedisPool};
 
-use crate::api_definition::ApiDefinitionId;
+use crate::api_definition::{ApiDefinitionId, HasIsDraft};
 use crate::repo::api_namespace::ApiNamespace;
 use crate::service::api_definition::ApiDefinitionKey;
 
@@ -26,6 +26,11 @@ pub trait ApiDefinitionRepo<Namespace: ApiNamespace, ApiDefinition> {
         &self,
         definition: &ApiDefinition,
         key: &ApiDefinitionKey<Namespace>,
+    ) -> Result<(), ApiRegistrationRepoError>;
+
+    async fn set_not_draft(
+        &self,
+        api_definition_key: &ApiDefinitionKey<Namespace>,
     ) -> Result<(), ApiRegistrationRepoError>;
 
     async fn get(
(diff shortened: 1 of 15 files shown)


golem-worker-service-base/src/service/api_deployment.rs
```diff
@@ -86,12 +89,22 @@ impl<Namespace: ApiNamespace, ApiDefinition> ApiDeploymentService<Namespace>
...
-                api_definition_key.namespace,
-                api_definition_key.id,
-            ));
+            })?
+            .ok_or(ApiDeploymentError::ApiDefinitionNotFound(
+                api_definition_key.namespace.clone(),
+                api_definition_key.id.clone(),
+            ))?;
+
+        if definition.is_draft() {
```
justcoon, line 98: there is existing deployment check,

i think we should update it only in case when it not fail

golem-worker-service-base/src/repo/api_definition_repo.rs
```diff
@@ -245,6 +275,31 @@ where
         }
     }
 
+    async fn make_immutable(
```
justcoon, line 278: maybe  `make_immutable` -> `set_not_draft`

golem-worker-service-base/src/service/api_deployment.rs
senia-psm: If there is an existing deployment for a draft definition than we have an invalid state and should fail any way.

golem-worker-service-base/src/repo/api_definition_repo.rs
senia-psm: I considered it, but I believe that `make immutable` better represents intention while `set_not_draft` describes implementation details.

golem-worker-service-base/src/service/api_deployment.rs
senia-psm: Fixed

golem-worker-service-base/src/service/api_deployment.rs
```diff
@@ -44,6 +46,7 @@ pub trait ApiDeploymentService<Namespace> {
 }
 
 pub enum ApiDeploymentError<Namespace> {
+    ApiDefinitionIsDraft(Namespace, ApiDefinitionId),
```
justcoon, line 49: is this used?

golem-worker-service-base/src/repo/api_definition_repo.rs
justcoon: i am not sure if this is implementation detail, 

basically there is `draft` flag in Api definition which at some point is synonym for immutable,
at the other side there is use of `immutable` , for example  `ApiRegistrationRepoError::Immutable` 

```rust
#[error("Immutable: ApiDefinition with id: {} and version: {} in namespace {} can not be updated", .0.id, .0.version, .0.namespace)]
    Immutable(ApiDefinitionKey<String>),
```
and then solution working with two words which more like mean same, 
and then this will require additional explanation what is relation between `draft` and `immutable`

if `immutable` state was something which is result of some computation or cover combination of multiple values or flags, then it could make sense to use different words

but otherwise, it is better to use just one term or word

golem-worker-service-base/src/service/api_deployment.rs
senia-psm: Good catch. Fixed.

golem-worker-service-base/src/repo/api_definition_repo.rs
senia-psm: Done

justcoon (approved):

PR #475, opened 2024-05-01
Evaluator Namespace, refined worker-response (no vector) and optimised workflow

### LINKED ISSUE #473: Make sure to distinguish between unit function result, named function results and single-function result when working with worker bridge

### LINKED ISSUE #494: Introduce EvaluationContext, that is well typed
Currently there is no concept of evaluation context, which can act as a global namespace for expression evaluator, to evaluate selection-field logic which can be arbitrary selection of a field or specific fields from global namespaces such as `request`, `worker metadata` , `worker-response` (which may or may not be a proper result) etc.

As of now, every input is just `TypeAnnotatedValue` which apparently got broken, after the introduction of refined worker response which can be unit, or multiple results or single result. Also, it has become difficult to maintain as everything got converted into TypeAnnotatedValue, such that these request details, worker details etc lost their type information as we became ignorant whether or not certain field belong to worker, or request. 

This, I think, will be the final touch required before milestone 1.0 in terms of supporting all kind of select-fields user can do.

afsalthaj:
Fix #473 
Fix #494

## Introduction of Evaluation Context
 
* Evaluation of an expression now takes an `EvaluationContext` that can hold the values of global namespaces such as worker,  request, or custom variables (foo), instead of an arbitrary type-annotated-value

* During evaluation of an expression in an `EvaluationContext`, the child values in these global namespaces get converted to `TypeAnnotatedValue` which is typed evaluation, allowing us to do things like `if request.user_id  == 1`. The difference here is, until evaluation we are never converting any of these details to `TypeAnnotatedValue`.

* The `request` namespace in EvaluationContext is basically an `enum` of `Http` or in future `Grpc` and if it is `Http`, it will consist of `HttpRequestDetails`.
 
* With this PR, more selection fields are supported. `worker.name`, `worker.component_id`, `worker.idempotency-key` , `worker.response` (implies `worker.response.foo`), `request.headers.idempotency-key`, `request.headers.foo`, `request.user-id` and so on.

## `RefinedWorkerResponse` - instead of response always being a vector

* After worker executor is executes the  resolved worker request, the response `WorkerResponse` is converted to `RefinedWorkerResponse` 

* `RefinedWorkerResponse` consist of refined results, by making  use of FunctionResult metadata to identify if they are named multiple results, or a single result, or unit function. This implies, the final http response goes from `RefinedWorkerResponse`. Therefore we have a `to_http_response` function under `RefinedWorkerResponse` given a `ResponseMapping` and  original`RequestDetails`  (in case mapping refers back to the original request details)

* The logic is also done with the _assumptions_ which I confirmed by going through WASM AST code which essentially result in the current model of `FunctionResult` in golem-services/model.rs

## Change in API Gateway Workflow

* Step 1: Given the above changes, the workflow also had to be simplified. Now for a custom http request hitting the worker-bridge, it looks up the API definition corresponding to the request path pattern.

* Step 2: Resolve the `worker-binding` (called `ResolvedWorkerBinding given the original http request, and `api-definition`. This is by making use of the instance of `WorkerBindingResolver` for `InputHttpRequest`. In future we will have `WorkerBindingResolver` instance for `GrpcRequest`.

* Step 3: Unlike before, `ResolvedWorkerBinding` has concrete `WorkerRequest`, `RequestDetails` as well as optional `ResponseMapping`. Previously, we kept propagating the entire binding, partial informations (as expressions) which is not super easy to reason about.

* Step 3: `ResolvedWorkerBinding` has `execute_with` function that executes the `WorkerRequest` and emit a Response (`A)` in the binding (given an executor) and instance of `ToResponse<A>`  for the `WorkerResponse`. Currently we have instances of `ToResponse<poem::Response>` for `WorkerResponse` as well as `WorkerExecutorError`. This allows us to easily add a GRPC version of these instances, without repeating the workflow.

### Minor: Spec changes
* Use `component_id ` instead of component
* Use `worker_name` instead of `worker_id` (it's a name and not the worker-id)


justcoon (approved):

golem-worker-service-base/src/api/custom_http_request_api.rs
```diff
@@ -91,52 +90,16 @@ impl CustomHttpRequestApi {
...
-                        }
-                    };
-
-                // Execute the request using a executor
-                match self
-                    .worker_to_http_response_service
-                    .execute(resolved_worker_request.clone())
+            Ok(resolved_worker_request) => {
+                resolved_worker_request
+                    .execute_with::<poem::Response>(&self.worker_request_executor_service)
```
afsalthaj, line 95: This specifying the type wasn't required.

golem-worker-service-base/src/http/http_request.rs
```diff
@@ -422,7 +193,7 @@ mod tests {
 
         let resolved_route = api_request.resolve(&api_specification).unwrap();
 
-        let result = WorkerRequest::from_resolved_route(resolved_route.clone());
+        let result = resolved_route.worker_request;
```
afsalthaj, line 196: This is an example that shows the workflow is better now. This is because, we are resolving the concrete worker request during the route resolution instead of a 2 step process (i.e, expr that is required to form the worker request gets evaluated as early as possible).

afsalthaj: Previously, the expr got propogated even further closer to the evaluation stage, but that's not needed.

golem-worker-service-base/src/evaluator/evaluator_context.rs
```diff
@@ -0,0 +1,69 @@
...
+use crate::merge::Merge;
+use crate::worker_binding::RequestDetails;
+use crate::worker_bridge_execution::{RefinedWorkerResponse, WorkerRequest};
+use golem_wasm_rpc::TypeAnnotatedValue;
+
+// Evaluator of an expression doesn't necessarily need a context all the time, and can be empty.
+// or contain worker details, request details, worker_response or all of them.
+
+#[derive(Clone)]
+pub struct EvaluationContext {
```
jdegoes, line 13: This is very poorly typed, in my opinion, because any combination of things can be `None`. Structures like this are usually better off as `enum`, but in this case, I will suggest something different for `evaluate` below.

golem-worker-service-base/src/evaluator/mod.rs
```diff
@@ -1,24 +1,59 @@
...
 use getter::GetError;
 use getter::Getter;
 use path::Path;
 
 use crate::expression::{Expr, InnerNumber};
-use crate::merge::Merge;
 
-use crate::tokeniser::tokenizer::{MultiCharTokens, Token, Tokenizer};
+pub trait Evaluator {
+    fn evaluate(
```
jdegoes, line 21: I think evaluation should take an input expression, which is nothing more than a record (`TypeAnnotatedValue`  or even more specific), together with a standard library -- i.e. functions which may be called by the expression (including, of course, worker functions, utility functions, etc.). A function can be defined by its formal definition (namespace, name, parameters and their types and names), and then an implementation, which is something the evaluator can call into, and which returns a type annotated value.

This is the way to enabling extensionality and also allowing Rib to be pulled out of this module and shipped as its own crate.

jdegoes: Actually, this is too poorly typed: the input to the evaluator should be a map of identifier name to `TypedAnnotatedValue` (yes, a record is basically the same thing, but a bunch of let-bindings is clearer and more type safe).

afsalthaj: The input to the evaluator is actually an expression.
The type-annotated-value is global namespace details, which are simply represented using type-annotated value. 

Logically what we have is the following

```
fn evaluate(input: Expr, context: EvaluationContext) -> Result<EvaluationResult, EvaluationError>

```

afsalthaj: The input expression is of the form `Expr`. It is the `evaluator` that converts it into `TypeAnnotatedValue`.

`TypeAnnotatedValue` is not further evaluated, meaning the input of the current evaluator is `Expr` and output is `TypeAnnotatedValue` (currently its called EvaluationResult)

afsalthaj: We can pair up on this for a quick discussion once I reach home

golem-worker-service-base/src/evaluator/evaluator_context.rs
afsalthaj: @jdegoes  Initially this was an enum, but it can be any combination at any point in time,including empty. We can call evaluator with just an input Expression `Expr` without passing any extra context (context::empty)

golem-worker-service-base/src/evaluator/mod.rs
afsalthaj: Current evaluationContext can be empty. As in, it can be any combination.  Users can ask the evaluator to evaluate an expression (to form a type annotated value) given optional worker info, optional request info etc. I don’t think we are ending up in representing any illegal states with this typing. But I would like to pair up with you and get more thoughts there.

jdegoes (approved):

afsalthaj:
We will be pairing up with John to make Expr more generic

PR #477, opened 2024-05-01
Idempotency key and exactly-once RPC

### LINKED ISSUE #360: RPC should have exactly-once semantics

vigoo:
Resolves #360 

- [x] Change terminology from `InvocationKey` to `IdempotencyKey`
- [x] Remove the `get-invocation-key` APIs, idempotency key is always generated by the caller
- [x] Make it possible to pass an optional idempotency key for both `invoke` and `invoke-and-await`
- [x] Removed the obsolete "JSON over Protobuf" invocation API which was only needed for the migration from Scala
- [x] Change the REST API to use idempotency headers instead
- [x] Change worker bridge to allow passing idempotency keys from the incoming headers and/or using explicit Rig expression
- [x] Generate a persistent idempotency key in the WASM RPC implementation before doing the invocation in order to make it exactly-once
- [x] Fix the "invocation key service" in worker executor base to be backed by a cached map of idempotency keys with pointers to the oplog for results, fully recoverable on startup from the oplog/metadata cache.
- [x] Remove the stdio-eventloop calling convention


golem-cli/src/model/text.rs
```diff
@@ -266,7 +266,7 @@ impl TextFormat for WorkerAddView {
     }
 }
 
-impl TextFormat for InvocationKey {
+impl TextFormat for IdempotencyKey {
```
senia-psm, line 269: You can remove this `impl`.

golem-cli/tests/worker.rs
```diff
@@ -195,7 +160,7 @@ fn worker_invoke_and_await(
         "TEST_ENV=test-value",
         "test-arg",
     ])?;
-    let args_key: InvocationKey = cli.run(&[
+    let args_key: IdempotencyKey = cli.run(&[
```
senia-psm, line 163: No such command

golem-cli/tests/worker.rs
```diff
@@ -222,7 +187,7 @@ fn worker_invoke_and_await(
 
     assert_eq!(args, expected_args);
 
-    let env_key: InvocationKey = cli.run(&[
+    let env_key: IdempotencyKey = cli.run(&[
```
senia-psm, line 190: No such command any more

golem-worker-executor-base/src/durable_host/mod.rs
```diff
@@ -1217,15 +1209,22 @@ impl<Ctx: WorkerCtx> PrivateDurableWorkerState<Ctx> {
...
     async fn get_oplog_entry_exported_function_invoked(
         &mut self,
-    ) -> Result<Option<(String, Vec<Value>, InvocationKey, Option<CallingConvention>)>, GolemError>
-    {
+    ) -> Result<
+        Option<(
+            String,
+            Vec<Value>,
+            IdempotencyKey,
+            Option<CallingConvention>,
```
afsalthaj, line 1217: Minor: Could have been better to be a struct?

golem-worker-service-base/src/service/worker/default.rs
```diff
@@ -325,55 +321,11 @@ where
...
-            })
-            .await?;
-
-        Ok(invocation_key)
-    }
-
     async fn invoke_and_await_function(
         &self,
         worker_id: &WorkerId,
+        idempotency_key: Option<IdempotencyKey>,
```
afsalthaj, line 327: So we generate one if its absent, and works as before?

afsalthaj: And at that point, will it break exactly once behaviour? Trying to understand

golem-worker-executor-base/src/durable_host/mod.rs
vigoo: This is not changed in this PR, you just see here because of the rename

golem-cli/src/model/text.rs
senia-psm: It would be better to remove this unused `impl`. It's a bit misleading.

golem-worker-executor-base/src/services/invocation_key.rs
```diff
@@ -1,274 +1,264 @@
...
-        let r1 = svc2.lookup_key(&worker_id, &key1);
-        let r2 = svc2.lookup_key(&worker_id, &key2);
-        let r3 = svc2.lookup_key(&worker_id, &key3);
-
-        assert_eq!(r1, LookupResult::Complete(Ok(vec![Value::U32(1)])));
-        assert_eq!(r2, LookupResult::Complete(Ok(vec![Value::U32(2)])));
-        assert_eq!(r3, LookupResult::Complete(Ok(vec![Value::U32(3)])));
-    }
-}
+// // Copyright 2024 Golem Cloud
```
justcoon, line 1: should not we remove this file if everything is commented?

jdegoes (approved):

golem-worker-executor-base/src/services/invocation_key.rs
vigoo: we should, thanks

jdegoes (approved):

PR #478, opened 2024-05-01
Make subdomain optional

### LINKED ISSUE #460: Make subdomain optional in ApiDeployment
This looks odd, and there must be some _reason_ for why this hack was done. We can uncover it once we make subdomain optional.

```scala
impl Display for ApiSite {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        // Need to see how to remove the need of subdomain for localhost , as subdomains are not allowed for localhost
        let host = &self.host;
        if host.contains("localhost") || host.contains("127.0.0.1") {
            write!(f, "{}", self.host)
        } else {
            write!(f, "{}.{}", self.subdomain, self.host)
        }
    }
```

It doesn't need to be this way - that's for sure. Users should be able to simply host their API in localhost:9006, for example without worrying about domain and subdomain.

afsalthaj:
Fixes #460


afsalthaj:
Once released need to see how it all chains together with golem-cloud

afsalthaj:
Generating client........ 
Will raise separate PR that speeds up this

justcoon (approved):

vigoo (approved):

Issue #324, opened 2024-03-26
Map idempotency headers to worker-executor invocation keys


vigoo:
The headers mentioned in https://github.com/golemcloud/golem/issues/323 could also be supported on the "server side", when providing a HTTP API on top of Golem workers.

Worker executors already support this with the *invocation key*.

justcoon:
i think that this issue is already implemented: [worker_binding_resolver](https://github.com/golemcloud/golem/blob/22f5d315dd50b1b7013f238d322c481a51125d84/golem-worker-service-base/src/worker_binding/worker_binding_resolver.rs#L220-L237)
cc: @afsalthaj, @vigoo

PR #486, opened 2024-05-08
worker bridge - idempotency key

justcoon:


afsalthaj (approved):

jdegoes:
@justcoon The worker bridge needs to pull idempotency keys from the HTTP header. 

See [here](https://datatracker.ietf.org/doc/draft-ietf-httpapi-idempotency-key-header/)

cc @vigoo

vigoo:
Currently you can set it using Rig any way you want (for example by forwarding that or any other header) but it's not happening automatically

afsalthaj:
Setting the idempotency key is per binding, and this part was already done in @vigoo 's big PR

vigoo:
I think @justcoon 's latest commits also implement https://github.com/golemcloud/golem/issues/324

golem-worker-service-base/src/worker_bridge_execution/mod.rs
```diff
@@ -77,6 +78,28 @@ impl WorkerRequest {
...
+            let idempotency_key = match idempotency_key_value {
+                TypeAnnotatedValue::Str(value) => value,
+                _ => return Err("Idempotency Key is not a string".to_string()),
+            };
+
+            Some(IdempotencyKey::new(idempotency_key))
+        } else {
+            resolved_route
+                .headers
+                .get("idempotency-key")
```
vigoo, line 98: Is this working case-insensitively?

vigoo (approved):

golem-worker-service-base/src/worker_bridge_execution/mod.rs
justcoon: headers should not be case sensitive, i added test

PR #492, opened 2024-05-10
golem-component-compilation-service - fix

justcoon:


vigoo (approved):

golem-component-compilation-service/src/lib.rs
```diff
@@ -91,6 +91,8 @@ async fn run(config: ServerConfig, prometheus: Registry) -> Result<(), Box<dyn s
 
     info!("Server started on port {}", config.grpc_port);
 
+    drop(http_server); // explicitly keeping it alive until the end
```
senia-psm, line 94: I really like using explicit `drop`, just want to inform you that in other places in code we are using `_` prefix instead of explicit drop (`let _http_server = ...`). Rust keeps all names variables alive till the end of code block.

senia-psm (approved):

PR #520, opened 2024-05-15
Worker executor storage abstractions

### LINKED ISSUE #288: Refactor data store access in worker executor to make it easier to implement new ones
Access to redis and S3 should be refactored in a way that makes it very easy to implement connectors to new data stores with similar capabilities.

vigoo:
Resolves #288 

Introduces three storage types:
- key-value storage (also supporting storing sets and sorted sets) 
- indexed storage
- blob storage

The key-value storage and indexed storage has Redis and in-memory implementations, the blob storage has S3, in-memory and local file system based implementations.

This simplifies the code as individual services no longer have different implementations (for example `PromiseService` used to have a redis and an in-memory implementation now there is only a single one depending on key-value storage, etc)

All three storage interfaces get _namespaces_ which are fixed enums, and they can use it in different ways to determine how to implement data. For example in Redis it affects key prefixes, and whether to use hashed objects or not, but other implementations may use it to use different tables, or choose to just use it as part of a key.

Cleaned up key paths in redis a bit (no longer using `instance:` for example)

Changed how blob-storage is implemented: previously it was creating an S3 bucket for each "container" in the WASI blob-storage interface which is not very good because the number of buckets is limited in an AWS account and it also requires more rights than desirable. So right now the S3 blob storage implementation only requires two pre-configured buckets, one for compilation cache and one for user data where the logical structure is determined by key prefixes (account-id/container-name)


golem-worker-executor-base/src/storage/mod.rs
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub mod blob;
+pub mod indexed;
+pub mod keyvalue;
```
jdegoes, line 17: Might be nice to have a sqllite version of this (and maybe `blob`) just for local development.

jdegoes (approved):

golem-worker-executor-base/src/storage/mod.rs
vigoo: I agree, just did not want to spend more time on implementing new backends now.

PR #521, opened 2024-05-15
Trying to speed up CI

vigoo:
- Using https://github.com/Swatinem/rust-cache instead of custom set up
- Using https://github.com/mozilla/sccache/
- Tweaks to the cargo make file to make things faster
- Restructured CI job flow trying to maximize reuse of the cache
- 5 parallel worker executor test jobs
- Tried `cargo-hikari` but decided to not use it


justcoon (approved):

afsalthaj (approved):

golem-worker-service/Cargo.toml
```diff
@@ -15,48 +15,48 @@ path = "src/main.rs"
...
+tokio = { workspace = true }
+tokio-stream = { workspace = true }
+tokio-util = { workspace = true }
+tonic = { workspace = true }
+tonic-health = { workspace = true }
+tonic-reflection = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
+url = { workspace = true }
+uuid = { workspace = true }
```
afsalthaj, line 62: Does these changes made it faster as well? As in making sure workspace=true is applied?

PR #523, opened 2024-05-16
Support all content type

### LINKED ISSUE #205: Make sure we can serve arbitrary content from workers
- Need to make sure we can use string and vec<u8> responses in API Gateway to directly set the body (not wrapping in JSON etc)
- and that we can set the content type header

afsalthaj:
Fixes #205 

Read through the requirements in the issue to reason about the logic applied


afsalthaj:
Closing this, as I realise more things to be done before draft PR

PR #526, opened 2024-05-19
Fix docker configurations

afsalthaj:


afsalthaj:
Merging this as this is a minor fix.

PR #532, opened 2024-05-22
Support Worker Function Invocation

afsalthaj:
### Main changes

* Allows a Rib script that can consist of worker invocation function to form the response

```rust

let status = 200;

let start_time = request.path.time;
let end_time = request.path.end_time;
let result = timeline:driver/api/run(start_time, end_time);

{ body: result, status: status }


```

* Make `EvaluationContext` independent of API Gateway use case. This implies, by the time this PR is ready, at least we can say, Rib is fully independent of Golem OSS API Gateway.

Now evaluation-context is

```scala
use golem_wasm_ast::analysis::AnalysedFunction;

#[derive(Clone)]
pub struct EvaluationContext {
    pub variables: Option<TypeAnnotatedValue>,
    pub analysed_functions: Vec<AnalysedFunction>,
}

```

* Make evaluator be able to run the invocation expressions in the Rib Expression Tree

* Use double-quotes instead of single quotes (this is a decent change from tokeniser itself) to align with warm-wave syntax
* Allow users to form variants/enums etc which are simple identifiers until evaluation. Depending on the info in evaluation context, they can be resolved to Enum, Variant, or even function-call.

@jdegoes  I thought I will have at least this decent shape before we start pairing up. It's even easier to reason about what's in there, and I will be more comfortable to pair up on this branch to bring in type information and fix things.


jdegoes (approved):

golem-cli/tests/text.rs
```diff
@@ -544,11 +544,11 @@ fn text_api_definition_import(
             "
             API Definition imported with ID {component_name} and version 0.1.0.
             Routes:
-            +--------+------------------------------+-------------+--------------------------------+--------------------------------+
-            | Method | Path                         | ComponentId | WorkerName                     | Function                       |
-            +--------+------------------------------+-------------+--------------------------------+--------------------------------+
-            | Get    | /{{user-id}}/get-cart-contents |    *{component_end} | worker-${{request.path.user-id}} | golem:it/api/get-cart-contents |
```
senia-psm, line 550: `{{` it is escaping. You have to preserve `{{user-id}}`

PR #535, opened 2024-05-26
Fully reuse worker service instead of trying to play with component metadata (for now) in API Gateway

afsalthaj:
As soon as we evaluate the function name, we simply call invoke function


afsalthaj:
Closing this as I will revisit after my current task (probably for the weekend)

Issue #502, opened 2024-05-14
Check again if we can generalize the AccountId in worker executor


vigoo:

afsalthaj:
This probably won't be taken up until @vigoo  is back

PR #538, opened 2024-05-30
Blob-store based oplog archive

### LINKED ISSUE #216: Multi layer oplog store, enabling an S3 backend
See https://www.notion.so/golemcloud/Oplog-format-v2-09508cb6f2a646d4ace1e27f8501c8e8?pvs=4

vigoo:
Resolves #216 

Contains refactorings related to #502 

Introduces `OwnedWorkerId`, a combination of `WorkerId` and `AccountId` and uses it in most places in the worker executor where previously worker and account IDs were passed separately. Also the number of places where this type is needed (the additional account ID) increased because now the _oplog_ requires to have an associated account ID for each operation. Although this change increases the required information to be transferred in many cases, it unified the API a bit as now everything requires a combination of worker id and account id: this can be explained by worker executor needs to know who owns the worker, but it does not want to query an external service to determine so - so this information must be passed from the outer layers where authorization already calculated the information anyway (or in case all worker belongs to the same account, it is passed as a constant value).

The primary reason for passing this ownership information down to the oplog service is to be able to store oplog archives grouped by accounts, similarly how custom blobs are already stored grouped by accounts. This is good for various security and privacy reasons and can support some planned features.

### DIFF
--- a/golem-worker-executor-base/src/storage/blob/mod.rs
+++ b/golem-worker-executor-base/src/storage/blob/mod.rs
@@ -12,19 +12,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::fmt::Debug;
+use std::path::{Path, PathBuf};
+
+use async_trait::async_trait;
+use bincode::{Decode, Encode};
+use bytes::Bytes;
+
+use golem_common::model::{AccountId, ComponentId, Timestamp, WorkerId};
+use golem_common::serialization::{deserialize, serialize};
+
 pub mod fs;
 pub mod memory;
 pub mod s3;
 
-use async_trait::async_trait;
-use bytes::Bytes;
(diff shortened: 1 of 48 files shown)


golem-worker-executor-base/src/storage/blob/mod.rs
```diff
@@ -155,9 +331,14 @@ pub enum BlobStorageNamespace {
         account_id: AccountId,
         worker_id: WorkerId,
     },
+    CompressedOplog {
+        account_id: AccountId,
+        component_id: ComponentId,
```
justcoon, line 336: is there reason why here is `component_id` and not `worker_id`?

vigoo: This way it was easy to implement the equivalent of redis SCAN without changing / extending the blob storage interface.

afsalthaj (approved):

Issue #403, opened 2024-04-15
Durability overhead benchmark


vigoo:
Create a benchmark that compares performance when persistence is disabled (through the transaction API) vs the normal persistent mode.

Issue #384, opened 2024-04-15
Suspend benchmark


vigoo:
Write a simple benchmark which can use the existing `clocks` test component to test how much time it takes to start workers which immediately get suspended for a while, but also get resumed relatively at the same time:

- set up cluster of `size`
- warmup: start `length` number of workers of the same component
- benchmark: call `sleep-for(10s)` on them in parallel (must be larger than the configured suspend threshold for sleep)
- measure how much time it took to complete the invocations (will be > 10s)

This ticket is only about writing a new benchmark and does not involve anything test-framework / infrastructure related.

Issue #383, opened 2024-04-15
Throughput benchmark


vigoo:
Implement a few benchmarks similar to each other to measure throughput.

All the benchmarks should have the following structure

- start a cluster of `1`
- start a `size` number of workers 
- warmup: call the benchmarked function a few times
- call a function `length` times

In this benchmark we want `size` to be low enough that all workers are continuously in memory - if not we will see that in the results. 

We should have the following variants (can be different exported functions of the same component), each forming a separate benchmark:

- simple parameter and return value, no significant work (string echo, for example)
- complex and big data structure passed as a function argument or returned as a result value (for example an array of relatively complex records and the endpoint needs to traverse it and calculate a result)
- simple in/out parameters but CPU computation heavy function body

This should be a new test component which is a multi-crate Rust project, implementing the actual logic as a library, and beside exposing it as a Golem copmonent there should also be a Rust http server with a JSON HTTP API that we can compare the throughput to (it's enough if we can compare it when running locally for now - no need to figure out a way to launch this custom http server in docker / k8s / etc).

This ticket is only about writing the benchmark and the test component / test server, should not involve benchmark framework or infrastucture related work.

Issue #382, opened 2024-04-15
General latency benchmark


vigoo:
Similar benchmark to the #379 but working with a fixed set of worker names instead of always starting fresh ones:

- generate N worker names
- warmup: invoke all
- benchmark: invoke the workers and measure invocation time

Note: what we will see will depend heavily on the worker executor's active worker cache size and how we manage memory constraints.

The invoked worker functions should be trivial (counter, or echo, etc) so it does not affect the measurement.

Note: the `simple_worker_echo` benchmark may already be this but make sure it is and create a few variants with different component sizes (because that will affect load time as well as the memory occupied per worker).

PR #545, opened 2024-06-04
Benchmark report

### LINKED ISSUE #507: Run benchmarks in CI
The benchmarks should be executed in CI so we continuously see any performance regressions.

Ideas/nice to have:
- trigger it on-demand in PRs (so we don't run all the time only when we think our change might have performance effects)
- report result in the PR as comments
- figure out a nice way to show results
- should show diff to some baseline

afsalthaj:
Fixes #507 

## Workflow Tested With Golem-Timeline

1. Developer raise a PR
2. Reviewer finds out that its worth running a benchmark, and sends a PR comment `/run-benchmark`
3. The benchmark job kicks off, and runs for all benchmark types producing a report as `Markdown` comment back to PR
    Example:
image
4. The benchmark data (that we used to create the above report) is auto-committed back to the branch (of PR). 
5. If the reviewer is happy, they click on `merge` button.
6. Now the master branch has baseline for comparing subsequent benchmark runs
7. For example, in the another PR, `/run-benchmark` will result in actually comparing the benchmark results and create a `comparison report` instead of simple report. See below (just an example)
image

## What happens if I add a new type of benchmark job in the `integration-tests/benchmark` in future?

1. This implies the master branch has only partial `benchmark-data` available and not for this new one
2. In this case running `/run-benchmark` results in 2 comments. One comment will be a comparison report for whatever it could compare and a standalone report for this new one
image

## What happens if I change the benchmark-report job or data-structures of the benchmarks in future?
1. This could lead to schema mismatches and failed comparison jobs. `/run-benchmark` will fail
2. At this point you can send a message `/run-benchmark-refresh` instead of `/run-benchmark` will run a report without comparing anything. See this
image

## Where to change the benchmark reporting and rendering etc?
1. All of this is in Rust, in `integration-tests` module, within `report` module of `benchmark`. So this is easy to change.
2. Obviously you need to be careful here as I spent a lot of time debugging on issues like the following:
   * You cannot form a json anywhere, because `issuing comment` to PR using Github APIs uses json, and escaping this json can lead to horrible UX. If you don't escape job will fail. This is why I chose markdown
   * You can see I have escaped newlines in markdown because, otherwise JSON parser error when sending messages to Github PR
   * In future you can plot and create image and upload to GitHub and send just the link to message, and the architecture now will make it easier for us to play with these in future.

## What if multiple jobs run benchmark?
1. Nothing much, the baseline for comparison is always based on `main` branch
2. The new benchmark data (the data we use to report) is always committed to this baseline (main branch) only when you merge.

## How this is architectured?
1. Tried to minimise the number of moving parts as small as possible
2. We have a single benchmark.yml as a new workflow, dedicated to run benchmark job, create report and send messages, and push the benchmark data back to the PR merge
4. The benchmark report job is within in benchmark module itself, so that we reuse the data structures to emit the data and parse the data
5. All benchmark jobs can now emit a json as it is more reliable to parse this back for report purposes. We use the json version to create the benchmark data, and run benchmark_report job that parses this (a simple command line parser there) and reports them together.

## Is this Tested ?

Refer to golem-timeline for all the tests conducted - dozens of PRs :)

https://github.com/afsalthaj/golem-timeline/pulls?q=is%3Apr+is%3Aclosed

This doesn't mean things will work straight away in Golem Repo. I tried best to have least surprises, but we can be 100% sure only when you merge this PR and then raise another PR to test these workflows. Please do expect failures and a couple of more reiterations


integration-tests/src/benchmarks/benchmark_service.rs
```diff
@@ -0,0 +1,68 @@
...
+    get, handler, listener::TcpListener, middleware::Tracing, post, web::Json, web::Path,
+    EndpointExt, Route, Server,
+};
+use golem_test_framework::config::CliParams;
+use golem_test_framework::config::cli::TestMode;
+use golem_test_framework::dsl::benchmark::{BenchmarkConfig, BenchmarkResult};
+use integration_tests::benchmarks::cold_start_large;
+
+#[handler]
+async fn run_cold_start_large() -> Json<BenchmarkResult> {
```
afsalthaj, line 12: This will be deleted./

golem-test-framework/src/config/mod.rs
```diff
@@ -45,6 +45,7 @@ pub trait TestDependencies {
     fn worker_executor_cluster(&self) -> Arc<dyn WorkerExecutorCluster + Send + Sync + 'static>;
 
     fn kill_all(&self) {
+        dbg!("By any chance");
```
justcoon, line 48: probably some leftovers from testing

justcoon (approved):

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,160 @@
...
+          ISSUE_NUMBER: ${{ github.event.issue.number }}
+          WORKFLOW_ID: ${{ github.run_id }}
+          COMMENT_BODY: ${{ github.event.comment.body }}
+        run: |
+          #! /bin/bash
+          
+          # Create a dummy benchmark result
+          mkdir -p ./benchmark-data-current
+          
+          ./target/debug/benchmark_cold_start_large --quiet --json spawned > benchmark-data-current/benchmark_cold_storage_large.json
```
vigoo, line 63: Based on the closed benchmark tickets we should have the following benchmarks:
- Suspend benchmark (https://github.com/golemcloud/golem/issues/384)
- 3 different cold start benchmarks (https://github.com/golemcloud/golem/issues/379)
- 3 different latency benchmarks (https://github.com/golemcloud/golem/issues/382)
- Durability overhead benchmark (https://github.com/golemcloud/golem/issues/403)
- 3 different cold-start benchmarks with compilation cache service enabled (https://github.com/golemcloud/golem/issues/381)
- Throughput benchmark (golem and native) (https://github.com/golemcloud/golem/issues/383)

Some of  these are missing, I hope only from this script ;)

afsalthaj: I can suspend and throughput. But there will definitely be iteration of 1 or 2 commits after the merge. I tested this only with golem-timeline, lets see how golem-repo is going to be testing my patience :)

justcoon: in relation to cold start benchmarks

compilation service is enabled by default
```
./target/debug/benchmark_cold_start_large  spawned
```
can be disabled with flag
```
./target/debug/benchmark_cold_start_large spawned --component-compilation-disabled
```

afsalthaj: So we are back to having 2 more required

PR #550, opened 2024-06-06
435 profiles

senia-psm:
termtosvg_u85ju84y


golem-cli/src/oss/main.rs
```diff
@@ -0,0 +1,111 @@
...
+        url,
+        worker_url,
+        allow_insecure,
+        config,
+    } = profile;
+
+    let format = cmd.format.unwrap_or(config.default_format);
+
+    let component_url = url;
+    let worker_url = worker_url.unwrap_or_else(|| component_url.clone());
```
justcoon, line 40: is there reason why property for component url is called once `url` and once `component_url`, and for worker just `worker_url` ?

justcoon (approved):

PR #555, opened 2024-06-07
Temporarily remove throughput benchmark

afsalthaj:
I have to remove this to see a complete workflow of benchmark

* Throughput benchmark kept failing in my machine. Firstly we need to ensure its running as consistent as other benchmarks before having this as part of the CI workflow.
* I am currently looking at this. 
* This PR is to unblock rest of the benchmark jobs
* Also as I mentioned earlier, it was a good idea to keep the number of jobs minimal when we test a workflows, and that's why I didn't have all of these jobs and it wasn't accidental


afsalthaj:
I am merging this as it is a straight forward change, not impacting any other part of the code.

PR #564, opened 2024-06-10
Final touch before release - make headers, status back to optional fields (and upgrade wasm-rpc?)

afsalthaj:
The main testing I wanted to do was upgrading wasm-rpc. I think the main contribution to wasm-rpc was wasm-rpc-stubgen (which is just my change allowing complex workflows)

Also also not sure whether we should release wasm-rpc before this [PR](https://github.com/golemcloud/wasm-rpc/pull/45) or simply give up on upgrading as such in this golem release

However, as part of the testing, I found the following behavior that got regressed in the API gateway.
This questioned the non existence of some unit tests. Therefore I added those tests.

The Rib expression always had optional headers etc when specifying response mapping. 
With Rib powerful in last couple of weeks, we also made it  mandatory (not quite intentional) to  specify status code, headers etc which can give horrible user experience with the next release. This is made back to being optional.

### DIFF
--- a/golem-worker-service-base/src/evaluator/math_op_evaluator.rs
+++ b/golem-worker-service-base/src/evaluator/math_op_evaluator.rs
@@ -1,4 +1,4 @@
-use crate::evaluator::{EvaluationError, EvaluationResult};
+use crate::evaluator::{EvaluationError, ExprEvaluationResult};
 use crate::primitive::{GetPrimitive, Primitive};
 use golem_wasm_rpc::TypeAnnotatedValue;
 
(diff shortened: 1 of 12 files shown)


golem-worker-service-base/src/evaluator/math_op_evaluator.rs
```diff
@@ -1,4 +1,4 @@
-use crate::evaluator::{EvaluationError, EvaluationResult};
+use crate::evaluator::{EvaluationError, ExprEvaluationResult};
```
afsalthaj, line 1: I myself got confused of `EvaluationResult`, mainly on evaluation of `what`?
This is why I renamed this as part of a `fix`

justcoon (approved):

PR #584, opened 2024-06-12
Fix API gateway to follow the same function syntax that worker service expets

afsalthaj:
Please note, the sole purpose of this PR to get as minimal changes as possible into the API Gateway's Rib, and unblock dynamic-resource-allocation feature, and have the existing function calls tested with Rib to follow the new syntax.  

I gave it some thoughts if I could do it in the best possible way within a day of time, before deciding to make this particular PR (which is not the best possible way).  The thoughts are tangible, but very raw documentations here mainly for myself: https://www.notion.so/golemcloud/Rib-Supporting-New-Function-Syntax-ed92a3f1b92e4dd2a768402127f9aa7f. I would like to continue with this as a separate work

## More details
The only function syntax Rib tested ever was of the structure `namespace:package/interface/function-name(param1, param2)`, or simple `function-name(param1, param2)`.

I have made just enough changes to change all these structures in tests to follow the new syntax which is `namespace:package/interface.{function-name}(param1, param2)`

## Summary

* we will document in Rib that the only supported function calls are only of the form `namespace:package/interface/function-name(param1, param2)` and that syntax consistent with rest of the `worker-service`, `cli` etc.
* For supporting the newer syntax properly for all possible functions with Rib, with cleaner changes, I would like it to be a separate work as mentioned above

### DIFF
--- a/golem-common/src/model/function_name.rs
+++ b/golem-common/src/model/function_name.rs
@@ -379,9 +379,11 @@ impl ParsedFunctionName {
                     ParsedFunctionName { site, function }
                 }),
         )
-        .or(identifier().map(|id| ParsedFunctionName {
-            site: ParsedFunctionSite::Global,
-            function: ParsedFunctionReference::Function { function: id },
+        .or(identifier().map(|id| {
+            ParsedFunctionName {
+                site: ParsedFunctionSite::Global,
+                function: ParsedFunctionReference::Function { function: id },
+            }
         }));
 
         let result: Result<(ParsedFunctionName, &str), easy::ParseError<&str>> =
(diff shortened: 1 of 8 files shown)


vigoo:
Ticket for the proper integration: https://github.com/golemcloud/golem/issues/582

golem-common/src/model/function_name.rs
```diff
@@ -440,7 +442,7 @@ mod tests {
         let parsed = ParsedFunctionName::parse("ns:name/interface.{fn1}").expect("Parsing failed");
         assert_eq!(
             parsed.site().interface_name(),
-            Some("ns:name/interface".to_string())
+            Some("ns:name/interace".to_string())
```
afsalthaj, line 445: Oops

Issue #506, opened 2024-05-14
Run, document and review all the benchmarks


vigoo:
In this task we should:

- review that all the benchmarks are doing what they are intended to
- run all the created benchmarks on both local and real infrastructure
- create a notion page with the results
- if anything is obviously wrong (too slow) create follow-up tickets

PR #594, opened 2024-06-17
Rpc benchmark

### LINKED ISSUE #385: RPC benchmark
We also need an initial benchmark testing the RPC feature, so we can measure the need / effect of #176.

I propose the following implementation:

- reuse the three test functions from #383
- but add another component which calls the throughput-test component via RPC and invoke that from the benchmark
- the benchmark should measure separately (and also together - so 3 metrics) the local (both workers are on the same executor) and remote (workesr are on different executors) case.

To be able to distinguish the local/remote case the benchmark will need to call the shard manager (test framework already provides gRPC connection to it) to understand where each spawned worker runs.

The benchmark should be something like:

- start a cluster of size `size`
- start `length` number of worker **pairs** 
- ask shard manager about each created worker's shad's assignment
- warmup: call the tested function a few times - the test component should forward the request to the other worker and wait its result and return
- run: call the same function many times and measure the invocation time, and store it in three metrics 

The result should be 3 distinct benchmarks for the 3 different functions so we don't mix them in one run.

This task does not involve any test framework or infrastructure work just the creating the benchmark itself.

afsalthaj:
Fixes #385 

* Make use of the 3 same functions in benchmark throughput
* Use shard manager to distinguish between worker to worker communication is within the same executor or not
* Warm up the parent workers
* Emit available functions in the error messages when the function invoked isn't available (debugging easier)

image

#### As part of #506 (Benchmark Analysis)
I have started doing some analysis here, that I wanted to check how does it perform if I directly call the existing rust_component_service that exposes exactly these 3 functions with the same implementation. 

However, I need to spend more time on it, which is tracked as #506  - https://www.notion.so/golemcloud/Benchmark-8b84c6ad14f54e85a14bd53b7de01a5f

##### Direct invocation of these 3 functions
```
Results for ‘worker-calculate-invocation’:
Duration:
  Avg: 3.886206222s
  Min: 3.710567s
  Max: 4.069081s
Results for ‘worker-echo-invocation’:
Duration:
  Avg: 24.197111ms
  Min: 20.557ms
  Max: 27.858ms
Results for ‘worker-process-invocation’:
Duration:
  Avg: 604.594666ms
  Min: 574.528ms
  Max: 683.213ms
```

##### Indirect invocation of these 3 functions with another worker re-routing

```
Results for ‘worker-calculate-invocation-local’:
Duration:
  Avg: 27.445728833s
  Min: 27.152319s
  Max: 27.914916s
Results for ‘worker-calculate-invocation-remote’:
Duration:
  Avg: 27.012445333s
  Min: 26.725519s
  Max: 27.58572s
Results for ‘worker-echo-invocation-local’:
Duration:
  Avg: 555.948666ms
  Min: 26.505ms
  Max: 1.803546s
Results for ‘worker-echo-invocation-remote’:
Duration:
  Avg: 507.976ms
  Min: 52.044ms
  Max: 1.413298s
Results for ‘worker-process-invocation-local’:
Duration:
  Avg: 1.368930666s
  Min: 1.30921s
  Max: 1.491955s
Results for ‘worker-process-invocation-remote’:
Duration:
  Avg: 1.548745333s
  Min: 1.542453s
  Max: 1.558883s
```


test-components/rust-service/rpc/Cargo.toml
```diff
@@ -0,0 +1,3 @@
+[workspace]
```
vigoo, line 1: Is there a reason for putting this _inside_ of the already existing `rust-service` test component? I find it confusing as all the other test components are directly in subdirectories of `test-components`

integration-tests/src/benchmarks/rpc.rs
```diff
@@ -0,0 +1,329 @@
...
+
+        let shard_manager = benchmark_context.deps.shard_manager();
+
+        let mut shard_manager_client = shard_manager.client().await;
+
+        let routing_table = shard_manager_client
+            .get_routing_table(GetRoutingTableRequest {})
+            .await
+            .expect("Unable to fetch the routing table from shard-manager-service");
+
```
afsalthaj, line 183: I should have moved this to test DSL

afsalthaj:
Closing this in favour of https://github.com/golemcloud/golem/pull/596

PR #595, opened 2024-06-17
565 split cli

senia-psm:


golem-cli/src/cloud.rs
```diff
@@ -12,10 +12,47 @@
...
-pub mod service;
+use chrono::{DateTime, Utc};
+use derive_more::{Display, FromStr, Into};
+use serde::{Deserialize, Serialize};
+use std::fmt::{Debug, Display, Formatter};
+use uuid::Uuid;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CloudAuthenticationConfig {
```
justcoon, line 23: what is the reason, why those configuration remaining in OSS?

profiles?

senia-psm: Yes. It is for shared profiles. I can hide it with `serde_json::Value`, but there is no good reason for this.

justcoon (approved):

Issue #284, opened 2024-03-20
shard-manager - worker executors without shard assignments


justcoon:
after rescaling worker-executor from 1 to 3 nodes (OSS golem, AWS kubernetes)

2 nodes ended without shard assignment 

```
2024-03-20T09:20:16.462509Z  INFO golem_shard_manager::shard_management: Executing shard assignments: [deployment-worker-executor-test-54d596d859-2cs85: [<0>..<100>, <460>..<664>], deployment-worker-executor-test-54d596d859-bfnsp: [], deployment-worker-executor-test-54d596d859-f295r: []]
2024-03-20T09:20:16.462536Z  INFO golem_shard_manager::worker_executor: Assigning shards {ShardId { value: 0 }, ShardId { value: 1 }, ShardId { value: 2 }, ShardId { value: 3 },  ... }
2024-03-20T09:20:16.462687Z  INFO golem_shard_manager::worker_executor: Assigning shards {} to pod Pod { host: "deployment-worker-executor-test-54d596d859-bfnsp", ip: 10.0.103.14, port: 9000, pod_name: Some("deployment-worker-executor-test-54d596d859-bfnsp") }
2024-03-20T09:20:16.462741Z  INFO golem_shard_manager::worker_executor: Assigning shards {} to pod Pod { host: "deployment-worker-executor-test-54d596d859-f295r", ip: 10.0.103.209, port: 9000, pod_name: Some("deployment-worker-executor-test-54d596d859-f295r") }

```

justcoon:
[shard_manager_logs.txt](https://github.com/golemcloud/golem/files/14670350/shard_manager_logs.txt)

Issue #271, opened 2024-03-19
Sharding test fails with docker


senia-psm:
Run this several times:
```
QUIET=true RUST_BACKTRACE=1 GOLEM_DOCKER_SERVICES=true cargo test --test sharding -- --nocapture
```
Sometimes the last step gets stuck:

Shard manager can see the worker
```
2024-03-19T14:51:58.760150Z DEBUG golem_shard_manager: Shard Manager scheduled to conduct health check
2024-03-19T14:51:58.760185Z DEBUG golem_shard_manager: Shard Manager checking health of registered pods...
2024-03-19T14:51:58.760212Z DEBUG golem_shard_manager::worker_executor: Health checking pod f161c4e18c35:9003 (172.18.0.10)
2024-03-19T14:51:58.760222Z DEBUG golem_shard_manager::model: Pod.address: http://172.18.0.10:9003/ => Ok(Endpoint)
2024-03-19T14:51:58.760240Z DEBUG hyper::client::connect::http: connecting to 172.18.0.10:9003
2024-03-19T14:51:58.760383Z DEBUG hyper::client::connect::http: connected to 172.18.0.10:9003
2024-03-19T14:51:58.760490Z DEBUG tower::buffer::worker: service.ready=true processing request
2024-03-19T14:51:58.760838Z DEBUG golem_shard_manager: All registered pods are healthy
2024-03-19T14:51:58.760845Z DEBUG golem_shard_manager: Golem Shard Manager finished checking health of registered pods
```
But worker service can't process requests:
```
2024-03-19T14:56:59.815468Z  INFO request{remote_addr=172.18.0.1 version=HTTP/1.1 method=POST uri=/v2/templates/4f85e992-38e8-44f0-ae2c-5ea456542f70/workers/echo-service-2/key}: poem::middleware::tracing_mw: error status=500 Internal Server Error error=500 Internal Server Error duration=11.257389642s
2024-03-19T14:57:02.887168Z DEBUG request{remote_addr=172.18.0.1 version=HTTP/1.1 method=POST uri=/v2/templates/4f85e992-38e8-44f0-ae2c-5ea456542f70/workers/echo-service-3/key}: tonic::transport::service::reconnect: reconnect::poll_ready: hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 113, kind: HostUnreachable, message: "No route to host" }))
```

PR #606, opened 2024-06-24
Fix initial sharding balancing

noise64:
Related issues:
 - https://github.com/golemcloud/golem/issues/271
 - https://github.com/golemcloud/golem/issues/284

Contains:
- some fixes around initial balancing (which was not fair and triggered rebalancing)
- a bit of cleanups for rebalancing tests
- some debug info for sharding tests.
Rechecking and reenabling of sharding tests will come in a followup PR.


integration-tests/tests/sharding.rs
```diff
@@ -278,22 +280,27 @@ async fn invoke_and_await_workers(workers: &[WorkerId]) -> Result<(), worker::wo
...
-                &worker_id_clone,
-                &idempotency_key,
-                "golem:it/api/echo",
-                vec![Value::Option(Some(Box::new(Value::String(
-                    "Hello".to_string(),
-                ))))],
-            )
-            .await
-        }));
+        tasks.push((
```
vigoo, line 283: Fyi: there is a nicer way to spawn and await many subtasks, called `JoinSet` - you `spawn` on the joinset itself and then do a `join_next` loop on it. I started using that in all the benchmarks.

vigoo (approved):

integration-tests/tests/sharding.rs
noise64: thanks! i will change it in the next PR

PR #617, opened 2024-06-25
Parser combinator for Rib and function_name parser reuse

### LINKED ISSUE #582: Support the new function invocation syntax and features in Rib

afsalthaj:
Fix #582 

* Complete revamp of parser logic, avoiding lexer and parser, using `combine` library
* Careful logic for certain parts such as select field/index, binary comparisons etc, to avoid stack overflow.
https://github.com/golemcloud/golem/blob/parser_comb/golem-rib/src/parser/select_field.rs#L71
https://github.com/golemcloud/golem/blob/parser_comb/golem-rib/src/parser/rib_expr.rs#L44
* Designed and written tests from the scratch - this is not a copy paste of old test cases. Parsing itself has more test coverage, and they are about 150 of them.
* golem-rib is kept independent and defined Rib Expr, ParsedFunctionName etc, its parsing, and write-back functionaltieis. However evaluator of Expr is not moved yet (I tried, and lost time there) since `EvaluationContext` has dependency to golem-service-base such as `ComponentMetadata`.

* Also note that I fixed the types with life time parameters to make things easier at call-site and some decisions are taken to get things compiling. It was a bit hard since I am using lot of parser! macros to get recursion going. The API docs of 'easy' encourages use of 'Input' type parameterisation instead of concrete &str but this is actually impossible if we need to write custom parser where we help inference with a Stream<&str>. Any other possibilities are super long types and it's definitely hard to work with in future. 

Out of all the tests, the below test sort of cover the function call feature:

https://github.com/golemcloud/golem/blob/parser_comb/golem-rib/src/parser/rib_expr.rs#L141

## TODO

- [x] Round trip tests has some more nuances after moving them to rib module. Just fixing them
- [x]  A total of 220 (Rib parser) + 150 (Rib evaluator) tests works

PS: Reviewers: There are a few large diffs


golem-rib/src/parser/call.rs
```diff
@@ -0,0 +1,1030 @@
...
+    let interface = many1(identifier()).message("interface");
+
+    let capture_resource_params = || {
+        parser(|input| {
+            let _: &mut easy::Stream<&str> = input;
+            let mut nesting = 1;
+            let mut current_param = String::new();
+            let mut result = Vec::new();
+            let mut result_committed: Option<Commit<()>> = None;
+            while nesting > 0 {
```
afsalthaj, line 51: I think this termination condition is wrong as matching brace may be missing forever and this never ends. I will see to it.

vigoo:
You will have to add version-override / publishing things into the `Makefile.toml` to make `cargo make publish` work

vigoo (changes requested):
Looks nice, just please set up the publishing in cargo-make

vigoo (approved):

Issue #173, opened 2024-03-17
Review logging in all services


vigoo:

noise64:
remaining work split into:
  - https://github.com/golemcloud/golem/issues/661
  - https://github.com/golemcloud/golem/issues/662
  - https://github.com/golemcloud/golem/issues/663

PR #620, opened 2024-06-26
Extract common metrics and logging

noise64:
Part of https://github.com/golemcloud/golem/issues/173

Done:
- extracted common grpc tracing and metrics from worker-executor to golem-common
- added grpc tracing and metrics to worker service
- switch to structured logging in
  - golem-common
  - worker-service (except openapi)
- added spans to golem-common retry and worker-service call_executor
- extracted common log and config handling
  - logging now can be configured for stdout and file output separately
  - added config loader:
    - configs now derived from the config struct defaults; added missing defaults and aligned them with the current toml local defaults
    - added common command line options to dump the default (and custom examples for covering enum config cases) and the loaded configs as toml or env vars
    - to avoid drifting code defaults and toml values (and to cleanup obsolete ones) added cargo make command to generate configs and sample env vars
 - added a new json formatter to flatten spans in logs:
   - spans are merged into the main message json object with the span name
   - in case of repeated spans only the last span is included
 - logging env var handling is extracted to common place for the local test env lib and now tracing env vars are passed through to the local golem services (e.g. one can enable json logging to files while running integ tests)

Todo in folllow up PR:
 - use structured logs everywhere
 - use grpc tracing everywhere
 - add OpenAPI tracing like we have for grpc


Makefile.toml
```diff
@@ -532,3 +540,22 @@ kill $shard_manager_pid || true
 kill $router_pid || true
 kill $redis_pid || true
 '''
+
+[tasks.generate-configs]
```
vigoo, line 544: As this is a manual step that we will forget to run, we could add a check to CI similar to how we check that the generated OpenAPI yaml is up-to-date.

golem-worker-executor-base/src/storage/blob/s3.rs
```diff
@@ -283,11 +282,11 @@ impl BlobStorage for S3BlobStorage {
     ) -> Result<Option<Bytes>, String> {
         let bucket = self.bucket_of(&namespace);
         let key = self.prefix_of(&namespace).join(path);
-        let description = format!("Downloading blob from {bucket}::{key:?}");
+
         let result = with_retries(
-            &description,
             target_label,
             op_label,
+            Some(format!("{} - {:?}", bucket, key)),
```
vigoo, line 289: fyi: these format string can be inlined which in my opinion is more readable (but probably it is a matter of taste :) (like `format!("{bucket} - {key:?}")`)

vigoo:
I like the changes! :)

Makefile.toml
noise64: https://github.com/golemcloud/golem/pull/620/commits/18093028d40a4bdf4e740ee1877e09a4e856eb99

golem-worker-executor-base/src/storage/blob/s3.rs
noise64: https://github.com/golemcloud/golem/pull/620/commits/d2a9f1e46a432d0293aa4cbaf123d840b9278d76

vigoo (approved):

vigoo (approved):

PR #625, opened 2024-06-27
Set ulimit for AWS mode

afsalthaj:


golem-test-framework/src/components/component_compilation_service/k8s.rs
```diff
@@ -90,6 +90,13 @@ impl K8sComponentCompilationService {
                 },
             },
             "spec": {
+                 "initContainers": [
+                    {
+                        "name": "init",
+                        "image": "busybox",
+                        "command": ["sh", "-c", "ulimit -n 1000000 && exec sleep 1"]
```
vigoo, line 97: Do we have to do this in our prod pods?

afsalthaj: I tried various other ways to set the ulimit such as being part of the Docker Cmd section, but that didn't work as far as I see.

This approach ensures we don't need to worry about changing code / image executor and services.

PR #629, opened 2024-06-29
Component metadata and ApiDefinition data serde

justcoon:
related to: #498 

- added namespace to component, metadata using proto for serialization, with version
- repository and service moved to `golem-component-service-base`
- fixed naming for `FunctionParameter`, `FunctionResult` - `tpe` -> `typ`
- api definition serialization changed from bincode to proto, proto for golem-rib


golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,467 @@
...
+        Self { db_pool }
+    }
+}
+
+#[async_trait]
+impl ComponentRepo for DbComponentRepo<sqlx::Sqlite> {
+    async fn upsert(&self, component: &ComponentRecord) -> Result<(), RepoError> {
+        let mut transaction = self.db_pool.begin().await?;
+
+        let result = sqlx::query(
```
senia-psm, line 148: This is not the method result. Please consider renaming it to something like `same_id` or inlining it.

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,467 @@
...
+            "SELECT count(component_id) as count FROM components WHERE namespace != $1 AND component_id = $2",
```
senia-psm, line 149: Can we just add a unique index on `component_id`?

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,467 @@
...
+
+        transaction.commit().await?;
+
+        Ok(())
+    }
+
+    async fn get(
+        &self,
+        namespace: &str,
+        component_id: &Uuid,
```
senia-psm, line 197: If component id should be unique - why would we need a namespace for `get`?

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,467 @@
...
+        )
+            .bind(namespace)
+            .bind(name)
+            .fetch_all(self.db_pool.deref())
+            .await?;
+
+        Ok(result.into_iter().map(|x| x.get("component_id")).collect())
+    }
+
+    async fn get_namespaces(&self, component_id: &Uuid) -> Result<Vec<(String, i64)>, RepoError> {
```
senia-psm, line 273: What is the purpose of namespace in the result? According to the `upsert` implementation multiple namespaces is an invalid state.

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,467 @@
...
+        sqlx::query_as::<_, (String, i64)>(
+            "SELECT namespace, max(version) FROM components WHERE component_id = $1 GROUP BY namespace",
+        )
+            .bind(component_id)
+            .fetch_all(self.db_pool.deref())
+            .await
+            .map_err(|e| e.into())
+    }
+
+    async fn delete(&self, namespace: &str, component_id: &Uuid) -> Result<(), RepoError> {
```
senia-psm, line 283: Same as for `get` - there should not be `namespace` in parameters.

senia-psm (changes requested):

golem-component-service-base/tests/db/migration/sqlite/001__init.sql
```diff
@@ -0,0 +1,14 @@
...
+    component_id        uuid    NOT NULL,
+    name                text    NOT NULL,
+    size                integer NOT NULL,
+    version             bigint  NOT NULL,
+    created_at          timestamp without time zone DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    user_component      text    NOT NULL,
+    protected_component text    NOT NULL,
+    protector_version   bigint,
+    metadata            blob    NOT NULL,
+    PRIMARY KEY (namespace, component_id, version)
```
senia-psm, line 13: Consider normalising the schema - create 2 tables: `components` with `component_id` primary key and `component_versions` with `(component_id, version)` primary key and a `component_id` foreign key. `namespace` and name should be in `components`.

Current implementation allows different names for different versions. It also requires backend check for unique `namespace`.

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,467 @@
...
+        }
+
+        sqlx::query(
+            r#"
+              INSERT INTO components
+                (namespace, component_id, version, name, size, user_component, protected_component, protector_version, metadata)
+              VALUES
+                ($1, $2, $3, $4, $5, $6, $7, $8, $9)
+              ON CONFLICT (namespace, component_id, version) DO UPDATE
+              SET name = $4,
```
senia-psm, line 169: This implementation allows different names for different versions on the same `component_id`. The only solution without races is to normalize the schema.

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,467 @@
...
+            "SELECT count(component_id) as count FROM components WHERE namespace != $1 AND component_id = $2",
+        )
+        .bind(component.namespace.clone())
+        .bind(component.component_id)
+        .fetch_one(&mut *transaction)
+        .await?;
+
+        let count: i64 = result.get("count");
+
+        if count > 0 {
```
senia-psm, line 158: There is a race here - you can create 2 different versions in different namespaces if you have 2 `upsert` running at the same same. The probability is low, but the race is there. The only solution without race is to normalise the schema.

golem-component-service-base/tests/db/migration/sqlite/001__init.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE components
+(
+    namespace           text    NOT NULL,
+    component_id        uuid    NOT NULL,
+    name                text    NOT NULL,
```
senia-psm, line 5: Please consider adding a unique index on `(name, namespace)` - in this order. To bi able to use it to search by `name` only.

golem-component-service-base/tests/db/migration/sqlite/001__init.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE components
+(
+    namespace           text    NOT NULL,
```
senia-psm, line 3: Please consider adding a non-unique index on `namespace`  for `get_all` method.

golem-component-service/db/migration/postgres/001__init.sql
```diff
@@ -1,13 +1,23 @@
 CREATE TABLE components
 (
-    component_id        uuid    NOT NULL,
+    namespace           text    NOT NULL,
+    component_id        uuid    NOT NULL PRIMARY KEY,
```
senia-psm, line 4: It is common to keep primary key field on top.

golem-component-service/db/migration/postgres/001__init.sql
```diff
@@ -1,13 +1,23 @@
...
     name                text    NOT NULL,
-    size                integer NOT NULL,
+    UNIQUE (namespace, name)
+);
+
+CREATE INDEX components_namespace_id_idx ON components (namespace, component_id);
```
senia-psm, line 9: I believe it should be `(namespace, name)`. Id is unique already

golem-component-service/db/migration/postgres/001__init.sql
```diff
@@ -1,13 +1,23 @@
...
+    size                integer NOT NULL,
     created_at          timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
     user_component      text    NOT NULL,
     protected_component text    NOT NULL,
     protector_version   bigint,
-    metadata            jsonb   NOT NULL,
-    PRIMARY KEY (component_id, version)
+    metadata            bytea   NOT NULL,
+    PRIMARY KEY (component_id, version),
+    FOREIGN KEY (component_id) REFERENCES components (component_id)
```
senia-psm, line 22: You can inline it with column declaration this way:
```sql
component_id        uuid    NOT NULL references components (component_id)
```

golem-component-service/db/migration/sqlite/001__init.sql
```diff
@@ -1,13 +1,24 @@
...
 (
-    component_id        uuid    NOT NULL,
+    namespace           text    NOT NULL,
+    component_id        uuid    NOT NULL PRIMARY KEY,
     name                text    NOT NULL,
-    size                integer NOT NULL,
+    UNIQUE (namespace, name)
+);
+
+CREATE INDEX components_namespace_id_idx ON components (namespace, component_id);
```
senia-psm, line 9: Same here - it should be `namespace, name`

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,510 @@
...
```
senia-psm, line 330: Impossible with `component_id` as primary key

golem-component-service-base/tests/db/migration/postgres/001__init.sql
```diff
@@ -0,0 +1,23 @@
+CREATE TABLE components
+(
+    namespace           text    NOT NULL,
+    component_id        uuid    NOT NULL PRIMARY KEY,
+    name                text    NOT NULL,
+    UNIQUE (namespace, name)
+);
+
+CREATE INDEX components_namespace_id_idx ON components (namespace, component_id);
```
senia-psm, line 9: This index is useless. For search by `namespace` there is already `UNIQUE (namespace, name)`. For search by ID - PK

golem-component-service-base/tests/db/migration/postgres/001__init.sql
```diff
@@ -0,0 +1,23 @@
+CREATE TABLE components
+(
+    namespace           text    NOT NULL,
+    component_id        uuid    NOT NULL PRIMARY KEY,
+    name                text    NOT NULL,
+    UNIQUE (namespace, name)
```
senia-psm, line 6: This index can't be used to search by name without namespace. If we have such operation - please add non-unique index by name

golem-component-service-base/tests/db/migration/postgres/001__init.sql
```diff
@@ -0,0 +1,23 @@
+CREATE TABLE components
```
senia-psm, line 1: Why do we have dedicated migrations for tests? I believe there should be only 1 set of migrations

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,510 @@
...
+            return Err(RepoError::Internal("Component id not unique".to_string()));
+        }
+
+        sqlx::query(
+            r#"
+              INSERT INTO components
+                (namespace, component_id, name)
+              VALUES
+                ($1, $2, $3)
+              ON CONFLICT (namespace, name) DO NOTHING
```
senia-psm, line 168: What about `ON CONFLICT (component_id)`?

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,510 @@
...
+        .execute(&mut *transaction)
+        .await?;
+
+        sqlx::query(
+            r#"
+              INSERT INTO component_versions
+                (component_id, version, size, user_component, protected_component, protector_version, metadata)
+              VALUES
+                ($1, $2, $3, $4, $5, $6, $7)
+              ON CONFLICT (component_id, version) DO UPDATE
```
senia-psm, line 183: Why do we have a version update at all? Should we keep it immutable?
What is the meaning of updating the size? We can't just replace component version code.

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,510 @@
...
+        Ok(())
+    }
+
+    async fn get(
+        &self,
+        namespace: &str,
+        component_id: &Uuid,
+    ) -> Result<Vec<ComponentRecord>, RepoError> {
+        sqlx::query_as::<_, ComponentRecord>(
+            "SELECT components.namespace, components.name, component_versions.* FROM components JOIN component_versions ON components.component_id = component_versions.component_id WHERE components.namespace = $1 AND component_versions.component_id = $2"
```
senia-psm, line 212: Could you please format long sql queries?

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,510 @@
...
+        ).bind(namespace)
+            .bind(component_id)
+            .fetch_optional(self.db_pool.deref())
+            .await
+            .map_err(|e| e.into())
+    }
+
+    async fn get_by_version(
+        &self,
+        namespace: &str,
```
senia-psm, line 245: namespace is never needed if you have `component_id` for query

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,510 @@
...
+            .map_err(|e| e.into())
+    }
+
+    async fn get_by_name(
+        &self,
+        namespace: &str,
+        name: &str,
+    ) -> Result<Vec<ComponentRecord>, RepoError> {
+        sqlx::query_as::<_, ComponentRecord>(
+            "SELECT components.namespace, components.name, component_versions.* FROM components JOIN component_versions ON components.component_id = component_versions.component_id WHERE components.namespace = $1 AND components.name = $2",
```
senia-psm, line 266: Please avoid using `*` in select. It is unreadable and the query will break after `ALTER TABLE` - you'll get a different columns order.
Please use explicit columns order instead.

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,510 @@
...
+        )
+            .bind(namespace)
+            .bind(name)
+            .fetch_all(self.db_pool.deref())
+            .await
+            .map_err(|e| e.into())
+    }
+
+    async fn get_ids_by_name(&self, namespace: &str, name: &str) -> Result<Vec<Uuid>, RepoError> {
```
senia-psm, line 275: It is not `Vec`  o it should be `Option` - `(namespace, name)` is a unique index

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,510 @@
...
+        )
+            .bind(namespace)
+            .bind(name)
+            .fetch_all(self.db_pool.deref())
+            .await?;
+
+        Ok(result.into_iter().map(|x| x.get("component_id")).collect())
+    }
+
+    async fn get_namespaces(&self, component_id: &Uuid) -> Result<Vec<String>, RepoError> {
```
senia-psm, line 287: It is not `Vec` - it should be `Option` - id is PK

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,510 @@
...
+        sqlx::query_as::<_, ComponentRecord>("SELECT components.namespace, components.name, component_versions.* FROM components JOIN component_versions ON components.component_id = component_versions.component_id WHERE components.namespace = $1")
+            .bind(namespace)
+            .fetch_all(self.db_pool.deref())
+            .await
+            .map_err(|e| e.into())
+    }
+
+    async fn get_latest_version(
+        &self,
+        namespace: &str,
```
senia-psm, line 231: namespace is not needed here

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,510 @@
...
+            .fetch_optional(self.db_pool.deref())
+            .await
+            .map_err(|e| e.into())
+    }
+
+    async fn get_by_name(
+        &self,
+        namespace: &str,
+        name: &str,
+    ) -> Result<Vec<ComponentRecord>, RepoError> {
```
senia-psm, line 264: It's not a `Vec` - it should be `Option`

golem-component-service-base/src/repo/component.rs
justcoon: fixed

golem-component-service-base/src/repo/component.rs
justcoon: thanks, fixed

golem-component-service/db/migration/postgres/001__init.sql
justcoon: thanks, fixed

golem-component-service/db/migration/postgres/001__init.sql
justcoon: moved to top

golem-component-service-base/src/repo/component.rs
justcoon: fixed

golem-component-service-base/src/repo/component.rs
justcoon: queries changed

golem-component-service-base/src/repo/component.rs
justcoon: it was upsert before, but I think you have good point, it should be just insert ..., i changed it to insert

golem-component-service-base/src/repo/component.rs
justcoon: formated

golem-component-service-base/src/repo/component.rs
justcoon: changed

golem-component-service-base/src/repo/component.rs
justcoon: changed

golem-component-service-base/src/repo/component.rs
justcoon: changed to insert only

golem-component-service-base/src/repo/component.rs
```diff
@@ -142,50 +142,41 @@ impl<DB: Database> DbComponentRepo<DB> {
...
+            .bind(component.component_id)
+            .fetch_optional(&mut *transaction)
+            .await?;
 
-        if count > 0 {
-            return Err(RepoError::Internal("Component id not unique".to_string()));
+        if let Some(result) = result {
+            let namespace: String = result.get("namespace");
+            if namespace != component.namespace {
+                return Err(RepoError::Internal("Component id not unique".to_string()));
```
senia-psm, line 156: This is not an error - you are creating a component version. Component might already exist.

You can (if you want to) compare `name` and `namespace` here.

I would suggest to use `get` here (without namespace - you want to get it from any namespace to compare all the fields).

golem-component-service-base/src/repo/component.rs
```diff
@@ -209,7 +200,14 @@ impl ComponentRepo for DbComponentRepo<sqlx::Sqlite> {
         component_id: &Uuid,
     ) -> Result<Vec<ComponentRecord>, RepoError> {
         sqlx::query_as::<_, ComponentRecord>(
-            "SELECT components.namespace, components.name, component_versions.* FROM components JOIN component_versions ON components.component_id = component_versions.component_id WHERE components.namespace = $1 AND component_versions.component_id = $2"
+            r#"
+                SELECT c.namespace AS namespace, c.name AS name, c.component_id AS component_id,
```
senia-psm, line 204: You don't need `t.name as name` - just `t.name` is enough.

Optional: I would suggest this code style:
```sql
select
    c.namespace,
    c.name,
    c.component_id,
    cv.version,
    cv.size,
    cv.user_component,
    cv.protected_component,
    cv.protector_version,
    cv.metadata
from components c
join component_versions cv on c.component_id = cv.component_id
where c.namespace = $1
  and c.component_id = $2
```
i.e. each column on a separate line (just like in a long case class creation), each join on a separate line, each `and` on a separate line.

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,648 @@
...
+    pub fn serialize<T: Encode>(routes: &T) -> Result<Bytes, String> {
+        serialize_with_version(routes, SERIALIZATION_VERSION_V1)
+    }
+
+    pub fn deserialize<T: Decode>(bytes: &[u8]) -> Result<T, String> {
+        let (version, data) = bytes.split_at(1);
+
+        match version[0] {
+            SERIALIZATION_VERSION_V1 => {
+                let (routes, _) = bincode::decode_from_slice(data, bincode::config::standard())
```
afsalthaj, line 640: This is not routes

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,648 @@
...
+    pub const SERIALIZATION_VERSION_V1: u8 = 1u8;
+
+    pub fn serialize<T: Encode>(routes: &T) -> Result<Bytes, String> {
+        serialize_with_version(routes, SERIALIZATION_VERSION_V1)
+    }
+
+    pub fn deserialize<T: Decode>(bytes: &[u8]) -> Result<T, String> {
+        let (version, data) = bytes.split_at(1);
+
+        match version[0] {
```
afsalthaj, line 638: I wouldn't do unsafe lookups

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,648 @@
...
+}
+
+pub mod record_metadata_serde {
+    use bincode::{Decode, Encode};
+    use bytes::Bytes;
+    use golem_common::serialization::serialize_with_version;
+    pub const SERIALIZATION_VERSION_V1: u8 = 1u8;
+
+    pub fn serialize<T: Encode>(routes: &T) -> Result<Bytes, String> {
+        serialize_with_version(routes, SERIALIZATION_VERSION_V1)
```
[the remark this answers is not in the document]
afsalthaj: Or should ComponentRecord (the db specific data) has ComponentMetadata with mostly optional fields and use annotations at that level?

golem-component-service-base/src/repo/component.rs
afsalthaj: Not sure if I understood the problem here :)

afsalthaj: Ok, isn't it handled by conflicts with in the sql data table? If not, let's revisit there so that you don't need to explicitly check if a component id exist in some other namespace?

golem-component-service-base/src/repo/component.rs
```diff
@@ -0,0 +1,648 @@
...
+        if let Some(result) = result {
+            let namespace: String = result.get("namespace");
+            if namespace != component.namespace {
+                return Err(RepoError::Internal("Component id not unique".to_string()));
+            }
+        } else {
+            sqlx::query(
+                r#"
+                  INSERT INTO components
+                    (namespace, component_id, name)
```
afsalthaj, line 162: if namespace and component_id are part of key, then you can avoid extra checks probably

golem-component-service-base/src/repo/component.rs
afsalthaj: I think data model should reflect access concerns. I kind of get your question though and its valid, but my opinion here is , I think Peter is right here with having namespace and it clearly reflects what's going on with the ownership and access.

golem-component-service-base/src/repo/component.rs
afsalthaj: May be namespace should be a key as well even if technically component_id is unique.

golem-component-service-base/src/repo/component.rs
senia-psm: Yes, ownership is quite important, but I believe we should keep all the authorization-related code in the service.
We might not even know the namespace on `get` - with the planned `URN` schema it should be `component_id` only. User can access component by id in other user account if he has such permissions. And the user should not provide project id for this.

golem-component-service-base/src/repo/component.rs
senia-psm: Component id is unique already (PK), so it can not exist in some other namespace.

If component id already exists in any namespace - the insert will fail with an error (there is no `ON CONFLICT` any more). So this check is excessive.

senia-psm: We should change this check - we should not check the count. We should select the actual component record and validate that the namespace and the name are the same as expected.

Because if we are trying to create the same component_id in the same namespace, but with the different name - it should also be treated as an error.

golem-component-service-base/src/repo/component.rs
senia-psm: @afsalthaj there is already a unique index on `(namespace, name). And `component_id` is a PK.

We are creating a new version here (in `component_version`). And for each new version we are creating `component` record first. It might exists already (if there are other versions) or not (if we are creating the first versions) - both options are not an error.

The only problem here is if we are trying to create a new version for some component id with different namespace or name comparing to the previous version.

golem-component-service-base/src/repo/component.rs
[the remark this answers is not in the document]
justcoon: serialization changed to proto

golem-component-service-base/src/repo/component.rs
justcoon: this was just typo, serialization changed to proto

golem-component-service-base/src/repo/component.rs
justcoon: changed to  return just optional value

golem-component-service-base/src/repo/component.rs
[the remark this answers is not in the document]
afsalthaj: Thanks 👍

afsalthaj (approved):

afsalthaj:
conflicts @justcoon

afsalthaj:
How exactly Expr is serde'd here?

golem-component-service-base/src/repo/component.rs
justcoon: changed to proto

golem-component-service-base/src/repo/component.rs
justcoon: changed

golem-component-service-base/tests/db/migration/postgres/001__init.sql
justcoon: tests are in base module, maybe we can have migration definition there, and then somehow copy them to service module

senia-psm: Alternatively you can explicitly specify path to migrations in the service module. It is ugly, but should work.

golem-component-service-base/tests/services_tests.rs
```diff
@@ -0,0 +1,453 @@
...
+            std::fs::remove_file(&self.db_path).unwrap();
+        }
+    }
+
+    #[tokio::test]
+    pub async fn test_with_postgres_db() {
+        let cli = Cli::default();
+        let (db_config, _container) = start_docker_postgres(&cli);
+
+        db::postgres_migrate(&db_config, "tests/db/migration/postgres")
```
senia-psm, line 64: You can specify path to real migrations here as `../golem-component-service/...`.

senia-psm (approved):
Unused index and no name validation on create. Otherwise perfect.

PR #631, opened 2024-07-01
config workaround

senia-psm:


golem-shard-manager/src/shard_manager_config.rs
```diff
@@ -61,11 +61,12 @@ pub struct HealthCheckConfig {
 #[derive(Clone, Debug, Deserialize)]
 #[serde(tag = "type", content = "config")]
 pub enum HealthCheckMode {
-    Grpc,
-    #[cfg(feature = "kubernetes")]
-    K8s {
-        namespace: String,
+    Grpc {
+        // Useless. Just to make config lib happy.
+        _dummy: Option<u8>,
```
justcoon, line 66: maybe better is to do `Grpc(golem_service_base::model::Empty)`, like here

https://github.com/golemcloud/golem/blob/main/golem-component-service-base/src/config.rs#L22

senia-psm: In this case I'll have to replace `K8s { namespace: String }` with `K8s(K8sMode)`. I'm not sure it worth the refactoring.
WDYT?

justcoon: not, sure, maybe we can do later ...

justcoon (approved):

golem-shard-manager/src/shard_manager_config.rs
senia-psm: Done. It was not that bad as I expected.

vigoo (approved):

PR #653, opened 2024-07-08
Tracking and enforcing worker memory limits

### LINKED ISSUE #613: Implement tracking and protecting available worker memory

vigoo:
Resolves #613

### DIFF
--- a/golem-worker-executor-base/src/worker.rs
+++ b/golem-worker-executor-base/src/worker.rs
@@ -23,7 +23,7 @@ use golem_wasm_rpc::Value;
 use tokio::sync::broadcast::error::RecvError;
 use tokio::sync::broadcast::Receiver;
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
-use tokio::sync::Mutex;
+use tokio::sync::{Mutex, MutexGuard, OwnedSemaphorePermit};
 use tokio::task::JoinHandle;
 use tracing::{debug, info, span, warn, Instrument, Level};
 use wasmtime::{Store, UpdateDeadline};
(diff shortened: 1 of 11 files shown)


golem-worker-executor-base/src/worker.rs
```diff
@@ -355,6 +352,31 @@ impl<Ctx: WorkerCtx> Worker<Ctx> {
...
+            Some(running) => {
+                if is_running_worker_idle(running) {
+                    if self.stopping.compare_exchange(
+                        false,
+                        true,
+                        Ordering::Acquire,
+                        Ordering::Relaxed,
+                    ) == Ok(false)
+                    {
+                        self.stop_internal_running(running_guard, false).await;
```
noise64, line 367: probably works, just interested: what is the for no-race-condition guarantee between
- is_running_worker_idle
- stopping mutex
- stop_internal_running?

and is there no way that the stop can fail? as in stopping a non-idle one

golem-worker-executor-base/tests/common/mod.rs
```diff
@@ -305,6 +305,11 @@ pub async fn start(context: &TestContext) -> anyhow::Result<TestWorkerExecutor>
             port: context.grpc_port(),
             access_token: "03494299-B515-4427-8C37-4C1C915679B7".to_string(),
         },
+        // memory: MemoryConfig {
```
noise64, line 308: is this needed or todo?

noise64 (approved):
added some question, but in general looks goods

golem-worker-executor-base/src/worker.rs
vigoo: I'm going to document this better in the next PR

golem-worker-executor-base/tests/common/mod.rs
vigoo: It makes it possible to test limited memory scenarios I left it there intentionally for the next step. In that I'm planning to have dedicated test cases doing this and remove the commented out lines here

PR #654, opened 2024-07-08
Fix benchmark report for comparison

afsalthaj:


integration-tests/src/benchmarks/report/benchmark_report.rs
```diff
@@ -276,13 +281,19 @@ mod internal {
...
     }
 
-    pub fn wrap_with_title(main_title: &str, content: &String) -> String {
+    pub(crate) fn wrap_with_title(main_title: &str, content: &String) -> String {
         format!(r#"\n# {}\n{}\n"#, main_title, content)
     }
+
+    pub(crate) fn display_optional_duration(duration: Option<Duration>) -> String {
+        duration.map_or("Unavailable".to_string(), |duration| {
+            format!("{:?}", duration)
```
afsalthaj, line 294: Unavailable shouldn't be in `debug` mode as it escapes string quotes, and that's the fix.
Before this, unavailable was printed as `\"unavailable\"` which conflicted within a json text

vigoo (approved):

PR #658, opened 2024-07-09
cli tests on release

senia-psm:


vigoo:
Note that CI is supposed to reuse the cached debug build in the CLI test step:
https://github.com/golemcloud/golem/blob/main/.github/workflows/ci.yaml#L403

So this change alone would break the current CI caching

PR #659, opened 2024-07-11
golem-host - get worker metadata

justcoon:
related to: https://github.com/golemcloud/golem/issues/497


golem-worker-executor-base/src/durable_host/golem/mod.rs
```diff
@@ -462,6 +463,26 @@ impl<Ctx: WorkerCtx> golem::api::host::Host for DurableWorkerCtx<Ctx> {
 
         Ok(())
     }
+
+    async fn get_self_metadata(&mut self) -> anyhow::Result<golem::api::host::WorkerMetadata> {
+        record_host_function_call("golem::api", "get_self_metadata");
+        let metadata = self.state.worker_service.get(&self.owned_worker_id).await;
```
vigoo, line 469: This should be something like `self.public_state.worker().get_metadata()` which will return the latest state in-memory instead of querying the last persisted one from redis

golem-worker-executor-base/src/durable_host/golem/mod.rs
```diff
@@ -462,6 +463,26 @@ impl<Ctx: WorkerCtx> golem::api::host::Host for DurableWorkerCtx<Ctx> {
...
+
+    async fn get_worker_metadata(
+        &mut self,
+        worker_id: golem::api::host::WorkerId,
+    ) -> anyhow::Result<Option<golem::api::host::WorkerMetadata>> {
+        record_host_function_call("golem::api", "get_worker_metadata");
+        let worker_id: WorkerId = worker_id.into();
+        let owned_worker_id = OwnedWorkerId::new(&self.owned_worker_id.account_id, &worker_id);
+        let metadata = self.state.worker_service.get(&owned_worker_id).await;
+        Ok(metadata.map(|m| m.into()))
```
vigoo, line 484: It should behave the same way as the gRPC get worker metadata endpoint, so the metadata here has to be actualized by doing 

```rust
        let latest_status =
            Ctx::compute_latest_worker_status(self, &owned_worker_id, &Some(metadata.clone()))
                .await?;
```
and then updating `metadata.last_known_status` before returning it

vigoo (approved):

PR #660, opened 2024-07-15
Renewed function invocation workflow to avoid metadata downloads 

afsalthaj:
# Optimizations around Component Metadata Downloads

`worker-service` mainly consist of two components. A `service` which acts as the entry point to invoke worker functions and `worker-gateway` . The `service` is mainly used by `golem-cli` as well as `worker-gateway`.  `worker-service` also consist of `grpc` service which is used by `worker-executor` mainly.

Any worker invocation (be it gateway call, or direct invocation through CLI, or the call back from worker-proxy in worker-executor), all of them led to a metadata download which in turn calls component-service. This document and PR describes how we avoided this, without compromising some of the type-safety that we have in golem-rib evaluation, and hopefully increase performance (as we reduce the number of RPC calls to component-service)

*Note that, this optimization is not just done for worker-gateway (or worker-bridge).*

### New invoke functions and Optimisations
* New invoke functions that takes precise-json (as a vector) of input, which gets converted to `wasm_rpc::Val` internally before calling worker-executor
* The internals of worker-executor now stores `TypeAnnotatedValue` so that there isn't a need to download the metadata at client site to interpret `Vec<wasm_rpc::Val>`
* In short, the two invoke-and-await functions `Vec<Val>` as input as well as `Vec<PreciseJson>`, and one returning `Vec<Val>` as output and the other a `TypeAnnotatedValue` which can be converted to JSON
* The golem-cli wasm wave syntax supports still downloads the metadata at cli code side. It was confirmed with @vigoo  we are not changing this
* As part of the PR, the drop functionality which was storing `empty` array was failing when converted to TypeAnnotatedValue. This is fixed in a separate PR. Take a look at this one: https://github.com/golemcloud/golem/pull/718
* As a summary, all interpretations are moved to worker-executor and client sites are light weight now.
* WorkerBridge is broken now until we fix https://app.zenhub.com/workspaces/golem-cloud-65f5b68988aba80f7d690199/issues/gh/golemcloud/golem/715
* Removed Metadata Download when Interpreting Rib in Worker Gateway

When evaluating Rib, we used to download the metdata for mainly two purposes.

- To evaluate whether an identifier that looks like a function call is really a function that’s part of the compoent-metadata, or is it constructing a Variant etc.
- To identify whether an empty `{}` could be an empty flag or an empty record etc.
- However we removed these downloads in Rib, and Rib is pushing this as it is to worker-service functionalities (mentioned above).

### Function Name Parsing - `ParsedFunctionName` and `Display` instance

- when evaluating a Rib expression, we already parse the function name as `ParsedFunctionName`. However the functions exposed by worker-service as well as GRPC interface of worker-executor, all work with function name as `String`. To make lesser changes in this PR, I have introduced a display instane for ParsedFunctionName, and worker-bridge simply reuses the existing functions without introducing a new function that accepts `ParsedFunctionName`.
- What this could imply is, we will have 2 times of parsing to form `ParsedFunctionName`, one during Rib evaluation, and one during execution of the worker function by worker-executor.
- All other parsing that was done in worker-service are removed.


golem-api-grpc/proto/golem/workerexecutor/worker_executor.proto
```diff
@@ -97,6 +103,18 @@ message CreateWorkerResponse {
   }
 }
 
+message InvokeAndAwaitWorkerRequestJson {
+  golem.worker.WorkerId worker_id = 1;
+  string name = 2;
+  repeated golem.common.JsonValue input = 3;
```
vigoo, line 109: Interesting that you created a protobuf encoding for JSON, can't we just pass down the json as string to this level?

afsalthaj: I did think through this, and ended up deciding not to parse JSON string, and instead keep the structure and deserialize and transform instead.

However, I can also see some advantages of using `String` (probably faster?, I am not sure if parsing is more faster than dserialisation and one-to-one transformation - may be string is faster). I can see the below old version, and happy to reuse.

https://github.com/golemcloud/golem/blob/main/golem-api-grpc/proto/golem/worker/invoke_result_json.proto

golem-worker-service-base/src/evaluator/mod.rs
```diff
@@ -7,9 +7,12 @@ mod math_op_evaluator;
 pub(crate) mod path;
 mod pattern_match_evaluator;
 
+mod internal;
+
 use golem_wasm_ast::analysis::AnalysedType;
 use golem_wasm_rpc::json::get_json_from_typed_value;
-use golem_wasm_rpc::TypeAnnotatedValue;
+use golem_wasm_rpc::protobuf::type_annotated_value::TypeAnnotatedValue;
+use golem_wasm_rpc::protobuf::TypedOption;
```
afsalthaj, line 15: Bigger changes mainly in data transformations, and no logical changes.

golem-api-grpc/proto/golem/workerexecutor/worker_executor.proto
justcoon: i think if we want to have structure, instead of just string, probably we should use https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/struct.proto#L52

vigoo: I don't really see the point in modelling the JSON structure with protobuf types when in the end it's transformed back to a JSON value anyway. And probably we don't even need now to parse the input JSON in the worker service side before forwarding it to worker executor. (But even if we do, I would not worry about that much before some benchmarks/profiling shows it is the bottleneck)

golem-worker-executor-base/Cargo.toml
```diff
@@ -20,6 +20,10 @@ mocks = []
 golem-api-grpc = { path = "../golem-api-grpc", version = "0.0.0" }
 golem-common = { path = "../golem-common", version = "0.0.0" }
 golem-rib = { path = "../golem-rib", version = "0.0.0" }
+
+# Avoid this dependency and simply move type checkers into worker executor base ensuring we don't r
```
vigoo, line 24: Let's move it

golem-worker-executor-base/src/services/component.rs
```diff
@@ -441,6 +444,22 @@ async fn get_metadata_via_grpc(
...
+                        .metadata
+                        .map(|metadata| {
+                            let export = metadata.exports;
+                            let vec: Vec<Result<Export, String>> = export
+                                .into_iter()
+                                .map(|proto_export| {
+                                    golem_service_base::model::Export::try_from(proto_export)
+                                })
+                                .collect();
+                            vec.into_iter().collect()
```
vigoo, line 457: You can do this whole thing in one run, by 
```rust
.map(|metadata| metadata.exports.into_iter().map(...).collect::<Result<Vec<_>, _>()
```

golem-worker-executor-base/src/worker.rs
```diff
@@ -86,6 +89,41 @@ pub struct Worker<Ctx: WorkerCtx> {
     oom_retry_config: RetryConfig,
 }
 
+impl<Ctx: WorkerCtx> Worker<Ctx> {
```
vigoo, line 92: Please don't open multiple `impl` sections for the same thing

golem-worker-executor-base/src/worker.rs
```diff
@@ -86,6 +89,41 @@ pub struct Worker<Ctx: WorkerCtx> {
...
+    pub async fn get_component_metadata(
+        worker: &Arc<Worker<Ctx>>,
```
vigoo, line 94: This should be a regular method, I don't see why it would take an `Arc<Worker<Ctx>>` (just take `&self`)

golem-worker-executor-base/src/worker.rs
```diff
@@ -86,6 +89,41 @@ pub struct Worker<Ctx: WorkerCtx> {
...
+                            UpdateDescription::SnapshotBased { .. } => "snapshot based",
+                        },
+                        worker_metadata.last_known_status.component_version
+                    );
+                    target_version
+                },
+            );
+        let (component, component_metadata) = worker
+            .component_service()
+            .get(&worker.engine(), &component_id, component_version)
```
vigoo, line 120: This is potentially download, compiling and returning the whole WASM component, not just the metadata! Must not be part of this function, only done once in the beginning of the invocation loop

golem-worker-executor-base/src/worker.rs
```diff
@@ -86,6 +89,41 @@ pub struct Worker<Ctx: WorkerCtx> {
...
 }
 
+impl<Ctx: WorkerCtx> Worker<Ctx> {
+    pub async fn get_component_metadata(
+        worker: &Arc<Worker<Ctx>>,
+    ) -> Result<(WorkerMetadata, Component, ComponentMetadata), GolemError> {
+        let component_id = worker.owned_worker_id.component_id();
+        let worker_metadata = worker.get_metadata().await?;
+
+        let component_version = worker_metadata
```
vigoo, line 99: This decision of whether we are attempting an update or not should not be part of the `get_component_metadata`, it is something we have to decide, and log, once when the component starts

golem-api-grpc/proto/golem/workerexecutor/worker_executor.proto
afsalthaj: We may not need structured data here, as there can be an overhead of type informations in the payload, and therefore payload size increases. And may be json parsing at the worker-executor side might be faster than deserialisin this proto struture with extra type information - may be. This implies, I am considering @vigoo 's opinion here (just fixing my tests)

However, on the below point:

>And probably we don't even need now to parse the input JSON in the worker service side before forwarding it to worker executor. 

On this point, we already get a serde_value::Value straight away in the request handlers, and it isn't a string. So in a way we have to transform this `Value` to `String` (if we are making this a String) instead f transforming it to another structure (the one in this PR)

Also,  I am not sure if we should end up having input as `String` in an open-api spec whatsoever. So if OpenAPI spec tells the user "please pass the JSON", then what we already have is `serde_json::Value` and instead of converting it to String, we convert it to `structured` data and send it to worker-executor. So my decision to use structure here is not an accident - but just another considerable solution which I may change today to `String` solely for performance reason as mentioned above

golem-worker-executor-base/src/worker.rs
afsalthaj: Did I? I searched for if there are duplicates. May be I missed out.

golem-worker-executor-base/src/worker.rs
afsalthaj: Why is it a download @vigoo ?
In the invoke functions in grpc, I am calling this after `start_if_needed`. So we should be using the cache by the time we call this function I suppose?

golem-api-grpc/proto/golem/workerexecutor/worker_executor.proto
vigoo: Still, trying to make poem pass an unparsed JSON to the request handler seems a better route to me than the protobuf encoding.

golem-worker-executor-base/src/worker.rs
vigoo: Yes.

vigoo: Basically most of the file is a `impl<Ctx: WorkerCtx> Worker<Ctx> ` below.

golem-worker-executor-base/src/worker.rs
vigoo: It's not an infinite cache. The currently configured component cache size is 32, while the component metadata cache size is 16k. And there is absolutely no need to return the `Component` (which is the compiled wasmtime thing!) in a function that is called `get_component_metadata`.

afsalthaj: Ok I understand. I fixed this bit. 

Yes, I understand it cannot be infinite cache.  I have changed this to use only component metadata cache and hopefully that goes to 16K elements cache where we could possibly have the metadata already downloaded once the instance is started (if it decides to start) and the next subsequent call of get_metadata will hopefully hit the cache. That's my assumption. Ofcourse if there is a `possible` download , but just wanted to ensure the miss rate is possibly high.

The new change:
```scala
 async fn invoke_worker_internal<Req: GrpcInvokeRequest>(
        &self,
        request: &Req,
    ) -> Result<(), GolemError> {
        let full_function_name = request.name();

        let calling_convention = request.calling_convention();

        let worker = self.get_or_create(request).await?;

        let component_metadata = worker::get_component_metadata(&worker).await?;
           ......
           
            .....
           
    }

```

Internally I can see it hits only component_metadata_cache, so hopefully this change would have fixed your concern.

```scala

 async fn get_metadata(
        &self,
        component_id: &ComponentId,
        forced_version: Option<ComponentVersion>,
    ) -> Result<ComponentMetadata, GolemError> {
        match forced_version {
            Some(version) => {
                let client = self.client.clone();
                let access_token = self.access_token;
                let retry_config = self.retry_config.clone();
                let component_id = component_id.clone();
                self.component_metadata_cache
                    .get_or_insert_simple(
                        &ComponentKey {
                            component_id: component_id.clone(),
                            component_version: version,
                        },
                        || {
                            Box
```

afsalthaj:
I am fiddling with a lot of errors in my local unfortunately (which doesn't happen in CI)

```
cargo-make] WARN - Build Failed.
 ✘ afsalthaj@Afsals-MacBook-Pro  ~/projects/golem   delete_metadata  cargo make build-release
[cargo-make] INFO - cargo make 0.37.10
[cargo-make] INFO - Calling cargo metadata to extract project info
[cargo-make] INFO - Cargo metadata done
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: build-release
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Skipping Task: set-version
[cargo-make] INFO - Execute Command: "cargo" "build" "--release"
   Compiling serde v1.0.203
   Compiling either v1.13.0
   Compiling indexmap v2.2.6
   Compiling semver v1.0.23
   Compiling serde_json v1.0.118
   Compiling ring v0.17.8
```

resulting in

```scala
Compiling golem-worker-service v0.0.0 (/Users/afsalthaj/projects/golem/golem-worker-service)
   Compiling golem-shard-manager v0.0.0 (/Users/afsalthaj/projects/golem/golem-shard-manager)
   Compiling golem-component-service v0.0.0 (/Users/afsalthaj/projects/golem/golem-component-service)
   Compiling golem-component-compilation-service v0.0.0 (/Users/afsalthaj/projects/golem/golem-component-compilation-service)
   Compiling golem-worker-executor v0.0.0 (/Users/afsalthaj/projects/golem/golem-worker-executor)
error: linking with `cc` failed: exit status: 1
  |
  = note: env -u IPHONEOS_DEPLOYMENT_TARGET -u TVOS_DEPLOYMENT_TARGET LC_ALL="C" PATH="/Users/afsalthaj/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/x86_64-apple-darwin/bin:/usr/bin:/Users/afsalthaj/.wasmtime/bin:/Users/afsalthaj/.cabal/bin:/Users/afsalthaj/.ghcup/bin:/Users/afsalthaj/.nix-profile/bin:/Users/afsalthaj/.nix-profile/bin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/usr/local/go/bin:/Users/afsalthaj/.cargo/bin:/Users/afsalthaj/bin:/Users/afsalthaj/.local/bin:/Applications/Postgres.app/Contents/Versions/13/bin" VSLANG="1033" ZERO_AR_DATE="1" "cc" "-arch" "x86_64" "-m64" "/var/folders/yq/7y70f7hn68bc_4h6cr2xqc4m0000gn/T/rustcBKy6FT/symbols.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.00.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.01.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.02.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.03.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.04.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.05.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.06.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.07.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.08.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.09.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.10.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.11.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.12.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.13.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.14.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.worker_executor.2f00dd851ecb2026-cgu.15.rcgu.o" "/Users/afsalthaj/projects/golem/target/release/deps/worker_executor-a2c8b0ff6ac5d10a.30zlfl8kcy5ozxum.rcgu.o" "-L" "/Users/afsalthaj/projects/golem/target/release/deps" "-L" "/Users/afsalthaj/projects/golem/target/release/build/wasmtime-d79c2a8b7a71f30f/out" "-L" "/Users/afsalthaj/projects/golem/target/release/build/ittapi-sys-bb5da18136dcb62a/out" "-L" "/Users/afsalthaj/projects/golem/target/release/build/zstd-sys-1b2e4ccefa58fdea/out" "-L" "/Users/afsalthaj/projects/golem/target/release/build/ring-203049e493763bee/out" "-L" "/Users/afsalthaj/projects/golem/target/release/build/libsqlite3-sys-d959280c4cde3a68/out" "-L" "/Users/afsalthaj/projects/golem/target/release/build/aws-lc-sys-39be0671a3e361f2/out" "-L" "/Users/afsalthaj/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/x86_64-apple-darwin/lib" "/Users/afsalthaj/projects/golem/target/release/deps/libgolem_worker_executor-02314f1b9ea1c4c0.rlib" "/Users/afsalthaj/projects/golem/target/release/deps/libgolem_worker_executor_base-703b2ad3a705b511.rlib" "/Users/afsalthaj/projects/golem/target/release/deps/libtonic_reflection-c2a631d2df08b7ff.rlib" "/Users/afsalthaj/projects/golem/target/release/deps/libtonic_health-6e9b57c563608a1b.rlib" "/Users/afsalthaj/projects/golem/target/release/deps/libasync_fs-b453b279fdc7202a.rlib" "/Users/afsalthaj/proj
  

```

image

golem-api-grpc/proto/golem/workerexecutor/worker_executor.proto
```diff
@@ -17,13 +18,17 @@ import public "golem/worker/worker_metadata.proto";
...
+
 import public "wasm/rpc/val.proto";
+import public "wasm/rpc/type_annotated_value.proto";
 
 package golem.workerexecutor;
 
 service WorkerExecutor {
   rpc CreateWorker(CreateWorkerRequest) returns (CreateWorkerResponse);
   rpc InvokeAndAwaitWorker(InvokeAndAwaitWorkerRequest) returns (InvokeAndAwaitWorkerResponse);
+  rpc InvokeAndAwaitWorkerJson(InvokeAndAwaitWorkerRequest) returns (InvokeAndAwaitWorkerResponseJson);
```
vigoo, line 30: We should not have this variant anymore.

golem-api-grpc/proto/golem/common/json_value.proto
```diff
@@ -0,0 +1,36 @@
+syntax = "proto3";
+
+package golem.common;
+
+message JsonValue {
```
vigoo, line 5: This can be deleted

golem-common/src/component_metadata.rs
```diff
@@ -0,0 +1,119 @@
...
+
+// Metadata of Component in terms of golem_wasm_ast types
+pub struct RawComponentMetadata {
+    pub exports: Vec<AnalysedExport>,
+    pub producers: Vec<Producers>,
+    pub memories: Vec<Mem>,
+}
+
+impl RawComponentMetadata {
+    pub fn from_data(data: &[u8]) -> Result<RawComponentMetadata, ComponentProcessingError> {
```
vigoo, line 33:
```suggestion
    pub fn analyse_component(data: &[u8]) -> Result<RawComponentMetadata, ComponentProcessingError> {
```

golem-common/src/component_metadata.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        Ok(RawComponentMetadata {
+            exports,
+            producers,
+            memories,
+        })
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
```
vigoo, line 66: why are we using this `thiserror` crate now?

golem-common/src/precise_json.rs
```diff
@@ -0,0 +1,511 @@
+use golem_wasm_rpc::protobuf::type_annotated_value;
+use golem_wasm_rpc::protobuf::type_annotated_value::TypeAnnotatedValue;
+use serde::Deserialize;
+use serde::Serialize;
+use serde_json::Value as JsonValue;
+
+use thiserror::Error;
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub enum PreciseJson {
```
vigoo, line 10: A couple of notes for this type
- Is there any difference from `wasm_rpc::Value`? I think there isn't?
- It misses resource handles 
- If we need a separate type (compared to `Value`) it still should be defined in `wasm_rpc` as all the other value types and conversions between them
- Why is there a custom decoder from json object - can't we use the default `serde_json` codecs (with some optional discriminator annotations if needed etc)?
- Where is the encoder

golem-worker-executor-base/Cargo.toml
vigoo: Can this be removed now?

golem-worker-executor-base/src/durable_host/mod.rs
```diff
@@ -1052,14 +1053,38 @@ impl<Ctx: WorkerCtx + DurableWorkerCtxView<Ctx>> ExternalOperations<Ctx> for Dur
...
                             }) => {
+                                let component_metadata =
+                                    store.as_context().data().component_metadata().clone();
+
+                                let function_results: Vec<AnalysedFunctionResult> =
+                                    exports::function_by_name(
+                                        &component_metadata.exports,
+                                        &full_function_name,
+                                    )
+                                    .unwrap()
```
vigoo, line 1064: no `unwrap`s in critical code please

golem-worker-executor-base/src/durable_host/mod.rs
```diff
@@ -1052,14 +1053,38 @@ impl<Ctx: WorkerCtx + DurableWorkerCtxView<Ctx>> ExternalOperations<Ctx> for Dur
                                 output,
                                 consumed_fuel,
                             }) => {
+                                let component_metadata =
+                                    store.as_context().data().component_metadata().clone();
```
vigoo, line 1057: I think this `clone` shouldn't be needed

golem-worker-executor-base/src/grpc.rs
```diff
@@ -1260,6 +1292,90 @@ impl<Ctx: WorkerCtx, Svcs: HasAll<Ctx> + UsesAllDeps<Ctx = Ctx> + Send + Sync +
         }
     }
 
+    async fn invoke_and_await_worker_json(
```
vigoo, line 1295: Let's remove this

golem-worker-executor-base/src/worker.rs
```diff
@@ -422,7 +427,7 @@ impl<Ctx: WorkerCtx> Worker<Ctx> {
         calling_convention: CallingConvention,
         full_function_name: String,
         function_input: Vec<Value>,
-    ) -> Result<Option<Result<Vec<Value>, GolemError>>, GolemError> {
+    ) -> Result<Option<Result<TypeAnnotatedValue, GolemError>>, GolemError> {
```
vigoo, line 430: As discussed in private, I think this should be `Vec<TypeAnnotatedValue>` everywhere  where we had `Vec<Value>`

golem-worker-executor-base/src/worker.rs
```diff
@@ -1390,13 +1398,35 @@ impl RunningWorker {
                                                 output,
                                                 consumed_fuel,
                                             }) => {
+                                                let function_results: Vec<AnalysedFunctionResult> =
```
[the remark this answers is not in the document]
afsalthaj: Hmm not sure if we have the exact code there. The implementation of on_invocation_success is

```scala

   async fn on_invocation_success(
        &mut self,
        full_function_name: &str,
        function_input: &Vec<Value>,
        consumed_fuel: i64,
        output: TypeAnnotatedValue,
    ) -> Result<(), GolemError> {
        let is_live_after = self.state.is_live();

        if is_live_after {
            if self.state.snapshotting_mode.is_none() {
                self.state
                    .oplog
                    .add_exported_function_completed(&output, consumed_fuel)
                    .await
                    .unwrap_or_else(|err| {
                        panic!("could not encode function result for {full_function_name}: {err}")
                    });
                self.state.oplog.commit().await;
                let oplog_idx = self.state.oplog.current_oplog_index().await;

                if let Some(idempotency_key) = self.state.get_current_idempotency_key() {
                    self.public_state
                        .worker()
                        .store_invocation_success(&idempotency_key, output.clone(), oplog_idx)
                        .await;
                }
            }
        } else {
            let response = self
                .state
                .get_oplog_entry_exported_function_completed()
                .await?;

            if let Some(function_output) = response {
                let is_diverged = function_output != output;
                if is_diverged {
                    return Err(GolemError::unexpected_oplog_entry(
                        format!("{full_function_name}({function_input:?}) => {function_output:?}"),
                        format!("{full_function_name}({function_input:?}) => {output:?}"),
                    ));
                }
            }
        }

        self.store_worker_status(WorkerStatus::Idle).await;

        debug!("Function {full_function_name} finished with {output:?}");

        // Return indicating that it is done
      
```

golem-worker-executor-base/src/grpc.rs
afsalthaj: It returns `JSON` result. If the user is needing only `JSON` result, I dont think we need to return `TypeAnnotatedValue` and a lager payload.

vigoo: We said we don't want the worker executor know about that JSON format at all, so I think we should not have such an invoke variant.

golem-worker-executor-base/src/worker.rs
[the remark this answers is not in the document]
vigoo: The exact same code was previously in the mentioned place in the PR, I even commented there to not do `unwrap`s. Can't find it now.

afsalthaj:
Group 4 is failing /timeout in CI, but works in my machine

golem-worker-executor-base/src/worker.rs
afsalthaj: For others: We both confirmed why it is TypeAnnotatedValue

golem-worker-executor-base/src/worker.rs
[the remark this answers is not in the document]
afsalthaj: We both confirmed this and resolved it

afsalthaj:
Finally, an important test is succeeding

```
cargo test --package golem-worker-executor-base --test integration api::counter_resource_test_1 -- --nocapture --test-threads=1
```

```
   INFO golem_worker_executor_base::http_server: Stopping Http server...
    at golem-worker-executor-base/src/http_server.rs:43
    in integration::api::counter_resource_test_1

ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 183 filtered out; finished in 1.20s

   INFO golem_test_framework::components::redis::spawned: Stopping Redis
    at golem-test-framework/src/components/redis/spawned.rs:92
```

afsalthaj:
Integration test log is terminated and is showing red, but works in my machine

```scala
-07-31T09:53:15.839285Z  INFO golem_test_framework::components: [workersvc] 2024-07-31T09:53:15.839196Z  INFO api_request{api="get_workers_metadata" api_type="grpc" component_id="f93a419e-4ef4-48ff-9f64-2bd0a7a43376"}: golem_common::metrics::api: API request succeeded elapsed_ms=3
test worker::get_workers ... ok
2024-07-31T09:53:15.849059Z  INFO golem_test_framework::components: [worker-9101] 2024-07-31T09:53:15.791625Z DEBUG api_request{api="create_worker" api_type="grpc" worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-14" component_version=0 account_id="-1"}:waiting-for-permits{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-14"}:invocation-loop{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-14"}:invocation{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-14" idempotency_key="637b7fb7-48b0-4248-8461-bdd9a1c062c4" function="golem:it/api.{start-polling}"}: golem_worker_executor_base::metrics::wasm: golem http::types::outgoing_request::set_method called
2024-07-31T09:53:15.849308Z  INFO golem_test_framework::components: [worker-9100] 2024-07-31T09:53:15.800828Z DEBUG api_request{api="create_worker" api_type="grpc" worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-2" component_version=0 account_id="-1"}:waiting-for-permits{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-2"}:invocation-loop{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-2"}:invocation{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-2" idempotency_key="3463d13a-82bb-4e2a-81ae-21764c3b6c29" function="golem:it/api.{start-polling}"}: golem_worker_executor_base::metrics::wasm: golem http::types::future_incoming_response::drop called

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out2024-07-31T09:53:15.850528Z  INFO golem_test_framework::components: [worker-9100] 2024-07-31T09:53:15.800827Z DEBUG api_request{api="create_worker" api_type="grpc" worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-4" component_version=0 account_id="-1"}:waiting-for-permits{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-4"}:invocation-loop{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-4"}:invocation{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-4" idempotency_key="19be4681-cf41-4a5f-b723-0414bec75325" function="golem:it/api.{start-polling}"}: golem_worker_executor_base::metrics::wasm: golem http::types::future_incoming_response::get called
2024-07-31T09:53:15.849923Z  INFO golem_test_framework::components: [worker-9101] 2024-07-31T09:53:15.791679Z DEBUG api_request{api="create_worker" api_type="grpc" worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-14" component_version=0 account_id="-1"}:waiting-for-permits{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-14"}:invocation-loop{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-14"}:invocation{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-14" idempotency_key="637b7fb7-48b0-4248-8461-bdd9a1c062c4" function="golem:it/api.{start-polling}"}: golem_worker_executor_base::metrics::wasm: golem http::types::outgoing_request::set_path_with_query called
; finished in 259.06s

2024-07-31T09:53:15.850823Z  INFO golem_test_framework::components: [worker-9100] 2024-07-31T09:53:15.800679Z DEBUG api_request{api="create_worker" api_type="grpc" worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-8" component_version=0 account_id="-1"}:waiting-for-permits{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-8"}:invocation-loop{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-8"}:invocation{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-8" idempotency_key="76d3637d-52b1-4538-b4c6-11f1e403d347" function="golem:it/api.{start-polling}"}: golem_worker_executor_base::metrics::wasm: golem http::types::future_incoming_response::drop called
2024-07-31T09:53:15.851221Z  INFO golem_test_framework::components: [worker-9100] 2024-07-31T09:53:15.804206Z DEBUG api_request{api="create_worker" api_type="grpc" worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-8" component_version=0 account_id="-1"}:waiting-for-permits{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-8"}:invocation-loop{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-8"}:invocation{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-8" idempotency_key="76d3637d-52b1-4538-b4c6-11f1e403d347" function="golem:it/api.{start-polling}"}: wasmtime::runtime::gc::enabled::rooting: Exiting GC root set LIFO scope: 0
2024-07-31T09:53:15.851092Z  INFO golem_test_framework::components::worker_executor_cluster::spawned: Killing all worker executors
2024-07-31T09:53:15.851265Z  INFO golem_test_framework::components: [worker-9100] 2024-07-31T09:53:15.801312Z DEBUG api_request{api="create_worker" api_type="grpc" worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-1" component_version=0 account_id="-1"}:waiting-for-permits{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-1"}:invocation-loop{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-1"}:invocation{worker_id="ab599159-f504-487a-9e32-a1effed6bd9a/worker-http-client-1" idempotency_key="d0e02879-6474-4b72-ab03-03d359412142" function="golem:it/api.{start-polling}"}: golem_worker_executor_base::metrics::wasm: golem http::types::future_incoming_response::get called
2024-07-31T09:53:15.851273Z  INFO golem_test_framework::components::worker_executor::spawned: Stopping golem-worker-executor 9100
2024-07-31T09:53:15.85099
```

justcoon (approved):

golem-common/src/precise_json.rs
afsalthaj: There is an encoder now.
II can try to reuse Value in a separate PR

afsalthaj:
I will address @vigoo 's comments in a separate PR. I think we have come a long way by this time, and would like to merge a working version.

PR #670, opened 2024-07-17
Removes http_02 dependency and updates tonic/prost

### LINKED ISSUE #634: Try to get rid of `http_02`
We are simultaneously using two versions of the `http` crate. Since then many dependencies has been updated, maybe we no longer need to.

vigoo:
Resolves #634 

Requires https://github.com/golemcloud/wasm-rpc/pull/63


vigoo:
Closed by #1172

PR #674, opened 2024-07-17
Unify test framework

senia-psm:
Allow reusing test-framework in cloud.

### DIFF
--- a/golem-test-framework/src/components/worker_executor/mod.rs
+++ b/golem-test-framework/src/components/worker_executor/mod.rs
@@ -26,7 +26,7 @@ use crate::components::component_service::ComponentService;
 use crate::components::redis::Redis;
 use crate::components::shard_manager::ShardManager;
 use crate::components::worker_service::WorkerService;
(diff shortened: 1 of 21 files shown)


golem-test-framework/src/components/worker_executor/mod.rs
```diff
@@ -71,85 +71,103 @@ async fn wait_for_startup(host: &str, grpc_port: u16, timeout: Duration) {
...
-        .with_str(
-            "GOLEM__SHARD_MANAGER_SERVICE__CONFIG__RETRIES__MULTIPLIER",
-            "2",
-        )
-        .with("GOLEM__PORT", grpc_port.to_string())
-        .with("GOLEM__HTTP_PORT", http_port.to_string())
-        .build()
+#[async_trait]
+pub trait WorkerExecutorEnvVars {
+    async fn env_vars(
```
noise64, line 76: is async needed here?

senia-psm: The goal is to reimplement this method on cloud. And on cloud it has to be `async`.

PR #676, opened 2024-07-17
Resource info in metadata

### LINKED ISSUE #181: Resource information into worker metadata
We support creating **resource instances** now but there is no way gather any information about the number of held resources. This ticket should:

- Add some basic numeric information into the worker metadata - number of instances per exported resource type. We only expose the count not the actual handles, so there is no way for “outside” manipulation of a created resource.
- Add some new metrics so we see resource creation/dropping/average count per worker

vigoo:
Resolves #181


golem-service-base/src/model.rs
```diff
@@ -2807,6 +2807,7 @@ pub struct WorkerMetadata {
     pub last_error: Option<String>,
     pub component_size: u64,
     pub total_linear_memory_size: u64,
+    pub owned_resources: HashMap<u64, ResourceMetadata>,
```
justcoon, line 2810: did not this changed also OpenApi spec? (if it needs to be regenerated)

vigoo: yes it did, now need to update the cli because of it :) one more commit

Issue #684, opened 2024-07-22
Sharding: rework routing retry loop and invalidation


noise64:
- in a retry loop only be "fast" on the first invalidation
- after that start doing backoff (and check if there are cases when it needs to be reseted)
- add global limits or throttling for sharding invalidation

PR #690, opened 2024-07-24
Sharding reworks

noise64:
Resolves:
 - https://github.com/golemcloud/golem/issues/684
 - https://github.com/golemcloud/golem/issues/683

Changes:
- added pod fields to call_on_executor logs, where available
- added generic jitter option to retry config
- ports retries that do not need special handling to use the common retry logic, as part of these:
  - refactored shard manager error types to be "more typed" 
  - extracted IsRetriableError to common, and added with_retriable_error variant for retries
- added configurable throttle for routing table cache invalidation in worker-service
- changed call_on_executor retry
  - to use the generic retry logic with backoff and jitter
  - to reset the retry logic and emit warn logs when max_attempts reached (so in effect we still retry "forever", but from time to time we will get warns)
- made worker starts and restarts parallel in the sharding test suite
- added logging tools
  - lnav config for golem JSON logs
  - elastic stack for the local logs
- change the sharding manager startup / lifecycle:
  - dropped the persisted rebalanace plan, given we need to do a full sharding broadcast anyway, and we cannot trust that the rebalance plan makes sense anymore
  - the new startup logic now only do an initial round of healtchchecks, then marks all the available nodes as new nodes and the unhealthy ones as removed, then it is using the normal worker loop to handle and apply an up-to-date rebalance plan

Some performance and behaviour results
 - In the sharding test examples the changed "call_on_executor retry" strategy decreased the number of worker and sharding-manager calls from a few thousands to a few dozens with the no increased times on the tests
 - The parallel starts / restarts sped up locally the env reset from 10 seconds to 4 seconds
 - Simplifying the shard manager startup reduced the typical coordinated sharding scenario test time from 60 sec to 40 sec (which was also increased since the quick fix done recently)


log-tools/elastic/docker-compose.yml
```diff
@@ -0,0 +1,52 @@
+services:
+  es01:
+    image: docker.elastic.co/elasticsearch/elasticsearch:${STACK_VERSION}
```
justcoon, line 3: is not better to use https://opensearch.org/ ?

noise64: we are not using any paid things and we are not hosting it. do you have any concerns with it?

justcoon: currently not a lot, but considering history of elasticsearch is hard to know what they do in future, so in general i think we should use rather opensearch, but we can change it later

golem-shard-manager/src/model.rs
```diff
@@ -79,11 +78,30 @@ impl Pod {
...
-        let resolved = pod.address()?.map(|sa| sa.ip()).collect::<Vec<_>>();
-        if !resolved.contains(&source_ip) {
-            warn!("Host mismatch between registration message and resolved message source. Provided: {resolved:?}; source: {source_ip:?}");
+        match pod.address() {
+            Ok(resolved_addresses) => {
+                let resolved_addresses = resolved_addresses.map(|sa| sa.ip()).collect::<Vec<_>>();
+
+                if !resolved_addresses.contains(&source_ip) {
+                    warn!(
+                        pod= ?pod,
```
justcoon, line 87: should not we rather use Display (instead of Debug) ?

noise64: 👍 https://github.com/golemcloud/golem/pull/690/commits/92d61924bd8af9d2652c3e59c04a432e3a3f23b4

golem-shard-manager/src/model.rs
```diff
@@ -79,11 +78,30 @@ impl Pod {
...
+                        resolved_adresses = ?resolved_addresses,
+                        source_ip = ?source_ip,
```
justcoon, line 89: Similar here and in some other cases, can we use Display instead of Debug ?

noise64: https://github.com/golemcloud/golem/pull/690/commits/56007858afaaa4410dfdd57c513c330b9327786a

i searched for `= ?`, did not found more

golem-shard-manager/src/shard_management.rs
```diff
@@ -129,63 +124,67 @@ impl ShardManagement {
...
-            persistence_service
-                .write(&current_routing_table, &rebalance)
-                .await
-                .expect("Failed to persist routing table");
-            drop(current_routing_table); // not holding the write lock while executing the rebalance
-
-            Self::execute_rebalance(worker_executors.clone(), &mut rebalance)
-                .await
-                .expect("Failed to execute rebalance");
+            debug!(rebalance=?rebalance, "Applying rebalance plan");
```
justcoon, line 173: here should be possible to use Display instead of Debug for arguments

golem-shard-manager/src/model.rs
```diff
@@ -79,11 +78,30 @@ impl Pod {
...
+                        source_ip = %source_ip,
+                        "Host mismatch between registration message and resolved message source"
+                    );
+                }
+
+                Ok(pod)
+            }
+            Err(error) => {
+                error!(
+                    pod=?pod,
```
justcoon, line 98: here should be possible to use Display instead of Debug for arguments

golem-shard-manager/src/model.rs
```diff
@@ -79,11 +78,30 @@ impl Pod {
...
-        if !resolved.contains(&source_ip) {
-            warn!("Host mismatch between registration message and resolved message source. Provided: {resolved:?}; source: {source_ip:?}");
+        match pod.address() {
+            Ok(resolved_addresses) => {
+                let resolved_addresses = resolved_addresses.map(|sa| sa.ip()).collect::<Vec<_>>();
+
+                if !resolved_addresses.contains(&source_ip) {
+                    warn!(
+                        pod= %pod,
+                        resolved_adresses = ?resolved_addresses,
```
justcoon, line 88: here should be possible to use Display instead of Debug for arguments

golem-shard-manager/src/shard_management.rs
```diff
@@ -129,63 +124,67 @@ impl ShardManagement {
...
-
-            // Getting a write lock while the rebalance plan is calculated and got persisted (but NOT applied)
-            let mut current_routing_table = routing_table.write().await;
-
-            for pod in removed_pods {
-                current_routing_table.remove_pod(&pod);
-            }
+            debug!(
+                new_pods=?new_pods,
+                removed_pods=?removed_pods,
```
justcoon, line 129: here should be possible to use Display instead of Debug for arguments for new_pods and removed_pods

golem-shard-manager/src/model.rs
noise64: thanks for the other comments, missed those

golem-shard-manager/src/shard_management.rs
noise64: https://github.com/golemcloud/golem/pull/690/commits/0a86347c73d43333b0b3c22abe0447b3fe41bebe

golem-shard-manager/src/model.rs
noise64: https://github.com/golemcloud/golem/pull/690/commits/0a86347c73d43333b0b3c22abe0447b3fe41bebe

golem-shard-manager/src/model.rs
noise64: https://github.com/golemcloud/golem/pull/690/commits/0a86347c73d43333b0b3c22abe0447b3fe41bebe

golem-shard-manager/src/shard_management.rs
noise64: https://github.com/golemcloud/golem/pull/690/commits/0a86347c73d43333b0b3c22abe0447b3fe41bebe

justcoon (approved):

Issue #688, opened 2024-07-23
Issue with adding headers to the response in API definitions


vigoo:
From https://discord.com/channels/1134448700572319785/1134448701260169249/1264231231017717840

Hi all! I've been playing with the API definition format, but I have a problem: when I try to add headers to the response, I get an empty body instead. Basically, I've defined my response like this:
```
        "response": "${ {body: golem:component/api.{get-character}() } }"
```

and it works (I get a JSON reponse from the worker); but when I change it to something like:
```
        "response": "${ {body: golem:component/api.{get-character}(), headers: { x-test: 'foobar' } } }"
```

then I get nothing:
```
> http GET http://localhost:9006/1/character Accept=application/json -phH
GET /1/character HTTP/1.1
Accept: application/json, */*;q=0.5
Accept-Encoding: gzip, deflate
Connection: keep-alive
Content-Length: 30
Content-Type: application/json
Host: localhost:9006
User-Agent: HTTPie/3.2.2

HTTP/1.1 200 OK
content-length: 0
date: Sat, 20 Jul 2024 14:13:08 GMT
```

Has anyone stubled upon this problem? I've checked the issues but I couldn't find anything related. Thanks!

(edit: forgot to add link to reproducible example: https://github.com/albx79/golem-experiments/tree/main/fate-api)

justcoon:
in relation to [Web Assembly Value Encoding](https://github.com/lann/wasm-wave), we changed `'` to `"`, 
parsing did not failed on `'`, but expression evaluation ended with empty expression, which resulted in Unit response

PR #696, opened 2024-07-25
expr eval - return error if parsing failed

justcoon:
related to: #688

### DIFF
--- a/golem-rib/src/text/mod.rs
+++ b/golem-rib/src/text/mod.rs
@@ -425,6 +425,25 @@ mod record_tests {
         let output_expr = from_string(expr_str.as_str()).unwrap();
         assert_eq!((expr_str, input_expr), (expected_record_str, output_expr));
     }
+
+    #[test]
+    fn test_round_trip_record_literal_invalid() {
+        let expr_str = r#"
+                 ${  {body: golem:component/api.{get-character}(), headers: { x-test: 'foobar' } } }
+            "#;
+
+        let result = from_string(expr_str);
+
+        assert!(result.is_err());
+
+        let expr_str = r#"
+                ${   {body: golem:component/api.{get-character}(), headers: { x-test: "foobar" } } }
+            "#;
+
+        let result = from_string(expr_str);
+
+        assert!(result.is_ok());
+    }
 }
 
 #[cfg(test)]
--- a/golem-worker-executor-base/src/invocation.rs
+++ b/golem-worker-executor-base/src/invocation.rs
@@ -265,7 +265,7 @@ async fn get_or_create_indexed_resource<'a, Ctx: WorkerCtx>(
         )),
     )?;
 
-    let constructor_param_types = resource_constructor.params(&store).iter().map(type_to_analysed_type).collect::<Result<Vec<_>, _>>()
+    let constructor_param_types = resource_constructor.params(store as &StoreContextMut<'a, Ctx>).iter().map(type_to_analysed_type).collect::<Result<Vec<_>, _>>()
         .map_err(|err| GolemError::invalid_request(format!("Indexed resource invocation cannot be used with owned or borrowed resource handles in constructor parameter position! ({err})")))?;
 
     let raw_constructor_params = parsed_function_name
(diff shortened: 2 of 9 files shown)


golem-rib/src/text/mod.rs
```diff
@@ -425,6 +425,25 @@ mod record_tests {
         let output_expr = from_string(expr_str.as_str()).unwrap();
         assert_eq!((expr_str, input_expr), (expected_record_str, output_expr));
     }
+
+    #[test]
+    fn test_round_trip_record_literal_invalid() {
+        let expr_str = r#"
+                 ${  {body: golem:component/api.{get-character}(), headers: { x-test: 'foobar' } } }
```
afsalthaj, line 432: The existence of this test is not a big problem, however I wouldn't personally include this test if was doing it. So its very opinionated statement.
 
All other tests in golem-rib is structured (as much as I can) and has a behavior. Example: testing a  `record`parser will test `record of primitive` or `record of sequence`, or `record of record` etc.  Infact, this is why I wrote things from scratch instead of copying old tests for this parser because old tests are all "randomish".

And on the other hand this particular, this test is very specific - "an invalid case of a record having a value which is having single quotes",  such tests can be theoretically infinite. This opens door to having "random" tests to exist in golem-rib.

golem-worker-service-base/src/api_definition/http/http_api_definition.rs
```diff
@@ -379,27 +379,27 @@ mod tests {
 
     #[test]
     fn expression_with_predicate0() {
-        test_string_expr_parse_and_encode("1>2");
+        test_string_expr_parse_and_encode("${ 1 < 2 }");
```
afsalthaj, line 382: why was this changed?

golem-worker-service-base/src/api_definition/http/http_api_definition.rs
```diff
@@ -379,27 +379,27 @@ mod tests {
...
     }
 
     #[test]
     fn expression_with_predicate1() {
-        test_string_expr_parse_and_encode("${request.path.user-id > request.path.id}");
+        test_string_expr_parse_and_encode("${ request.path.user-id > request.path.id }");
```
afsalthaj, line 387: why changing thse tests?

golem-worker-service-base/src/api_definition/http/http_api_definition.rs
```diff
@@ -379,27 +379,27 @@ mod tests {
...
     #[test]
     fn expression_with_predicate3() {
-        test_string_expr_parse_and_encode("${request.path.user-id}=2");
+        test_string_expr_parse_and_encode("${ request.path.user-id == 2 }");
     }
 
     #[test]
     fn expression_with_predicate4() {
-        test_string_expr_parse_and_encode("${request.path.user-id}<2");
+        test_string_expr_parse_and_encode("${ request.path.user-id < 2 }");
```
afsalthaj, line 402: Should there be a space now?  or the actual fix is the curly bracce had to go out?

golem-worker-executor-base/src/invocation.rs
```diff
@@ -265,7 +265,7 @@ async fn get_or_create_indexed_resource<'a, Ctx: WorkerCtx>(
         )),
     )?;
 
-    let constructor_param_types = resource_constructor.params(&store).iter().map(type_to_analysed_type).collect::<Result<Vec<_>, _>>()
+    let constructor_param_types = resource_constructor.params(store as &StoreContextMut<'a, Ctx>).iter().map(type_to_analysed_type).collect::<Result<Vec<_>, _>>()
```
afsalthaj, line 268: Clippy driven?

afsalthaj (approved):

golem-worker-service-base/src/api_definition/http/http_api_definition.rs
justcoon: because it was failing on
```
called `Result::unwrap()` on an `Err` value: "Failed to parse: \"1 < 2\""
```

golem-worker-service-base/src/api_definition/http/http_api_definition.rs
justcoon: curly brace has to go out

otherwise
```
called `Result::unwrap()` on an `Err` value: "Failed to parse: \"${request.path.user-id}<2\""
```

PR #697, opened 2024-07-25
Typescript RPC Tests

nicoburniske:
Closes https://github.com/golemcloud/golem/issues/537


vigoo (changes requested):
Can you update it so it does not have the `wit/deps` but uses `wit-deps` as the other test components now? Also if possible add to the build-components.sh script, or at least have a dedicated one in the `ts-rpc` dir that copies the wasm files to the right location (in `test-components`) (maybe this exists just missed).

vigoo (approved):

Issue #693, opened 2024-07-25
Multiple issues with invoking bad python WASM


vigoo:
Python component trying to use Golem Runtime API:

```wit
package golem:demo;

interface api {
  resource counter {
    constructor(name: string);
    add: func(value: u64);
    get: func() -> u64;
  }
}

world py-example {
  import golem:api/host@0.2.0;
  export api;
}
```

```python
from bindings.py_example import exports

class Api(exports.Api):
    pass
```

```python
import bindings
import bindings.py_example
import bindings.py_example.imports
import bindings.py_example.imports.types
import bindings.py_example.imports.host
import bindings.py_example.exports
import bindings.py_example.exports.api
from bindings.py_example.exports import api

class Counter(api.Counter):
    def __init__(self, name: str):
        self.name = name
        self.state = 0

    def add(self, value: int):
        self.state = self.state + value

    def get(self) -> int:
        from bindings.py_example.imports.host import get_oplog_persistence_level, set_oplog_persistence_level, PersistenceLevel_PersistNothing

        old = get_oplog_persistence_level()
        try:
            set_oplog_persistence_level(PersistenceLevel_PersistNothing())
            # ...
        finally:
            set_oplog_persistence_level(old)

        return self.state
```

- gets into an infinite error loop when invoking `get` for example with `golem-cli worker invoke-and-await --component-name py_example --worker-name test6 --function 'golem:demo/api.{counter("c1").get}'`
- worker interruption is not stopping it

PR #730, opened 2024-08-06
Fix of the previous fix for invoking drop via the JSON invoke API

vigoo:
- Exposes the JSON invocation through the gRPC Worker Service API as well for testability (we had this once before many iterations and refactors ago)
- Adds new integration tests based on that
- Changes the JSON invocation API to return a `PreciseJson` instead of the old format
- Reverts the changes in worker executor by #718 
- Fixes the signature of the `drop` functions in `ComponentMetadata`
- Adds a `details` field to `ParamTypeMismatch` executor error to help debugging similar cases in the future

Note that the infinite retry loop caused by the original issue is the same as reported in #693 and will be fixed as part of that ticket.

In addition to the above:
- Switch to a new invocation JSON format (serialized version of `TypeAnnotatedValue`)
- Dropped duplicated types and using their updated upstream versions from `golem-wasm-ast` 

The new JSON format is the following:

```json
{
  "typ": <value's type>,
  "value": <value>
}
```

where `<value>` is the same as the existing (original) human-readable JSON value representation, and `<value's type>` has the same format as the types in the **component metadata** returned by the component REST API. 

For example:
```json
{
  "typ": {
    "type": "Tuple",
    "items": [
      {
        "type": "List",
        "inner": {
          "type": "Tuple",
          "items": [ 
            { "type": "Str" },
            { "type": "U64" }
          ]
        }
      }
    ]
  },
  "value": [
    [
      ["counter1", 5],
      ["counter2", 3]
    ]
  ]
}
```

The same format is used for both parameters and results.
In code, everything that was previously either `TypeAnnotatedValue` or `PreciseJson` or `serde_json::Value` is now just `TypeAnnotatedValue`.

Of course the direct type-less `Val` based invocation API through gRPC still exists.

Note that `golem-cli`'s invocation commands when used in JSON mode can be either called with the new JSON format (type+value) or with the old one (just value). If it is called with just value, CLI gets the latest metadata and uses it to interpret the value.


afsalthaj (dismissed):

golem-worker-service-base/src/service/worker/default.rs
```diff
@@ -85,24 +82,13 @@ pub trait WorkerService<AuthCtx> {
         auth_ctx: &AuthCtx,
     ) -> WorkerResult<()>;
 
-    // Accepts Json Params and Returns Json (with no type information)
+    // Accepts Vec<TypeAnnotatedValue> and returns TypeAnnotatedValue
     async fn invoke_and_await_function_json(
```
justcoon, line 86: should not it be invoke_and_await_function_json -> invoke_and_await_function_typed ?

vigoo: I don't like the `_typed` postfix because it's always typed so chose the `_json` one (previously we had both). This function is called when the REST API is called with JSON parameters, which are `TypeAnnotatedValue`'s.

golem-api-grpc/proto/golem/worker/v1/worker_service.proto
```diff
@@ -114,6 +116,21 @@ message InvokeAndAwaitResponse {
   }
 }
 
+message InvokeAndAwaitJsonRequest {
```
justcoon, line 119: is this still used?

vigoo: yes, added it so these things (like the json-invoke drop error) can be tested through "regular" integration tests not just via cli.

justcoon (approved):

Issue #647, opened 2024-07-07
golem-cli first-time initialization is confusing to users


vigoo:
See https://discord.com/channels/1134448700572319785/1176446164393013269/1259237497687380148

Not sure what the best solution is, but some ideas:

- list all commands in `--help` but mark the ones that need init somehow
- do not list any commands in `--help` and just display more detailed information about the need to init
- make it more clear how to init for the default golem-oss docker (have a simple to find default option, don't ask hosts/ports the user does not know)
- Use default docker compose for OSS CLI profile configuration

afsalthaj:
Managing the profile after the first initialisation can be confusing too. For example: Let' say I created an OSS profile and I was using it and then later down the line, I decided to create a cloud profile and switch to it. I have to do that using

```scala
golem-cli profile add golem-cloud something
Unexpected client error: error sending request for url (http://localhost:8083/version): error trying to connect: tcp connect error: Connection refused (os error 61)
Profile created
```

It's good if it doesn't try to connect to the servers in the OSS profile, if all we need is a new cloud profile. It also says profile is created. I think important to solve this before 1.0.

senia-psm:
Use John repo fork for doc updates

PR #734, opened 2024-08-08
647 docker compose profile init

senia-psm:
- Additional option for `init` command - default docker compose port
 - Prevent version check warning during profile manipulations
 - Fake commands in `--help` with no profile

Related #647

### DIFF
--- a/golem-cli/src/init.rs
+++ b/golem-cli/src/init.rs
@@ -12,15 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::command::api_definition::ApiDefinitionSubcommand;
+use crate::command::api_deployment::ApiDeploymentSubcommand;
+use crate::command::component::ComponentSubCommand;
 use crate::command::profile::{ProfileSubCommand, UniversalProfileAdd};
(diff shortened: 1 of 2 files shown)


vigoo (approved):

golem-cli/src/init.rs
```diff
@@ -193,14 +239,24 @@ fn validate_profile_override(
...
     GolemCloud,
 }
 
 impl Display for ProfileType {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
-            ProfileType::Golem => write!(f, "{}. For stand-alone installations", "Golem".bold()),
+            ProfileType::GolemDefault => write!(
+                f,
+                "{}. For default docker compose configuration",
```
justcoon, line 252: `default docker` , should dot we call it just `default` ?

senia-psm: It is not `default docker` it is `default "docker compose"`

senia-psm: i.e. is don the default CLI confiduration. it is configuration for the default docker compose file

PR #739, opened 2024-08-09
Fix worker infinite retry

### LINKED ISSUE #693: Multiple issues with invoking bad python WASM

vigoo:
Resolves #693 

- Skip hint entries when calculating retry count (the number of errors at the end of the oplog) as they can interleave the errors (example: new incoming invocation)
- Not writing resource create/describe/drop entries in reply mode (this with the lack of the first point caused the infinite retry loop)
- Fixes an issue where a worker ends up idle when invoked with a non-existing function, as the error was returned earlier than writing the bad invocation to the oplog.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -462,7 +462,7 @@ jobs:
           SCCACHE_GHA_ENABLED: "true"
           RUSTC_WRAPPER: "sccache"
         run: cargo make sharding-tests
-        timeout-minutes: 20
+        timeout-minutes: 30
   publish:
     needs:
       [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3731,8 +3731,9 @@ dependencies = [
 
 [[package]]
 name = "golem-wasm-ast"
-version = "0.0.0"
-source = "git+https://github.com/golemcloud/golem-wasm-ast?branch=analysis-type-features#4d1c8d6c21a62fd5379151f34300b5bf7b44388e"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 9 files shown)


vigoo:
Made it draft because I realized that the last point makes workers permanently failed if they get invoked with a wrong function name which we don't want. New fix coming to the PR soon.

justcoon (approved):

PR #740, opened 2024-08-10
Made stdout/err/log writes persistent

### LINKED ISSUE #738: Persist stdout/err/log

vigoo:
Resolves #738 
Required wasmtime change: https://github.com/golemcloud/wasmtime/pull/3/commits/adae4ab236f1eb4216dab8f7e1db6f34470de9e7

Introduces a new oplog entry called `Log`, which in general is treated like other _hint entries_ (got skipped during recovery when looking for the next non-hint entry), but there is a special logic that remembers the _hash_ of all the log entries seen **since the last non-hint entry**. So when reading an oplog entry describing a side effect, now we not only skip the next hint entries, but also get a set of hashes of all the logs that were persisted after that side effect, before the next one (or before going live).

With this set up, the new rules of emitting log events (writing to stdout/err or using wasi::logging) is the following:
- In replay mode we never emit logs
- In live mode, if the log's hash is in the "seen logs set", we don't emit the log, and remove this hash from the set
- In live mode if the log's hash is not in the set, we emit the log and write an oplog entry

This solution is not interfering with any existing properties of the system like the previous attempt (transactions, updates, etc).

##  PREVIOUSLY (not true anymore):
Emitting logs is currently considered `WriteLocal` so it does not interfere with things like http requests.
See https://github.com/golemcloud/golem/issues/741 about making this configurable.

Note that one probably surprising side effect of this change is that adding logging to a component is no longer compatible with automatic hot-update. 
I think we can live with this if we do two things:
- mention this in the documentation and as a workaround recommend to do non-persistent prints in such cases
- help doing non-persistent prints in the language-specific libraries (basically a print surrounded with changing the persistence level)


vigoo:
Tests are failing because hot update test need to be updated to not use persisted prints

vigoo:
Alternatively we could introduce a **dedicated oplog entry** for this.
That would avoid the effects on transactional and update behavior but it would require explicit handling this new oplog entry in several places in the code.

jdegoes (approved):

PR #745, opened 2024-08-12
Worker connect improvements

### LINKED ISSUE #704: CLI `--connect` parameter for invoke-and-await
The invoke-and-await command should have a `--connect` parameter, which automatically connects to the worker and shows its output during the time of the invocation.

It must also work in the dynamic invocation case (when the targeted worker does not exist yet) which can be implemented by silent connection retries)

vigoo:
Fixes many issues regarding worker connect streams.
Resolves #704 

In details:
- Fixed the ring buffer in worker executor for log events
- Fixed cancellation on all levels; if a worker is dropped in the executor, the top level web-socket stream closes, if the web-socket stream is closed, the connection is removed from the broadcast channel
- Introduced timestamp to the log events
- Introduced a configurable renderer for log events in the CLI supporting colors, timestamp and log-level prefixes in text mode and also json and yaml
- Automatically retry the worker connection in CLI if the connection is dropped
- Added a `--connect` flag to `invoke` and `invoke-and-await` to show log lines together with the invocation

### DIFF
--- a/golem-worker-service-base/src/service/worker/connect_stream.rs
+++ b/golem-worker-service-base/src/service/worker/connect_stream.rs
@@ -21,7 +21,7 @@ use futures::{Stream, StreamExt};
 use tokio::sync::mpsc;
 use tokio_util::sync::CancellationToken;
 use tonic::{Status, Streaming};
-use tracing::{error, info, Instrument};
+use tracing::{error, Instrument};
 
 use golem_api_grpc::proto::golem::worker::LogEvent;
 use golem_common::metrics::api::{


justcoon (approved):

jdegoes (approved):

golem-cli/src/command/worker.rs
```diff
@@ -402,44 +460,94 @@ impl<ComponentRef: clap::Args, WorkerRef: clap::Args> WorkerSubcommand<Component
...
+                            worker_uri,
+                            idempotency_key,
+                            function,
+                            parameters.parameters,
+                            parameters.wave,
+                            project_id,
+                        )
+                        .await;
+
+                    connect_handle.abort();
```
noise64, line 489: is this and unconditional connect /  abort concurrently to the invoke? i'm not sure i understand the error handling and side effect of this

vigoo: we stream the worker output concurrently to the invoke, and when the invoke returns we abort it.

noise64: what confuses me: is there a guarantee that connect succeed and streaming is started before invoke starts (and ends)?

noise64: and also i'm not sure what is the diff between how it handled here (invoke and await) and in invoke

golem-worker-executor-base/src/services/worker_event.rs
```diff
@@ -173,20 +97,26 @@ impl WorkerEventService for WorkerEventServiceDefault {
 
             let _ = self.sender.send(event.clone());
         }
-        let _ = unsafe { Producer::new(&self.ring) }.push(event);
+
+        let mut ring_prod = self.ring_prod.lock().unwrap();
+        while ring_prod.try_push(event.clone()).is_err() {
```
noise64, line 102: does this mean we keep dropping events until we have a free slot in the buffer?
is that something that should be logged?
is it possible that this will need to be run more then once? (as in is this while just for simplicity, or is there an actual possibility for looping?  because of the lock i guess not)

golem-worker-service-base/src/service/worker/connect_proxy.rs
```diff
@@ -106,8 +110,9 @@ async fn forward_worker_message<E>(
 where
     ConnectProxyError: From<E>,
 {
-    let message = message?;
+    let message: WorkerEvent = message?.try_into().map_err(ConnectProxyError::Proto)?;
     let msg_json = serde_json::to_string(&message)?;
+    info!("forward_worker_message {msg_json}"); // TODO: remove
```
noise64, line 115: // TODO: remove

golem-worker-service-base/src/service/worker/connect_stream.rs
```diff
@@ -41,42 +41,44 @@ impl ConnectWorkerStream {
...
-                            );
+                loop {
+                    tokio::select! {
+                        _ = cancel_clone.cancelled() => {
+                            info!("WorkerStream cancelled");
                             break;
                         }
+                        message = streaming.next() => {
+                            if let Some(message) = message {
+                                info!("ConnectWorkerStream received message {message:?}");
```
noise64, line 58: do we want to info log this, or was this just for debug?

noise64 (approved):
added some minor comments

PR #748, opened 2024-08-14
Ability to try update workers after component update, and component redeploy command

### LINKED ISSUE #701: CLI idea: support rapid componen testing
During developing a component, developers continuously recompile the WASM and want to try them out. This currently consists of the following steps:

- compile the component
- `component add` or `component update` 
- (if the worker needs configuration) `worker add`
- connect to the worker
- invoke it

This requires either always creating a new worker with a new name, and because of that updating that name or passing it through variables, which is very inconvenient in practice.

I propose to have a `redeploy` (or similar) command, implemented on the CLI side, which does something like the following:

- get the metadata of all workers of component
- delete all workers of a component 
- update the component
- recreate the workers with the same names / env vars / args (optionally should be skippable by a cli option)

vigoo:
Resolves #701 

Implements two new features into `golem-cli`:

`golem-cli component update` gets a new optional parameter, `--try-update-workers` which triggers a **worker update** (auto or manual) to each existing worker of the updated component.

`golem-cli component redeploy` **deletes** all existing workers of a component, but stores their arg/env properties and names, and then **recreates** the same set of workers with the same set of configuration using the latest component version (starting from a clean state, unlike with upgrade). This is useful during development to quickly test changes in the component under development. Otherwise it's very destructive, so in _text mode_ it requires a user confirmation, in _json_ and _yaml_ modes it requires an explicit confirmation flag to be passed.


afsalthaj (approved):

golem-cli/src/factory.rs
```diff
@@ -133,6 +144,22 @@ pub trait ServiceFactory {
             clients: self.health_check_clients(auth)?,
         }))
     }
+
+    fn deploy_service(
+        &self,
+        auth: &Self::SecurityContext,
+    ) -> Result<
+        Arc<dyn DeployService<ProjectContext = Self::ProjectContext> + Send + Sync>,
```
senia-psm, line 152: All the services are stateless - why would we need `Arc`?

golem-cli/src/factory.rs
```diff
@@ -161,7 +188,7 @@ impl<PR: Send + Sync, PC: Display + Send + Sync, S: Clone + Send + Sync> Compone
 {
     fn build(
         &self,
-    ) -> Result<Box<dyn ComponentService<ProjectContext = PC> + Send + Sync>, GolemError> {
+    ) -> Result<Arc<dyn ComponentService<ProjectContext = PC> + Send + Sync>, GolemError> {
```
senia-psm, line 191: All the services are stateless - why would we need `Arc`?

golem-cli/src/service/deploy.rs
```diff
@@ -0,0 +1,195 @@
...
+#[async_trait]
+impl<ProjectContext: Display + Send + Sync> DeployService for DeployServiceLive<ProjectContext> {
+    type ProjectContext = ProjectContext;
+
+    async fn try_update_all_workers(
+        &self,
+        component_uri: ComponentUri,
+        project: Option<Self::ProjectContext>,
+        mode: WorkerUpdateMode,
+    ) -> Result<GolemResult, GolemError> {
```
justcoon, line 63: it could be better if we had api at worker service, which will process it at server side, but probably ok for now

senia-psm (approved):

golem-cli/src/factory.rs
senia-psm: The service creation is quite cheap. It's just a clone for 3-4 strings. No need to keep shared service instances in `Arc` - we can just create multiple instances.

golem-cli/src/service/deploy.rs
vigoo: I don't agree with that, these are higher level workflows built on the worker service API, including the need to enumerate through all workers through the "slow" worker enumeration. I would not put it in the service side.

golem-cli/src/factory.rs
vigoo: But for example for the `--connect` it made it simpler to implement as I could clone the service to pass to the background fiber. (But that was in a previous PR)

senia-psm: It is 100% up to you. And it is not realy so important any way. I just prefer to avoid Arc since it hints state sharing.

PR #762, opened 2024-08-15
Convenience fallback mode for component add and update

### LINKED ISSUE #699: CLI idea: add and update interactive mode
Idea: in "human readable mode", `component add` and `component update` could offer to act as the other variant in case of the component already / does not exist, so the user does not have to retype the proper commend in case of a mistake. Some kind of `-y` (accept automatically) command could make it non-interactive.

vigoo:
Resolves #699 

Small convenience feature that allows fallback `component update` to add and `component add` to update, only in `text` mode.

Note that the "error matching" is ugly (string based) but I did not want to do bigger changes now in error handling.

### DIFF
--- a/golem-cli/src/command/component.rs
+++ b/golem-cli/src/command/component.rs
@@ -39,6 +39,10 @@ pub enum ComponentSubCommand<ProjectRef: clap::Args, ComponentRef: clap::Args> {
         /// The WASM file to be used as a Golem component
         #[arg(value_name = "component-file", value_hint = clap::ValueHint::FilePath)]
         component_file: PathBufOrStdin, // TODO: validate exists
+
+        /// Do not ask for performing an update in case the component already exists
+        #[arg(short = 'y', long)]
+        non_interactive: bool,
     },
 
     /// Updates an existing component by uploading a new version of its WASM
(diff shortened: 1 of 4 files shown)


golem-cli/src/command/component.rs
```diff
@@ -59,6 +63,10 @@ pub enum ComponentSubCommand<ProjectRef: clap::Args, ComponentRef: clap::Args> {
         /// Update mode - auto or manual
         #[arg(long, default_value = "auto", requires = "try_update_workers")]
         update_mode: WorkerUpdateMode,
+
+        /// Do not ask for creating a new component in case it does not exists
+        #[arg(short = 'y', long)]
```
noise64, line 68: why is it 'y'? looks like as if it would imply saying yes to the interactive question, but it seems to be like the opposite

vigoo: I don't like short parameters :D Copied it from the other places where the same flag existed already (added yesterday)
What would you use?

noise64: i would either change it to behave like "yes", or change it to --no? not sure :D

noise64 (approved):

PR #793, opened 2024-08-20
Bug fixes with more tests in place for type pull up phase

afsalthaj:
Part of incremental fixes targeting the issues 781 and 783, making sure we have a decent push down and pull up logic.

### DIFF
--- a/golem-rib/src/parser/literal.rs
+++ b/golem-rib/src/parser/literal.rs
@@ -59,10 +59,10 @@ mod internal {
 
     fn static_part<'t>() -> impl Parser<easy::Stream<&'t str>, Output = Expr> {
         many1(
-            letter()
-                .or(space())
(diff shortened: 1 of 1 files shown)


jdegoes (approved):

golem-rib/src/type_inference/type_push_down.rs
```diff
@@ -245,9 +245,8 @@ pub fn push_types_down(expr: &mut Expr) {
                 }
             }
 
+            // {bohy
```
justcoon, line 248: is this some typo?

justcoon (approved):

afsalthaj:
This PR was held up because I faced a stack overflow error, when I corrected the input to worker function. But that's a different problem and not something this PR introduced.

So I will be working to get this in.

Issue #764, opened 2024-08-15
Api Deployments should have timestamp


nicoburniske:
```typescript
type ApiDeploymentsResponse = {
    apiDefinitions: {
        id: string
        version: string
    }[]
    projectId: string
    deploymentTime: string
    site: {
        host: string
        subdomain?: string | undefined
    };
}[]
```

Right now we have no timestamp for when the apis were actually deployed. this would be a useful field to keep track of

PR #802, opened 2024-08-21
Api definition and deployment createdAt timestamp

justcoon:
related to: #764
for OpenaAPI and gRPC API:
* added createdAt to ApiDefinition and ApiDeployment - data are already there in DB, they were not just propagated
* added ApiDefinitionRequest and ApiDeploymentRequest - alternative to ApiDefinition and ApiDeployment, without createdAt


nicoburniske (approved):
LGTM

openapi/golem-service.yaml
```diff
@@ -2901,6 +2917,28 @@ components:
...
+          type: boolean
+        createdAt:
+          type: string
+          format: date-time
+      required:
+      - id
+      - version
+      - routes
+      - draft
+      - createdAt
```
vigoo, line 2940: I think this new field needs to be optional to not break the API

justcoon: it is used just in responses, not in requests, 
in such case it can break, if somebody use new client version, but will have old version of server, if we want to avoid that scenario then it had to be optional

justcoon: changed to optional

afsalthaj (approved):

golem-cli/src/oss/clients/api_definition.rs
```diff
@@ -82,13 +83,13 @@ async fn create_or_update_api_definition<
...
         Action::Create => {
-            let value: HttpApiDefinition = serde_json::from_str(definition_str.as_str())
+            let value: HttpApiDefinitionRequest = serde_json::from_str(definition_str.as_str())
                 .map_err(|e| GolemError(format!("Failed to parse HttpApiDefinition: {e:?}")))?;
 
             Ok(client.create_definition(&value).await?)
         }
         Action::Update => {
-            let value: HttpApiDefinition = serde_json::from_str(definition_str.as_str())
+            let value: HttpApiDefinitionRequest = serde_json::from_str(definition_str.as_str())
```
afsalthaj, line 92: I wouldn't rename something in a PR that's concerned with something else unless if it's mandatory.

golem-worker-service-base/src/api_definition/http/http_api_definition.rs
```diff
@@ -17,14 +17,76 @@ use crate::parser::{GolemParser, ParseError};
...
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CompiledHttpApiDefinitionRequest {
+    pub id: ApiDefinitionId,
+    pub version: ApiVersion,
+    pub routes: Vec<CompiledRoute>,
+    #[serde(default)]
+    pub draft: bool,
```
afsalthaj, line 89: Compiled version doesn't need to be a request anymore. Compiled is created from a fully formed http definition

golem-cli/src/oss/clients/api_definition.rs
justcoon: this is related to OpenAPI spec changes

golem-worker-service-base/src/api_definition/http/http_api_definition.rs
justcoon: thanks, removed

PR #804, opened 2024-08-21
Fix unit returns

afsalthaj:
Making sure, given a function `foo` that returns `unit`, it responds with a http response with empty body.

Handling `Unit` was done before, and therefore this is more of a bug fix,  with more tests at worker-service side which decides what to do with a result from `golem-rib` that can either be a `Unit` or `TypeAnnotatedValue`

Following is possible

```scala
// foo returns unit
let x = foo("bar");
x

```

This will return a response with empty response body. I think we cannot expect the user to never use the result from a function which was `Unit`. Intuitively,

```scala
foo("bar")
```

is isomorphic to

```
let x = foo("bar");
x
```

### DIFF
--- a/golem-worker-service-base/src/worker_service_rib_interpreter/mod.rs
+++ b/golem-worker-service-base/src/worker_service_rib_interpreter/mod.rs
@@ -1053,6 +1053,7 @@ mod tests {
         assert_eq!(&value1, &expected);
     }
 
+
     #[tokio::test]
     async fn test_evaluation_for_function_returning_unit() {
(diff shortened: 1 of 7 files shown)


golem-worker-service-base/src/worker_service_rib_interpreter/mod.rs
```diff
@@ -182,7 +182,7 @@ mod tests {
         async fn evaluate_with_worker_response(
             &self,
```
afsalthaj, line 183: Tests!

jdegoes (approved):

Issue #806, opened 2024-08-21
Component should have timestamp


justcoon:
similar to #764, propagate createdAt timestamp for component

PR #807, opened 2024-08-21
component with created_at

justcoon:
related to #806, and #802

 OpenaAPI and gRPC API 
 - added createdAt timestamp to Component- data are already there in DB, they were not just propagated


vigoo:
Make sure to apply the same decision here as to the other (for example making the field optional)

nicoburniske (approved):

PR #821, opened 2024-08-23
Sharding not ready

noise64:


golem-worker-executor-base/src/services/scheduler.rs
```diff
@@ -68,16 +68,22 @@ impl SchedulerServiceDefault {
...
+            tokio::spawn(async move {
+                loop {
+                    tokio::time::sleep(process_interval).await;
+                    if svc.shard_service.is_ready() {
+                        let r = svc.process(Utc::now()).await;
+                        if let Err(err) = r {
+                            error!(err, "Error in scheduler background task");
+                        }
+                    } else {
+                        warn!("Skipping schedule, shard service is not ready")
```
vigoo, line 82: This will produce a lot of warnings in "normal" scaling up/down situations I think, but we can leave it for now

golem-worker-executor-base/src/services/shard.rs
```diff
@@ -120,25 +171,47 @@ impl ShardServiceMock {
...
 
-    fn current_assignment(&self) -> ShardAssignment {
-        ShardAssignment::default()
+    fn opt_current_assignment(&self) -> Option<ShardAssignment> {
+        tracing::info!("ShardServiceMock::opt_current_assignment");
+        None
+    }
+}
+
+fn sharding_not_ready_error() -> GolemError {
```
vigoo, line 213: So as the `WorkerService` below can fail with this error now, the PR as is is not good enough yet - because the worker service (the actual service) needs to retry invocations until the sharding gets properly assigned. 

So basically I think the system needs to behave the same way as currently with `GoelmError::InvalidShardId` when sharding is not ready. Otherwise we will end up bubbling this error up to the user. 

Unfortunately we cannot just emit `InvalidShardId` because that contains a `ShardId` field which we cannot compute without knowing the number of shards.

So I propose to:
- hold this until the serialization tests are done, so we feel safer
- introduce a new error case somehow
- and as part of this PR, also handle it in the worker service
- additionally, probably the sharding 'readiness' should be checked in the beginning of the gRPC handlers, not randomly failing somewhere in the middle of handling them.

vigoo (changes requested):

vigoo (approved):
Good to merge once it's green

PR #822, opened 2024-08-26
 Resolve variable conflicts, fix select field type inference issues, and generic handling of type extraction and type annotation

### LINKED ISSUE #786: Type Annotation:  Better type inference, avoid flexibility (with InferredType::Unknown) by allowing users to annotate types
This is mainly to support dynamic input `reliably`

Most probably we need to be able to allow users to write complex annotations

```scala

 let x = { body : { id: "bId", name: "bName", titles: request.body.titles, address: request.body.address } };
 call("foo");
 x
```

Here you can see Rib Compilation produces an Expr that consist of `InferredType::Unknown` for titles and address. In reality, address is a `Str` and titles is a `List`. But this is impossible to infer. The best Rib could do as of now is request is a record with body field, which is a record that has titles and address whose types are InferredType::Unknown.

There are various ways to handle this:

## Type Annotation strategy:

This is the most reliable strategy as far as I see.

Example: Users can write the above Rib and try and upload to see an error which says titles and address types are unknown, or couldn't be resolved, and this allows users, and ask them to be explicit with types. All we need as far as I see are the support to every type in `AnalysedType`. 

```scala
   let x = { body : { id: "bId", name: "bName", titles: request.body.titles as list<Str>, address: request.body.address as Str } };
 call("foo");
 x

```

When it comes to record there are various ways users might need to write this

Say address was a record
```scala

 let address = request.body.address as record<("street" -> Str, "house_number" -> u64)>
 let x = { body : { id: "bId", name: "bName", titles: request.body.titles, address: request.body.address  as record<()>} };
 call("foo");
 x
```

Most probably type unification phase becomes  way simpler that it doesn't have `InferredType::Unknown` to handle. Type checking phase also can take an easier decision that it will choose to simply phase. After all the scanning to-and-fro multiple times with push-down and pull-up and identifier-inference etc, there isn't a reason to have InferredType::Unknown

## Flexibility Strategy

We don't fail when users upload the raw expression without types. But that expression will have uninferred types. 
This implies the `RibInputTypeInfo` (the input requirements of the Rib script) will be in terms of InferredType and not in terms of `AnalysedType`. Currently `RibInputTypeInfo`is `HashMap<String, AnalysedType>` (this is part of an output of compilation along with rib-byte-code). This is simply impossible because, as mentioned above, `titles` and `address` types are unknown.

* So make input requirement in terms of `InferredType` as the first step, so that compilation don't fail and we allow Rib to return this requirement back to user.
* During an actual API request, the request body can take any form. Rib can try to use this RibInputTypeInfo to form the real RibInput (HashMap -> TypeAnnotatedValue) and if this fails, it can infer from the JSON, what could be the TypeAnnotatedValue. This has issue when it comes to optional fields etc. But probably Option is not a thing at all when it comes to such an inference 
* Now somehow we made the request input a HashMap<String, TypeAnnotatedValue> and we feed this to rib-interpreter along with the byte code. During instruction execution, Rib knows these types anyway which exist in the environment and playing with this value shouldn't fail in the stack based interpreter.

The issue here is, such a code is very hard to maintain. There are possibilities of failure still, and the compiler will find it really hard to properly compile, or in other words, its way better to ask users to annotate the types if there exist an uninferred type during compilation.

### LINKED ISSUE #810: Make sure variables in pattern match don't conflict with others

### LINKED ISSUE #832: Avoid defaulting to U64 (instead use type annotation)

afsalthaj:
Fixes #810,   #825,  #826, #783, #797, #828, #829, #830, #832, and #786

## Major logical changes

- Handle variable conflicts in match binding variables
- Make sure users can annotate types (record etc are remaining) 
- Fix bugs in variable reassignment type inference
- Make sure block expressions are inferred. 
- Introduce type refinement and generic way to extract inner types for push down and pull up phases
- More tests in type inference module, type push down and type pull up
- Earlier compile time error for obvious errors when passing arguments to functions
- Fix stack safety issues with nested records and improve performance of parser
- Slight cleanups in compiler error messages

It's not into 100 percent perfection, but managed to improve significantly. Other `TODO` are ticketed properly.

## TODO
 

- [x] FIx #832 

- [x] Probably fix #786


afsalthaj:
Having a failing test in rib module. Fixing this properly requires a few more hours

jdegoes (approved):

PR #824, opened 2024-08-26
Golden tests for worker executor serialization

### LINKED ISSUE #811: Golden tests for worker serialization
We need to write golden tests to guarantee we are not breaking worker serialization backward compatibility from 1.0.0.

For this, we have to store binary data serialized with 1.0.0 of various data structures, and have deserialization tests for them.

An initial list of what have to be included:

- every case of the `Oplog` type
- `WorkerStatusRecord` 
- `RedisPromiseState`
- `ScheduledAction`
- Types used in oplog payloads. These should be collected as part of the task, but some of them: 
  - `blob_store::ObjectMetadata`
  - `SerializableError`
  - `SerializableStreamError`
  - `SerializableIpAddresses`
  - `WitValue`

vigoo:
Resolves #811


golem-worker-executor-base/src/durable_host/wasm_rpc/serialized.rs
```diff
@@ -15,11 +15,27 @@
...
     Completed(Result<WitValue, RpcError>),
 }
+
+impl PartialEq for SerializableInvokeResult {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Failed(e1), Self::Failed(e2)) => e1 == e2,
+            (Self::Pending, Self::Pending) => true,
+            (Self::Completed(Ok(r1)), Self::Completed(Ok(r2))) => {
+                let v1: Value = r1.clone().into();
```
noise64, line 33: is clone unavoidable here? maybe *r1 == *r2 or some other workaround?

golem-worker-executor-base/tests/compatibility/v1.rs
```diff
@@ -0,0 +1,1890 @@
...
+        )
+        .unwrap();
+    let encoded = serialize(&value).unwrap();
+    file.write_all(&encoded).unwrap();
+    file.flush().unwrap();
+}
+
+#[test]
+pub fn worker_status() {
+    let ws1 = WorkerStatus::Running;
```
noise64, line 126: are there also tests for not introducing new enum cases?

golem-worker-executor-base/src/durable_host/wasm_rpc/serialized.rs
vigoo: No idea did not try to optimize this code :)

golem-worker-executor-base/tests/compatibility/v1.rs
vigoo: I don't think we need to explicitly deny creating new enum cases. The test covers all the ones that were known for 1.0. So what we are testing is that we have to be able to deserialize all the possible cases created by 1.0. This is covered here. So if you can introduce a new case without breaking this, that's fine (I believe with bincode we can do it if we add the case to the end, but haven't tried).

PR #833, opened 2024-08-30
Type binding

afsalthaj:
The PR is not showing enough changes since the base branch is https://github.com/golemcloud/golem/pull/822, however shows the fundamental logic addition and test changes  (that verifies the change) for easier review.

* Able to specify if there is an ambiguity in types (type annotation), as our stack based interpreter needs to know the actual types and cannot guess (we made some hacks during release 1.0 such as fall back to u64 if user specified a number which can be u8, u16 etc but this leads to a lot of corner cases, and we decided to take this path). Example returning an `empty list` is never possible without this strategy.

```scala
let my_list: list<u8> = [];
{status: 200, body: my_list}
```

* There is no need to specify the type if it can infer from any function calls to worker. The following will work

```scala
let x = 1;
call_my_function(x) 

```

* Limited support, as you need to write a let binding to specify the types. But this PR will unblock users to a great extent. 

Obviously, it will be a bit more user friendly for simpler expressions to work without type annotation. Example, with this PR, we see we have to specify the type of x. 

```
let x : u8 = 1;
x
```
Avoiding u8 would make it sound user friendly, but making ways to allow such thing will introduce a lot of hacks into our type inference logic, and thereby the future maintainability of Rib will be in question, unless we find a non hacky structured solution to bring such user friendliness. Not to mention, Rib is mainly used to call worker functions and to a relatively good extent types will be inferred.

We had some flexibility in the code previously, to get release 1.0 going with whatever worker bridge used to do until then, but introduced corner cases.

### DIFF
--- a/golem-rib/src/type_inference/mod.rs
+++ b/golem-rib/src/type_inference/mod.rs
@@ -612,7 +612,8 @@ mod type_inference_tests {
         #[test]
         fn test_select_field_type_inference() {
             let rib_expr = r#"
-          let x = { foo : 1 };
+          let n: u64 = 1;
+          let x = { foo : n };
           x.foo
 
           "#;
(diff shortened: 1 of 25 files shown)


golem-rib/src/type_inference/mod.rs
```diff
@@ -477,7 +477,7 @@ mod type_inference_tests {
         #[test]
         fn test_identifier_type_inference_multiple_re_assign() {
             let rib_expr = r#"
-          let x = 1;
+          let x: u64 = 1;
           let y = x;
```
afsalthaj, line 481: You can see its ambiguous in these expressions as we don't make any function calls, to resolve the types on its own.

afsalthaj:
There are things to clean up, but this is definitely 10x improved version compared to release 1.0.0

The following needs to be done:

- [ ] Add TypeName node in Let Expr binding, its Protobuf versions etc (I didn't do this because most of the tests don't read from any persistent expr store. Now we need to ensure these annotations in Expr is kept intact when writing back to DB etc
- [ ] Make sure CLI tests pass

PR #834, opened 2024-08-30
Fix top level exported function names in golem-cli

noise64:


golem-cli/src/model/component.rs
```diff
@@ -169,7 +169,7 @@ pub fn show_exported_function(prefix: &str, f: &AnalysedFunction) -> String {
         let name = &f.name;
 
         DisplayNamedFunc {
-            name: format!("{prefix}{{{name}}}"),
+            name: format!("{prefix}{name}"),
```
vigoo, line 172: We need the `{}` for non-toplevel ones. See the https://github.com/golemcloud/golem/blob/main/golem-rib/src/function_name.rs file for examples of everything that's possible as a function name.

vigoo: (It is always needed when it's not a top-level function because the package/interface part can have a version, which can be anything according to the semver spec, which includes arbitrary prefixes and dots)

PR #837, opened 2024-08-30
Remove hacks around compiling worker name and idempotency key expressions (rib)

afsalthaj:
### Context
Worker name expressions when compiled gives the global input requirements (similar to any Rib expression), however has the chances over Uninferred types (Example: "my-worker-${request.path.id}", where we have no clue what is the type of `id`. Obviously, we considered it as a Str, but just to do that we had to do more hacks. Idea of falling back to Str and the corresponding hacks are removed in this PR

#### More details 

On a high level, this implies compilation will fail and it "falls back" to "Str" as a requirement (replacing this Unknown). This happens only when there is an unknown type. These expressions to form the text names doesn't have a valid metadata to rely on either as it doesn't make any function calls to help infer the types.

This fallback strategy was an interception to original compilation logic using the function `compile_pure` (simply means, this will never fail, and will always infer as a Str for whatever it tries to grab from Http input request path or body. 

Not to mention, situations like these: worker-name Rib expression compilation consider request.path.id as a Str while in response-mapping expression, it can be used as a U64, and this is all very confusing and hard to maintain.

This PR is not mandatory as this is a fall-back strategy (i.e, if types are uninferred), and now with https://github.com/golemcloud/golem/pull/822, it never goes to this fallback codebase.

### Type Annotation
Thanks to the introduction of type annotation, now that we don't need to keep this hack anymore

 Now users can annotate the type


golem-worker-service-base/src/http/http_request.rs
```diff
@@ -777,7 +777,7 @@ mod tests {
 
         let api_specification: HttpApiDefinition = get_api_spec(
             "foo/{user-id}",
-            "${let userid: u64 = request.path.user-id; let max: u64 = 100; let zero: u64 = 0; let one: u64 = 1;  let res = if userid>max then zero else one; \"shopping-cart-${res}\"}",
+            "${let userid: str = request.path.user-id; let max: u64 = 100; let zero: u64 = 0; let one: u64 = 1;  let res = if userid == \"bar\" then one else zero; \"shopping-cart-${res}\"}",
```
afsalthaj, line 780: We have other tests that consider them as a number

jdegoes (approved):

PR #838, opened 2024-08-31
A few more tests and simplify type binding for numbers

afsalthaj:
More of a suggestion PR
- [x] Have TypeName in Number expr


golem-api-grpc/proto/golem/rib/expr.proto
```diff
@@ -73,6 +73,7 @@ message LiteralExpr {
 
 message NumberExpr {
   double float = 1;
+  optional TypeName type_name = 3;
```
justcoon, line 76: should not it have index 2 ?

justcoon (approved):

PR #854, opened 2024-09-03
Rib Type inference Improvement 

### LINKED ISSUE #864: Implement a well defined fix point of type inference
Scanning to and fro and inferring types are not optimal now. Just with a proper addition of fix point for inference and extract expr, we willl cross this final hurdle and then on can handle any complexities.

afsalthaj:
Fixes #864  #865  and #866 

- [x] Introduce a single `fix-point` for a subset of compilation phases w.r.t `Expr`

    * Break at `f(expr) = expr`, and send `expr` to `unification`, where `f` is a subset of compilation phases that requires multiple scanning
    * This approach is better than a well defined termination condition (such as try until fully inferred, because it can be an invalid code and we need to stop obviously)

- [x]  Separate Type Bind compilation phase. 
    * Type binding is the binding of user specified type to the actual expression, before it begins with rest of the compilation including name binding. This step was squashed along with parser (note that we are persisting these info using grpc types, and these `From` and `To` between grpc types also had binding phase logic) making it all complex. I wanted to unblock users for hackathon and therefore this was the shortest pathway, and now all of that is fixed, and we have a separate phase of binding.

- [x] Avoid using random type push down in other compilation phases (explained further down below).

- [x]  Simplify identifier inference (propagating types of identifiers with specific ids across the code - top down and bottom up)
        
    * This  is not a bug fix, but code improvement which is detailed below.    
    * It had very nuance/subtle logic which differs between top-down and bottom-up scans to inform type info between each identifiers in the Rib program . However,  it doesn't matter whether you are going top down or bottom up as we are simply accumulating types using `AllOf` and `OneOf` and except for the direction of traversal, there shouldn't be any other difference. 

   * There was a call to type-push-down in this basic phase, and I managed to remove it after the above simplification. This slightly slows down compilation though. It is because after propagation of types between identifier,  we are not fully pushing down the type info to the structures identified by these identifiers yet. And it requires another pass to make that happen. However, that's managed by above fix-point logic which is nothing but a `loop`. This is a better approach for maintainability of this complex algorithms.    
 
- [x]  Better modularisation of compilation phases: Here is the current state

An example of better inference here:

image

## TODO

- [x] 4 flaky tests due to the order of the types in the list (easy to fix)
- [x] Ensuring fast termination

Compilation taking 2-5 seconds for the following. This can get exponential,  and this must be compiler trying its best while accumulating big data structures. I will need some time to dig into this!

```scala

 let x = { body : { id: "bId", name: "bName", titles: request.body.titles, address: request.body.address } };
 let result = foo(x);
  match result { 
     some(value) => "personal-id", 
    none =>  x.body.titles[1] 
  }

```

Ideally this shouldn't have happened, as the number of iterations shouldn't change.

```

running 1 test
test type_inference::type_inference_tests::record_tests::test_record_type_inference_select_with_function_call ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 380 filtered out; finished in 2.03s
```


golem-rib/src/parser/let_binding.rs
```diff
@@ -178,7 +175,7 @@ mod tests {
                 Expr::let_binding_with_type(
                     "foo",
                     TypeName::U8,
-                    Expr::Identifier(VariableId::global("bar".to_string()), InferredType::U8)
+                    Expr::Identifier(VariableId::global("bar".to_string()), InferredType::Unknown)
```
afsalthaj, line 178: Tests that shows parser doesn't do much with InferredType

golem-worker-service-base/src/worker_service_rib_interpreter/mod.rs
```diff
@@ -905,7 +905,7 @@ mod tests {
 
         let _expected = TypeAnnotatedValue::Str("bStreet".to_string());
 
-        assert!(result.contains("Types do not match. Inferred to be both Record([(\"street\", Str), (\"city\", Str)]) and List(Unknown)"));
+        assert!(result.contains("Types do not match. Inferred to be both List(Str) and Record([(\"street\", Str), (\"city\", Str)])"));
```
afsalthaj, line 908: Tests that shows inference is way better

golem-rib/src/type_inference/mod.rs
```diff
@@ -956,17 +977,13 @@ mod type_inference_tests {
...
               let x: u64 = 1;
 
-              let fallback: option<option<u64>> = none;
-
               match err(x) {
-                err(_) => fallback,
+                err(_) => none,
                 ok(_) => some(some(x))
               }
             "#;
```
afsalthaj, line 989: Test that shows better type inference. With multiple passes of pushing down and propagating types `none` inferred to be `Option(Option(str))`

justcoon (approved):

Issue #586, opened 2024-06-13
Cache worker metadata in worker bridge


vigoo:
See #585 for measuring the effect

afsalthaj:
More details; 

ExecutionContext can be cached with  component metadata mentioned in a binding of an API definition. 


Note that API gateway or worker bridge can deal with multiple API Definiton IDs (pretty much multi tenancy) especially when it comes to cloud. This implies we will have loads to cache. Also, ARC is the best approach here. It's adaptive and take into consideration of situations where there is large gaps between accesses but it's accessed for sure at this time gaps. If we don't do this, the gateway users will feel the slowness whenever they hit their endpoint to invoke worker function at regular (but larger) intervals. ARC will be automatically LRU and size aware. Not sure how many Rust libraries have this support. But it's easy enough to implement (max 3 hours). 

There should also be a function in worker service trait that accepts arguments (mainly useful for worker bridge) such that there is no need to look up function after loading component metadata again and again because that information is preloaded in evaluation-context of worker bridge. This will also avoid double (duplicate) parsing of function names when calling worker through worker bridge.

afsalthaj:
Please find the explanations in PR and in slack: https://github.com/golemcloud/golem/pull/630
and in here: https://zivergeteam.slack.com/archives/C057S2E4XT5/p1720013155994449

afsalthaj:
As the next step, I did try to remove the download of metadata in Rib evaluation, however bumped or uncovered following bugs with the latest OSS release:


https://github.com/golemcloud/golem/issues/638 (in progress)
https://github.com/golemcloud/golem/issues/641 (under review)
https://github.com/golemcloud/golem/issues/639 (not mandatory for now)

PR #871, opened 2024-09-05
move dbconfig to golem-common module

irach-ramos:
A bit of refactoring related to [#586](https://app.zenhub.com/workspaces/golem-cloud-65f5b68988aba80f7d690199/issues/gh/golemcloud/golem/856)


golem-service-base/src/db/mod.rs
```diff
@@ -13,17 +13,32 @@
...
+use std::ops::Deref;
 use std::path::Path;
 
 use sqlx::postgres::{PgConnectOptions, PgPoolOptions};
 use sqlx::sqlite::{SqliteConnectOptions, SqlitePoolOptions};
 use sqlx::{Connection, Executor, PgConnection, Pool, Postgres, Sqlite, SqliteConnection};
 use tracing::info;
 
-use crate::config::{DbPostgresConfig, DbSqliteConfig};
+use golem_common::config::{DbPostgresConfig, DbSqliteConfig};
```
noise64, line 24: are these wrappers useful? maybe we should just add `SqliteConnectOptions::from_config`, then we could get rid of them.

irach-ramos: You mean to add into `sqlx::sqlx_sqlite::SqliteConnectOptions` a method `from_config` ?

irach-ramos: This is what the wrapper and Deref are doing, private to this mod only.

noise64: technically it does, i just see no point in having `DbPostgresConfigWrapper`. from this name alone one cannot find out that this is a wrapper created only to provide "config to opts" as a From instance, and this could be helped with more precise naming, i just think it does not worth it now.

when config was not extracted, then i think this was okay, but now it is just plumbing (wrapping for the purpose of unwrapping) with no good reason (at least for me).

i would simply just add `xxx_options_from_config`, or `xxx_config_to_options` for sqlite and pg

irach-ramos: You mean to define something like 

```
Db[Postgres|Sqlite]Config:: xxx_options_from_config -> [Pg|Sqlite]ConnectOptions
Db[Postgres|Sqlite]Config:: xxx_config_to_options -> [Pg|Sqlite]ConnectOptions 
```
In `Db[Postgres|Sqlite]Config` ?

If it is that, we need to add sqlx dependency in golem-common, which I understood we want to avoid.

noise64: https://github.com/golemcloud/golem/pull/875
i just meant using simple functions :)

noise64 (approved):

Issue #844, opened 2024-09-02
Add "diagnose" command to golem-cli


noise64:
The command should check for required and optional tooling for general and language specific golem development.
The check should validate the existence and versions of the tooling (and could even suggest install steps / links).

PR #877, opened 2024-09-06
Golem-cli diagnose command

noise64:
Resolves: https://github.com/golemcloud/golem/issues/844

Add golem-cli diagnose command which checks the required tooling for a specific language (toolchain).

Some examples:
image
image


afsalthaj (approved):

golem-cli/src/diagnose.rs
```diff
@@ -0,0 +1,893 @@
...
+                .find_map(|e| e.ok().and_then(|e| detect_in_dir(e.path())));
+
+            if language.is_some() {
+                return language;
+            }
+        }
+
+        // Searching - up
+        {
+            let starting_dir = std::env::current_dir().expect("Failed to get current dir");
```
afsalthaj, line 139: Trying to understand things here. 

This is for being able to run diagnose command from anywhere in the project ?

golem-cli/src/diagnose.rs
```diff
@@ -0,0 +1,893 @@
...
+            Tool::Python => vec![],
+            Tool::Rustc => vec![Tool::Rustup],
+            Tool::Rustup => vec![],
+            Tool::RustTargetWasm32Wasi => vec![Tool::Rustc],
+            Tool::TinyGo => vec![Tool::Go],
+            Tool::WasiSdk => vec![],
+            Tool::WasmTools => vec![Tool::Cargo],
+            Tool::WitBindgen => vec![Tool::Cargo],
+            Tool::Zig => vec![],
+        }
```
afsalthaj, line 553: Very minor: I would have personally called it dependency-graph (because i see it as full dependency graph represented using adjacency list matrix strategy). 



It implies this adjacency-list is kept mainly to help out with looping through all dependencies in the graph and collect all of the dependencies. It would have been easier to understand if we kept the real graph structure as it is

afsalthaj: Ok, I see why you did this  (and named it that way) later down the line

golem-cli/src/diagnose.rs
noise64: i wanted to make it work for our current guides / examples, and there are 3 main kind of projects currently:
- simple ones with only one component,
- multi components examples with a root project
- lower level doc examples with a project folder, where every subfolder is a separate project

also, i wanted to make it work in case you are inside of some project src folder, for some examples we have those, for some not.

so in the end it search in both directions for hints

golem-cli/src/diagnose.rs
noise64: it is a simple dep list

PR #885, opened 2024-09-09
Indexed storage test suite and fixes

### LINKED ISSUE #861: Implement a generic test suite for IndexedStore implementations

vigoo:
Resolves #861

### DIFF
--- a/golem-worker-executor-base/src/services/oplog/primary.rs
+++ b/golem-worker-executor-base/src/services/oplog/primary.rs
@@ -76,7 +76,7 @@ impl PrimaryOplogService {
     }
 
     pub fn get_worker_id_from_key(key: &str, component_id: &ComponentId) -> WorkerId {
-        let redis_prefix = format!("worker:oplog:{}:", component_id.0);
+        let redis_prefix = format!("{}:", component_id.0);
         if key.starts_with(&redis_prefix) {
             let worker_name = &key[redis_prefix.len()..];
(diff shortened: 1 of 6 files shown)


golem-worker-executor-base/src/services/oplog/primary.rs
```diff
@@ -76,7 +76,7 @@ impl PrimaryOplogService {
     }
 
     pub fn get_worker_id_from_key(key: &str, component_id: &ComponentId) -> WorkerId {
-        let redis_prefix = format!("worker:oplog:{}:", component_id.0);
+        let redis_prefix = format!("{}:", component_id.0);
```
noise64, line 79: is this breaking / okay?

vigoo: Previously `scan` was returning full redis "keys" instead of the logical keys within the selected namespace which is a bad API. So I changed that, now `scan(oplog, pattern)` will not return `[worker:oplog:key1]` just `[key1]`. The prefix which is derived from the namespace is an implementation detail of the storage implementation. 

We were previously dropping this prefix here, now we don't have to. (just the component-id part which is not part of the namespace but the actual key)

vigoo: So it's not breaking anything just makes the storage interface more reasonable.

noise64 (approved):

PR #886, opened 2024-09-09
sqlite kv-storage implementation

irach-ramos:


golem-common/src/metrics.rs
```diff
@@ -405,3 +405,81 @@ pub mod api {
         };
     }
 }
+
+pub mod sqlite {
```
justcoon, line 409: i would consider to do it like generic 'db' metrics

with `type` label

so later we do not have to create set of root level metric for different DB/SQL like backend

golem-worker-executor-base/src/storage/keyvalue/sqlite.rs
```diff
@@ -0,0 +1,751 @@
...
+
+        SqlitePool::init(&pool).await?;
+
+        Ok(SqlitePool { pool })
+    }
+
+    async fn init(pool: &SqlitePoolx) -> Result<(), Error> {
+        sqlx::query(
+            r#"
+            CREATE TABLE IF NOT EXISTS kv_storage (
```
justcoon, line 313: is there reason why this is not in db migrations script but in code ?

vigoo: Not sure how it would look like with db migration scripts, but I would like to have these storage trait implementations self contained - so you can drop in different implementations without any external requirements.

irach-ramos: I think the idea is to allows the database to be automatically initialise when the application starts, without needing a separate step to run migration scripts, which is faster and without external dependencies.
For this small-scale sqlite usage (for POCs or local development), and  without having complex schema definitions, I think migrations aren’t needed.
Probably we can go with this approach and I think we can later on add the migration part if needed.

justcoon (approved):

PR #888, opened 2024-09-10
Rib type info

### LINKED ISSUE #787: Return `RibInputTypeInfo` when user uploads API definition

afsalthaj:
Fixes #787


openapi/golem-service.yaml
```diff
@@ -2929,9 +2929,32 @@ components:
...
+        workerNameInput:
+          $ref: '#/components/schemas/RibInputTypeInfo'
+        idempotencyKeyInput:
+          $ref: '#/components/schemas/RibInputTypeInfo'
+      required:
+      - componentId
+      - workerName
+      - response
+      - responseMappingInput
+      - workerNameInput
```
justcoon, line 2954: responseMappingInput, workerNameInput - those look like are new required attributes, which will break backward compatibility

afsalthaj: Its on the response and not on the input

afsalthaj: How can adding a required field on response of every API breaks compatibility?

afsalthaj: No change is done on the input data types (HttpApiDefinitionRequest or HttpApiDefinition)

justcoon: I know, I was discussing same situation with createdAt fields also in respinses only, and general opinion was it shold be optional

justcoon (approved):

Issue #835, opened 2024-08-30
Fix version api


justcoon:
version endpoint to not return proper version (based on release)
```
% curl http://localhost/version
{"version":"0.0.0"}
```

Issue #819, opened 2024-08-22
CLI should try to pretty-print callstacks when invocation fails


vigoo:
Example failure:

```
[...]

PR #926, opened 2024-09-15
Golem-cli improvements

noise64:
Related issues:
 - https://github.com/golemcloud/golem/issues/835
 - https://github.com/golemcloud/golem/issues/924
 - https://github.com/golemcloud/golem/issues/819

- use git tags for version info with fallback to cargo pkg and 0.0.0
- unified, colorized and formatted text output
- extracted some common main logic, so error handling are the same
  
  
Some examples of text formatting:

image

image

image

image

image

image


golem-common/src/golem_version.rs
```diff
@@ -0,0 +1,12 @@
+pub use git_version::git_version;
+
+#[macro_export]
+macro_rules! golem_version {
+    () => {
+        golem_common::golem_version::git_version!(
+            args = ["--tags", "--abbrev=0"],
+            cargo_prefix = "",
+            fallback = "0.0.0"
```
justcoon, line 9: should not be fallback 1.0.0 (as we have 1.0.+ released)

noise64: i would keep it as "not known", given:
 - when git cloned repo is used, we can assume git is installed and git describe will work
 - when using cargo install it should use the CARGO_PKG_VERSION fallback (the cargo_prefix arg triggers this, but we should test this)
 
so in general the fallback is a "should not happen" value, but if it happens i would rather be able to see a "zero" value for debug, then assuming any version

golem-cli/tests/text.rs
```diff
@@ -143,35 +144,19 @@ fn text_component_add(
...
-    assert_eq!(*lines.get(2).unwrap(), "Exports:");
-    assert_eq!(
-        *lines.get(3).unwrap(),
-        "\tgolem:it/api.{get-environment}() -> result<list<tuple<string, string>>, string>"
-    );
-    assert_eq!(
-        *lines.get(4).unwrap(),
-        "\tgolem:it/api.{get-arguments}() -> result<list<string>, string>"
-    );
+    assert!(regex_contains(
```
vigoo, line 155: I think the chang of this file is my favorite part :) it was so hard to maintain

vigoo (approved):

vigoo (approved):

PR #937, opened 2024-09-17
Introducing ephemeral component type

### LINKED ISSUE #932: Introduce the concept of ephemeral components
In this step *component metadata* receives a new property, component type which can be `Durable` or `Ephemeral`. Old components are treated as `Durable`.

The REST API, gRPC API and the CLI is updated to support providing and querying this property.

vigoo:
Resolves #932

### DIFF
--- a/golem-component-service/db/migration/postgres/002__add_component_type.sql
+++ b/golem-component-service/db/migration/postgres/002__add_component_type.sql
@@ -1,2 +1,2 @@
 ALTER TABLE component_versions
-    ADD COLUMN IF NOT EXISTS integer component_type NOT NULL DEFAULT 0;
+    ADD COLUMN IF NOT EXISTS component_type integer NOT NULL DEFAULT 0;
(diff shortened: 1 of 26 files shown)


openapi/golem-service.yaml
```diff
@@ -1608,6 +1610,8 @@ paths:
               properties:
                 name:
                   type: string
+                component_type:
```
noise64, line 1613: shouldn't this be also optional, like in grpc?

noise64: ah ignore me, forgot that it is in the required  parts

noise64 (approved):

golem-component-service/db/migration/postgres/002__add_component_type.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE component_versions
+    ADD COLUMN IF NOT EXISTS component_type integer NOT NULL DEFAULT 0;
```
justcoon, line 2: can `component_type` change with version? if not this column should be added to `components` table

justcoon: ok, i see now that there is component type in update, so it can be updated, then it is ok ...

vigoo: it can

PR #940, opened 2024-09-18
Precise Rib Parsing & Errors

afsalthaj:
## Precise error messages 

For an invalid syntax such as the following:

```rust
match foo
  ok(x) => x,
  err(x) => x
}
```

Main branch behaviour:

```scala
Failed to parse: "${match request.id ok(x) => x, err(x) => x }}"
```

## PR branch behaviour

```scala
Parse error at line: 2, column: 1
Unexpected `o`
Expected whitespace or `{`
Invalid syntax for pattern match
```

* Points to the exact line number and column number using Stream Input `position::Stream`
* Better errors (as much as possible). There are corner cases or ambiguous cases such as an invalid expression  `{a` could be  a flag or a record, and it cannot say the exact error but just saying `missing :` or `missing }` in specific location in the code. Also I am stopping it here as this is far better than what we have in main branch.
* Proper use of `eof` instead of partial parsing and fail if there is anything remaining.
* A few minor issues  are raised as part of fixing this issue. Example: https://github.com/golemcloud/golem/issues/943

## Performance improvement (although not the concern of this PR)
While this PR wasn't really concerned with optimization (and therefore no specific benchmark tests), just putting some numbers here to give the reviewers about the importance of the subtle changes done in this PR. Mainly this is an outcome that I had to reduce the backtracking as much as possible to avoid suppression of errors. Avoidance of backtracking involves specific decisions to be made in different parts of the parser.

|Rib Module |Total tests| Time Taken (Main branch) | Time Taken (PR branch) | Remarks |
|--|--|--|--|--|
|text(read+write)|87|~10s|~0.1s| All combinations - Ex: record of sequence, record of record, sequence of record etc. This is why significant difference compared to rest of the modules|
|parser(read)|171|~1s |~0.1s| These are mostly simple Rib expressions compared to roundtrip tests, yet there is difference here
| across-all-modules|400|14s |~0.3s|  Includes more tests for compilation and final execution of Rib.|

 In general, backtracking heavily slowed down Rib parsing before. That said, avoiding backtracking now resulted in subtle/nuance precise `attempts` in only certain parts of the code.  Even if this is all part of pre-compilation, 15s vs 0.1s just for test cases is significant and shouldn't be regressed in future as it is a sign of major issue in the logic. Therefore benchmark tests for Rib (separate task) are required to avoid  performance regression in future.

### TODO
- [x] Support position::Stream
- [x] Reduce backtracking
- [x] All existing test pass
- [x] Precise error tests
- [x] More negative tests
- [x] Some syntax errors in `Record` is showing error as `unable to parse block`. Need to see why.

### DIFF
--- a/golem-rib/src/parser/literal.rs
+++ b/golem-rib/src/parser/literal.rs
@@ -15,12 +15,12 @@
 use crate::expr::Expr;
 
 use crate::parser::literal::internal::literal_;
-use combine::{easy, parser, Stream};
+use combine::{parser, Stream};
 
 parser! {
-    pub fn literal['t]()(easy::Stream<&'t str>) -> Expr
+    pub fn literal[Input]()(Input) -> Expr
     where [
-        easy::Stream<&'t str>: Stream<Token = char>,
+        Input: Stream<Token = char>
     ]
     {
         literal_()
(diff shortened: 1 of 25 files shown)


golem-rib/src/parser/literal.rs
```diff
@@ -15,12 +15,13 @@
 use crate::expr::Expr;
 
 use crate::parser::literal::internal::literal_;
-use combine::{easy, parser, Stream};
+use combine::{parser, Stream};
 
+//  impl Parser<Input, Output = Expr> where Input : combine::Stream<Token = char>
```
justcoon, line 20: is this some leftover ?

vigoo (approved):

PR #945, opened 2024-09-18
Introducing ephemeral workers

### LINKED ISSUE #933: Introduce the concept of ephemeral workers
This step starts differentiating workers based on their component type. 

The following changes need to be made:

- Worker executor uses the new component type field and stores the component type in the worker state
- A new type of `WorkerId` has to be introduced and used in the invocation APIs. This will be called `TargetWorkerId` and supports to NOT specify the name part of the `WorkerId`.
    - Technically this can be a new type introduced in code and in OpenAPI, having the same structure but the name part optional
    - In the protobuf version of `WorkerId`  gRPC the name is not marked as `optional` so the generated Rust binding is not optional either. I think making it optional does not break the wire format, and then the protobuf `WorkerId` can be converted to either `WorkerId` or `TargetWorkerId` . This needs to be verified (including the binary compatibility)
- `TargetWorkerId` is passed down to worker executor - if the component type is `Ephemeral`, `TargetWorkerId` is resolved to a `WorkerId` with a random-generated name part (if the name is not explicitly given). If `Durable`, and the name is missing, it fails with an invalid request error.
- CLI needs to be updated to allow omitting the worker name (or accept component uri, etc) if the component type is ephemeral, otherwise give a descriptive error.
- Worker executor immediately drops the instance out of the active workers after invocation if the component type is `Ephemeral`
- Worker executor skips the replay in `prepare_instance` and just jumps to the end of the oplog if the component type is `Ephemeral`
- At least one new worker executor test is added that calls an ephemeral component without passing a worker name
- At least one new integration test is added that calls an ephemeral component without passing a worker name
- At least two new CLI tests are added that calls an ephemeral component without passing a worker name, and one that verifies the error handling

vigoo:
Resolves #933 

User facing changes:

- In the gRPC APIs, every invocation-related endpoint now allows passing `None` as worker name. This is binary compatible with the existing API (and this is verified by a test)
- On the HTTP APIs there are new variants of the two invocation endpoints for the case where the user is not specifying a worker name. So for example beside of `/v1/components/{component_id}/workers/{worker_name}/invoke-and-await` we have also `/v1/components/{component_id}/invoke-and-await`
- In the CLI, worker url and worker urn can be specified without a worker name now, (for example `worker://<component-id>`) and it is accepted for invoke commands. For other commands where specifying an exact worker by name is necessary, it ends up in an error.


vigoo:
@afsalthaj isn't the linked routing configurations already routing the new endpoints to worker service (based on their first rule?)

afsalthaj:
current worker-service apis (prior ephemeral workers) have paths like
`/v1/components/<component-id>/workers*`
if there is not */workers*  part, it is taken like component endpoint
new APIs are:
 `/v1/components/{component_id}/invoke-and-await`
`/v1/components/{component_id}/invoke`
witch will not match `/v1/components/<component-id>/workers*`

afsalthaj (approved):

PR #960, opened 2024-09-20
Support dynamic resource parameters in Rib

### LINKED ISSUE #872: Rib Robustness: Make sure Rib works with functions that are part of resource
* Make sure to be able to parameterise arguments to resource constructors
* Make sure to test functions that are part of resources with Rib

afsalthaj:
Fixes #872 

There were several ways to solve this problem. Here is how I approached it, and had more subtle details to handle than I expected. 

 * The introduction of `DynamicParsedFunctionName` (a concern that's only related to the Rib module) may contain Expr nodes within it. The `Expr::call` will now hold `DynamicParsedFunctionName`. 
 
 * Some parts might sound a repetition but they are not unsafe repetition, because we need to successfully form the `ParsedFunctionName` with the details in `DynamicParsedFunctionName` during Rib Instruction Execution. 

* `ParsedFunctionName` and associated functions are untouched, and therefore the change is _NOT_ affecting any other modules such as the worker executor. 

* `ParsedFunctionName` implies that it's a fully formed function name, making it all type-safe. 

* `Rib IR` is now responsible for creating the function names based on parameters. As you can imagine, it internally forms the `ParsedFunctionName`which can be used to invoke the function in worker-service. 

* The Instruction set has enough details on how to form the `ParsedFunctionName` safely. For example `rib::ir::FunctionReferenceType::IndexedResourceMethod(String, usize, String))` has details such as resource names, how many parameters to retrieve from the stack etc. 

* The resource parameters in the stack are `TypeAnnotatedValue`, and we convert them to `wasm-wave syntax` string and then create the `ParsedFunctionName`. T

* This may be slightly over-detailing, yet good to point out. `DynamicParsedFunctionName::parse` parsing is basically the mostly the same previous implementation of `ParsedFunctionName::parse(str)` , and obviously the difference there is resource parameters are `Expr` and not `String`. `ParseFunctionName::parse` simply delegates to  `DynamicParsedFunctionName::parse`, and modules such as worker-executor is using it and by that time, the `str` doesn't have any dynamic parameters (or unknown identifiers). This implies the result of DynamicParsedFunctionName::parse` will have theirs `Expr`s containing fully formed values. This can be safely converted to `str` using `rib::to_string(expr)` which is  `wasm-wave` format. 

* In other words, `ParsedFunctionName::parse` tests work as before. The only change in those text is in spaces. Example `{ foo: "bar" }` is changed to `{foo: "bar"}`. 

* `InvokeFunction` instruction is changed to first retrieve the the function name which is in stack created by `CreateFunctionName` instruction, before popping up the resource/method parameters from the stack and then do the actual invocation. 

* Another implication of all these change is, ensuring the type inference phase is aware of these `Exprs` that exist in the `DynamicParsedFunctionName`.

* I had to also understand certain [details](https://www.notion.so/golemcloud/First-Class-Resources-286810fb4b214c47b3476c58a1a42a0e?pvs=4) to jump through some more hurdles and get things working, which took 1.5 more days than expected.

### TODO

- [x] [More tests](https://github.com/golemcloud/golem/blob/dynamic_resource_parameters/golem-rib/src/interpreter/rib_interpreter.rs#L1432C1-L1597C6)
- [x] Cleanup code (as much as I could)
- [x] Fix tests (I had to make slight changes
- [x]  Make sure backward compatible
- [x] [Add test](https://github.com/golemcloud/golem/blob/dynamic_resource_parameters/golem-rib/src/interpreter/rib_interpreter.rs#L1603) for methods that are part of resource with zero constructor parameters. Fixed a [bug]([170dbcf](https://github.com/golemcloud/golem/pull/960/commits/170dbcfffed057fe38ebb5bb7ea753db00d8f4ac))
- [x] Other clean ups, such as unnecessary requirements of `Serialize` and `Deserialize` for `RibByteCode`

### Test cases

[Testing all the functions in the `cart` resource in a shopping cart component](https://github.com/golemcloud/golem/blob/dynamic_resource_parameters/golem-rib/src/interpreter/rib_interpreter.rs#L1432C1-L1597C6)

#### Note:
I had a different approach before (an approach where the resource parameter could be static or dynamic, contained within `ParsedFunctionName`, which broke many things, affected many other modules, broke many tests to later realise that the above approach is the safest and robust approach.


afsalthaj:
Somewhere I messed up with stack interpreter

```

---- interpreter::rib_interpreter::interpreter_tests::test_interpreter_for_pattern_match_on_tuple_with_wild_pattern stdout ----
thread 'interpreter::rib_interpreter::interpreter_tests::test_interpreter_for_pattern_match_on_tuple_with_wild_pattern' panicked at golem-rib/src/interpreter/rib_interpreter.rs:1422:64:
called `Result::unwrap()` on an `Err` value: "Expected 10 value on the stack"

---- interpreter::rib_interpreter::interpreter_tests::test_interpreter_for_pattern_match_on_tuple_with_all_types stdout ----
thread 'interpreter::rib_interpreter::interpreter_tests::test_interpreter_for_pattern_match_on_tuple_with_all_types' panicked at golem-rib/src/interpreter/rib_interpreter.rs:1389:64:
called `Result::unwrap()` on an `Err` value: "Expected 10 values on the stack"


failures:
    interpreter::rib_interpreter::interpreter_tests::test_interpreter_for_pattern_match_on_tuple_with_all_types
    interpreter::rib_interpreter::interpreter_tests::test_interpreter_for_pattern_match_on_tuple_with_wild_pattern

test result: FAILED. 408 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.36s
```

afsalthaj:
I focussed on getting Rib aware of dynamic parts in a function name which is "ns:name/interface.{resource1(\"hello\", 1, true).drop}", but now I can't see in metadata, what exactly are the types of those parameters in the `resource1`. May be I overlooked/underestimated something here.

golem-worker-service/src/worker_bridge_request_executor.rs
```diff
@@ -92,7 +92,7 @@ mod internal {
             .invoke_and_await_function_json(
                 &worker_id.into_target_worker_id(),
                 worker_request_params.idempotency_key,
-                worker_request_params.function_name.to_string(),
```
afsalthaj, line 95: Before, we had a string, which we then parsed  it to `ParsedFunctionName` and then converted back to `String` to call the invoke function in worker-service. A redundant roundtrip, and here we just avoid it.

afsalthaj: Making it elegant would make every function receive a nice `ParsedFunctionName` instead of `String` but then has 2 parsing and display call involved in hot path.

I haven't gone deeper into what if invoke function accepts a ParsedFunctionName until it reach worker-executor. But that's not the concern of this PR. Also `String` would be optimal especially there is a GRPC call and I can see why it is the way it is now.

golem-rib/src/compiler/ir.rs
```diff
@@ -60,6 +61,127 @@ pub enum RibIR {
     Negate,
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Encode, Decode)]
+pub enum FunctionReferenceType {
```
vigoo, line 65: These look very ugly with no field names.

Also, why do you need this "copy" of the `ParsedFunctionName` type?

golem-rib/src/function_name.rs
```diff
@@ -41,6 +43,26 @@ pub enum ParsedFunctionSite {
 #[derive(PartialEq, Eq, Clone)]
 pub struct SemVer(pub semver::Version);
 
+impl Serialize for SemVer {
```
vigoo, line 46: Where are we serializing this to json?

afsalthaj: This is re-used in the instruction set, and the instruction set (RibByteCode) is given `Serialize` and `Deserialise`.
However, I realise, why are we json serde RibByteCode. Let me see if I can remove this requirement.

golem-rib/src/compiler/ir.rs
afsalthaj: It's part of the instruction, and they are not copy of ParsedFunctionName.
I can add field names though

golem-api-grpc/proto/golem/rib/expr.proto
```diff
@@ -175,7 +175,7 @@ message CallExpr {
 
 message InvocationName {
    oneof name {
-     golem.rib.ParsedFunctionName parsed = 1;
+     golem.rib.DynamicParsedFunctionName parsed = 1;
```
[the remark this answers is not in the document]
afsalthaj: how? These are grpc types only for internal persistence.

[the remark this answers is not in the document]
afsalthaj: Ok, discard my question. In terms of backward compatibility, well the function name is different now because has `Exprs` in it, and it's a different structure. I need to test backward compatibility.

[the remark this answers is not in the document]
vigoo: What is internal persistence? If it is stored in any database, then this is a breaking change.

golem-rib/src/function_name.rs
vigoo: Yeah let's not have unnecessary derived serde instances, it just slows compilation down

afsalthaj:
There are a couple of more important clean ups to do, as I did the self review + add 1 more test

golem-rib/src/function_name.rs
afsalthaj: Removed

golem-rib/src/compiler/ir.rs
afsalthaj: Added !

golem-api-grpc/proto/golem/rib/expr.proto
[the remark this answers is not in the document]
afsalthaj: Need to test/fix backward compatibility , which is what is remaining.

afsalthaj:
Fixes #872

afsalthaj:
I will push 1 more change to avoid further "expects" that I found.

golem-rib/src/type_inference/function_type_inference.rs
```diff
@@ -68,8 +40,156 @@ pub fn infer_function_types(
...
+                        if let Some(AnalysedType::Handle(_)) = parameter_types.first() {
+                            parameter_types.remove(0);
+                        }
+                    }
+
+                    if parameter_types.len() == args.len() {
+                        for (arg, param_type) in args.iter_mut().zip(parameter_types) {
+                            check_function_arguments(&param_type, arg)?;
+                            arg.add_infer_type_mut(param_type.clone().into());
+                            arg.push_types_down()?
```
afsalthaj, line 145: This is not required.

golem-api-grpc/proto/golem/rib/expr.proto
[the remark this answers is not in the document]
jdegoes: Another way to handle this would be to detect breakage and recompile from source. Since the IR is not exposed to the user.

PR #961, opened 2024-09-20
Adding cassandra key value storage

irach-ramos:


golem-worker-executor-base/src/cassandra.rs
```diff
@@ -0,0 +1,678 @@
...
+                )),
+                &[],
+            )
+            .await
+            .map_err(|e| e.to_string())
+            .map(|_| ())
+    }
+
+    pub fn with(&self, svc_name: &'static str, api_name: &'static str) -> CassandraLabelledApi {
+        CassandraLabelledApi {
```
noise64, line 140: What is the intended relation between CassandraSession, CassandraLabelledApi, CassandraKeyValueStorage?

iIm a bit confused, as the CassandraSession and CassandraLabelledApi suggest some generic functionality, but AFAIU all the specific key values storage implementation is in CassandraLabelledApi, which wraps a CassandraSession, which contains the specific key-value schema, and CassandraKeyValueStorage is calling CassandraSession, and wrapping it before every call into a CassandraLabelledApi.

Maybe I'm missing something, but
 - I see no reason for a separate CassandraSession, the schema is tied to CassandraKeyValueStorage
 - I think the work done by CassandraLabelledApi could be a generic wrapper method in CassandraKeyValueStorage, instead of duplicating all the API
 - and I think the queries should live in CassandraKeyValueStorage

noise64 (changes requested):

golem-worker-executor-base/src/cassandra.rs
noise64: So I do see that we have in RedisLabbelled in common, but that is not "IndexedKeyValue" level, rather a generic redis wrapper (ofc because redis is a key value too in general, the two layer matches somewhat). I think we can have CassandraLabelled, but that should be in common, and on the level of cassandra operations (e.g. execute query with the common labels and a custom "query-name" label)

noise64: @vigoo wdty?

golem-worker-executor-base/src/cassandra.rs
```diff
@@ -0,0 +1,678 @@
...
+        let session = session_builder.build().await.map_err(|e| e.to_string())?;
+
+        Ok(CassandraSession {
+            session: Arc::new(session),
+            keyspace: config.keyspace.clone(),
+            set_tracing: config.tracing,
+        })
+    }
+
+    pub async fn create_docker_schema(&self) -> Result<(), String> {
```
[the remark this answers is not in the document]
irach-ramos: Yes, it is. 
I guess the user will have installed Cassandra, this is just for test.

golem-worker-executor-base/src/cassandra.rs
irach-ramos: The problem to move CassandraLabelled into common is that we are going to have new dependencies in CLI, which I remember we want to avoid, that's why I kept everything in executor-worker modules until the refactor comes to be able to move all common stuff into common without impacting CLI or others.

golem-test-framework/src/config/env.rs
```diff
@@ -461,6 +470,8 @@ impl EnvBasedTestDependencies {
 
         let redis_monitor = redis_monitor_join.await.expect("Failed to join");
 
+        let cassandra = Self::make_cassandra(config.clone());
```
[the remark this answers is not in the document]
irach-ramos: I think we want to test all code regardless of if it will be used or not, at the usage level , yes it should be optional, but we want to make sure that it works, so IMO we should test it all the time, or probably I'm missing somthing.

golem-worker-executor-base/src/cassandra.rs
noise64: well, we also have service base and / or we can use features.
which refactor is that?

golem-test-framework/src/config/env.rs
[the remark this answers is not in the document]
vigoo: This test configuration (`EnvBasedTestDependencies`) is also used for integration and CLI tests, and there we don't want to have cassandra. Having to download it's image and start it would just slow down the already slow CI further.

So please make it lazy initialized, at least, so it only gets created in the storage implementation tests when needed.

golem-worker-executor-base/src/cassandra.rs
```diff
@@ -0,0 +1,678 @@
+// Copyright 2024 Golem Cloud
```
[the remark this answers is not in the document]
noise64: sorry, i maybe still not understanding this, as this file contains specific Cassandra schemas and queries tied to our key value store usages, and not just generic Cassandra helpers. am i missing something?

golem-worker-executor-base/src/cassandra.rs
vigoo: The redis code in common adds metrics and logging to the redis client and it just directly wraps the redis operations, independent of any actual use case.

I see that in this top-level cassandra module there is key-value store specific queries, which is definitely not something we want. Everything key-value store specific should be in the `storage/keyvalue/cassandra` module. 

If you want to have common metrics and logging for cassandra, like we have for redis, then you can wrap the cassandra library but that should just instrument (wrap) the library's functions and not add any logic on top if it.

Where this "common" cassandra wrapper is is another question, top-level worker-executor-base is definitely not a good place for it. Until we need it in any other service, we can keep it in worker-executor-base, but let's move it to the storage module at least.

golem-worker-executor-base/src/cassandra.rs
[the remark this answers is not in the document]
vigoo: It's ok to only run this in tests, but that does not make it docker specific, this is code that creates the schema in any cassandra connection. 

Also it is key-value-store specific, so we should have it in the keyvalue/cassandra module.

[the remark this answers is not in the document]
irach-ramos: Well, for now it is just the kv implementation that is done, that's why we have kv specific, but we'll add indexed and blob too. 

I think we need to move, for now, cassandra.rs from top level golem-executor-base to storage as we have done for sqlite_types.rs, which contains all sql related statement for the 3 types of storage. 

@vigoo @noise64 WDYT ?

golem-worker-executor-base/src/cassandra.rs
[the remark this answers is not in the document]
noise64: discussed in other comments, closing this one

golem-worker-executor-base/src/storage/cassandra.rs
```diff
@@ -0,0 +1,675 @@
...
+use std::{num::NonZeroUsize, sync::Arc};
+
+#[derive(Debug, Clone)]
+pub struct CassandraSession {
+    pub session: Arc<Session>,
+    keyspace: String,
+    set_tracing: bool,
+}
+
+impl CassandraSession {
```
noise64, line 41: this struct and impl is still specific to our key value store schema, and not a generic cassandra session. anything that is using the kv_store table etc. tables should be named and placed as part of the KVStore implementation, and only things that are using generic cassandra primitives should be named as such cassandra

irach-ramos: Yes, as I tried to explain yesterday, this is a common struct as CassandraLabelledApi, which will contain all the kv and indexed storage queries, tables, and functions.
This PR is KV related, and I have the Indexed related changes for this file and other files. 
Once this PR is approved I'll create the Indexed storage PR, probably you will see what I mean at that time.

vigoo: What we are trying to ask is to follow the structure of how it was done for Redis:

- One layer wraps the 3rd party library and adds metrics and logging - without adding any higher level operations (no actual queries, just the same operations provided by the 3rd party library, wrapped)
- KV store implementation built on this, only containing KV store related queries
- Indexed store implementations built on this, only containing indexed store related queries

irach-ramos: I see, probably we can create another ticket to fix sqlite implementation as it is done in the same way as Cassandra.
There is a struct like this
`SqliteLabelledApi {
    svc_name: &'static str,
    api_name: &'static str,
    pool: SqlitePoolx,
}`
to avoid passing common parameters like `svc_name` & `api_name` for every method to the metrics and logging functionality, so this Labelled is holding the 3 storage kind of queries for simplicity, but it seems we are going to change that in Cassandra, then I think we need to change also in sqlite.

WDYT ?

noise64: yes, the sqlite one should also use similar naming and structuring

golem-test-framework/src/config/mod.rs
```diff
@@ -54,6 +56,7 @@ pub trait TestDependencies {
         self.rdb().kill();
         self.redis_monitor().kill();
         self.redis().kill();
+        self.cassandra().kill();
```
noise64, line 59: now that cassandra is lazy, will this boot up cassandra just to kill it?
if so, i think we should use RwMutex Option instead of lazy cell

golem-worker-executor-base/src/storage/cassandra.rs
```diff
@@ -0,0 +1,284 @@
...
+        let session = session_builder.build().await.map_err(|e| e.to_string())?;
+
+        Ok(CassandraSession {
+            session: Arc::new(session),
+            keyspace: config.keyspace.clone(),
+            set_tracing: config.tracing,
+        })
+    }
+
+    pub async fn create_schema(&self) -> Result<(), String> {
```
noise64, line 68: i would still move the (test) schema creations out of the common cassandra (and sqlite) code, and even separate them by usage (for sqlite), but if we create an issue for it, then i'm okay with it for now

noise64 (commented):
added some notes about the cassandra layzness and for creating followup issues

golem-worker-executor-base/src/storage/cassandra.rs
irach-ramos: I think there is only one common schema for the storage, to where you want to move the schema creation ? , 
how you want to separate them ?

noise64: based on the storage type: kv, blob, indexed; and move to that package.
with ideally we can select separately what to use, and only install what is needed (even if it is only for tests)

irach-ramos: I don't think it makes sense, because I don't think we will never use  kv storage with sqlite and indexed storage in Cassandra or some approach like that, and I don't see neither any advantage in performance or space usage or something useful, but I did it because I think we need to move on with this PR.

vigoo: The three storage "types" (kv, indexed, blob) should be completely separate and separately configurable. Even though sqlite + cassandra for example is not a likely combination, it is possible that we will have more backend implementations in the future that make sense to combine them in ways we don't see right now. In general, using something different for kv-store (basically caching data) and indexed-store (our primary storage layer for durable execution) completely makes sense to me.

afsalthaj:
Can we close this PR?
cc @vigoo  @noise64 ?

vigoo:
Closing this for now, as the base changed too much since then. Will use as inspiration if we decide to add cassandra support

PR #972, opened 2024-09-25
Special oplog implementation for ephemeral workers

### LINKED ISSUE #935: Optimize ephemeral worker oplog writes
Worker executor needs to select a different oplog configuration based on the component type. For ephemeral workers it needs only a primary oplog layer that directly writes chunks on each commit to the blob storage. 

This requires implementing a blob-storage based primary oplog layer (currently only the oplog archive layers are implemented) and checking the component type when constructing the instance dependencies.

vigoo:
Resolves #935 

Introduces a new oplog implementation for ephemeral workers, with the following properties:
- It directly writes to the last oplog archive (targeting the blob store with the default config)
- It has a separate "max entries before commit" configuration, which is bigger by default than the original one
- In many cases where there is a forced commit for durable workers (for example after remote writes), it does not commit 
- Even if it commits, it does that in the background, except in some "forced commit" situations.


golem-worker-executor-base/src/services/oplog/mod.rs
```diff
@@ -132,6 +135,17 @@ pub trait OplogService: Debug {
...
 
+/// Level of commit guarantees
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum CommitLevel {
+    /// Always commit immediately and do not return until it is done
+    Immediate,
+    /// Always commit, both for durable and ephemeral workers, no guarantees that it awaits it
+    Always,
+    /// Only commit immediately if the worker is durable
+    DurableOnly,
```
noise64, line 146: is this also async / no guartantees? just thinking about the naming

vigoo: For durable workers all commits are as before - they are "async blocking" until the entries are written.
For ephemeral workers, the three levels are:
- like the above
- commit, but on a spawned background fiber
- ignore

tried to bring in names that makes sense, maybe failed :)

jdegoes (approved):

PR #978, opened 2024-09-30
Comprehensive test for interpreter, new global input type info tests, support match on all types, restricted global variables and other fixes

### LINKED ISSUE #876: Migrate Rib specific tests from worker-service-base to golem-rib

### LINKED ISSUE #881: Rib Robustness: Make sure Rib doesn't condsider random variables as global variables

### LINKED ISSUE #954: Rib Robustness: Support pattern match for all types
Currently variants, option, some, none, ok, and err, enums are supported, but not `record` for example.
With the change, the following should be supported

```
match {foo: "bar"} {foo: b}
```

afsalthaj:
Fixes #876 
Fixes #881 
Fixes #954 

* Added a [comprehensive test ](https://github.com/golemcloud/golem/blob/migrate_tests/golem-rib/src/interpreter/tests/mod.rs#L8) in rib interpreter module replacing random tests that existed in worker-service module. This is a complete replacement and _not migration_. Tests in worker-service turned out to be overly repetitive and deleted (almost deleting 2000 lines of code including some support functions). The new comprehensive test is just 1 test case but it covers almost all possibilities of function inputs and function outputs in terms of types. It's another 2000 lines, however, more structured. This test exist along with other fine grained 400 tests in Rib. 

* Added [tests](https://github.com/golemcloud/golem/blob/migrate_tests/golem-rib/src/compiler/byte_code.rs#L988) specifically covering the scenarios of global-inputs in compiler module.
* Support pattern matching on all types
* Restricted Global Variables in Rib, that worker service can satisfy only `request` variable when using Rib.
* Other minor fixes


afsalthaj:
It was difficult to satisfy wasm-wave syntax here: https://github.com/golemcloud/golem/blob/migrate_tests/golem-rib/src/interpreter/comprehensive_test/mock_data.rs#L155, mainly due to abstract error messages and not mentioning the exact reason: `invalid token at 31..32`. 

If we are using the error message of https://github.com/golemcloud/wasm-rpc/blob/main/wasm-rpc/src/text.rs#L25 (which in turn comes from wasm-wave crate) back to users, then we need to be careful. Hopefully they will improving since it got [migrated](https://github.com/bytecodealliance/wasm-tools/tree/main/crates/wasm-wave).

PR #982, opened 2024-09-30
Fixes related to capturing stderr with failures

### LINKED ISSUE #930: Last error should contain the related stdout / stderr
Invoke-and-await now shows the stack trace and related outputs, but later the output cannot be retrieved, it should be part of last error, so it is visible when using worker get.

vigoo:
Resolves #930

### DIFF
--- a/golem-test-framework/src/dsl/mod.rs
+++ b/golem-test-framework/src/dsl/mod.rs
@@ -82,7 +82,7 @@ pub trait TestDsl {
     async fn get_worker_metadata(
         &self,
         worker_id: &WorkerId,
-    ) -> crate::Result<Option<WorkerMetadata>>;
+    ) -> crate::Result<Option<(WorkerMetadata, Option<String>)>>;
     async fn get_workers_metadata(
         &self,
         component_id: &ComponentId,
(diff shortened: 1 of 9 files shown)


golem-test-framework/src/dsl/mod.rs
```diff
@@ -82,15 +82,15 @@ pub trait TestDsl {
     async fn get_worker_metadata(
         &self,
         worker_id: &WorkerId,
-    ) -> crate::Result<Option<WorkerMetadata>>;
+    ) -> crate::Result<Option<(WorkerMetadata, Option<String>)>>;
```
noise64, line 85: can we create a struct for this tuple or make it part of the WorkerMetadata?

vigoo: It is intentionally not part of the worker metadata (because that's the actual serialized struct that is stored in redis, and the error information is gathered from oplog entries) .

I don't like using tuples anywhere in general, but in these tests, I felt that having 
``` let (metadata, _) = ...` where we don't care about the last failure is less annoying than completely changing the result type to something that wraps both. 

But I'm open to change my mind :)

noise64 (approved):

PR #987, opened 2024-10-02
Fixes #966, #797 and other clean ups

### LINKED ISSUE #966: Avoid `Enum` and `Variant` types to be part of `type_registry`
There isn't a need to have those types in the `type_registry`, that everything can be considered as a function indeed.
Somewhere I added to branch out specific logic, and after making significant changes later on with Rib, I can't see the value-add in it.

afsalthaj:
Fixes #966, #797, and other cleanups:

* We already have a `WorkerServiceRibInterpreter` service in the worker-service module, which handles the details of the worker-gateway (worker-service). To maintain consistency, this PR introduces `WorkerServiceRibCompiler`, responsible for compiling Rib `Expr` in a way specific to the worker-gateway. Example: `WorkerServiceRibCompiler` ensures that the  only allowed global variable in the Rib scrip is a request.

* Removed the `pure_evaluate` function in the worker-service module. It was slightly harder to reason about, as it primarily dealt with evaluating Rib expressions without function calls (e.g., resolving a worker-name expression). This has been replaced with directly calling the `rib::interpret_pure` function. This is useful anytime we need to evaluate a Rib expression anywhere in Golem.

* The function-type registry is a map of `RegistryKey` to `RegistryValue`. RegistryKey is a `FunctionName`, but it included includes `Enum`, `Variant`, etc.  As I was improving `Rib`, I find this part confusing, so I cleaned it up. To begin with, I wanted to treat everything as a generic function call and delegate the responsibility of differenting whether it is a Variant/Enum or actual worker-function call to the `Rib Interpreter`. However, that turned out to be a long-term goal with many brittle logic points. So it is still almost the same except that `RegistryValue` will hold the information whether or not it is a `Variant`. This allows safer pattern match and logical decisions during type inference.  This adjustment allows for more context-aware error messages. Example: Instead of "wrong number of arguments to register-user," we can have "wrong number of arguments to `variant` `register-user`". Likewise, we have the flexibility to add more information in the error messages.

* Renaming of some important variables, adding an fixing comments making it easier to understand the workflow of serving a custom `http` request for a new developer.

* Early error messages for invalid function calls before going to other phases of compilation which makes the error message more opaque due to extra details.

```scala
Unknown function call: `foo`
```

```scala
Unknown resource constructor call: `golem:it/api.{cart0(user_id).add-item}`. Resource `cart0` doesn't exist
```

```scala
Unknown resource method call golem:it/api.{cart(user_id).foo}. `foo` doesn't exist in resource `cart`

```

```scala
Incorrect number of arguments for function `foo`. Expected 1, but provided 2
```

```scala
Incorrect number of arguments for resource constructor `cart`. Expected 1, but provided 2
```

```scala
Invalid type for the argument in resource method `golem:it/api.{cart(user_id).add-item}`. 
Expected type `record`, but provided argument `\"foo\"` is a `str`

```

etc


afsalthaj:
The error messages in the PR description is a candidate for review

golem-rib/src/type_registry.rs
```diff
@@ -226,7 +229,44 @@ mod internal {
                 }
             }
 
-            _ => {}
```
afsalthaj, line 229: As I do further tickets, I will ensure this `_` is removed. Even if it results in some duplicate looking code, it ensures correctness and accidentally missing out on required cases.

golem-worker-service-base/src/worker_bridge_execution/worker_request_executor.rs
```diff
@@ -38,17 +38,3 @@ impl<T: AsRef<str>> From<T> for WorkerRequestExecutorError {
...
-impl WorkerRequestExecutor for NoopWorkerRequestExecutor {
-    async fn execute(
-        &self,
-        _worker_request_params: WorkerRequest,
-    ) -> Result<WorkerResponse, WorkerRequestExecutorError> {
-        Err(WorkerRequestExecutorError(
-            "NoopWorkerRequestExecutor".to_string(),
-        ))
-    }
-}
```
afsalthaj, line 54: There is no need of using `NoopWorkerRequestExecutor` anywhere in code. It's dangerous.

golem-worker-service-base/src/worker_service_rib_compiler/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+
+impl WorkerServiceRibCompiler for DefaultRibCompiler {
+    fn compile(rib: &Expr, export_metadata: &[AnalysedExport]) -> Result<CompilerOutput, String> {
+        rib::compile_with_limited_globals(
+            rib,
+            &export_metadata.to_vec(),
+            Some(vec!["request".to_string()]),
+        )
+    }
+}
```
afsalthaj, line 20: To make it more consistent with `WorkerServiceRibInterpreter`

golem-worker-service-base/src/worker_service_rib_interpreter/mod.rs
```diff
@@ -26,12 +26,6 @@ pub trait WorkerServiceRibInterpreter {
         rib_byte_code: &RibByteCode,
         rib_input: &RibInputValue,
     ) -> Result<RibInterpreterResult, EvaluationError>;
-
-    async fn evaluate_pure(
-        &self,
-        expr: &RibByteCode,
-        rib_input: &RibInputValue,
-    ) -> Result<RibInterpreterResult, EvaluationError>;
```
afsalthaj, line 34: `evaluate_pure` is not a concern of worker-service.

jdegoes (approved):

PR #1029, opened 2024-10-24
Clean up and remove unnecessary constraints

afsalthaj:


golem-rib/src/compiler/byte_code.rs
```diff
@@ -844,7 +872,10 @@ mod compiler_tests {
             InferredType::Str,
```
afsalthaj, line 872: Typesafe approach forced me to fix the wrong test inputs :)

noise64 (approved):

PR #1030, opened 2024-10-25
Improved logs

### LINKED ISSUE #938: Make sure we have errors and spans on each level where needed in each service
We need to see what went wrong juts by analysing logs. This means we need to always log errors on repo/service/api layers, independently of the error propagation and/or information hiding in the error channel.

vigoo:
Resolves #938


golem-common/src/client.rs
```diff
@@ -98,42 +111,60 @@ pub struct MultiTargetGrpcClient<T: Clone> {
...
             + Send,
     {
         let mut retries = RetryState::new(&self.config.retries_on_unavailable);
+        let span = debug_span!(
+            "gRPC call",
+            target_name = self.target_name,
+            endpoint = endpoint.to_string(),
+            description = description.as_ref()
+        );
+        let _span = span.enter();
```
noise64, line 148: i think this will create async span leak problems, AFAIU:
 - in_current_span should be used with spawn, not in general async code
 - in code which uses await no span should be entered, as that will leak to other random tasks
 - rather the span should be only entered for sync blocks (and be dropped before using await) and use instrument for async ones

vigoo: Thanks, forgot that it works like that

noise64 (approved):

PR #1032, opened 2024-10-27
Initial File System - Bounty-to-Hire

### LINKED ISSUE #1004: Initial File System - Bounty-to-Hire
This is the official ticket for Golem Cloud's Bounty-to-Hire event!

This is your chance to contribute to open source, potentially win a $15,000 bounty, and maybe even land a new 100% Rust job building Golem, a next-generation open source platform for building and deploying bulletproof backends.

The specification for this ticket is attached as a PDF:

[Initial File System Spec v3.pdf](https://github.com/user-attachments/files/17370325/Initial.File.System.Spec.v3.pdf)

## Important Websites

1. [Discord](https://discord.gg/WgQuEZgr7X): On the official Discord for the Golem open source project, you may chat with fellow participants and learn more about Golem, but remember, the Golem engineering team will not assist you in any way with building the project or becoming familiar with the code base.
2. [Github](https://github.com/golemcloud/golem): On the Github repository for the Golem open source project, you can fork the repo, browse the source code, and find links to other resources, such as the [Golem Cloud website](https://golem.cloud).

## Timeline

- **October 11**: Coding begins. You can see the requirements and begin working on the ticket.
- **October 28**: Judging begins. You may open pull requests that satisfy the requirements. We may pick a winner anytime **after** October 28, but not before. So even if you finish early, you should not submit your solution until October 28th. We do not guarantee to pick a winner on October 28th, and may require that contributors revise their pull requests in order to qualify.
- **November 4th**. The winner is officially announced.
- **November 11th**. Interviews with all willing and qualified participants will be complete and one or more job offers will be sent out.

## The Structure

 * **One Winner**. Whether there is a single complete solution or five, only one will be awarded the $15,000 bounty. In the unlikely event that no solutions satisfy the requirements, no bounty will be awarded, no pull requests will be merged, and no job offers extended.
 * **Guaranteed Interview**. If you submit an original solution that is reasonably complete, you are guaranteed an interview for our open position. If we end up with 5 solutions, then each of those five participants will be given an interview for the opening.
 * **Job Offer**. We intend to hire at least one developer from the pool of original and complete solutions. Our preference is to hire the winner, but this depends on many factors, and in some cases (e.g. the winner is not looking for a job), we might extend job offers to others.

## The Rules

1. **No Assistance**. We won't help you in any way. If you need clarification on the specification only, then ask on THIS Github issue. Do not ask in the Discord. You are free to ask questions to fellow participants on the Discord, or to ask general questions about Golem, not related to the source code or contributing to the project.
2. **No Cheating**. Absolutely no copying or deriving solutions from other developers. Though you should feel free to use Copilot-like solutions when you develop, if you use AI to generate a full solution (which definitely won't work and probably won't even compile), you will be banned from participation in the repository.
3. **Have Fun**. Bounties can get competitive but remember to enjoy the process. There will be more employment and bounty opportunities ahead for anyone who ends up submitting a meaningful pull request.

## Note on Plagiarism 

We recommend developing in private, so no other developers can copy or derive from your work. Before opening a pull request on October 28, we recommend **first** sending a ZIP archive of your complete solution to `john@ziverge.com`, to verify original authorship of the source code in the event of contested authorship.

**Good luck to everyone!**

mschuwalow:
# Initial Component Files

/claim #1004 
/resolves #1004 

---

Features:
- [x] Uploading initial component files during creation via cli
- [x] Uploading initial component files during update via cli
- [x] Initial component files get loaded by workers during execution
- [x] Permissions of initial component files are respected by workers
- [x] Support for initial component files is added to the test framework
- [x] Worker files are exposed via worker executor api
- [x] Worker files are exposed via worker api
- [x] Worker files are served via the gateway
- [x] Read-only files are directly loaded from the blob storage when accessed via the gateway

---
Note the comment I left regarding updating workers with intial component files present. There are test present for correct behavior in automatic update mode.

For manual update I would argue that we should give control to the user and have them explicitly store / restore files from the blob storage instead of trying to do it for them.

I.e. at the time when load-snapshot gets called the user will have a fully initialized filesystem with the files declared for the new component version and can restore whatever is needed from the old files.

Alternatively, if we want to simplify this for users as much as possible, we could always restore the previous r+w files to a special, reserved descriptor in read-only mode after a manual update. I.e. something like '/@previous'. This will not show up in normal directory listings for '/' if it's added as a seperate preopen, so it should not interfere with normal operations.
If desired I can add this behavior here / as a followup.

---

Obviously this is a very chunky boy. I tried to make it somewhat reviewable commit by commit, maybe that will help a bit.


golem-cli/src/command/component.rs
```diff
@@ -50,6 +53,25 @@ pub enum ComponentSubCommand<ProjectRef: clap::Args, ComponentRef: clap::Args> {
         non_interactive: bool,
     },
 
+    /// Create a new component, taking all parameters from the app manifest
+    AddWithManifest {
```
mschuwalow, line 57: I'm not sure what your approach to backwards compat is, so I added this and update as new commands for now.

golem-cli/src/command/component.rs
```diff
@@ -78,6 +100,32 @@ pub enum ComponentSubCommand<ProjectRef: clap::Args, ComponentRef: clap::Args> {
         non_interactive: bool,
     },
 
+    /// Updates an existing component by uploading a new version of its WASM and functions.
+    /// All parameters are taken from the app manifest.
+    #[command()]
+    UpdateWithManifest {
```
mschuwalow, line 106: I'm not sure what your approach to backwards compat is, so I added this and update as new commands for now.

golem-cli/src/model/application_manifest.rs
```diff
@@ -0,0 +1,919 @@
...
+use serde::{Deserialize, Serialize};
+use url::Url;
+use std::collections::{BTreeMap, BTreeSet};
+use std::path::{Path, PathBuf};
+use golem_wasm_rpc_stubgen::model::oam::TypedTraitProperties;
+use tracing::warn;
+
+use super::GolemError;
+
+pub fn load_app(app_resolve_mode: &ApplicationResolveMode) -> Result<ApplicationManifest, GolemError> {
```
mschuwalow, line 34: This is mostly taken from wasm-rpc-stubgen and extended. I think it would be good to merge and move them into some common crate so the validation / import / etc. logic is consistent across projects.

golem-common/src/model/mod.rs
```diff
@@ -2482,6 +2483,243 @@ impl FromStr for ComponentType {
     }
 }
 
+/// Key that can be used to identify a component file.
+/// All files with the same content will have the same key.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, NewType)]
+pub struct InitialComponentFileKey(pub String);
```
mschuwalow, line 2489: The design that I went for is that all initial are stored based on their hash in the blob storage.
This way we can have good sharing of data without making the implementation more complex.

The hashes are computed server-side so we are resistant to any attacks.
Collisions for sha-256 should basically not exist in practice.

golem-component-service-base/Cargo.toml
```diff
@@ -16,17 +16,22 @@ golem-common = { path = "../golem-common", version = "0.0.0" }
 golem-service-base = { path = "../golem-service-base" }
 golem-rib = { path = "../golem-rib" }
 golem-wasm-ast = { workspace = true }
+golem-worker-executor-base = { path = "../golem-worker-executor-base", version = "0.0.0" }
```
mschuwalow, line 19: This dependency is here for the blob storage. It feels very bad to have this dependency instead of some common crate though.

golem-component-service/src/api/component.rs
```diff
@@ -188,16 +219,57 @@ impl ComponentApi {
...
+                .map_err(|e| e.into())
+                .map(|response| Json(response.into()))
+        };
+        record.result(response)
+    }
+
+    /// Update a component
+    #[oai(
+        path = "/:component_id/updates",
+        method = "post",
```
mschuwalow, line 247: Update (the existing one too) is not idempotent, so imo post is a better fit then push.

This old one is left for backwards compat, as we now need to be multipart here too.

golem-worker-executor-base/src/durable_host/filesystem/types.rs
```diff
@@ -52,11 +52,12 @@ impl<Ctx: WorkerCtx> HostDescriptor for DurableWorkerCtx<Ctx> {
...
     fn write_via_stream(
         &mut self,
-        self_: Resource<Descriptor>,
+        fd: Resource<Descriptor>,
         offset: Filesize,
     ) -> Result<Resource<OutputStream>, FsError> {
+        self.fail_if_read_only(&fd)?;
         record_host_function_call("filesystem::types::descriptor", "write_via_stream");
-        HostDescriptor::write_via_stream(&mut self.as_wasi_view(), self_, offset)
+        HostDescriptor::write_via_stream(&mut self.as_wasi_view(), fd, offset)
```
mschuwalow, line 60: I played around a bit with wasmtime and I saw a few issues with preopened directories:

1. Preopened directories in the root don't work well. If I have a preopen for '/' and '/data' and list descriptors for '/' I don't see the descriptor for /data.
2. Preopen only works on the level of directories, meaning that we would have to make the whole dir readonly. This does not really work with the format that is specced for the files properties in the application manifest and is very confusing for users imo.
3. When playing around with complex setups I saw bad file descriptor errors a few times, but I'm not sure how to consistently trigger them.

Based on these I decided to implement this slightly differently.
1: Each worker maintains a shared cache of read only files /tmp/foo/{hash}. When a worker gets scheduled all required files are hard linked from the cache to the worker tmpdir. All read-write files are copied into the appropriate location.
2: The worker maintains a hashset of readonly paths that were loaded. All mutating operations in the HostDescriptor resolve the host path via the ResourceTable and check the hashset. If the path is present we fail with permission denied. This is very similar to how the preopen feature is [implemented](https://github.com/bytecodealliance/wasmtime/blob/af968461b0800325d61353e065bf5fe71abbefd4/crates/wasi/src/host/filesystem.rs#L4) in wasmtime.

golem-worker-executor-base/src/durable_host/mod.rs
```diff
@@ -155,6 +158,8 @@ impl<Ctx: WorkerCtx> DurableWorkerCtx<Ctx> {
             owned_worker_id.worker_id, worker_config.deleted_regions
         );
 
+        let read_only_paths = prepare_filesystem(file_loader, temp_dir.path(), &component_metadata.files).await?;
```
mschuwalow, line 161: My proposal for handling files during updates replays is the following:

1. Automatic: We replay from the beginning using the new initial file set. If we encounter errors the user will have to use a snapshot based update. This is similar to how other incompatibilities in the component are being handled from what I can tell.
2. Snapshot: We rely on the user-defined save-snapshot function to save all data that the users wants to persist across updates. This could either be implemented by using the blob storage apis directly, or we could allow returning paths to files that should be persisted from the snapshot function. The user-defined load-snapshot function is then responsible to reintialize state on top of the new initial filesystem. From what I can tell this is consistent with how other state in the component is handled, so this should be least surprising to users.

jdegoes:
@mschuwalow Congratulations on submitting your solution for our Bounty-to-Hire challenge (#1004)!

We have reviewed your pull request and believe that it is very promising. While not yet satisfying all of the requirements of #1004, we believe there is a short and well-defined path to get there, and so we are delighted to announce that you are a finalist in our Bounty-to-Hire event!

Next steps:

1. You are **guaranteed** an interview based on your work thus far.
2. Should you continue to work on this pull request, and get it into a state that can be merged to satisfy all requirements, you stand a reasonable chance of winning the bounty.

If you no longer wish to work on this pull request, please close it, and we will reach out about the interview. On the other hand, if you wish to continue pushing this pull request forward, then please let us know here and continue your work.

Congratulations again on being a finalist for Bounty-to-Hire!

golem-api-grpc/proto/golem/component/v1/component_service.proto
```diff
@@ -64,6 +65,8 @@ message CreateComponentRequestHeader {
   golem.common.ProjectId projectId = 1;
   string componentName = 2;
   optional ComponentType componentType = 3;
+  // All files need to be uploaded to the blob storage before providing them here
+  repeated InitialComponentFile files = 4;
```
mschuwalow, line 69: It feel a bit wasteful to upload the files through grpc if we are already in the 'trusted' context of the application and have direct access to the blob_storage, so I went with this approach

golem-test-framework/src/config/cli.rs
```diff
@@ -425,6 +430,10 @@ impl CliTestDependencies {
                 .await,
             );
 
+
+        let blob_storage = Arc::new(FileSystemBlobStorage::new(&PathBuf::from("/tmp/ittest-local-object-store/golem")).await.unwrap());
```
mschuwalow, line 434: This is consistent with what the other services use for their blob_storage, but this looks a bit broken to me. I think it would be better to have somethink like minio running and use the s3 based blob storage for all services

golem-service-base/src/service/initial_component_files.rs
```diff
@@ -0,0 +1,85 @@
...
+                "get",
+                BlobStorageNamespace::InitialComponentFiles,
+                &PathBuf::from(key.0.clone()),
+            )
+            .await
+    }
+
+    pub async fn put_if_not_exists(
+        &self,
+        bytes: &Bytes,
```
mschuwalow, line 52: In a followup this should be changed to instead stream the data so we can handle files that are larger than memory

golem-test-framework/src/dsl/mod.rs
```diff
@@ -1184,50 +1203,18 @@ pub fn to_worker_metadata(
     )
 }
 
-fn dump_component_info(path: &Path) -> golem_common::model::component_metadata::ComponentMetadata {
```
mschuwalow, line 1187: This got fully moved to the file system implementation of the component service, so doing it as part of the dsl is unneeded now

golem-api-grpc/proto/golem/component/v1/component_service.proto
vigoo: I did not think of the gRPC API as something that is only available for internal communication (but it is right that nothing external uses it at the moment).

golem-component-service-base/Cargo.toml
vigoo: That's a better place :)

golem-worker-executor-base/src/durable_host/filesystem/types.rs
vigoo: Maybe you can get rid of the need to do hard links as well by adding an extra path mapping in each filesystem host function. (Just an idea, not necessarily the way to go)

golem-test-framework/src/config/cli.rs
vigoo: The test framework is used for multiple things:
- Worker executor tests (`golem-worker-executor-tests/tests`) for testing various properties of the executor (only). For this minimal dependencies and as fast startup time as possible are the most important for me so here we use file system / in memory implementations
- Integration tests involving all services (`integration-tests/tests`). For these I agree that it would be better to use the s3 based blob storage. 
- CLI tests (`golem-cli/tests`) are like integration tests but doing operations through invoking the CLI itself. For historical reasons many things we should have as integration tests are implemented as CLI tests now - and for those I would also use S3 instead of file system storage, but if we think about it as just tests for the CLI layer itself it should not matter
- Benchmarks - we are running those either against locally a started golem cluster where I think using minio instead of the local file system would not be very useful (not interested in the performance of minio itself). And of course we plan to be able to run them against a real cluster deployed somewhere (but currently not doing it)

That said there are various ways to set up the test dependencies, as you see, and the `cli.rs` one is only used for benchmarks at the moment.

golem-test-framework/src/dsl/mod.rs
vigoo: I don't think we want this dump happen in the prod service, but that can be achieved by using log levels or some configurations so I'm ok with it.

mschuwalow: I meant the test-framework ComponentService, not the prod one :)

https://github.com/golemcloud/golem/blob/main/golem-test-framework/src/components/component_service/mod.rs#L51

golem-test-framework/src/config/cli.rs
mschuwalow: What I mean with broken here is that we setup the blob storage as using a filesystem in the containers /tmp/ittest-local-object-store/golem, but I don't see any volume mounts that would allow sharing this directory between containers.

golem-common/src/model/mod.rs
```diff
@@ -2482,6 +2483,380 @@ impl FromStr for ComponentType {
...
+}
+
+
+/// Path inside a component filesystem. Must be
+/// - absolute (start with '/')
+/// - not contain ".." components
+/// - not contain "." components
+/// - use '/' as a separator
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct ComponentFilePath(Utf8UnixPathBuf);
```
mschuwalow, line 2504: As this guy is getting serialized and we might have windows binaries at some point I went with this instead of a normal path.

golem-service-base/src/config.rs
```diff
@@ -58,3 +59,77 @@ impl Default for WorkerExecutorClientCacheConfig {
         }
     }
 }
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(tag = "type", content = "config")]
+pub enum BlobStorageConfig {
```
mschuwalow, line 66: No changes here, just moved to the service-base crate

golem-worker-executor-base/src/durable_host/filesystem/types.rs
mschuwalow: This hope was that with the hard links and the permission overlay we can keep the implementation on our side as simple as possible. I also saw that they have very similar infrastructure in wasmtime, so maybe we can get them to expose an api that allows us to just give the file permission to them.

golem-worker-executor-base/src/durable_host/mod.rs
```diff
@@ -1387,6 +1418,123 @@ impl<Ctx: WorkerCtx + DurableWorkerCtxView<Ctx>> ExternalOperations<Ctx> for Dur
     }
 }
 
+#[async_trait]
+impl<Ctx: WorkerCtx + DurableWorkerCtxView<Ctx>> FileSystemReading for DurableWorkerCtx<Ctx> {
+    async fn list_directory(&self, path: &ComponentFilePath) -> Result<ListDirectoryResult, GolemError> {
```
mschuwalow, line 1423: We don't need to bother with any locking here. There is some machinery higher up that will ensure these are only called when the worker is not executing anything (and potentially mutating the files)

mschuwalow:
@vigoo could you kick of ci for me later? I would like to make sure I have no regressions (and running the full suite locally sucks :D)

golem-cli/tests/api_deployment_fileserver.rs
```diff
@@ -0,0 +1,324 @@
...
+
+    let res = reqwest::blocking::get(format!("http://{host}/files/{component_id}/file"))?;
+    assert_eq!(res.status().as_u16(), 201);
+    assert_eq!(res.headers().get("content-type").unwrap(), "application/json");
+    assert_eq!(res.text()?, "foo\n");
+
+    Ok(())
+}
+
+fn api_deployment_fileserver_stateful_worker(
```
mschuwalow, line 188: this guy is probably the single most interesting test in the pr as it touches almost everything I added.

vigoo:
> @vigoo could you kick off ci for me later? I would like to make sure I have no failures (and running the full suite locally sucks :D)

Feel free to open a "fake" PR (changing a line in the readme or something:) that I can merge so we don't need to approve every attempt

mschuwalow:
> > @vigoo could you kick off ci for me later? I would like to make sure I have no failures (and running the full suite locally sucks :D)
> 
> Feel free to open a "fake" PR (changing a line in the readme or something:) that I can merge so we don't need to approve every attempt

Looks like it's green except for some formatting stuff, will quickly fix that and leave this one alone for now. WIll follow your suggestions if I do more work on this though.

jdegoes:
@mschuwalow 

We have reached November 4th, which is the day we must announce the winner of the Bounty-to-Hire program!

Given the scope of the challenge, it's incredibly impressive that we have had not one, but THREE finalists, who each demonstrated high competence and skill in tackling a problem that would stump most engineers on the planet.

However, ultimately, there can be only ONE WINNER to the Bounty-to-Hire challenge. So after MUCH analysis by Golem engineers, thoughtful consideration, and spirited discussion, we have decided to pick a winner.

That winner will be announced on our LIVE STREAM event TODAY at 12 NOON ET:

https://www.youtube.com/watch?v=at8EPqLWIRE

Without question, all three finalists are highly qualified Rust engineers and we look forward to interviewing and getting to each of them, with the intention of making at least one job offer to fill our recently opened position.

While you await the announcement of the WINNER, please shoot an email to john@ziverge.com to schedule your interview!

Stay tuned for more.

jdegoes:
@mschuwalow If you plan to join for the Live stream today, please do send me a message on Discord (you can find me on the Golem Discord), or an email at john@ziverge.com. Thank you!

jdegoes:
@mschuwalow Congratulations on winning the Bounty-to-Hire event!

You will be paid directly by Algora.io when your pull request is merged.

In order to get your pull request merged:

1. Allow time for the team to comment on necessary & suggested revisions
2. Implement revisions until the team has no further comments
3. Ensure that CI is green, and that all original and new tests complete successfully
4. Notify @vigoo or other maintainer that you are ready for final review

Once merged, your $15,000 bounty will be deposited into your Algora account, where you can cash out in the ways provided to you by the platform.

Thanks for your work on this feature and congratulations on winning what turned out to be a highly competitive event!

golem-api-grpc/proto/golem/apidefinition/api_definition.proto
```diff
@@ -79,4 +80,10 @@ message CompiledWorkerBinding {
   optional golem.rib.RibByteCode compiled_idempotency_key_expr = 9;
   optional golem.rib.RibInputType idempotency_key_rib_input = 10;
   optional golem.rib.WorkerFunctionsInRib worker_functions_in_response = 11;
+  WorkerBindingType type = 12;
```
afsalthaj, line 83: Could we make sure it is backward compatible?

afsalthaj: I would make it `optional` just to be safe :)

golem-cli/src/clients/component.rs
```diff
@@ -42,11 +43,15 @@ pub trait ComponentClient {
         file: PathBufOrStdin,
         project: &Option<Self::ProjectContext>,
         component_type: ComponentType,
+        files_archive: Option<&Path>,
+        files_permissions: Option<&ComponentFilePathAndPermissionsList>,
```
afsalthaj, line 47: I would suggest a rename if you like it.
`ComponentFilePathWithPermissions`

golem-cli/src/command/component.rs
afsalthaj: How about existing `Add` command with optional `app`?
Same goes for `Update` .

cc @vigoo  What's your take?

golem-cli/src/oss/clients/component.rs
```diff
@@ -82,17 +84,35 @@ impl<C: golem_client::api::ComponentClient + Sync + Send> ComponentClient
...
 
+        let files_archive_file = match files_archive {
+            Some(fa) => {
+                let file = File::open(fa)
+                    .await
+                    .map_err(|e| GolemError(format!("Can't open component files archive: {e}")))?;
+                Some(file)
+            }
+            None => None,
+        };
```
afsalthaj, line 100: Wouldn't `files_archive.map(|fa| ...).transpose?` work?

golem-cli/src/oss/clients/component.rs
```diff
@@ -116,17 +142,35 @@ impl<C: golem_client::api::ComponentClient + Sync + Send> ComponentClient
...
```
afsalthaj, line 158: If you like it, I think `map` and `transpose` would make it feel less `duplicated` between `add` and `update`

golem-cli/src/oss/factory.rs
```diff
@@ -74,6 +80,9 @@ impl OssServiceFactory {
             worker_url,
             service_http_client_config: HttpClientConfig::new_for_service_calls(allow_insecure),
             health_check_http_client_config: HttpClientConfig::new_for_health_check(allow_insecure),
+            file_download_http_client_config: HttpClientConfig::new_for_file_download(
+                allow_insecure,
+            ),
```
afsalthaj, line 85: May be, it's a good idea to rename `new_for_file_download`?. It doesn't really expose the purpose of that function.

golem-cli/src/service/component.rs
```diff
@@ -64,6 +76,7 @@ pub trait ComponentService {
         uri: ComponentUri,
         project: &Option<Self::ProjectContext>,
     ) -> Result<ComponentUrn, GolemError>;
+    async fn resolve_component_name(&self, uri: &ComponentUri) -> Result<String, GolemError>;
```
afsalthaj, line 79: Does `resolving` component_name based on `component_uri` depend on the instance of `ComponentService`? Should this be really abstract?

golem-cli/src/service/component.rs
```diff
@@ -77,6 +90,145 @@ pub trait ComponentService {
...
+                let mut read_dir_stream = ReadDirStream::new(read_dir);
+
+                while let Some(entry) = read_dir_stream.next().await {
+                    let entry = entry.map_err(|err| {
+                        GolemError(format!("Error reading directory entry: {}", err))
+                    })?;
+                    let next_path = entry.path();
+
+                    let file_name = entry.file_name().into_string().map_err(|_| {
+                        GolemError("Error converting file name to string".to_string())
```
afsalthaj, line 137: May be it's a good idea to use the actual error. This is because, given it's a corner case to fail, it would be good to know, what exactly made it fail.

golem-cli/src/service/component.rs
```diff
@@ -77,6 +90,145 @@ pub trait ComponentService {
...
+        Ok(LoadedFile {
+            content,
+            target: component_file.target,
+        })
+    }
+
+    async fn build_files_archive(
+        &self,
+        component_files: Vec<InitialComponentFile>,
+    ) -> Result<ComponentFilesArchiveAndProperties, GolemError> {
```
afsalthaj, line 178: I think I would rename it to `ComponentFilesArchiveWithProperties`. If its a `product` / `struct`, may be its just enough to say `ComponentFilesArchive` allowing it to hold more information in future, without the need to rename.

golem-cli/src/service/component.rs
```diff
@@ -77,6 +90,145 @@ pub trait ComponentService {
...
+            for LoadedFile { content, target } in self.load_file(component_file).await? {
+                // zip files do not like absolute paths. Convert the absolute component path to relative path from the root of the zip file
+                let zip_entry_name = target.path.to_string();
+                let builder = ZipEntryBuilder::new(zip_entry_name.into(), Compression::Deflate);
+
+                zip_writer
+                    .write_entry_whole(builder, &content)
+                    .await
+                    .map_err(|err| GolemError(format!("Error writing to archive: {}", err)))?;
+                properties_files.push(target);
```
afsalthaj, line 202: May be fail it here the moment it breaks the uniqueness by `f.path`.

golem-cli/tests/api_deployment_fileserver.rs
```diff
@@ -0,0 +1,344 @@
...
+    (deps, name, cli): (
+        &(impl TestDependencies + Send + Sync + 'static),
+        String,
+        CliLive,
+    ),
+) -> anyhow::Result<()> {
+    // the port is hardcoded in the env based test dependencies. Do not change this.
+    let host = "localhost:9093";
```
afsalthaj, line 203: There should be a way to make this available from test dependencies. Worth making that change to avoid future surprises.

golem-cli/tests/api_deployment_fileserver.rs
```diff
@@ -0,0 +1,344 @@
...
+        - method: Get
+          path: "/files/{component_id}/{{file}}"
+          binding:
+            bindingType: file-server
+            componentId:
+              componentId: '{component_id}'
+              version: 0
+            response: 'let file: str = request.path.file; "/files/${{file}}"'
+    "#,
+    ))?)
```
afsalthaj, line 299: 👍

golem-common/src/model/mod.rs
```diff
@@ -2482,6 +2482,412 @@ impl FromStr for ComponentType {
...
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Encode, Decode, Enum, Default)]
+#[serde(rename_all = "kebab-case")]
+#[oai(rename_all = "kebab-case")]
+pub enum WorkerBindingType {
+    #[default]
+    Default,
+    FileServer,
+}
```
afsalthaj, line 2851: 👍

golem-component-service-base/src/service/component.rs
```diff
@@ -152,6 +233,19 @@ pub trait ComponentService<Namespace> {
...
+
+    // Files must have been uploaded to the blob store before calling this method
+    async fn create_internal(
+        &self,
+        component_id: &ComponentId,
+        component_name: &ComponentName,
+        component_type: ComponentType,
+        data: Vec<u8>,
+        // None signals that files should be reused from the previous version
+        files: Vec<InitialComponentFile>,
```
afsalthaj, line 248: Probably for consistency with update_internal, use `Option<Vec>`

golem-component-service-base/src/service/component.rs
```diff
@@ -152,6 +233,19 @@ pub trait ComponentService<Namespace> {
         component_name: &ComponentName,
         component_type: ComponentType,
         data: Vec<u8>,
+        files: Option<InitialComponentFilesArchiveAndPermissions>,
+        namespace: &Namespace,
+    ) -> Result<Component<Namespace>, ComponentError>;
+
+    // Files must have been uploaded to the blob store before calling this method
+    async fn create_internal(
```
afsalthaj, line 241: Could you quickly explain the difference here with `create`. I feel this is slightly a complicated workflow.
Can't `create` handle the situation of making sure the files are already in blob-store?

golem-component-service-base/src/service/component.rs
```diff
@@ -304,6 +413,142 @@ impl ComponentServiceDefault {
...
+
+        let new_type_registry = FunctionTypeRegistry::from_export_metadata(&metadata.exports);
+
+        if let Some(constraints) = constraints {
+            let conflicts =
+                Self::find_component_metadata_conflicts(&constraints, &new_type_registry);
+            if !conflicts.is_empty() {
+                return Err(ComponentError::ComponentConstraintConflictError(conflicts));
+            }
+        }
```
afsalthaj, line 504: Is this repeated anywhere else? May be in `update` method?

golem-component-service-base/src/service/component.rs
```diff
@@ -304,6 +413,142 @@ impl ComponentServiceDefault {
...
+            .filter(|c| c.namespace == namespace.to_string())
+            .ok_or(ComponentError::UnknownComponentId(component_id.clone()))
+            .and_then(|c| {
+                c.try_into()
+                    .map_err(|e| ComponentError::conversion_error("record", e))
+            })
+            .map(Component::next_version)?;
+
+        // Fallback to the files from the previous version if no files are provided.
+        let files_to_use = files.unwrap_or_else(|| next_component.files.clone());
```
afsalthaj, line 519: Please remind me of the spec while I review the PR. Is it about merging with the previous files (if none of the new files with the ones of previous version) ?

golem-component-service-base/src/service/component.rs
```diff
@@ -404,106 +650,123 @@ where
...
-            .clone()
-            .try_into()
-            .map_err(|e| ComponentError::conversion_error("record", e))?;
-
-        self.component_repo.create(&record).await?;
-
-        self.component_compilation
-            .enqueue_compilation(component_id, component.versioned_component_id.version)
-            .await;
-
```
afsalthaj, line 505: Ok I got the answer for my own question in PR, if the constraint check is not duplicated. Thanks

golem-component-service/db/migration/postgres/004__add_component_files.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE component_files
+(
+    component_id uuid NOT NULL,
+    version bigint NOT NULL,
+    file_key VARCHAR(255) NOT NULL,
+    file_path VARCHAR(255) NOT NULL,
+    file_permissions VARCHAR(255) NOT NULL,
+    PRIMARY KEY (component_id, version, file_path),
+    FOREIGN KEY (component_id, version) REFERENCES component_versions (component_id, version)
```
afsalthaj, line 9: 👍

golem-component-service/src/api/component.rs
afsalthaj: May be you are right, but I am having my doubts :) 
So let me confirm, the existing `/:component_id/upload` with `PUT` doesn't work here?
I guess update should be merging with the previous files if there are no conflicts.

golem-service-base/src/config.rs
afsalthaj: Just to give you a history, `golem-service-base` is probably a carry forward of a very early version of golem, and personally I was wishing everything that 's common just stays within golem-common. This moving may hinder the avoidance of `golem-service-base` in near future. Is this mandatory to have it here? 

Thanks for the comments though, that it clarifies the context.

golem-service-base/src/storage/blob/s3.rs
```diff
@@ -124,6 +127,9 @@ impl S3BlobStorage {
                         .to_path_buf()
                 }
             }
+            BlobStorageNamespace::InitialComponentFiles => {
+                Path::new(&self.config.object_prefix).to_path_buf()
+            }
```
afsalthaj, line 132: Are we sure we have to move this from worker-executor-base to service-base? I can see the need, but in that case, probably to golem-common.

golem-component-service/src/api/component.rs
vigoo: @afsalthaj the IFS is tied for a specific component version. No merging.

golem-service-base/src/config.rs
vigoo: It's good to move this here for now.

golem-service-base/src/storage/blob/s3.rs
vigoo: It makes sense to share the storage abstractions among services, and it also makes sense to not put it in golem-common which is referenced by things like the generated client library and CLI as well. 

Yes we talked about earlier that we could move all these things into a single crate and use features to avoid having all these backend stuff on the client side, but I'm not sure it would be better and would not make the build flow even slower, so at the current state moving these to `golem-service-base` totally makes sense and is the best choice.

golem-cli/src/command/component.rs
noise64: For now I would use the current commands with optional flags for either:
 - explicitly specifying the manifest files
 - or to auto detect the manifest files to be used
 
 Later we can make the second default "true".  This way most our docs should be up to date, and we can keep adding the new flags, also I would expect that eventually every command that works with components will have the "auto detect".

golem-cli/src/model/application_manifest.rs
noise64: wasm-rpc-stubgen is a dependency even currently (the rpc cli commands are even "embedded" into the other golem-cli commands), and most of the things should be public for reuse. The intention is to have the generic oam and rpc manifest model available and working in wasm-rpc, and only call and extend it (for the extra properties and validation) here. It might be happen that we have to make some thing more generic of public in wasm-rpc, but currently the plan is to reuse from that lib directly.

golem-worker-executor-base/src/services/file_loader.rs
```diff
@@ -0,0 +1,131 @@
...
+use tracing::debug;
+
+use crate::error::GolemError;
+
+use golem_service_base::service::initial_component_files::InitialComponentFilesService;
+
+/// Interface for loading files and making them available to workers.
+/// This will hardlink to a temporary directory to avoid copying files between workers. Beware
+/// that hardlinking is only possible within the same filesystem.
+// TODO: add bookkeeping for files that are no longer needed.
```
vigoo, line 31: This should be done otherwise we fill up the executor's disk eventually. 

It could be just something like a reference counting on the shared directories, decremented when a `Worker` is stopped.

golem-worker-executor-base/tests/blob_storage.rs
```diff
@@ -16,20 +16,17 @@ use aws_config::meta::region::RegionProviderChain;
...
+use golem_service_base::storage::blob::{fs, memory, s3, BlobStorage, BlobStorageNamespace};
+use golem_service_base::storage::sqlite::SqlitePool;
 use sqlx::sqlite::SqlitePoolOptions;
+use tempfile::{tempdir, TempDir};
 use testcontainers::runners::AsyncRunner;
 use testcontainers::ContainerAsync;
+use testcontainers_modules::minio::MinIO;
+use uuid::Uuid;
 
 macro_rules! test_blob_storage {
```
vigoo, line 31: I think these tests (and the other two suite) should be moved to the crate where the implementation is now

vigoo: Ah I realized you only moved the blob storage not the other two. In that case I'm only talking about this file.

golem-service-base/src/storage/blob/mod.rs
```diff
@@ -328,6 +328,7 @@ impl<'a, S: BlobStorage + ?Sized + Sync> LabelledBlobStorage<'a, S> {
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub enum BlobStorageNamespace {
     CompilationCache,
+    InitialComponentFiles,
```
vigoo, line 331: Let's add `AccountId` to this namespace too (and use it in the path just like with the oplog related blob stores). This makes it easier to collect / delete a specific account's data.

golem-api-grpc/proto/golem/apidefinition/api_definition.proto
mschuwalow: Do you care about backwards compat of the grpc api? I was under the impression it was intended only for use between golem components (and I assume you don't support running mixed versions of golem)

Just double checking. If yes, I can absolutely change this (and the component api as well -- I think there is another case like this there)

golem-cli/src/oss/clients/component.rs
mschuwalow: I have an Option<Future<Result<?>>> here while I think transpose would only work with Option<Result<?>>

golem-cli/src/oss/clients/component.rs
mschuwalow: same as above

golem-cli/src/oss/factory.rs
mschuwalow: Can you provide a suggestion? This is used to build a client config specific to fetching remote files, so I'm struggling a bit to come up with a better name.

golem-cli/src/service/component.rs
mschuwalow: Docs for into_string say
```
Converts the OsString into a String  if it contains valid Unicode data.

On failure, ownership of the original OsString is returned.
```

So in the case of error we would not get an actual error only a string which we cannot even display without conversion first. Adjusted the error to better reflect this.

golem-cli/tests/api_deployment_fileserver.rs
```diff
@@ -0,0 +1,344 @@
...
+        let mut api_file = std::fs::File::create_new(&api_path)?;
+        let mut writer = std::io::BufWriter::new(&mut api_file);
+        serde_json::to_writer(&mut writer, &api_definition)?;
+        writer.flush()?;
+    }
+
+    let definition: HttpApiDefinitionWithTypeInfo =
+        cli.run(&["api-definition", "add", api_path.to_str().unwrap()])?;
+
+    drop(temp_dir);
```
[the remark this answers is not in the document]
mschuwalow: I just added these to be more explicit until when the resource is (at least) needed. Removed.

golem-component-service-base/src/service/component.rs
mschuwalow: This comment was a broken leftover, removed. There is no semantic difference between Some(vec![]) and None here, so I think it makes sense to not have more possible values than needed here.

golem-component-service-base/src/service/component.rs
mschuwalow: The problem is that I want to accept both an archive and a list of paths in create.

In case of internal create I want to take the files from blob storage (meaning that I need to be able to set the archive to None, to avoid a useless update). On the other hand for create through the rest api I always want to have either both an archive and a list or neither, so I don't really want to have two separate optional parameters here.

This way was the best I could come up with.

golem-component-service-base/src/service/component.rs
mschuwalow: There are there things now for both update and create:
* update: This is used by the rest api and uploads all files to the blob_storage + checks consistency of the archive and permissions list. Afterwards it's forwarded to update_unchecked
* update_internal: This is used by the grpc api and just checks that all referenced files are in the blob_store, then it forwards to update_unchecked
* update_uncheck: Does the actual logic of uploading files. Everything that is duplicated between the above two code paths lives in here.

I admit the update_internal name is a bit bad, suggestions welcome :)

golem-component-service-base/src/service/component.rs
mschuwalow: Yeah, when no files are uploaded the set from the previous versions should be used.
To make my live simpler everywhere else here I just copy the data over from the previous version here (it's only the file paths + permissions, so quite small. All the actual data is deduplicated and in the blob_store)

golem-service-base/src/storage/blob/mod.rs
mschuwalow: Done. Note that this will increase storage use potentially as we won't be able to share data between accounts.

golem-cli/src/command/component.rs
mschuwalow: The main problem I see here is that the non-manifest version takes a required positional  argument for the component wasm file:

```
❯ ./target/debug/golem-cli component add --help
Creates a new component with a given name by uploading the component WASM

Usage: golem-cli component add [OPTIONS] --component-name <COMPONENT_NAME> <component-file>

Arguments:
  <component-file>  The WASM file to be used as a Golem component

Options:
  -c, --component-name <COMPONENT_NAME>
          Name of the newly created component
  -v, --verbose...
          Increase logging verbosity
      --ephemeral
          Create an Ephemeral component. If not specified, the command creates a Durable component
  -q, --quiet...
          Decrease logging verbosity
      --durable
          Create a Durable component. This is the default
  -F, --format <FORMAT>
          
  -y, --non-interactive
          Do not ask for confirmation for performing an update in case the component already exists
  -h, --help
          Print help
```

I think it would be quite confusing for users if that argument would be required or unused depending on whether a flag for the manifest file is provided.

Additionally, wasm-rpc (and the new commands too) try to search for a manifest file automatically if none was provided through the cli and fail if none was found. IMO we definitely don't want to introduce that behaviour in the existing add command as it would break user's flows.

Could you please provide some pointers how you want this to work?

golem-cli/src/model/application_manifest.rs
mschuwalow: The problem is that a lot of the stuff used here is private. The main offenders are

* [load-app](https://github.com/golemcloud/wasm-rpc/blob/06e127aeb10bc602a7920c53b5c03f6accd54421/wasm-rpc-stubgen/src/commands/declarative.rs#L322C4-L322C19)
* [the manifest itself](https://github.com/golemcloud/wasm-rpc/blob/06e127aeb10bc602a7920c53b5c03f6accd54421/wasm-rpc-stubgen/src/model/wasm_rpc.rs#L57) -- This guy has all the validation logic and the functionality to import referenced wasmbuild projects. Sadly the underlying [oam manifest with extra properties](https://github.com/golemcloud/wasm-rpc/blob/06e127aeb10bc602a7920c53b5c03f6accd54421/wasm-rpc-stubgen/src/model/oam.rs#L150) is not exposed from there. We should absolutely refactor this part once those are public though.

golem-cli/src/oss/clients/component.rs
afsalthaj: Okay 👍

golem-cli/src/service/component.rs
afsalthaj: Sure. Makes sense.

golem-cli/tests/api_deployment_fileserver.rs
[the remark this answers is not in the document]
afsalthaj: 👍

golem-component-service-base/src/service/component.rs
afsalthaj: Okay 👍

golem-component-service-base/src/service/component.rs
afsalthaj: I am thinking. @vigoo @nicoburniske @noise64  May be jump in for some suggestions if you have any. 
I hope this won't block the merge though. After the explanation I understand why you did this.

golem-service-base/src/storage/blob/mod.rs
vigoo: Yeah it is a bit tricky because there are no accounts in the OSS part of Golem. But if you check worker executor, it propagates an account id everywhere, and various upper layers are trying to abstract it. (It's a bit of a mess right now). 

For example https://github.com/golemcloud/golem/blob/main/golem-worker-service/src/lib.rs#L11 and in the component service it's represented by the "namespace" (but the `DefaultNamespace` here https://github.com/golemcloud/golem/blob/main/golem-service-base/src/auth.rs does not contain anything, as there are no accounts in the open-source version).

As you probably already figured out, the reason for all these abstractions, and different -base crates etc is that we have some advanced features built on top of this codebase which require these features. 

Of course it's hard to figure out the best way to do this as you don't see the other half. 
For this PR is suggest the following:

- add `AccountId` to this namespace, because it already has it for `OplogPayload`, for example 
- in worker executor you already have account id available everywhere (sometimes hidden in an `OwnedWorkerId`)
- in component service, you need to extract it from the namspace which is propagated to the service level. It does not have a way to extract it right now, but you can define a trait with an `Self -> Option<AccountId>` extractor and require it for the namespace here: https://github.com/golemcloud/golem/blob/main/golem-component-service-base/src/service/component.rs#L390
- if there is no account id, use "-1" (because worker service does it already in one of the above links)

golem-component-service-base/src/service/component.rs
vigoo: I think the names are "fine". 
I see the logic in having these two different ways of updating in the REST vs gRPC APIs, and right now it does not really matter because the component update gRPC API is not really used (only by the test framework). 

We did not really do such separate behaviors in REST vs gRPC (I think we not necessarily treat the gRPC API as internal, we are just not exposing it by default in our setup). That said it makes sense to have more internal endpoints on the gRPC APIs, like this one, and once we have it, we can always add anoter gRPC endpoint which incorporates the upload too (like the REST one), if needed.

So for me this is not a blocking issue and I'm good with adding it like this.

golem-cli/src/model/application_manifest.rs
noise64: We have a problem of code duplication already in some other places, and we are working on getting rid of them, so I would rather have this fixed first in wasm-rpc and applied here. Even if it's another repo, it is part of the project.

golem-cli/src/command/component.rs
noise64: I don't think that making a positional argument optional is an uncommon thing, especially if it's controlled by a new flag (or existence of new config files). Also, while we are aiming in general for backward compatibility, in the case of cli and tooling it is expected that we will deprecate / refactor things in the coming releases, given the new app manifest approach. So we should not increase too much the number of soon to be deprecated commands, I would rather keep adding flags to existing thing, and maybe later deprecate them together (or flags of them)

So I would make the wasm file optional depending the flags, given it is not breaking anything out there, and I think having so is quite common in tools (e.g. taking input from stdin vs file vs a config which points to a file).

(It also a good question that why we have this positional arg in the first place, when mostly everything else is by named flags, but this is already out :) )

mschuwalow: How should it behave in regards to automatic discovery of manifests?
In wasm-rpc it always does so, failing the command if it cannot find one.

Should we be inconsistent with wasm-rpc here and only support explicitly passed manifests?

golem-cli/src/model/application_manifest.rs
mschuwalow: I understand your reasoning here 👍.
Though if possible, I would like to do this as a followup, given that it doesn't change the behaviour and the pr is already huge (what I mostly am a bit afraid of is other things getting merged and having to rebase this monstrosity). Doing a dance with releases of multiple repos would further complicate it.

noise64: In that case can you please add a few TODOs for it (to any source that was copied)?
As for a followup, I think you can let that go, as there will be soon a bigger batch of wasm-rpc-stubgen updates, so it can be part of that.

golem-cli/src/command/component.rs
noise64: I would probably do this fallback logic:
- if the component-file is specified then use no manifests at all
- if the component-file is not specified, check for manifest flags, if none, do auto discovery
- if both component-file and manifest flags are defined then show an error that they are mutually exclusive

mschuwalow: I'm struggling to do this one with clap. I think the correct approach to do this is:

* Have an arg_group with two nested arg groups for explicit component file version and manifest based version. This one is required, the two nested ones are optional.
* In the nested ones, set requires to false but explicitly requires all fields in the arggroup, i.e.
```
#[derive(clap::Args, Debug, Clone)]
#[group(required = true, multiple = true)]
pub struct ExplicitComponentOrManifestArgGroup {
    #[command(flatten)]
    explicit_component: Option<ExplicitComponentArgGroup>,

    #[command(flatten)]
    manifest: Option<ManifestArgGroup>,
}

#[derive(clap::Args, Debug, Clone)]
#[group(required = false, multiple = true, requires_all = vec!["component_file", "component-type-flag"])]
pub struct ExplicitComponentArgGroup {
    /// The WASM file to be used as a Golem component
    #[arg(value_name = "component-file", value_hint = clap::ValueHint::FilePath)]
    component_file: PathBufOrStdin, // TODO: validate exists

    /// The component type. If none specified, the command creates a Durable component.
    #[command(flatten)]
    component_type: ComponentTypeArg,
}

#[derive(clap::Args, Debug, Clone)]
#[group(required = false, multiple = true, requires = "app")]
pub struct ManifestArgGroup {
    /// List of application manifests. Can be specified multiple times
    #[arg(long, short)]
    app: Vec<PathBuf>,
}
```
This way, if any of the args in one of the nested groups is present, all of the other ones are also requires. The outer arggroup ensures that no two conflicting arguments are present.


There are two problems
* This currently doesn't work with the positional arguments. Clap assumes there is an extra positional argument, but I'm not sure where it's coming from.
* The help messages for this are really bad. We need to have some custom docstring for sure that explains all these mutually exlusive options to users

golem-service-base/src/storage/blob/mod.rs
mschuwalow: Done. Please take a look whether this what you had in mind:
https://github.com/golemcloud/golem/pull/1032/commits/81155567c18f945c8c025e47d7edb09790f3e719

mschuwalow:
I think I addressed most of the comments. I'll require some guidance what to do with this guy https://github.com/golemcloud/golem/pull/1032#discussion_r1831303395, as I'm struggling to find a way to nicely extend the existing commands with the new options vs. introducing new ones.

golem-api-grpc/proto/golem/apidefinition/api_definition.proto
afsalthaj: For a variaty of reasons, this proto definition is used to store in repo.

Example:

```rust
    pub fn deserialize(bytes: &[u8]) -> Result<Vec<CompiledRoute>, String> {
        let (version, data) = bytes.split_at(1);

        match version[0] {
            SERIALIZATION_VERSION_V1 => {
                let proto_value: CompiledHttpApiDefinition = Message::decode(data)
                    .map_err(|e| format!("Failed to deserialize value: {e}"))?;

                let value = proto_value
                    .routes
                    .into_iter()
                    .map(CompiledRoute::try_from)
                    .collect::<Result<Vec<CompiledRoute>, String>>()?;

                Ok(value)
            }
            _ => Err("Unsupported serialization version".to_string()),
        }
    }
    
        pub fn serialize(value: &[CompiledRoute]) -> Result<Bytes, String> {
        let routes: Vec<CompiledHttpRoute> = value
            .iter()
            .cloned()
            .map(CompiledHttpRoute::try_from)
            .collect::<Result<Vec<CompiledHttpRoute>, String>>()?;

        let proto_value: CompiledHttpApiDefinition = CompiledHttpApiDefinition { routes };

        let mut bytes = BytesMut::new();
        bytes.put_u8(SERIALIZATION_VERSION_V1);
        bytes.extend_from_slice(&proto_value.encode_to_vec());
        Ok(bytes.freeze())
    }
```

afsalthaj: I think everything related to grpc for worker gateway (Api Definition etc) has to be made backward compatible

golem-service-base/src/storage/blob/mod.rs
vigoo: I'm surprised it "leaked" into so many places like API definition models. Is that necessary? Worker service was already able to address workers somehow, which means it had the mechanism to get the necessary context (accountid etc) alredy without these changes.

I would like to revisit this and if possible make less spread out change.

Other than this from my side the PR is ready to go.

vigoo (changes requested):
I have one last request (https://github.com/golemcloud/golem/pull/1032#discussion_r1832279299) then it's good to go.

We should merge when each reviewer approves it as we reviewed different parts.

golem-service-base/src/storage/blob/mod.rs
mschuwalow: The main reason this leaked into so many places is that I need to use the correct accountId for an api deployment if I understand the code correctly.

The way it was done before was that namespace information was included when creating the ApiDefinition, but it was serialized to a string and never deserialized again. Deserializing and carrying the information around as a proper type makes it show up in a lot of places https://github.com/golemcloud/golem/blob/b861a50c1a9bcab8a45eec4ad70acb9ac59802e1/golem-worker-service-base/src/repo/api_definition.rs#L27

Let me try to limit it to only the compiled versions of the definitions.

mschuwalow: @vigoo Done. I think this looks much better https://github.com/golemcloud/golem/pull/1032/commits/85736967003e06fe3719c2794b712a4f6fa3ad34.

noise64:
all comments are resolved on cli side 👍

mschuwalow:
@vigoo what is missing here / anything to do from my side to unblock merging?

vigoo:
> @vigoo what is missing here / anything to do from my side to unblock merging?

Hopefully nothing, I just want an approve from @afsalthaj too

afsalthaj:
https://github.com/golemcloud/golem/pull/1032#discussion_r1832212517

If you could just make this small change, I think we good to merge

mschuwalow:
> [#1032 (comment)](https://github.com/golemcloud/golem/pull/1032#discussion_r1832212517)
> 
> @mschuwalow If you could just make this small change, I think we good to merge May be proto3 is internally optional by default, but we have been following a safer approach by making things explicitly optional to keep backward compatibility.

ah sorry, missed that one. Will adjust 👍

mschuwalow:
> [#1032 (comment)](https://github.com/golemcloud/golem/pull/1032#discussion_r1832212517)
> 
> @mschuwalow If you could just make this small change, I think we good to merge May be proto3 is internally optional by default, but we have been following a safer approach by making things explicitly optional to keep backward compatibility.

@afsalthaj Done, please take a look https://github.com/golemcloud/golem/pull/1032/commits/625786aa0dcc1d23e9250af2d5aadf2cb62011e3
Only CompiledHttpApiDefinition and ComponentMetadata seem to be getting stored as a protobuf, so the rest should be fine.

I was thinking to add a comment to all grpc definitions that (directly or transitively) get stored in the database so other people don't run into the same problem. This would affect quite a lot of types like rib expressions though (due to how big CompiledHttpApiDefinition is). Something like
```
// Note: This message is used to store serialized data in the database.
// make sure to not break backward compatibility when changing this message.
```
Want me to go for it?

afsalthaj (approved):

afsalthaj:
We are still thinking as to how to manage this: `bincode`  vs `proto`. Something for us to improve. 
The advantage of `proto` is we throw the backward compatibility pollutions there. We will see.

Thanks for addressing the review comments.

PR #1045, opened 2024-10-30
Plugin model

### LINKED ISSUE #1034: Plugin model
Define the plugin model and it's associated operations in Golem OSS.

- CRUD for defined plugins
- Ability to set a prioritized list of enabled plugins for each component

Every new feature is exposed as:
- gRPC and HTTP APIs 
- Host functions
- CLI commands

## Oplog Processor Plugin Definition

The definition of an oplog processor plug-in involves the following elements:

- **Name**. For example, “MongoDB”.
- **Description**. A Markdown description of the plug-in.
- **Icon**. A square picture representing the plug-in, suitable for low-resolution and high-resolution environments.
- **Version**. A version string. The combination of name and version must be unique.
- **Homepage**. A URL that points to the home of the plug-in on the web.
- **Binary**. The WASM Component.

## Component Transformer Plugin Definition

The definition of a component transformer plug-in involves the following elements:

- **Name**. For example, “MongoDB”.
- **Description**. A Markdown description of the plug-in.
- **Icon**. A square picture representing the plug-in, suitable for low-resolution and high-resolution environments.
- **Version**. A version string. The combination of name and version must be unique.
- **Homepage**. A URL that points to the home of the plug-in on the web.
- **Providing WIT**. The (optional) WIT that the transformer will provide to the component. This helps users understand what sort of WIT API is provided by the plug-in (if any).
- **JSON Schema**. JSON Schema describing the “typed input” of the plug-in, if any.
- **Validate URL**. The URL of a validator.
- **Transform URL**. The URL of the component transformer.

vigoo:
Resolves #1034


golem-component-service/db/migration/postgres/005__add_plugins.sql
```diff
@@ -0,0 +1,32 @@
...
+    homepage             text     NOT NULL,
+    plugin_type          smallint NOT NULL,
+    scope_component_id   uuid REFERENCES components (component_id),
+    provided_wit_package text,
+    json_schema          text,
+    validate_url         text,
+    transform_url        text,
+    component_id         uuid REFERENCES components (component_id),
+    component_version    bigint,
+    deleted              boolean  NOT NULL DEFAULT FALSE,
```
noise64, line 16: it seems that we do need to use this in almost every query, so i would add a covering index for the primary key columns and this together

golem-component-service-base/src/repo/component.rs
```diff
@@ -419,12 +632,13 @@ impl ComponentRepo for DbComponentRepo<sqlx::Postgres> {
             let namespace: String = result.get("namespace");
             let name: String = result.get("name");
             if namespace != component.namespace || name != component.name {
+                transaction.rollback().await?;
```
noise64, line 635: is there a higher level transaction construct in sqlx? manually doing the rollbacks on errors seems fragile

noise64: I would change this select to be part of the insert (or rather, an upsert, and check if there were any affected rows, otherwise i think there are some race conditions here.

golem-component-service-base/src/repo/component.rs
```diff
@@ -478,9 +701,209 @@ impl ComponentRepo for DbComponentRepo<sqlx::Postgres> {
...
+        namespace: &str,
+        component_id: &Uuid,
+        data: Vec<u8>,
+        metadata: Vec<u8>,
+        component_type: Option<i32>,
+        files: Option<Vec<FileRecord>>,
+    ) -> Result<ComponentRecord, RepoError> {
+        let mut transaction = self.db_pool.begin().await?;
+
+        let result = sqlx::query("SELECT namespace FROM components WHERE component_id = $1")
```
noise64, line 716: i would probably make this part of the update through a join, and make them a predicate.
As for the "WITH prev selects", i think those are not reliable at all without golding explicit locks for the affected rows (or use a pattern where before any component version changes we do a single row lock on the components table)

noise64 (commented):
Added a few DB related comments and questions, let's disccuss them in more detail IRL

golem-component-service/db/migration/postgres/005__add_plugins.sql
```diff
@@ -0,0 +1,32 @@
...
+(
+    installation_id   uuid    NOT NULL PRIMARY KEY,
+    plugin_name       text    NOT NULL,
+    plugin_version    text    NOT NULL,
+    priority          integer NOT NULL,
+    parameters        bytea   NOT NULL,
+    component_id      uuid REFERENCES components (component_id),
+    component_version bigint,
+
+    FOREIGN KEY (plugin_name, plugin_version) REFERENCES plugins (name, version)
```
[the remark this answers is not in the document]
vigoo: No, I intentionally allowed that. I can imagine cases that installing the same plugin multiple times with different configurations makes sense. (For example, some log shipping plugin and you want to ship to multiple targets).

golem-component-service/db/migration/sqlite/005__add_plugins.sql
```diff
@@ -0,0 +1,32 @@
...
+(
+    installation_id   uuid    NOT NULL PRIMARY KEY,
+    plugin_name       text    NOT NULL,
+    plugin_version    text    NOT NULL,
+    priority          integer NOT NULL,
+    parameters        blob    NOT NULL,
+    component_id      uuid REFERENCES components (component_id),
+    component_version bigint,
+
+    FOREIGN KEY (plugin_name, plugin_version) REFERENCES plugins (name, version)
```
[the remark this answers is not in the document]
vigoo: So what do I do? :)

golem-component-service-base/src/repo/component.rs
vigoo: It should be rolled back on drop, my tests indicated that it is was not working properly, and adding an explicit rollback fixed the test. I don't like it either.

golem-component-service/db/migration/sqlite/005__add_plugins.sql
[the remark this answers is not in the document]
noise64: As discussed, will be done in an followup PR

noise64 (approved):

PR #1053, opened 2024-11-07
WIP : Introduce CORS and make ways to add authentication

### LINKED ISSUE #921: CORS support in worker gateway
An ability to support CORS from the Worker Gateway without having to do so through Rib code

afsalthaj:
Fixes #921 

Code design subject to change for a bit more time. This is because I am parallely looking at WorkerGateway Authentication design [doc](https://www.notion.so/golemcloud/Authn-Worker-Gateway-105a4cb355ec80d7b844c57a4e886e64) to see where does that fit in.

Also please read: https://www.notion.so/golemcloud/CORS-in-Worker-Gateway-Draft-105a4cb355ec80c58253c80b483a08af?pvs=25#136a4cb355ec80eb8116d26521bb5a3b

* Restructure - so that we have clear introduction of  new plugins and authentication modules.  Find the image below. The prefix `gateway` is helping as a developer, however will disappear once we can move this out of worker-service-base to `gateway` module. worker-service can of course reuse this module (i.e, not introducing another network traffic between gateway and worker). I understand restructuring could make the PR review harder, but I will make more comments within PR to make the review easy later on.
* Introduce proper plugins (Kong inspired). [After comments from @jdegoes  in [NOTION docs, ](https://www.notion.so/golemcloud/CORS-in-Worker-Gateway-Draft-105a4cb355ec80c58253c80b483a08af#136a4cb355ec8002bbcdd78c8d0d4e6a), I am changing the name to middleware)
* Plugins can be part of a binding (example: worker binding), or plugin by  itself can be a binding. Both are needed.  Example certain pre-flight requests just need to know only a plugin and shouldn't know anything else on the backend. This is well typed now. Tomorrow FileServer (IFS PR) can be part of the same `enum`. Probably we should rename types with prefix or suffix as `binding`  to `integration`. AWS call it `integration`.
* Support both concise and verbose versions of CORS - OpenAPI spec supports both and native Api Definition supports only concise. This is the best I could come up with.
* Ensuring backward compatibility ensured both from REST API perspective and GRPC proto. This was a bit of work
* OpenAPI spec fix to support binding at Operator level
*  CORS workflow has already been tried out and made it work, however, I made more changes in design afterwards. Setting up CORS was the easiest part of the PR

image


afsalthaj:
Closing this due to huge conflicts in favor of another one

PR #536, opened 2024-05-27
Nixfying

countoren:

PR #1055, opened 2024-11-08
Development Env Using Nix Shell

khajavi:


vigoo:
See https://github.com/golemcloud/golem/pull/536#issuecomment-2372277658

jdegoes:
Closing in favor of [this](https://github.com/NixOS/nixpkgs/pull/344289), but if we need to do something here to accomodate the Nix pkg PR, then we should discuss that.

PR #1058, opened 2024-11-09
Initial file system followup: Fix race condition in file-loader

mschuwalow:
I saw the [test failure on master](https://github.com/golemcloud/golem/actions/runs/11753754003/job/32746966597) and looked into the cause.

The problem is that the file-loader uses the key of the intial file as the path of the file in it's local cache. This can run into the following sequence when multiple workers are using the same file:

1. Worker A requires file a
2. file loader downloads the file and stores a copy in ${cache_dir}/a. It returns an Arc to the worker that will delete the file when it's dropped
3. Worker A is shutting down and drops the last Arc for file a. This removes it from the cache as weak.upgrade is now going to fail. The file is not yet deleted.
3. Worker B requires file a
4. file loader checks the cache and doesn't find it. It creates a new cache entry and starts downloading the file again to ${cache_dir}/a
5. Drop for the entry owned by Worker A runs, the underlying file ${cache_dir}/a is deleted while Worker B is still using it
6. Boom

I resolved this in this pr by additionally having a counter in the file_loader and just giving each locally created file a unique id, instead of using the key as the name. This way worker B will just create a new file in the above scenario while worker A deletes the old one.

Ran the integration_test suite 10 times locally to check that it works fine, hopefully the tests are not flaky now.

### DIFF
--- a/golem-worker-executor-base/src/services/file_loader.rs
+++ b/golem-worker-executor-base/src/services/file_loader.rs
@@ -17,6 +17,7 @@ use async_lock::Mutex;
 use golem_common::model::{AccountId, InitialComponentFileKey};
 use std::collections::HashMap;
 use std::path::Path;
+use std::sync::atomic::AtomicU64;
 use std::sync::Weak;
 use std::{path::PathBuf, sync::Arc};
 use tempfile::TempDir;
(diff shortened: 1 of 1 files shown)


golem-worker-executor-base/src/services/file_loader.rs
```diff
@@ -128,28 +147,31 @@ impl FileLoader {
...
-                        );
-                        tokio::fs::copy(&path_handle.path, target).await?;
-                        drop(cache_entry_guard);
-                        drop(cache_entry);
-                        return Ok(());
-                    }
-                    Err(e) => {
-                        return Err(anyhow!("Failed to download file: {}", e));
-                    }
+            let cache_entry = cache_guard.get(key).and_then(|weak| weak.upgrade());
```
mschuwalow, line 150: unrelated cleanup. There is no reason to hold the lock while we are copying

vigoo (approved):
Thanks for noticing the failing test and fixing it!

Issue #1063, opened 2024-11-12
Implement all Identity Providers 


afsalthaj:
This ticket ensures we can seamlessly add more identity providers into the existing code base.
We will get a clear picture of whether this is achievable 

Only by the end of this ticket, OSS users will really get to use the feature of authentication for its first version. Until then the code may be quite internal and can be tried out internally.

This should be done once the following tickets is finished. 

1. https://github.com/golemcloud/golem/issues/1060
2. https://github.com/golemcloud/golem/issues/1061
3. https://github.com/golemcloud/golem/issues/1062

The idea is here to get the use case running such that we have a breathing space to get the design right. By the time design is right and reliable, users get to really use this feature.

Issue #1062, opened 2024-11-12
Try integrate the workflow as part of Middlewares in API Gateway


afsalthaj:
This work is mainly making the previous steps production grade for Github Identity Provider (as an example)
This can be part of https://github.com/golemcloud/golem/issues/1060, however, making a general workflow working with 1060 (probably in a branch) doesn't necessarily production grade code design. However with this ticket, we are making the workflow production grade.

Probably we will see some hiccups (from a code level perspective/design perspective) to include this seamlessly with a middlewares.

Issue #1061, opened 2024-11-12
Make sure the authentication workflow works with the standard OpenAPI spec


afsalthaj:
Once #1060  is done, decide the configuration setup.  The above workflow will be slightly different in terms of its configuration set up compared to making a normal application work with the workflow, especially configuring the redirections. Setting up the workflow is quite flexible in normal applications, and that may not be the case when it comes to API Gateway, especially the redirections. Note that user is not writing any code using Rib or injecting logic within a component. This would imply most of the heavy lifting is done within the OpenAPI spec.

The bare minimum that we expect is the following:https://swagger.io/docs/specification/v3_0/authentication/oauth2/
Here we can either map the name of the security scheme (instead of being arbitrary) to an actual name that's registered with the golem application, where the scheme is owned by a project

It's a very good idea to try this out with  Github.

```

# Step 1 - define the security scheme
components:
  securitySchemes:
    oAuthSample:    # <---- arbitrary name
      type: oauth2
      description: This API uses OAuth 2 with the implicit grant flow. [More info](https://api.example.com/docs/auth)
      flows:
        implicit:   # <---- OAuth flow(authorizationCode, implicit, password or clientCredentials)
          authorizationUrl: https://api.example.com/oauth2/authorize
          scopes:
            read_pets: read your pets
            write_pets: modify pets in your account

# Step 2 - apply security globally...
security:
  - oAuthSample:
    - write_pets
    - read_pets

# ... or to individual operations
paths:
  /pets:
    patch:
      summary: Add a new pet
      security:
        - oAuthSample:
          - write_pets
          - read_pets
      ...
```

The purpose of this ticket, is to ensure #1060  works with a proper OpenAPI spec. With #1060, in OSS version, you may need a dictionary of security schemes with no references to project while in cloud, this scheme is owned per project. 

The callback auth URL is part of the OAuth2 app settings in Github, as an example, and this would imply these need to be stored in Golem as part of the details of the security scheme

<img width="681" alt="image" src="https://github.com/user-attachments/assets/cbd0e7b7-225a-49e5-a3e7-f36cc4a5b408">

Issue #1060, opened 2024-11-11
Introduce Github identity provider implementation in API Gateway - with a workflow that works


afsalthaj:
This should be dead simple and can be done in the following

*Let's say user specified the security scheme and registered with `golem` under a project.  For this task, let's assume this is done already and security scheme is already available for a particular in GitHub.  In OSS, this security scheme can be a table not owned by any domain such aa project. This assumption implies, we will have a single security scheme stored in the table.
* From the perspective of API Gateway, when a user hits a configured secured endpoint, it understands that it is a secured endpoint probably because its gateway binding is configured with security scheme middleware. Before it goes on and evaluate the rib (hitting the worker), it checks if the request has an active session or access token, and if so go on with serving the request. If not the API Gateway redirects the user to GitHub for authentication, and in the request URL, it includes a state parameter that holds information about the original API endpoint they wanted to access.
* User authenticate themselves with the GitHub. and GitHub redirects to a dedicated API Gateway endpoint with the state parameter included in the query string.
* API Gateway now needs to exchange authorization code for an access token,  and once it receives, it stores this information into the session and the redirects back to the original URL, and this time, the sessions will be active, and it goes on and execute the rib script with the details in the session passed along with the other inputs to rib.

This basic workflow is the first step to implementing worker authentication

PR #1067, opened 2024-11-16
Add gateway authentication

afsalthaj:
Add worker gateway authentication resolving. Please excuse the verbose PR description, but I request you to go through it before reviewing code.

#1061 
#1062 
#1063 
#1064  
#1060 
#1065

As part of these tickets, there are some important changes done here - mainly focussing on user experience with authentication, CORS and similar middleware(s)  in API gateway, that can be very difficult to fix in future if not taken care now. 

## Notes on OpenAPI

This was also a learning as to how much we can keep up with the real OpenAPI spec.  The native API definition (user-facing part) is now mostly inspired from OpenAPI spec itself. 

1) API Gateway is powerful enough, that you don't need all the details required in OpenAPI spec all the time. Trying to add them to align strongly with OpenAPI spec can be really confusing. Example:  There is no need of `openIdConnectUrl` which is part of the OpenAPI spec - just the security scheme id is enough. The reasoning here is, not many API gateways really support this level of authentication as far as I see. While OpenAPI spec is designed and geared to be part of the actual application and not the gateway. Another example is, there is no clear path on how to setup cors in a standard OpenAPI spec - while it is aware of some headers, but not entirely. 

2) API Gateway doesn't have all of the features that an advanced OpenAPI spec may refer to. However, by the time we support the major ones (which we are already in progress with), we will end up having the above situation. Example: We support only OpenID based authentication, while OpenAPI can list a set of securities to be applied. I haven't done the sequential application of security yet. Supporting a list right now can only confuse developers.  

## Hints on what's in PR
* `HttpApiDefinitionRequest` (user facing internal API  definition structure which is formed either through direct native-api-definition upload or OpenAPI import). As an example, this will have concise information about security (there is more to it, which is further explained below)
* Explicit [Transformation Layer](https://github.com/golemcloud/golem/tree/add_gateway_auth/golem-worker-service-base/src/gateway_api_definition_transformer) which makes extra routes, extra middlewares etc based on the user request to form the core internal `HttpApiDefinition` (explained below). Transformations are idempotent.
* `HttpApiDefinition` . This will have detailed information about security, added auth call back endpoints based on the redirect-url of user's security scheme with Google (as an example), cors-header-additions based on preflight endpoints etc.
* `CompiledHttpApiDefinition`(not much change here except compiler driven changes). This is HttpApiDefinition itself, but with pre compiled rib, input type information for each Rib used (worker name and response mapping)
* `HttpApiDefinitionResponse` - This is user facing response (unexposed byte code, unexposed identity provider metadata, unexposed secrets etc). There is an extra HttpApiDefinitionRequestData which is almost 1-1 with HttpApiDefinitionRequest, however, I didn't focus on fixing this. Duplicate data came into existence for some reason, but such details can be dealt later, as this is not affecting user's hot path.
* Security Scheme Create and Get. Delete and Update has implications and I would want to do this later.
* Session Store mainly to correlate requests that went through multiple redirections as part of authentication. Currently a a redis also backed by a simple in-memory structure backed by existing cache mechanisms withe eviction policies. Currently session store is mainly for authentication, so I used the original state information (a random value) produced by the openidconnect crate
* Session management are skipped for requests that doesn't require it. i.e, only those requests that goes through a auth middleware becomes aware of the need of the session information, and the response from middleware will have a session id, which is later used to lookup the [auth details which gets fed to rib input](https://github.com/golemcloud/golem/blob/add_gateway_auth/golem-worker-service-base/tests/api_gateway_end_to_end_tests.rs#L473). If requests don't go through this middleware, this sessionId is always a `None`. It can be `None` for certain other cases too such as if the middleware results in `Unauthorized` response or a `redirect` response.
* Auth call back static endpoint - this is part of static binding which was introduced as part of cors
* MiddlewareIn and MiddlewareOut. Authentication Middleware has and instance of MiddlewareIn (inspecting request) and Cors has an instance of MiddlewareOut (transforming response)
* Authentication Middleware implementation. If the request doesn't have tokens (id token or access token, i.e, unauthenticated calls), then it sends a redirect response back to identity provider. Similarly, if the session was expired, it sends a redirect response. If the signature is mismatched, it is unauthorized response rather than redirect. 
* [IdentityProvider trait ](https://github.com/golemcloud/golem/blob/add_gateway_auth/golem-worker-service-base/src/gateway_security/identity_provider.rs) implementations. There Is a default implementation and it's reused for various providers. The identity provider trait is focussed on easier testing:
* Explicit Tests: [Signature Mismatch](https://github.com/golemcloud/golem/blob/add_gateway_auth/golem-worker-service-base/tests/api_gateway_end_to_end_tests.rs#L140), [Expired Tokens](https://github.com/golemcloud/golem/blob/add_gateway_auth/golem-worker-service-base/tests/api_gateway_end_to_end_tests.rs#L356), [Expired Session](https://github.com/golemcloud/golem/blob/add_gateway_auth/golem-worker-service-base/tests/api_gateway_end_to_end_tests.rs#L245) and actual [Successful authentication](https://github.com/golemcloud/golem/blob/add_gateway_auth/golem-worker-service-base/tests/api_gateway_end_to_end_tests.rs#L462). This includes redirect flows, and this was a time killer however, making it work will allow escape weeks of debugging in future.
* The generics such as `ToResponse<R>` was hindering development to a great extent and it only got worse in last couple of days, so I made it explicit for `Http` last day. Basically the workflow is much more easier to understand now. History: This was mainly done to reuse the workflow (resolve the binding and execute the binding while handling middlewares etc) between different protocols long time back, however, it was giving more pain than needed. We will figure out how other protocols can be incorporated later. It's not a concern of this PR. In other words, `gateway_input_executor` is changed to [gateway_http_input_executor](https://github.com/golemcloud/golem/blob/add_gateway_auth/golem-worker-service-base/src/gateway_execution/gateway_http_input_executor.rs)
* I haven't given much love to golem-cli part yet, but works for adding a security scheme and use it in API definition. This is more of mechanical work, and we can keep improving
* I tried to fix a few clones that happens at hot path. But I am hesitant to do it  in a hurry (I tried, and it resulted in some changes that I am not convinced of yet). We are aware of it, and I raised this [ticket](https://github.com/golemcloud/golem/issues/1088)

## Experimental in nature

Being able to tag a security which is a one-liner in configuration (OpenAPI as well as native API definition), and then specify things like `request.auth.email` in a Rib script makes the whole workflow production grade. That is, not anyone can simply invoke a worker even in a publicly exposed API.

The changes done are still experimental as there are numerous possibilities of configurations and usage of these features. It' less ROI  to go through all of those possibilities in the first attempt, and therefore the PR is more of getting _one_ thing right with focus on extensibility, and not focussing much on minor details which can be added/fixed based on testing in coming weeks. Example: Adding cookies is just inlined straight away for the auth call back responses - we are not making it configurable and complicating it further - because we don't know if this is worth it yet. 

I took an extra week for this work, but I strongly believe the details figured out are worth it, and offloading them to future would result in weeks of downtime.


golem-worker-service-base/src/api/common.rs
```diff
@@ -25,7 +24,7 @@ pub struct MessagesErrorsBody {
 
 #[derive(Clone, Debug, Object)]
 pub struct ValidationErrorsBody {
-    errors: Vec<RouteValidationError>,
```
afsalthaj, line 28: This was giving more pain and less reliability than it looks like.

golem-worker-service-base/src/gateway_execution/gateway_binding_executor.rs
```diff
@@ -138,50 +166,167 @@ impl<N> DefaultGatewayBindingExecutor<N> {
...
+            Err(err) => {
+                Some(err.to_response_from_safe_display(|_| StatusCode::INTERNAL_SERVER_ERROR))
+            }
+        }
+    }
 }
 
 #[async_trait]
-impl<N: Send + Sync, R: Debug + Send + Sync> GatewayBindingExecutor<N, R>
+impl<N: Send + Sync + Clone, R: Debug + Send + Sync> GatewayBindingExecutor<N, R>
```
vigoo, line 244: Please don't use single letter type parameters - it is completely unreadable (for me). Let's stick to the verbose ones as before

afsalthaj: Done, fixed :+1
Thanks for spotting it

vigoo:
First question, just based on the PR description. You mention:

> Session Store mainly to correlate requests that went through multiple redirections as part of authentication. Currently a simple in-memory structure with an eviction policy. 

How does this work with multiple load-balanced worker service nodes (which is the normal way to set up a real golem cluster)?

afsalthaj:
Yes. Absolutely. As I mentioned in the PR description the plan is to add redis backend. Happy to push that into the current PR. Just stopped committing once the tests becoming green after fixing other corner cases.

golem-worker-service-base/src/api/register_api_definition_api.rs
```diff
@@ -239,15 +423,15 @@ pub struct MiddlewareData {
 #[oai(rename_all = "camelCase")]
 pub struct GatewayBindingWithTypeInfo {
```
mschuwalow, line 424: I think we should consider making this guy an enum when we can break backwards compatibility next time

golem-worker-service-base/src/gateway_api_definition/http/http_api_definition.rs
```diff
@@ -50,26 +49,114 @@ pub struct HttpApiDefinition {
...
+            .collect()
+    }
+
+    pub async fn from_http_api_definition_request<Namespace>(
+        namespace: &Namespace,
         request: HttpApiDefinitionRequest,
         created_at: chrono::DateTime<chrono::Utc>,
-    ) -> Self {
-        HttpApiDefinition {
+        security_scheme_service: &Arc<dyn SecuritySchemeService<Namespace> + Send + Sync>,
```
mschuwalow, line 76: Personal opinion, but I don't like passing in services as part of data constructors. At the use-site it will look like the service will become part of the struct, which is not at all what is actually happening.

I think either a top-level function or a method on the SecuritySchemeService would look better here

golem-worker-service-base/src/gateway_api_definition_transformer/mod.rs
```diff
@@ -0,0 +1,36 @@
+use crate::gateway_api_definition::http::HttpApiDefinition;
+pub use api_definition_transformer::*;
+pub use auth_transformer::*;
+pub use cors_transformer::*;
+
+mod api_definition_transformer;
+mod auth_transformer;
+mod cors_transformer;
+
+pub fn auth_transformer() -> Box<dyn ApiDefinitionTransformer> {
```
mschuwalow, line 10: Let's maybe get rid of the trait and just use plain functions here? Looks like this  is not extensible anyway

golem-worker-service-base/src/gateway_binding/mod.rs
```diff
@@ -37,17 +37,25 @@ mod worker_binding_compiled;
 pub enum GatewayBinding {
     Default(WorkerBinding),
     FileServer(WorkerBinding),
-    Static(StaticBinding),
+    Static(Box<StaticBinding>),
```
mschuwalow, line 40: Why do we need the box here?

golem-worker-service-base/src/gateway_execution/gateway_session.rs
```diff
@@ -0,0 +1,225 @@
...
+
+        // Start the eviction task in the background
+        tokio::spawn(async move {
+            let session = InMemoryGatewaySession {
+                data: data_clone,
+                eviction_strategy: eviction_strategy_clone,
+            };
+
+            session.start_eviction_task().await;
+        });
```
mschuwalow, line 135:
```suggestion
        }.in_current_span());
```

golem-worker-service-base/src/gateway_execution/gateway_session.rs
```diff
@@ -0,0 +1,225 @@
...
+impl Default for SessionData {
+    fn default() -> Self {
+        SessionData {
+            value: HashMap::new(),
+            created_at: Instant::now(),
+        }
+    }
+}
+
+pub struct InMemoryGatewaySession {
```
mschuwalow, line 97: I don't believe there is ever a reason to have an in-memory version of this. Let's build this directly on top of the key-value store (and add a ttl there)

golem-worker-service-base/src/gateway_middleware/middleware_in.rs
```diff
@@ -0,0 +1,63 @@
...
+// limitations under the License.
+
+use crate::gateway_execution::gateway_http_input_executor::GatewayHttpInput;
+use crate::gateway_execution::gateway_session::SessionId;
+use crate::gateway_middleware::HttpAuthenticationMiddleware;
+use async_trait::async_trait;
+use golem_common::SafeDisplay;
+
+#[async_trait]
+pub trait HttpMiddlewareIn<Namespace> {
```
mschuwalow, line 22: can we replace this with a normal function? Doesn't look like this needs to be extensible

golem-worker-service-base/src/gateway_middleware/middleware_out.rs
```diff
@@ -0,0 +1,60 @@
...
+// While some middlewares are  specific to `MiddlewareIn` other are specific to `MiddlewareOut`.
+// This ensures orthogonality in middleware usages, such that, at type level, we distinguish whether it is
+// used only to process input (of api gateway) or used only to manipulate the output (of probably rib evaluation result)
+// These middlewares are protocol-independent. It's input `GatewayRequestDetails`,
+// with is an enum of different types (protocols) of input the protocol type.
+// An `enum` over type parameter is used merely for simplicity.
+// The middleware decides which protocol out of `GatewayRequestDetails` to process.
+// This approach centralizes middleware management, and easy to add new middlewares without worrying about protocols.
+#[async_trait]
+pub trait HttpMiddlewareOut<Response> {
```
mschuwalow, line 30: same

mschuwalow (commented):
will continue tomorrow

golem-worker-service-base/src/gateway_execution/gateway_session.rs
afsalthaj: Removed but not committed yet

golem-worker-service-base/src/gateway_api_definition/http/http_api_definition.rs
afsalthaj: Hmmm. I think it's debatable. A top level function is not easily observable. 
`HtttpApidefinition::from...` is easily seen, and it indicates, to construct it we will need to pass a function that talks about how to get the full scheme given an identifier. I don't see it as a service, but as a higher order function.

golem-worker-service-base/src/gateway_binding/mod.rs
afsalthaj: Clippy.

afsalthaj: Due to size it holds. So basically clippy driven change.

golem-worker-service-base/src/gateway_execution/gateway_session.rs
afsalthaj: We should have a in-memory session (sort of acting as a cache) on top of real kv store

golem-worker-service-base/src/gateway_middleware/middleware_in.rs
afsalthaj: We have a lot of middlewares. and  the `in` and `out` middlewares are clearly distinguishable with this approach. And the next developers don't need to think where to place how to process a middleware type. It's grouped into 1 single place

golem-worker-service-base/Cargo.toml
```diff
@@ -44,6 +44,7 @@ lazy_static = { workspace = true }
 mime_guess = "2.0.5"
 nom = { workspace = true }
 openapiv3 = { workspace = true }
+openidconnect = "3.5.0"
```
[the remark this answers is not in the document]
afsalthaj: Why? Any other services using it?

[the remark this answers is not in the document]
mschuwalow: for consistency. It should also make it easier to avoid having more than one version of the dependency around, which should help with compile times.

afsalthaj:
@mschuwalow  Tried to resolve many of your comments thanks for the review. Transformers are top level functions and so for middleware in and out (initially I had troubles, propagation of some history there, but fixed it). I think these are useful comments.

I would prefer doing the rest (if any later), especially IdentityProvider. Initially provider is a trait, but not each one of them. Initially they were in the commit history, and later to moved to a single IdentityProvider implementation. If we have diverging implementations we can consider it as each provider as a trait. Or we can think about it later

golem-worker-service-base/Cargo.toml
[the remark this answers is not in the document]
vigoo: Previously I was "guarding" and cleaning up our repo several times by manually checking if there is any dependencies that are shared and moving them to the workspace level, thinking that it's nice to see what dependencies are shared and what not (keeping them in the inner cargo files). But I no longer really believe this worth the effort so I'm ok if we move all dependencies to the workspace if there are no other known side effects of that (I don't know any).

golem-worker-service-base/src/api/register_api_definition_api.rs
afsalthaj: User facing response site, and I didn't want to bother too much about it.
Anything core logic that's done internally is type safe and its in enums as far as I believe.

jdegoes (approved):

PR #1085, opened 2024-11-25
Plugins in worker executor

### LINKED ISSUE #1036: Active oplog processor plugins per worker
Worker executor must track the active set of oplog processor plugins for each worker. 
In details:

- The initial set comes from the component's enabled set (and new optional parameters if the worker was explicitly started)
- The Worker Executor gRPC API and Golem host functions allow adding/removing plugins for a worker
- New oplog entries need to be added when the set of plugins change
- The current active set is cached in the worker status record and is recomputable from the oplog

The outcome of this task is that there is full support for managing what oplog processor plugins are enabled for a worker but there is no actual oplog broadcast to them yet.

vigoo:
Resolves #1036


golem-common/src/model/mod.rs
```diff
@@ -1031,6 +1031,62 @@ pub struct WorkerStatusRecord {
...
+            failed_updates: Decode::decode(decoder)?,
+            successful_updates: Decode::decode(decoder)?,
+            invocation_results: Decode::decode(decoder)?,
+            current_idempotency_key: Decode::decode(decoder)?,
+            component_version: Decode::decode(decoder)?,
+            component_size: Decode::decode(decoder)?,
+            total_linear_memory_size: Decode::decode(decoder)?,
+            owned_resources: Decode::decode(decoder)?,
+            oplog_idx: Decode::decode(decoder)?,
+            active_plugins: Decode::decode(decoder).or_else(|err| {
```
mschuwalow, line 1054: This doesn't look so nice. Can we include some versioned metadata or something like this past this point so we can evolve this in the future?

I understand that you have a lib that can handle the evolution, maybe we can serialize everything past oplog_idx using your lib and put it in here.
During deserialization we can check for UnexpectedEnd like now, otherwise start reading data with your decoder including schema evolution.

golem-worker-executor-base/src/durable_host/mod.rs
```diff
@@ -573,6 +585,7 @@ impl<Ctx: WorkerCtx + DurableWorkerCtxView<Ctx>> DurableWorkerCtx<Ctx> {
                                         .on_worker_update_succeeded(
                                             target_version,
                                             component_metadata.size,
+                                            HashSet::new(), // TODO: get it from component_metadata
```
mschuwalow, line 588: Part of this pr or as a followup?

golem-worker-executor-base/src/services/plugins.rs
```diff
@@ -0,0 +1,476 @@
...
+use std::sync::Arc;
+use tonic::codec::CompressionEncoding;
+use tonic::transport::Channel;
+use uuid::Uuid;
+
+/// Part of the `Plugins` service for recording observed information as a way to pre-cache
+/// data. It is in a separate trait because it does not have to be parametric for the Owner/Scope
+/// types.
+#[async_trait]
+pub trait PluginsObservations {
```
mschuwalow, line 43: Do we need to care about invalidating this cache or will we always end up calling observe_plugin_installation with uncached data when (re) running a worker?

golem-worker-executor-base/src/durable_host/mod.rs
vigoo: No, I'm fixing these - just marked as "ready" too soon

golem-worker-executor-base/src/services/plugins.rs
vigoo: The installations are immutable so no need for invalidation. (And old entries will be evicted when the cache is full)

golem-worker-executor-base/src/worker.rs
```diff
@@ -822,6 +823,28 @@ impl<Ctx: WorkerCtx> Worker<Ctx> {
         receiver.await.unwrap()
     }
 
+    pub async fn activate_plugin(
```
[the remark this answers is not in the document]
vigoo: That logic is now one layer above (in the gRPC request handler) - I think I would keep it there for now for consistency with the existing code

golem-worker-executor-base/src/grpc.rs
```diff
@@ -1303,31 +1311,177 @@ impl<Ctx: WorkerCtx, Svcs: HasAll<Ctx> + UsesAllDeps<Ctx = Ctx> + Send + Sync +
...
+
+        match metadata {
+            Some(metadata) => {
+                // Worker exists
+
+                if worker_status
+                    .active_plugins
+                    .contains(&plugin_installation_id)
+                {
+                    Err(GolemError::invalid_request("Plugin is already activated"))
```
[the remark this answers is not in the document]
vigoo: I was thinking of that, and could not decide.

PR #1114, opened 2024-12-03
Docker image fixes to avoid runtime errors in ubuntu/linux machines

afsalthaj:
### Why now?
I had to do this to try golem as a user before its 1.1 release to see if anything obvious is missing and [patch the code](https://github.com/golemcloud/golem/pull/1115) straight away. On top of that the main purpose of this PR is as follows:

_**In case anyone reports issues with 1.1 release, we got this PR  to come back to and see if it solves their problem. As discussed within the team, if this doesn't resolve the issue, we will not be spending much time trying to fix it though**_

### Context 
This is _NOT_ regarding cross compilation _builds_ or docker-compose _builds_, but actually running golem docker  images in latest ubuntu, and that resulted in incompatibility errors at _runtime_ (which we don't test anywhere). I don't think anyone tested running golem other than Mac (or if at all they ran, it accidentally worked)

I haven't spent much time finding out the right version of ubuntu or the cleanest approach (or the slimmest image with the right version since this task is unplanned), and therefore simply using `latest` to avoid incompatibility issues.

#### Environment
Ubuntu with [multipass](https://multipass.run), with the following version

```
ubuntu@foo:~$ ldd --version
ldd (Ubuntu GLIBC 2.39-0ubuntu8.3) 2.39
Copyright (C) 2024 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper
```

##### Commands Run

`cargo make --profile ci build-release` with necessary platform overrides, within ubuntu, and then built images with the current docker-files successfully, however simply running any of them failed with lots of errors.

##### Errors without this change:
Most importantly trying to run them in an ubuntu, resulted in glibc compatibilities (not to mention other errors encountered before narrowing it down to glibc incompatibilities). The cause was Debian slim image was at `2.34` while the required one for our app built in [ubuntu-latest ](https://github.com/golemcloud/golem/blob/main/.github/workflows/ci.yaml#L23) needed a specific version.

With the changes in the PR, there are no more runtime issues running golem images in Ubuntu.


golem-component-compilation-service/docker/Dockerfile
```diff
@@ -16,8 +16,8 @@ WORKDIR /app
 COPY /target/$RUST_TARGET/release/golem-component-compilation-service ./
 COPY /golem-component-compilation-service/config/ ./config/
 
-RUN apt-get update && apt-get install -y libssl-dev
-RUN apt-get update && apt-get install -y ca-certificates
+RUN apt-get update --allow-releaseinfo-change && apt-get install -y libssl-dev
```
kmatasfp, line 19: not sure if  `--allow-releaseinfo-change` is needed here. Could be be needed if we used pinned version of ubuntu, but as we use ubuntu:latest it will handle the repo information change like suite, codename, version etc

CONTRIBUTING.md
```diff
@@ -86,16 +86,17 @@ The `cargo make run` command will start all the services and show a unified view
...
 
-```bash
-# Clone golem
-cd golem
+### Example: 
 
-# Find more info below if you are having issues running this command(example: Running from MAC may fail)
-# Target has to be x86_64-unknown-linux-gnu or aarch64-unknown-linux-gnu-gcc
-cargo build --release --target x86_64-unknown-linux-gnu
+Use `multipass` or similar platforms, with a reasonable disk space (since it's the easiest to get a ubuntu)
```
kmatasfp, line 93: might be worth mentioning that we recommend ubuntu box with following dependencies installed on top of rust toolchain to compile golem:
```
sudo apt install pkg-config libssl-dev libfontconfig1-dev
```

As they might use their own tooling to get ubuntu vm (vagrant for example is really popular) or already have access to some other ubuntu box

afsalthaj:
@kmatasfp  Feel free to raise PRs on what you did on your end to get these things working like https://about.gitea.com/

golem-component-compilation-service/docker/Dockerfile
afsalthaj: I think I needed `--allow-releaseinfo-change` to get something working.

kmatasfp: Just verified with fresh ubuntu vm that `--allow-releaseinfo-change` was not needed, also doing `apt-get update` twice is not needed, all of this can be replaced with
```
RUN apt-get update && apt-get install -y libssl-dev ca-certificates
```

kmatasfp:
After you merge this PR I will create a followup PR that describes how to tag and publish built container images to registry, be it self hosted like gitea or dockerhub, github, ecr so you can actually use these built images on your mac which is end goal for most devs

CONTRIBUTING.md
kmatasfp: also dependent on you starting point, this might be needed
```
sudo apt install build-essential

```

kmatasfp: Also this 
```
sudo apt-get install protobuf-compiler
```

PR #1115, opened 2024-12-03
Final clean ups, bug fixes, additional tests (due to bugs) based on UAT

afsalthaj:
* Found bugs, therefore added missing tests further
* A few scenarios where we exposed internal endpoints related to security is removed. As a user, the definition I registered should be the one I get back in `list` and `get` endpoints, otherwise it confuses.
* Minor confusing response schema names (`WithTypeInfo`) is replaced with `ResponseData` which is consistent across the code.
* Added a few tickets based on a few observations as a user, but not critical: #1116  #1119  #1118


golem-worker-service-base/src/api/register_api_definition_api.rs
```diff
@@ -163,17 +163,27 @@ pub struct HttpApiDefinitionResponseData {
...
 
-        Self {
+        for route in value.routes {
+            // We shouldn't expose auth call back binding to users
+            // as it is giving away the internal details of the call back system that enables security.
+            if !route.binding.is_security_binding() {
+                let v = RouteWithTypeInfo::try_from(route)?;
+                routes.push(v);
+            }
+        }
```
afsalthaj, line 178: In case this if less condition taken off by anyone, tests will fail with error messages, that the auth call back endpoint shouldn't be exposed.

vigoo (approved):

PR #1145, opened 2024-12-06
Easy Dev env setup for a contributer

kmatasfp:
### Abstract

Inspired by now closed PR https://github.com/golemcloud/golem/pull/1114 which proposed to use [multipass](https://multipass.run/) to spin up a vm, install necessary dependencies and then do the docker image building within the vm and then use some external registry to push the images so one can use them on their mac.

With power of [Declarative Development Environments](https://nixos.wiki/wiki/Development_environment_with_nix-shell) powered by Nix we can just  `cd` to our `golem` directory and we will have everything we need to compile and even cross compile to aarch64.

#### Native to native dev env
```
cd golem
nix develop
``` 

#### Native to  linux aarch64
```
cd golem
nix develop .#cross-arm64
```

#### Extra credit
Use power of  [Nix direnv](https://github.com/nix-community/nix-direnv) we can just `cd golem` and execute `cargo build` without needing to install anything ourselves

#### Requirements
Nix installed and flakes enabled in `~/.config/nix/nix.conf` like so
```
experimental-features = nix-command flakes
``` 

#### Future possibilities
Can also be used to provision CI servers and build docker images that are published for everyone to use

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ golem-client/*
 logs
 data
 .DS_Store
+.direnv
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,82 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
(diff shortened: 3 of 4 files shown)


vigoo:
Thanks, but as discussed earlier, we don't want to maintain nix support in the golem repo itself.

kmatasfp:
To clarify this was not for building/packaging golem using nix like some of the previous PRs, but for setting up toolchain in a declarative way to be able to work on golem, but fair enough

PR #1154, opened 2024-12-06
WIP: Workload Identities

smndtrl:
This is a WIP!

The goal is for a worker component to be able to attest to a remote entity (API, STS) that it is part of the golem environment. The remote entity can make their decisions based on the attested claims to e.g. allow API calls or in case of a secrets vault unlock the secret just for a specific component in a specific account.

It adds a API for components to retrieve a signed JWT with claims about the workers account, component id, worker name.

* worker-executor has the signing capabilities and `golem:api/identity.{get-token}`
* worker-service has a JWKS endpoint + for compatibility a openid-configuration discovery endpoint

Both must share the configuration with the keys. The worker service just needs to get the public information to display it on the endpoints while the worker executor actually needs the private parts to sign the JWT.

There is a lot to be done and cleaned up. I wanted to make it public for awareness and to see if this is of interest to the project.


vigoo:
Closing this as Golem changed a lot since the draft. Please reopen or just open a feature request ticket if you are still interested!

PR #1157, opened 2024-12-09
Bug fixes in gateway

afsalthaj:
* Make sure worker name can look up athentication details, and not just response body.

* JWKset were never serialized (its "skipped" in openidconnect library), and provider-metadata reuse for performance reasons resulted in signature mismatch errors. So now for claim verification, we discover the keyset on the fly, and is now consistently working.
* 2 test additions in Rib, which I thought was missing (but there is no bug here)

Here is a simple end user test with [this](https://github.com/afsalthaj/golem-worker-gateway-examples/blob/main/html/api-definition.yaml#L33) api-definition, where both worker name and response mapping looks up authentication fields.

image


golem-rib/src/interpreter/rib_interpreter.rs
```diff
@@ -1445,6 +1445,68 @@ mod interpreter_tests {
...
+
+            let result = interpreter
+                .run(compiled.byte_code)
+                .await
+                .unwrap()
+                .get_val()
+                .unwrap();
+
+            assert_eq!(result, TypeAnnotatedValue::Str("foo, bar".to_string()));
+        }
```
afsalthaj, line 1508: Added these tests, but there are no bugs in logic

golem-worker-service-base/src/gateway_execution/gateway_binding_resolver.rs
```diff
@@ -36,30 +42,37 @@ pub trait GatewayBindingResolver<Namespace, ApiDefinition> {
...
-    RouteNotFound,
+pub enum ErrorOrRedirect {
+    Error(GatewayBindingResolverError),
+    Redirect(poem::Response),
 }
 
-impl GatewayBindingResolverError {
-    pub fn internal(err: &str) -> Self {
-        GatewayBindingResolverError::Internal(err.to_string())
+impl ErrorOrRedirect {
```
afsalthaj, line 54: During the binding resolution phase itself (resolving a binding based on a http request, where the worker name is also resolved (Ex: worker-${request.auth.given_name})), middlewares such as authentication is applied, and can decide to redirect or not. That's the main change.

golem-worker-service-base/tests/api_gateway_end_to_end_tests.rs
```diff
@@ -2490,6 +2512,15 @@ nUhg4edJVHjqxYyoQT+YSPLlHl6AkLZt9/n1NJ+bft0=
...
         )
+        .set_jwks({
+            let public_key = rsa::RsaPublicKey::from_public_key_pem(TEST_PUBLIC_KEY)
+                .expect("Failed to parse public key");
+
+            let n = public_key.n().to_bytes_be();
+            let e = public_key.e().to_bytes_be();
+            let kid = JsonWebKeyId::new("my-key-id".to_string());
+            JsonWebKeySet::new(vec![CoreJsonWebKey::new_rsa(n, e, Some(kid))])
+        })
```
afsalthaj, line 2523: Now the test case is better, that the provider metadata will have the public key in it.
And the corresponding id_token will have its private key. Much easier to understand tests

noise64 (approved):

PR #1193, opened 2024-12-14
Make running grpc api configurable

afsalthaj:
Also made the service creation to a separate function for better readability


vigoo (changes requested):
This does not seem right to me - if the gRPC API is disabled, it does not create any of the internal services; but the goal is to have everything properly initialized (the result of `create_services`) and just don't run the gRPC server itself.

vigoo:
So now this implements what the PR says (and what I hinted in the related task) but it won't really allow us to build an alternative server API on top of it.

Basically it has the following issues:
- We are creating this `WorkerExecutorImpl` struct - but that's actually the type implementing the gRPC API which we wanted to make optional. So creating it and then not starting a gRPC API does not make much sense.
- When we will build another API that replaces the worker executor gRPC API we will need access to all the services created in the boostrap code - but this is not possible now - it just "hides" them in the `WorkerExecutorImpl` and then that's not accessible in any way.
- We want the `run` function of boostrap to not start the default gRPC API - this is achieved - but also we will need to start our own without duplicating the whole boostrapping logic.

So I propose the following implementation instead of the one in the PR:

- We don't need the `run_grpc_server` method
- Instead add a `run_server` method that takes `All<Ctx>` as a parameter and its default implementation in the trait starts the default gRPC server, but by implementing it in the new service, we will be able to overwrite it and start our own server while still having access to `All` the dependencies we created.

PR #1194, opened 2024-12-15
App manifest more cleanups

noise64:


vigoo:
I see the composed WASM files' name has changed (lost the `-composed` part). Can we change it back (override it in the manifest, I guess)? 
Reason is I already added the uncomposed versions of these WASMs in https://github.com/golemcloud/golem/pull/1188 to have tests both with linked stub and without. 

We probably will remove the "stubbed" ones once everything is done but for now there is no option yet to skip the compose part in the stub generator, so I just wanted to have the manually created "non-stubbed" wasms as "caller.wasm" and the stubbed as "caller-composed.wasm" to be able to have them both in the tests.

noise64:
@vigoo will change those back with manifest, sorry for dropping them

wasm-ast/wasm/auction_registry_composed.wat
vigoo, line 1: I don't think we need this file, was itin the repo already?

noise64: image
it is on master

wasm-rpc-stubgen/src/rust.rs
```diff
@@ -394,6 +394,7 @@ fn generate_function_stub_source(
     );
 
     let rpc = match mode {
+        // TODO: these restrictions on static seems to be arbitrary, let's recheck
```
vigoo, line 397: If you check what it is used for, you see that it is a real limitation. The stub implementation requires an `rpc` value (which is a host owned resource of the wasm-rpc implementation). For a constructor, it is created in the stub's constructor so it is the local variable `rpc`. For methods, it is coming from `self.rpc`. For a static method in general, there is no way to get the reference to it. So this is a workaround making it at least somewhat usable, by supporting static functions that are in reality "methods" just not marked as such.

A better solution would be to construct a new rpc resource temporarily for these calls (the whole thing was designed in a way that the RPC resource is kept open to allow optimizations like keeping an open channel between workers). 

But as we are going to drop the generated stubs anyway, I think we can live with this limitation for now and solve it only in the dynamic linked stubs.

wasm-rpc-stubgen/src/rust.rs
```diff
@@ -476,10 +477,10 @@ fn generate_function_stub_source(
                 Span::call_site(),
             );
             let root_name = Ident::new(
-                &format!("{}_stub", def.source_package_name.name.to_snake_case()),
+                &format!("{}_client", def.source_package_name.name.to_snake_case()),
```
vigoo, line 480: Should not all these format!-d names coming from the `names` module instead?

noise64: they should

noise64: i will do another round on searching for them

noise64: updated the PR with more naming things

wasm-ast/wasm/auction_registry_composed.wat
noise64: deleted

vigoo (approved):

golem-cli/src/command/component.rs
```diff
@@ -703,7 +704,10 @@ impl ApplicationComponentContext {
             .application_context
             .application
             .component_wasm_rpc_dependencies(component_name)
-            .clone();
+            .iter()
```
vigoo, line 707: I actually did not do this intentionally. But ok to change it.

(It does not really matter because the dynamic linker is only handling the interfaces that were not linked statically)

vigoo (approved):

golem-cli/src/command/component.rs
noise64: it was confusing for me that it was adding static deps into the dynamic_linking field

noise64: especially not that we show them in the cli

Issue #1189, opened 2024-12-13
Create Open Source GUI for Golem in TypeScript


jdegoes:
Golem only has two methods of interaction:

 - The REST API
 - The Command-Line Interface (`golem-cli`)

Both of these are useful for programmatic purposes, while the CLI is designed mostly for developers in mind. However, neither of these methods are discoverable or provide a high-level graphical user-interface with a wonderful developer experience. While Golem Cloud comes with [Console](https://console.golem.cloud), a high-level graphical user-interface, Console is deeply integrated into the Cloud offering and relies on APIs that don't even exist in OSS or make sense there.

In order to provide developers a new graphical way of interacting with Golem, we wish to sponsor a ticket that is primarily intended for TypeScript developers -- with a shout-out to developers familiar with Effect TS (though use of this library or any other is not required). The end result will be a GUI for Golem that ships with the single executable (even though you could use it elsewhere) and makes it easy for developers to do local development and testing.

In order to claim this bounty, you need two do two things:

1. Build a TypeScript application that allows graphical management of Golem, supporting all the APIs that Golem supports. This requires knowledge of TypeScript, front-ends, cross-browser development, user-experience, user-interface design (including usability), and developer experience.
2. Serve the application as part of the Single Executable build of Golem (which bundles everything together in a single executable). This may require a tiny amount of Rust knowledge, but you can probably get by with Google and Copilot.

In particular, the application must support:

1. **Component management**. Developers can create, update, and delete components visually. All backend APIs should have a way to perform the function visually.
2. **Worker management**. Developers can create, update, manage, and delete workers visually. All backend APIs should have a way to perform the function visually.
3. **API management**. Developers can create, update, manage, and delete APIs visually. All backend APIs should have a way to perform the function visually.
4. **Plugin management**.  All backend APIs should have a way to perform the function visually (chiefly about installing / uninstalling plug-ins inside components and workers).

Although one can take inspiration from [Console](https://console.golem.cloud), this is not necessary, as Cloud is designed for use in a multi-tenant environment, which does not apply to Golem. However, the breadth of support for backend APIs, as well as the polish and attention to aesthetics, should be similar to Console. You are free to make your own decisions on design, user-interface, user-experience, etc. However, if you choose poorly, and your user-interface is ugly, difficult to use, inconsistent, incomplete, or buggy, then your PR won't be merged.

A winning pull-request will include screenshots of every major area of the user-interface and attach pre-compiled single-executable binaries for Linux / MacOS or both.

[...]

PR #1195, opened 2024-12-16
Add Web UI for managing Golem

Myestery:
/claim #1189 

Screenshot 2024-12-16 at 02 16 56
Screenshot 2024-12-16 at 02 16 27
Screenshot 2024-12-16 at 02 16 12
Screenshot 2024-12-16 at 02 15 39
Screenshot 2024-12-16 at 02 15 33
Screenshot 2024-12-16 at 02 15 20
Screenshot 2024-12-16 at 02 15 14
Screenshot 2024-12-16 at 02 15 01
Screenshot 2024-12-16 at 02 14 54
Screenshot 2024-12-16 at 02 14 43


jdegoes:
@Myestery

Thanks for submitting your proposed solution to #1189!

I have not reviewed the code very closely, because first the UI must be in a state where it meets the requirements for the ticket. Once the UI meets the requirements, I will proceed to perform a code review, which may ask for additional revisions.

Overall, I want to congratulate you on so quickly developing a user-interface for the REST API! It was easy for me to find my way around the user-interface, and for the most part, it feels snappy an modern. Great job!

Now, I have to say the UI still need a lot of work before I can consider merging it. Let me review the main areas:

 - **Branding**. The UI is not branded with Golem logo or name, and in fact, the title is "Vite + React + TS" from the template used. The color scheme is not designed with https://golem.cloud in mind.
 - **Workers**. Aside from a few toolbar actions, there is no user-interface for workers. While ephemeral workers do not need a worker page, each durable worker needs its own page, which can be used to interact with the worker. All functionality related to the worker should be accessible from the worker page, including updating, logs, system events, configuration, files in the file system, etc. This is one of the biggest omissions of your user-interface. See Console for some inspiration around this.
 - **Browser-based Routing**. Here I mean routing in the sense of URL mapping to a user-interface view--if you refresh each page (for example, a given API Definition page), sometimes you get "not found", which is a sign routing is not implemented properly.
 - **Route Editing**. Route editing does not support all the binding types (for example file server, which was recently added, and allows serving static files at an endpoint; see Initial File System Issue for documentation). Further, it is not designed with Rib in mind. Rib is how one enters the worker name and the response. Right now, these are not text areas but text fields, and there is no support for inserting function names in Rib or specifying the component. It feels very incomplete and was clearly not used to build anything significant.
 - **Error Management**. In the happy path, most things work (adding API definition, uploading component, etc.), but in the unhappy path, no inside-UI error is reported, one has to check the JS error console. For example, add a route using invalid response or worker-name. In general, every action must provide inline help when bad data leads to failure of the API. That's not being done generally or at all right now.
 - **File System**. Components may have attached files, which seems to be supported; but the read/read-write status of these files cannot be changed in the interface. In addition, files cannot be browsed on a running worker (related to missing worker UI). 

There are a number of other missing features and usability improvements that are required, but I will wait until if and when you decide to address these issues. I do think the pull request can be pushed forward to completion, but what I **strongly recommend** is building a TODO app in one of the Golem-supported languages, including with a full REST API and simple UI served by the backend. Once you write the code, force yourself to upload the component, create all API definitions, and serve the static content, using only the front-end you built. This will force you to add missing features and fix bugs that will end up making the UI production-ready and mergeable (right now, though a promising start, the UI is not really production-ready and does not replicate or extend upon the features of Console).

Please re-open when you get a chance to work on these issues!

Issue #1178, opened 2024-12-12
Develop OpenAPI Export for API Definition


jdegoes:
Golem is a platform for rapidly building highly reliable backends. Powered by durable computing, Golem's serverless platform automatically provides fault-tolerance, so that running software becomes impervious to hardware and software faults; together with automatic durable state (in-memory structures are durable) and exactly-once execution semantics. Together, these features allow rapidly building robust solutions to all mission-critical problems, such as AI agents, financial transaction processing, onboarding, business processes, and long-running workflows.

The process of building a backend on Golem looks like this:

 - Develop a new component in a supported programming language, using [WIT](https://component-model.bytecodealliance.org/design/wit.html) to describe the public API of the component. This API can be called from other components in a type-safe way, but it is also the basis of higher-level HTTP / gRPC APIs added atop.
 - Although there exist APIs to invoke running instances of a component (called *workers*) using their WIT-based API, this would not be a pleasant development experience. So in general, you would use Golem's _Worker Gateway_ to add a custom API atop the component(s) of the backend
 - The Worker Gateway allows creation of API Definitions, which have a number of routes. Each route is defined by path (which may include variables), query variables, and then "worker name" and "response", which are defined using  _Rib_, a tiny scripting language. Using Rib for the worker name allows directing specific API calls to specific workers; while using Rib for the response allows you to define which function (defined in WIT and exported by your component) you want to invoke when that particular endpoint is hit from the front-end (or other backend, in a microservice architecture).

This ticket involves adding a way to export API Definitions into OpenAPI Schemas, which must be valid and faithful representations of the APIs created by the Worker Gateway. In addition, the ticket also specifies a way for developers to add a Swagger UI endpoint to their custom APIs, so they can have self-describing and testable APIs (this feature is in fact powered by the OpenAPI Schema generation feature). Finally, the ticket specifies a way to launch Swagger UI during local development and testing, for purposes of interacting with an existing API being served by the Worker Gateway.

The following sections describe these features in more detail.

-------------------------------

In order to make it easy to standup REST APIs atop workers, Golem has a concept of "API Definition", which consists of a number of routes, with binding types, and implemented by the Worker Gateway (worker bridge + worker services).

Currently, two binding types are supported in the Worker Gateway:

[...]

PR #1202, opened 2024-12-18
Develop OpenAPI Export for API Definition #1178

zelosleone:
# API Definition OpenAPI Export & Swagger UI Integration

### OpenAPI Export
- Added conversion logic to transform Golem API Definitions into OpenAPI Spec format
- Implemented type mapping between Golem types and OpenAPI schemas
- Preserved request/response types, security schemes, and CORS configurations

### Swagger UI Integration 
- Created new SwaggerGenerator to manage Swagger UI assets and configurations
- Added SwaggerUI binding type for API routes
- Implemented automatic Swagger UI generation and file serving

### Testing
- Added comprehensive test coverage including:
  - Unit tests for OpenAPI conversion
  - Integration tests validating Swagger UI serving
  - End-to-end tests verifying generated client usage

## Technical Details
```rust
// New SwaggerUI binding support
impl SwaggerGenerator {
    pub fn new<P: AsRef<Path>>(output_dir: P) -> Self { ... }
    pub async fn generate(&self, api_id: &str, spec_path: &str) -> Result<()> { ... }
    pub async fn clean(&self, api_id: &str) -> Result<()> { ... }
}

// OpenAPI conversion & validation 
#[tokio::test]
async fn test_openapi_export_flow() {
    let api = create_test_api().await;
    let spec = OpenAPIConverter::convert(&api);
    assert!(validate_openapi(&spec).is_ok());
}
```

## Key Features
- Full OpenAPI 3.0 spec generation preserving all API features
- Automated Swagger UI integration
- Support for security schemes and CORS 
- Client library generation compatibility

### CI Configuration Updates:
- This should be ignored, i mistakenly thought i can use the same thing as my local in github without noticing it goes over disk space of github. I deleted the file from commit but please make sure to not to accept that part.

### Dependency Updates:
* Added new dependencies `heck` and `validator` and updated the `wasmtime` dependencies to version `21.0.1` in `Cargo.toml`. [[1]](diffhunk://#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R75-R76) [[2]](diffhunk://#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542L216-R229)
* Updated `golem-wit` dependency version to `1.1.0` in `golem-worker-executor-base/Cargo.toml`.

### Code Refactoring:
* Moved the `ParsedFunctionSite` and `ParsedFunctionReference` enums to a new location in `golem-rib/src/function_name.rs` and added a new `ParsedFunctionName` struct. [[1]](diffhunk://#diff-a622e8a3e91db3324af3447284a3f159b67e4b11ab3b460b617eda28a54a9f84R25-R38) [[2]](diffhunk://#diff-a622e8a3e91db3324af3447284a3f159b67e4b11ab3b460b617eda28a54a9f84L97-L110) [[3]](diffhunk://#diff-a622e8a3e91db3324af3447284a3f159b67e4b11ab3b460b617eda28a54a9f84R132-R170) [[4]](diffhunk://#diff-a622e8a3e91db3324af3447284a3f159b67e4b11ab3b460b617eda28a54a9f84L281-L319) [[5]](diffhunk://#diff-a622e8a3e91db3324af3447284a3f159b67e4b11ab3b460b617eda28a54a9f84R494-R499) [[6]](diffhunk://#diff-a622e8a3e91db3324af3447284a3f159b67e4b11ab3b460b617eda28a54a9f84L568-L573)
* Added a `clone` method to `init_tracing` call in `golem-cli/src/lib.rs`.
* Added async function permits in various methods of `DurableWorkerCtx` in `golem-worker-executor-base/src/durable_host/blobstore/container.rs`. [[1]](diffhunk://#diff-edde0121fe3037d155db686d57872255bc9f31e2fdca3861c8ee850129730f76R38) [[2]](diffhunk://#diff-edde0121fe3037d155db686d57872255bc9f31e2fdca3861c8ee850129730f76R52) [[3]](diffhunk://#diff-edde0121fe3037d155db686d57872255bc9f31e2fdca3861c8ee850129730f76R72) [[4]](diffhunk://#diff-edde0121fe3037d155db686d57872255bc9f31e2fdca3861c8ee850129730f76R116) [[5]](diffhunk://#diff-edde0121fe3037d155db686d57872255bc9f31e2fdca3861c8ee850129730f76R152) [[6]](diffhunk://#diff-edde0121fe3037d155db686d57872255bc9f31e2fdca3861c8ee850129730f76R190) [[7]](diffhunk://#diff-edde0121fe3037d155db686d57872255bc9f31e2fdca3861c8ee850129730f76R222)

This implementation provides seamless integration between Golem's API definitions and standard OpenAPI tooling while maintaining all the power of Golem's type system and binding features.

Docs Pull Request: https://github.com/golemcloud/docs/pull/105

/claim #1178


golem-worker-service/src/api/openapi/converter.rs
```diff
@@ -0,0 +1,1011 @@
...
+                },
+                security: None,
+                tags: Some(vec![route.template_name.clone()]),
+            },
+        }
+    }
+
+    fn extract_path_parameters(path: &str) -> Option<Vec<Parameter>> {
+        let params: Vec<Parameter> = path
+            .split('/')
```
jdegoes, line 213: It is a bad idea to parse this yourself, because it can diverge from the official parser in the worker service (which already supports extensions such as `{foo..}` for multi-segment matches; and will eventually support all RFC-driven extensions).

golem-worker-service/src/api/types.rs
```diff
@@ -0,0 +1,190 @@
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct OpenAPISpec {
```
jdegoes, line 5: Why not use one of the Rust crates that already models OpenAPI v3?

For example:

 - <https://github.com/kurtbuilds/openapiv3>
 - <https://github.com/glademiller/openapiv3>

(I am not endorsing these libraries.)

golem-worker-service/src/api/validation.rs
```diff
@@ -0,0 +1,93 @@
...
+                validate_wit_type(input_type)?;
+                validate_wit_type(output_type)?;
+            }
+            _ => {}
+        }
+    }
+    Ok(())
+}
+
+fn validate_wit_type(wit_type: &str) -> Result<(), String> {
```
jdegoes, line 17: I haven't seen where this function is used yet, but if we are ever validating WIT types represented as plain strings, we have taken a wrong turn somewhere!

There are strongly-typed representations of WIT types in the project that can be used in place of strings.

golem-worker-service/src/api/openapi/converter.rs
zelosleone: You are right then, I am gonna use `AllPathPatterns::parse()` to properly parse paths then extract path parameters from the parsed patterns. And map the path parameters to OpenAPI Parameter objects.

golem-worker-service/src/api/types.rs
zelosleone:
```
// Custom OpenAPI types
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct OpenAPISpec {
    pub openapi: String,
    pub info: Info,
    pub paths: HashMap<String, PathItem>,
    pub components: Option<Components>,
    pub security: Option<Vec<SecurityRequirement>>,
}
```

As you can see, we can achieve a custom implementation by going with this instead. How about just adding them together for validation against openapiv3 crate's model?

golem-worker-service/src/api/validation.rs
zelosleone: You are absolutely right! It's my mistake to do it like this! Thanks for pointing out! Changing this too.

golem-worker-service/src/api/types.rs
zelosleone: After careful consideration, I am gonna go with your suggestion on this too.

zelosleone:
The code has been updated to incorporate feedback from @jdegoes, utilizing a custom OpenAPIv3 implementation and glademiller crate for validation.  Further review and feedback are appreciated.

golem-rib/src/function_name.rs
```diff
@@ -22,6 +22,20 @@ use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 use std::fmt::Display;
 
+#[derive(Debug, PartialEq, Eq, Clone, Encode, Decode)]
+pub enum ParsedFunctionSite {
+    Global,
+    Interface {
+        name: String,
+    },
```
jdegoes, line 30: To avoid spurious conflicts in a project like this, it's best to avoid moving code and definitions around within a file. If this is your IDE doing this, I recommend turning off the feature.

zelosleone: I appreciate your suggestion.  This has happened to me before too, so thank you for your help!

golem-worker-executor-base/src/durable_host/blobstore/container.rs
```diff
@@ -35,6 +35,7 @@ impl<Ctx: WorkerCtx> HostContainer for DurableWorkerCtx<Ctx> {
         &mut self,
         container: Resource<Container>,
     ) -> anyhow::Result<Result<String, Error>> {
+        let _permit = self.begin_async_host_function().await?;
```
jdegoes, line 38: Why the introduction of this change? It increases the boilerplate required, and also will be difficult to enforce, making it error-prone. Finally, if we did need such a change, we would want to make `record_host_function_call` always return the permit, since that must be the first thing done in these functions, anyway.

zelosleone: The primary reason was my OpenAPI3 implementation testing. My tests couldn't handle the minimal boilerplate, requiring an increase. However, I can remove it if needed.

golem-worker-executor-base/src/durable_host/blobstore/container.rs
```diff
@@ -410,3 +422,112 @@ impl<Ctx: WorkerCtx> HostStreamObjectNames for DurableWorkerCtx<Ctx> {
 
 #[async_trait]
 impl<Ctx: WorkerCtx> Host for DurableWorkerCtx<Ctx> {}
+
+#[async_trait]
+impl<Ctx: WorkerCtx> HostContainer for &mut DurableWorkerCtx<Ctx> {
```
jdegoes, line 427: I still don't know the purpose of these traits but there's a lot of boilerplate here, if possible it would be nice to avoid it (will look for other implementations that may explain).

zelosleone: You might be right on the current permit system for host function calls having flaws.  It might add unnecessary boilerplate and is prone to developer errors due to the separate permit acquisition step.

Maybe we can just integrate permit acquisition into `record_host_function_call`.

This improved function could:

* Reduce redundant code.
* Guarantee permit acquisition.
* Maintain existing synchronization.
* Eliminate the possibility of forgetting to acquire a permit.

golem-worker-executor-base/src/durable_host/filesystem/types.rs
```diff
@@ -75,16 +75,28 @@ impl<Ctx: WorkerCtx> HostDescriptor for DurableWorkerCtx<Ctx> {
         length: Filesize,
         advice: Advice,
     ) -> Result<(), FsError> {
+        let _permit = self
```
jdegoes, line 78: Ditto for this comment. All these async permits will add overhead and boilerplate and are difficult to enforce.

golem-worker-executor-base/src/durable_host/filesystem/types.rs
```diff
@@ -484,6 +584,266 @@ impl<Ctx: WorkerCtx> Host for DurableWorkerCtx<Ctx> {
...
+    async fn get_type(&mut self, self_: Resource<Descriptor>) -> Result<DescriptorType, FsError> {
+        (*self).get_type(self_).await
+    }
+
+    async fn set_size(
+        &mut self,
+        self_: Resource<Descriptor>,
+        size: Filesize,
+    ) -> Result<(), FsError> {
+        (*self).set_size(self_, size).await
```
jdegoes, line 639: Lots of pure delegations here that are adding a lot of boilerplate. What's the purpose of this indirection (again, will look for other implementations to understand better)?

golem-worker-service/src/api/definition/binding.rs
```diff
@@ -0,0 +1,48 @@
...
+impl From<BindingType> for base::BindingType {
+    fn from(binding: BindingType) -> Self {
+        match binding {
+            BindingType::Default { input_type, output_type, function_name } => 
+                base::BindingType::Default { input_type, output_type, function_name },
+            BindingType::FileServer { root_dir } => 
+                base::BindingType::FileServer { root_dir },
+            BindingType::SwaggerUI { .. } => 
+                // SwaggerUI bindings are handled specially by the worker service
+                panic!("Cannot convert SwaggerUI binding to base binding type"),
```
jdegoes, line 45: I would do `TryFrom` here rather than `From`.

zelosleone: For converting SwaggerUI bindings, it's better to use panics than errors.  A panic clearly shows a mistake in the code that should be found during review or testing, unlike an error which might only be noticed while the program is running. This makes the code easier to understand and helps find problems during testing. But i can go with your approach if you want.

golem-worker-service/src/api/definition/patterns.rs
```diff
@@ -0,0 +1,48 @@
...
+    Var(VarInfo),
+    CatchAllVar(CatchAllVarInfo),
+}
+
+#[derive(Debug)]
+pub struct AllPathPatterns {
+    pub path_patterns: Vec<PathPattern>,
+}
+
+impl AllPathPatterns {
```
jdegoes, line 23: Is this simply reorganization of the existing code? Or are you introducing new parsing here?

zelosleone: Its reorganization of the existing code

golem-worker-service/src/api/openapi/converter.rs
```diff
@@ -0,0 +1,1019 @@
...
+                    request_body: Self::create_request_body(route),
+                    responses: {
+                        let mut map = Self::create_responses(route, "*");
+                        // Add CORS headers to all responses
+                        for response in map.values_mut() {
+                            response.headers = Some(Self::create_cors_headers("*"));
+                        }
+                        map
+                    },
+                    security: None,
```
jdegoes, line 84: You don't know there is no security -- or are you deferring that to later?

zelosleone: Security configuration is handled dynamically by the runtime middleware, not statically within the OpenAPI specification.  This approach provides flexible and adaptable security management without requiring API specification modifications.

golem-worker-service/src/api/definition/binding.rs
jdegoes: Can you link to the code that is using `From`? I just wonder why any part of Golem that could handle `FileServer` couldn't also handle `SwaggerUI`.

golem-worker-service/src/api/definition/types.rs
```diff
@@ -0,0 +1,85 @@
...
+    pub description: String,
+    pub routes: Vec<Route>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Route {
+    pub path: String,
+    pub method: HttpMethod,
+    pub description: String,
+    pub template_name: String,
```
jdegoes, line 58: Older code survives here! This would be called `component_name` nowadays.

golem-worker-service/src/api/openapi/converter.rs
jdegoes: Yes, but we need to document the security in the OpenAPI specification. It's fine to produce `None` here but then in a later pass, you would have to fill it in based on what type of security the middleware is providing (it's driven by other details in the API definition).

zelosleone: Working on it at the moment.

zelosleone:
Pushed the commit according to latest reviews.

golem-worker-service/src/api/openapi/converter.rs
```diff
@@ -0,0 +1,1019 @@
...
+                    tags: Some(vec![route.template_name.clone()]),
+                }
+            },
+            BindingType::FileServer { .. } => {
+                Operation {
+                    summary: Some(route.description.clone()),
```
jdegoes, line 99: Overall, the file server needs to expand to an OpenAPI schema that one would expect to find for a file server (at the specified route, and using a multi-segment path variable, e.g. `{+foo}`, as per RFC). I haven't checked this (yet, still reviewing) but I see delegation to `create_request_body` here, which is also being used for the default routes defined by Rib scripts.

golem-worker-service/src/api/openapi/converter.rs
```diff
@@ -0,0 +1,1019 @@
...
+                            response.headers = Some(Self::create_cors_headers("*"));
+                        }
+                        map
+                    },
+                    security: None,
+                    tags: Some(vec![route.template_name.clone()]),
+                }
+            },
+            BindingType::SwaggerUI { .. } => {
+                Operation {
```
jdegoes, line 124: For SwaggerUI, I would expect it to be documented as any endpoint which hosts a SwaggerUI (probably similar to the file server, but where the files being served are constrained to SwaggerUI files).

golem-worker-service/src/api/openapi/converter.rs
```diff
@@ -0,0 +1,1019 @@
...
+                operation_id: Some(format!("{}_{}",
+                    route.template_name.to_snake_case(),
+                    route.method.to_string().to_lowercase())),
+                parameters: {
+                    let mut params = Self::extract_path_parameters(&route.path).unwrap_or_default();
+                    params.extend(Self::extract_query_parameters(route));
+                    params.extend(Self::extract_header_parameters(route));
+                    if params.is_empty() { None } else { Some(params) }
+                },
+                request_body: Self::create_request_body(route),
```
jdegoes, line 185: This is the "main" binding type whose path parameters, query parameters, request body, and response body, are all known and can be encoded precisely through conversion of Rib typing information.

golem-worker-service/src/api/openapi/converter.rs
zelosleone: I see what you mean about the FileServer needing a more detailed OpenAPI schema. Right now, it’s just delegating to create_request_body I see, which might not fully capture what a file server should do—especially around multi-segment paths like {+foo}, yeah, missing that is fatal you are right.

So, thinking about it, Maybe I should add a dedicated schema builder for the FileServer. That would let us:

- Properly handle file paths – for example, specifying the path parameter as /{+filePath} or something similar, so OpenAPI clients can see that the path can have multiple segments.
- Generate operations for file retrieval and possibly file upload – making sure we describe the content types (e.g. application/octet-stream for binary files, etc.).
- Show file data – specify that the response can be binary data and maybe define a multipart/form-data request body if we support uploads.

Do you think something like this would address the core need, or is there a different approach you’d prefer? Let me know, and I can get started on a commit that implements these changes.

golem-worker-service/src/api/openapi/converter.rs
zelosleone: Yes, that is correct. The BindingType::Default handles comprehensive parameter binding and type conversion. Specially:

- Extracts and converts path parameters from the route path
- Handles query parameters from the route definition
- Processes header parameters
- Converts request/response bodies using the Rib type system
- All parameter types are precisely encoded through the OpenAPI converter, maintaining type safety from Golem's internal types to the OpenAPI spec.

zelosleone:
Added more detailed documentation.

golem-worker-service/src/api/openapi/converter.rs
```diff
@@ -0,0 +1,1019 @@
...
+        // Extract parameters from path patterns
+        let params: Vec<Parameter> = path_pattern.path_patterns
+            .iter()
+            .filter_map(|pattern| match pattern {
+                PathPattern::Var(info) => Some(Parameter {
+                    name: info.key_name.clone(),
+                    r#in: ParameterLocation::Path,
+                    description: None, 
+                    required: Some(true),
+                    schema: if info.key_name.ends_with("_id") {
```
jdegoes, line 323: For path parameters, Rib has inferred types, which we should be using to generate the schema.

In general, we will need, in a separate module (so it's easy to test), a way to convert a Rib type into an OpenAPI schema. Then we will use that for the path parameters, query variables, and request / response headers & bodies.

golem-worker-service/src/api/openapi/converter.rs
```diff
@@ -0,0 +1,1019 @@
...
+        }
+    }
+
+    fn extract_query_parameters(route: &Route) -> Vec<Parameter> {
+         let mut params = Vec::new();
+
+        if route.path.contains("/workers") && route.method == HttpMethod::Get {
+            params.push(
+                Parameter {
+                    name: "filter".to_string(),
```
jdegoes, line 366: See comment on Rib type conversion required here.

golem-worker-service/src/api/openapi/converter.rs
```diff
@@ -0,0 +1,1019 @@
...
+                    r#in: ParameterLocation::Query,
+                    schema: Schema::Boolean,
+                    style: Some("form".to_string()),
+                    explode: Some(true),
+                    required: Some(false),
+                    description: None,
+                }
+            );
+        }
+        if route.path.contains("/invoke-and-await") || route.path.contains("/invoke") {
```
jdegoes, line 414: What is the point of this special-case logic here? Also, I am not sure what this is trying to do because in general, we should be generically converting the user-defined API into OpenAPI (not our own REST API that we expose for workers). Perhaps this is arising from the reuse of the same set of functions across all binding types--if so we should use different functions for different binding types because there's no way to unify them at this level.

golem-worker-service/src/api/openapi/converter.rs
```diff
@@ -0,0 +1,1019 @@
...
+                    r#in: ParameterLocation::Query,
+                     schema: Schema::String { format: None, enum_values: None },
+                    style: Some("form".to_string()),
+                    explode: Some(true),
+                    required: Some(false),
+                    description: None,
+                }
+            );
+        }
+        if route.path.contains("/api/definitions") && route.method == HttpMethod::Get {
```
jdegoes, line 514: Again, all this special-case logic doesn't begin here. We want to generate OpenAPI for user-defined APIs (which are defined with file server, swagger, and default / worker binding types), not for Golem REST APIs.

golem-worker-service/src/api/openapi/converter.rs
```diff
@@ -0,0 +1,1019 @@
...
+                }
+            );
+         }
+         params
+    }
+
+
+   fn extract_header_parameters(route: &Route) -> Vec<Parameter> {
+        let mut params = Vec::new();
+        if route.path.contains("/invoke-and-await") || route.path.contains("/invoke") {
```
jdegoes, line 592: I am convinced all this special-case logic around the Golem REST APIs is not correct and not satisfying the requirements of the ticket, which are to convert the user-defined API into OpenAPI schema.

golem-worker-service/src/api/openapi/converter.rs
```diff
@@ -0,0 +1,1019 @@
...
+                    explode: Some(true),
+                    required: Some(false),
+                    description: None,
+                }
+            );
+        }
+        params
+    }
+
+    fn create_request_body(route: &Route) -> Option<RequestBody> {
```
jdegoes, line 608: This implementation should be re-evaluated in light of the above comments.

Essentially, what you need to be doing here is looking at the inferred query, path, request, and response types (which is something you get from the Rib compiler metadata associated with worker binding type), and then generating OpenAPI for them.

That's only for worker binding type. For file server, it's going to be much simpler, and for swagger UI binding type, it will be even simpler. The most complex type is when a user has defined a route with Rib.

jdegoes:
@zelosleone I stopped reviewing for now because I discovered the core of the problem is not solved, which is: when the user builds a custom API, using Rib (worker binding), file server, and of course the new Swagger UI binding types for different routes, an OpenAPI Schema needs to be generated that precisely describes how to interact with this API (including CORS and Security, and having full information on request and response, including headers, bodies, and query and path variables). As far as I can tell, the essential logic to go from a Rib inferred type to some element of OpenAPI schema is not present, which means this solution does not satisfy the requirements. However, you overall have the right architecture and are looking in the right places, so I look forward to taking another look when you have a chance to address these issues. Please re-open when re-architected!

PR #1203, opened 2024-12-19
OpenAPI Export for API Definition (Issue1178)

uurl:
Tasks completed:

- Full and lossless conversion of any API Definition into a valid OpenAPI Specification.
- Tests at the unit and integration level which demonstrate the correctness of pieces and the whole.
- System tests that generate client libraries from the OpenAPI Spec, and interact to verify the client library works.
- Manual testing for complex scenarios.
-  A new binding type for routes called "Swagger UI", which will use serve Swagger UI at the specified route, which are generated during API Definition update (if necessary) and stored for subsequent serving of the endpoint.

/claim #1178


golem-worker-service/src/api/api_definition.rs
```diff
@@ -291,6 +317,134 @@ impl RegisterApiDefinitionApi {
...
+            // Generate OpenAPI Specification
+            let openapi_spec = generate_openapi(&http_api_definition, &version.0.to_string());
+
+            Ok(Json(openapi_spec))
+        };
+
+        record.result(response)
+    }
+
+    #[oai(path = "/swagger-ui/*path", method = "get")]
```
jdegoes, line 378: Can you explain this endpoint?

golem-worker-service/src/api/api_definition.rs
```diff
@@ -291,6 +317,134 @@ impl RegisterApiDefinitionApi {
...
+            RegisterApiDefinitionApi::new(services.definition_service.clone()),
+            ApiDeploymentApi::new(services.deployment_service.clone()),
+            SecuritySchemeApi::new(services.security_scheme_service.clone()),
+            HealthcheckApi,
+        ),
+        "API Service",
+        "1.0",
+    ));
+
+    Route::new()
```
jdegoes, line 401: I can only see the relative mounting point here. What's the absolute mounting point?

golem-worker-service/src/api/api_definition.rs
```diff
@@ -291,6 +317,134 @@ impl RegisterApiDefinitionApi {
...
+                middleware: None,
+            })
+        })
+        .collect::<Result<Vec<HttpRoute>, ApiEndpointError>>()?;
+    Ok(HttpApiDefinition { routes })
+}
+
+fn http_method_to_i32(method: &str) -> Result<i32, ApiEndpointError> {
+    match method.to_uppercase().as_str() {
+        "GET" => Ok(0),
```
jdegoes, line 436: This seems an error-prone encoding. Why `i32`?

golem-worker-service/src/api/openapi_generator.rs
```diff
@@ -0,0 +1,437 @@
...
+
+use golem_api_grpc::proto::golem::apidefinition::{HttpApiDefinition, HttpRoute};
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use std::collections::HashMap;
+use poem_openapi::Object;
+
+/// Represents an OpenAPI Specification document.
+#[derive(Serialize, Deserialize, Object)]
+pub struct OpenApiSpec {
```
jdegoes, line 23: It could be useful to use a third-party crate that already models OpenAPI 3.0 (I found two with a quick search).

golem-worker-service/src/api/openapi_generator.rs
```diff
@@ -0,0 +1,437 @@
...
+            4 => "PUT".to_string(),
+            5 => "PATCH".to_string(),
+            6 => "OPTIONS".to_string(),
+            7 => "TRACE".to_string(),
+            8 => "HEAD".to_string(),
+            _ => "UNKNOWN".to_string(),
+        },
+        summary: None,                         // Populate if available
+        description: None,
+        parameters: vec![], // Add conversion logic for parameters
```
jdegoes, line 193: There is no conversion logic here or elsewhere. No code that forms OpenAPI schemas based on the structure of requests/responses inside the different binding types (especially the default / worker binding type which defines requests and response in terms of Rib).

jdegoes:
@uurl This is missing any conversion logic and does not appear to be very close to a solution. I will close the pull request for now. Feel free to re-open once the essential problem has been solved.

PR #1209, opened 2024-12-24
Implement fork

afsalthaj:


vigoo:
I will do a proper review later, but some initial mismatch from the ticket:

- The forking logic should not be implemented in the gRPC request handlers. That should just be a thin layer as we want to have the actual forking logic used later in the debugging service for example. So as I wrote in the ticket, find a place (one of the "service layers" or create a new one) for it and from the gRPC request handler just call it.
- The ticket explicitly asks for opening the worker in suspended mode and use the `Oplog` instance of it to read the source oplog - and not the "global" one in `OplogService`. I am not sure exactly why did I ask for that but I suspect there was a reason :) By using `Oplog` it is properly synchronized in case the worker is running etc. So let's just do it in the way as the ticket asks.

afsalthaj:
I used get_or_create_suspended and used oplog from there. If you look at some previous commits you can see that to later change to this one - but I thought it's the same. Anyway I will revert.

vigoo:
> Probably validate the forking further using the details in WorkerStatusRecord such as if pending invocations or pending updates are present, probably not allow forking until finished?

I don't think this is necessary. 

>  resume would fail if the oplog corresponds to an entry which is tagged as “running” while calculating the latest status.

let's add a `force: bool` parameter to the resume gRPC call. I think it already does everything needed if that check would not happen in it (so when force is true, it must always do a `Worker::get_or_create_running` no matter what the status is)

> Fail forking if component was uploaded as an ephemeral one?

I don't think this is necessary. For example you could have a failed ephemeral worker that you load into the debugger, change something and fork it so it may run to completion.

>  Compute worker-status-record when copying the oplog with a bit more care (to be discussed with vigoo)

I suspect there is a general problem here of not using the buffered entries but I will check and think it through in details and get back to you.

golem-worker-executor-base/src/grpc.rs
```diff
@@ -141,6 +142,7 @@ pub struct WorkerExecutorImpl<
 > {
     /// Reference to all the initialized services
     services: Svcs,
+    worker_fork: Arc<dyn WorkerFork + Sync + Send>,
```
vigoo, line 145: The idea of `HasAll` is that it has all the services. So you should add the new one to those traits according to how it is done for the existing services, and then you don't need a new field here.

golem-worker-executor-base/src/grpc.rs
```diff
@@ -418,6 +426,75 @@ impl<Ctx: WorkerCtx, Svcs: HasAll<Ctx> + UsesAllDeps<Ctx = Ctx> + Send + Sync +
         Ok(())
     }
 
+    async fn get_worker_ids(
```
vigoo, line 429: A couple of problems with this:

- It's an internal step of the fork logic, but it has a very generic name (`get_worker_ids`) in a non-fork specific place
- The validations (whether the target worker ID exists, and the account IDs match) should be in the fork implementation (in the service layer) not here in the request handler
- In the gRPC layer the only thing that should happen is getting values out of the protobuf structures and fail if something is missing

So I would move the validations to the fork service, and then just move back the remaining few lines into the `fork_internal` and delete this function.

golem-worker-executor-base/src/services/mod.rs
```diff
@@ -43,6 +43,9 @@ pub mod worker_activator;
 pub mod worker_enumeration;
 pub mod worker_event;
 pub mod worker_proxy;
+
+pub mod worker_fork;
```
vigoo, line 47: alphabetic order please :'(

golem-worker-executor-base/src/services/worker_fork.rs
```diff
@@ -0,0 +1,178 @@
...
+    ) -> Result<(OwnedWorkerId, OwnedWorkerId, WorkerMetadata), GolemError> {
+        let second_index = OplogIndex::INITIAL.next();
+
+        if oplog_index_cut_off < second_index {
+            return Err(GolemError::invalid_request(
+                "oplog_index_cut_off must be at least 2",
+            ));
+        }
+
+        // We assume the target worker is also owned by the same account
```
vigoo, line 57: I see this validation that I asked for in a comment above is already here. So I guess it just needs to be deleted from the other place.

golem-worker-executor-base/src/services/worker_fork.rs
```diff
@@ -0,0 +1,178 @@
...
+        let owned_source_worker_id = OwnedWorkerId::new(account_id, source_worker_id);
+
+        let metadata = self
+            .all
+            .worker_service()
+            .get(&owned_source_worker_id)
+            .await
+            .ok_or(GolemError::worker_not_found(source_worker_id.clone()))?;
+
+        Ok((owned_source_worker_id, owned_target_worker_id, metadata))
```
vigoo, line 79: I think there is a missing `calculate_latest_worker_status` here. The result of `get` is not the latest status just the latest cached one.

(You can say this is a bad API and I agree but let's not change it now :)

golem-worker-executor-base/src/services/worker_fork.rs
```diff
@@ -0,0 +1,178 @@
...
+
+        let new_oplog = self
+            .all
+            .oplog_service()
+            .create(
+                &owned_target_worker_id,
+                target_initial_oplog_entry,
+                target_worker_metadata,
+                Arc::new(RwLock::new(ExecutionStatus::Suspended {
+                    last_known_status: WorkerStatusRecord::default(), // default is idle, TODO: check if we need to update this or derive from the source last known status
```
vigoo, line 147: I think this does not matter because we are going to drop this value here and the "resume" will recreate it (potentially on another executor)

golem-worker-executor-base/src/services/worker_fork.rs
```diff
@@ -0,0 +1,178 @@
...
+                    component_type: ComponentType::Durable, // Probably forking should fail if component type is ephemeral, or not?
```
vigoo, line 148: I don't think it should fail, but here you should use the real component type just to avoid surprises :) It can be read from the component metadata in `source_worker_instance`

golem-worker-executor-base/src/services/worker_fork.rs
```diff
@@ -0,0 +1,178 @@
...
+                    timestamp: Timestamp::now_utc(),
+                })),
+            )
+            .await;
+
+        let second_index = u64::from(OplogIndex::INITIAL.next());
+        let cut_off_index = u64::from(oplog_index_cut_off);
+
+        for index in second_index..=cut_off_index {
```
vigoo, line 157: I saw you created some oplog index range iterator, but not using it? Either use it or delete it or am I missing something?

integration-tests/tests/worker.rs
```diff
@@ -1326,6 +1326,304 @@ async fn search_oplog_1(deps: &EnvBasedTestDependencies, _tracing: &Tracing) {
     assert_eq!(result3.len(), 2); // two invocations
 }
 
+#[test]
+#[tracing::instrument]
+#[timeout(120000)]
+async fn fork_worker_1(deps: &EnvBasedTestDependencies, _tracing: &Tracing) {
```
vigoo, line 1332: As there are many tests I would put them in a separate `fork.rs` (and make sure in the Makefile.toml it is listed in the integration-tests task)

golem-worker-executor-base/src/grpc.rs
afsalthaj: @vigoo  I tried this (a few days ago), but if I remember correctly, the worker_fork implementation pretty much needs of all what's in HasAll.  So I was a bit confused in this area. I think worker_fork has the same type of behavior as `DefaultWorkerActivator` in terms of dependency requirements.

golem-worker-executor-base/src/services/worker_fork.rs
afsalthaj: Sure.

vigoo:
About `calculate_last_worker_status` not using the "uncommitted" changes. After checking the code I now remember that this is actually how it is supposed to be, and the solution is to not use it in this case.

Because we already load a `Worker` instance in fork, you can call `Worker::get_metadata` on that instance to get the latest metadata. This works currectly even for running workers.

golem-worker-executor-base/src/grpc.rs
afsalthaj: > The validations (whether the target worker ID exists, and the account IDs match) should be in the fork implementation (in the service layer) not here in the request handler

@vigoo  Uh! That validation was a redundant check which I forgot to delete when refactoring. The validation already exist in service layer under the function `validate_worker_forking`

golem-worker-executor-base/src/services/worker_fork.rs
afsalthaj: That was reusing an existing range functionality in `OplogService` somehow, which essentially doesn't require `get_or_suspended` if used

```
 async fn read_range(
        &self,
        owned_worker_id: &OwnedWorkerId,
        start_idx: OplogIndex,
        last_idx: OplogIndex,
    ) -> BTreeMap<OplogIndex, OplogEntry> {
```

afsalthaj: Ok, I forgot what I did 1.5 weeks back, I see what you mean. I will use the iterator I implemented :)

golem-worker-executor-base/src/services/worker_fork.rs
afsalthaj: I didn't get this part. During forking why would we need the latest worker status of the source worker?

afsalthaj: Is it to fail the fork if the status is say "exited" or "failed" or "interrupted"?

vigoo: Because you are copying it as the target worker metadata, so I was thinking it is best to copy the latest version and not some random old one :) (Maybe there is no real difference, but easier if you don't have to think it through).

Btw notice later I realized that instead of calculate_latest_worker_status you can get the latest worker status through the worker instance which is better.

afsalthaj:
As seen in the commits, dealing with flaky tests in CI (although it works all the time in my machine)

golem-worker-executor-base/src/grpc.rs
vigoo: I think it must be similar to `Rpc` (it does both local and remote worker invocations).

This change is a must to keep consistency of how worker executor is built.

golem-worker-executor-base/tests/lib.rs
```diff
@@ -92,8 +91,6 @@ tag_suite!(guest_languages3, group7);
 tag_suite!(ts_rpc2, group8);
 tag_suite!(ts_rpc2_stubless, group8);
 
-tag_suite!(fork, group9);
```
vigoo: Let's not introduce a new group for this. The reason for these groups is only to run them in parallel on CI and they are grouped in a way to optimize the CI run time. You can add the new suite to `group3` for example.

vigoo: Sorry I see this was removed later :) Please tag the new suite as `group3`.

wasm-rpc-stubgen/tests-integration/tests/compose.rs
```diff
@@ -31,6 +31,8 @@ use wasm_rpc_stubgen_tests_integration::{test_data_path, wasm_rpc_override};
 
 test_r::enable!();
 
+// TODO: disabled until getting newer glibc on CI
+#[ignore]
```
vigoo, line 35: Don't forget to remove this

golem-worker-executor-base/src/grpc.rs
```diff
@@ -418,6 +419,58 @@ impl<Ctx: WorkerCtx, Svcs: HasAll<Ctx> + UsesAllDeps<Ctx = Ctx> + Send + Sync +
...
+        request: ForkWorkerRequest,
+    ) -> Result<ForkWorkerResponse, GolemError> {
+        let account_id_proto = request
+            .account_id
+            .clone()
+            .ok_or(GolemError::invalid_request("account_id not found"))?;
+
+        let account_id = account_id_proto.into();
+
+        // Check if target worker id exists already and if so fail
```
vigoo, line 433: This comment is not true anymore

golem-worker-executor-base/src/services/worker_fork.rs
```diff
@@ -0,0 +1,472 @@
...
+        let target_worker_id = owned_target_worker_id.worker_id.clone();
+        let account_id = owned_target_worker_id.account_id.clone();
+
+        let source_worker_instance =
+            Worker::get_or_create_suspended(self, &owned_source_worker_id, None, None, None, None)
+                .await?;
+
+        let source_worker_metadata = source_worker_instance.get_metadata().await?;
+
+        // Not sure if we should copy the metadata or not, or stick on to just default
```
vigoo, line 412: Let's remove this comment :)

golem-worker-executor-base/tests/lib.rs
vigoo: Or there are no worker executor tests at all. Sorry got confused by trying to review commit by commit

vigoo (approved):
Added a few more comments (just cleaning up stuff), after that good to merge

golem-worker-executor-base/tests/lib.rs
afsalthaj: Yes, I had to go through a bit of round trip there, just to realize, what I really wanted is the way it is now

golem-worker-executor-base/src/services/worker_fork.rs
afsalthaj: Added that comment myself, before understanding the entire code

PR #1220, opened 2025-01-07
Introduced new static-wasm-rpc dependency type and made wasm-rpc the stubless one

vigoo:
Resolves https://github.com/golemcloud/wasm-rpc/issues/106

### DIFF
--- a/wasm-rpc-stubgen/src/commands/app.rs
+++ b/wasm-rpc-stubgen/src/commands/app.rs
@@ -7,7 +7,7 @@ use crate::log::{
 };
 use crate::model::app::{
     includes_from_yaml_file, AppBuildStep, Application, ComponentName,
-    ComponentPropertiesExtensions, ProfileName, DEFAULT_CONFIG_FILE_NAME,
+    ComponentPropertiesExtensions, DependencyType, ProfileName, DEFAULT_CONFIG_FILE_NAME,
 };
 use crate::model::app_raw;
 use crate::stub::{StubConfig, StubDefinition};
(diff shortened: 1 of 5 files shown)


wasm-rpc-stubgen/src/model/app.rs
```diff
@@ -272,18 +299,27 @@ impl<CPE: ComponentPropertiesExtensions> Application<CPE> {
                 validation.push_context("component", component_name.to_string());
 
                 for dependency in component_dependencies {
-                    if dependency.type_ == "wasm-rpc" {
+                    if dependency.type_ == "static-wasm-rpc" || dependency.type_ == "wasm-rpc" {
```
noise64, line 302: now that we have multiple values maybe we could add these as consts to app_raw (or we could make it an enum there, but this way error reporting is more lenient)

wasm-rpc-stubgen/src/commands/app.rs
```diff
@@ -394,8 +394,11 @@ async fn gen_rpc<CPE: ComponentPropertiesExtensions>(
         for component_name in ctx.wit.component_order_cloned() {
             create_generated_base_wit(ctx, &component_name)?;
         }
-        for component_name in &ctx.application.all_wasm_rpc_dependencies() {
-            build_stub(ctx, component_name).await?;
+
+        for dep in &ctx.application.all_wasm_rpc_dependencies() {
```
noise64, line 398: maybe we should add a getter named "all_static_wasm_rpc_dependencies" (or we might want to report skipping dynamic ones, which could be useful for debugging, although it would make logs more versbose)

wasm-rpc-stubgen/src/commands/app.rs
```diff
@@ -614,18 +620,17 @@ pub fn clean<CPE: ComponentPropertiesExtensions>(config: Config<CPE>) -> anyhow:
         log_action("Cleaning", "component stubs");
         let _indent = LogIndent::new();
 
-        for component_name in app.all_wasm_rpc_dependencies() {
+        for dep in app.all_wasm_rpc_dependencies() {
```
noise64, line 623: i think we also should filter for static ones

wasm-rpc-stubgen/src/model/app.rs
vigoo: ok

wasm-rpc-stubgen/src/commands/app.rs
noise64: 👌

noise64 (approved):

mschuwalow (approved):

noise64 (approved):

PR #1236, opened 2025-01-13
Defined the DurabilityHost interface and migrated Durability to it

### LINKED ISSUE #1223: Extract a clean set of durability host functions
After #1222, by a further refactoring step we separate the `Durability` implementation to a set of host functions (with access to `DurableWorkerCtx`) and a set of abstractions built on top of them. At the end of this step we have three layers of functionalities (still everything in `golem-worker-executor-base`:

- the core set of durability-related functions using `DurableWorkerCtx`
- the higher level abstractions built on them, used in the host functions
- the host function implementations using these higher level abstractions _and_ calling underlying wasmtime host functions

vigoo:
Resolves #1223

The primary change is introducing https://github.com/golemcloud/golem/pull/1236/files#diff-bcdc1292ac93ac52e1b1b02d1320cae842ccfa169d1b62a97b5aba1abdc44a4eR49 which is going to be moved to Golem's WIT definitions in the next step.

### DIFF
--- a/golem-worker-executor-base/src/durable_host/durability.rs
+++ b/golem-worker-executor-base/src/durable_host/durability.rs
@@ -18,62 +18,262 @@ use crate::metrics::wasm::record_host_function_call;
 use crate::model::PersistenceLevel;
 use crate::services::oplog::{CommitLevel, OplogOps};
 use crate::workerctx::WorkerCtx;
+use async_trait::async_trait;
 use bincode::{Decode, Encode};
 use bytes::Bytes;
-use golem_common::model::oplog::{OplogEntry, OplogIndex, WrappedFunctionType};
-use golem_common::serialization::try_deserialize;
+use golem_common::model::oplog::{DurableFunctionType, OplogEntry, OplogIndex};
+use golem_common::model::Timestamp;
+use golem_common::serialization::{serialize, try_deserialize};
+use golem_wasm_rpc::{IntoValue, IntoValueAndType, ValueAndType};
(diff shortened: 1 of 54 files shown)


golem-worker-executor-base/src/durable_host/durability.rs
```diff
@@ -18,62 +18,262 @@ use crate::metrics::wasm::record_host_function_call;
...
+    pub is_live: bool,
+    pub persistence_level: PersistenceLevel,
+}
+
+#[derive(Debug)]
+#[allow(dead_code)]
+pub struct PersistedDurableFunctionInvocation {
+    timestamp: Timestamp,
+    function_name: String,
+    response: Vec<u8>,
```
mschuwalow, line 43: It might make sense to not load the response by default but instead allow streaming them in when needed. I believe we can have quite easily have a situation where response_size > host_memory, especially with streaming responses.

mschuwalow (approved):

golem-worker-executor-base/src/durable_host/durability.rs
vigoo: It is not in this level - when persisting a streamed wasi-http response (which I think you have in mind) each chunk read is persisted as a separate oplog entry.
This type is only used in host functions made durable, so basically what they do with it is that they read from the oplog what they otherwise would return when not in replay mode.

It is of course interesting to think of what happens when WASI (and WIT etc) itself will support streams - but to persist functions returning streams will require many changes / improvements anyway.

PR #1240, opened 2025-01-14
Composable durability

### LINKED ISSUE #1225: Move the durable WASI wrappers into a separate WASM crate
After this step we move our custom WASI implementation into a separate subcrate which itself compiles to WASM, and both imports and exports all WASI interfaces, as well as the host functions defined in #1224.

This subcrate must be compiled to WASM and the resulting WASM must be embed into the component service, and it must compose every component with it automatically similar to plugins.

A new flag must be introduced so we know if an old component was not composed with durability yet - and then we can apply the transformation first time it gets downloaded as a migration strategy.

vigoo:
Resolves #1225


vigoo:
We decided to not move durability to the composable `durable-wasi` module, and just keep the rest of the changes of this PR, moved to https://github.com/golemcloud/golem/pull/1278

PR #1242, opened 2025-01-17
Resume replay for debugging service

afsalthaj:


golem-worker-executor-base/src/worker.rs
```diff
@@ -452,6 +452,24 @@ impl<Ctx: WorkerCtx> Worker<Ctx> {
...
+        match &*self.instance.lock().await {
+            WorkerInstance::Running(running) => {
+                running
+                    .sender
+                    .send(WorkerCommand::ResumeReplay)
+                    .map_err(|_| GolemError::unknown("Failed to resume command"))
+                //Probably just `expect(..)` would be better
+            }
+            WorkerInstance::Unloaded | WorkerInstance::WaitingForPermit(_) => {
+                debug!("Worker is initializing, persisting pending invocation");
```
vigoo, line 465: Wrong log line

golem-worker-executor-base/src/worker.rs
```diff
@@ -1485,6 +1503,13 @@ impl RunningWorker {
                 while let Some(cmd) = receiver.recv().await {
                     waiting_for_command.store(false, Ordering::Release);
                     match cmd {
+                        WorkerCommand::ResumeReplay => {
+                            let mut store = store.lock().await;
+
+                            Ctx::resume_replay(&mut *store, &instance)
+                                .await
+                                .expect("resume_replay failed");
```
vigoo, line 1511: We cannot panic here - this is a completely valid outcome (when trying to recovering a worker that failed in the past and will fail again).

The error of `resume_replay` must be handled exactly the same way as it is handled when it is coming from `prepare_instance` at the beginning of the invocation loop.

golem-worker-executor-base/src/workerctx.rs
```diff
@@ -366,6 +366,11 @@ pub trait ExternalOperations<Ctx: WorkerCtx> {
         metadata: &Option<WorkerMetadata>,
     ) -> Result<WorkerStatusRecord, GolemError>;
 
+    async fn resume_replay(
```
vigoo, line 369: Doc comment please - everything else in this module has.

afsalthaj:
@vigoo  I think I did address all the review comments.

vigoo (approved):

PR #1249, opened 2025-01-18
Support invoking wasi:http/incoming-handler

### LINKED ISSUE #1184: Expose a new invoke function via gRPC from Worker Executor for incoming HTTP requests
This is the first step in supporting workers implementing the wasi-http `incoming-handler` interface (https://wa.dev/wasi:http#incoming-handler).

The primary challenge is that this interface takes an `incoming-request` handle to a host resource, and returns with a `outgoing-response` handle to a host resource, which is not something we can directly invoke using Golem's existing invokation API.

Instead, as part of this ticket, we define:

- A gRPC definition of HTTP request->response handler, using gRPC streaming for the request and response body
- A special invocation mode in worker executor, wrapping the existing one (on the level of `invocation.rs`) that does the following:
  - Take the request information (coming via gRPC) and set up a host `incoming-resource` instance 
  - Invoke the worker's exported `wasi:http/incoming-handler:{handle}` function with this resource
  - Take the returned `outgoing-response` resource and convert it back to the protobuf response structure
- This new invocation mode needs to be implemented on all layers (`Worker`, including the invocation loop) and wired to the gRPC handler
- To make the scope of the task smaller, it is enough for this ticket to NOT do actual request/response body streaming
- Likely requires changes to the oplog too (to be decided by the implementor), some options:
  - Maybe it is possible to not introduce new invocation start/end entries, but instead use the existing ones, converting the request/response data to `Value`. Or maybe it is easier to have dedicated new ones. Think of the current `ExportedFunctionInvoked`, `ExportedFunctionCompleted` and `PendingWorkerInvocation` entries.
- Support this new call in the test DSL
- Write worker executor test(s) proving this works

With this ticket done we can call worker executor using a representation of a "http request" and get back a representation of a "http response", while the worker itself exports only the wasi-http incoming handler interface. 

Follow-up tickets add streaming support and wires this functionality into the worker bridge.

mschuwalow:
resolves #1184


golem-common/src/model/exports.rs
```diff
@@ -16,24 +16,24 @@ use golem_wasm_ast::analysis::{AnalysedExport, AnalysedFunction, AnalysedInstanc
 
 use rib::{ParsedFunctionName, ParsedFunctionReference, ParsedFunctionSite};
 
-pub trait AnalysedExportExtensions {
```
mschuwalow, line 19: These were unused

afsalthaj: 👍

golem-worker-executor-base/src/virtual_export_compat/http_incoming_handler.rs
```diff
@@ -0,0 +1,125 @@
...
+use crate::error::GolemError;
+use bytes::Bytes;
+use golem_common::virtual_exports::http_incoming_handler::*;
+use golem_wasm_rpc::Value;
+use http::{HeaderName, HeaderValue};
+use http_body_util::combinators::BoxBody;
+use http_body_util::BodyExt;
+use wasmtime_wasi_http::bindings::http::types::ErrorCode;
+
+pub fn input_to_hyper_request(
```
afsalthaj, line 26: Not that it significantly matters, yet if the module name was adapter instead of a handler, I will automatically think of these functions to exist - both on input as well as output. When it is `http_incoming_handler`, it confuses between which "input" and what exactly is being handled.

afsalthaj, line 26: Not that it significantly matters, yet I feel like the module name should be  `*adapter`  (that converts to hyper request (required by wasihttpviewhandler::new_incoming_request) instead of a `handler`.

golem-worker-executor-base/src/invocation.rs
```diff
@@ -407,6 +425,118 @@ async fn invoke<Ctx: WorkerCtx>(
...
+            .new_incoming_request(scheme, hyper_request)
+            .unwrap();
+        let outgoing = store_context
+            .data_mut()
+            .durable_ctx_mut()
+            .as_wasi_http_view()
+            .new_response_outparam(sender)
+            .unwrap();
+
+        unsafe {
```
afsalthaj, line 471: I couldn't figure out from reading it the need of unsafe here. Would you mind explaining it here?

mschuwalow: this comes from the async_scoped crate. https://docs.rs/async-scoped/latest/async_scoped/

The invariant they want here is just that:
```
An asynchronous function that creates a scope and immediately awaits the stream. The outputs of the futures are collected as a Vec and returned along with the output of the block.
Safety

This function is not completely safe: please see cancellation_soundness in [tests.rs](https://github.com/rmanoka/async-scoped/blob/master/src/tests.rs) for a test-case that suggests how this can lead to invalid memory access if not dealt with care.

The caller must ensure that the lifetime ’a is valid until the returned future is fully driven. Dropping the future is okay, but blocks the current thread until all spawned futures complete.
```

golem-worker-executor-base/src/virtual_export_compat/http_incoming_handler.rs
mschuwalow: I tried to express this using the compat part of the module naming.

I.e "this module is for compatibility with the http_incoming_handler virtual export"

afsalthaj: 👍

golem-common/src/virtual_exports/http_incoming_handler.rs
```diff
@@ -0,0 +1,524 @@
...
+use semver::Version;
+
+// The following wit is modelled here:
+//
+// type fields = list<tuple<string, list<u8>>>;
+// type body = list<u8>;
+//
+// record body-and-trailers {
+//   body: body,
+//   trailers: option<simple-fields>
```
vigoo, line 29:
```suggestion
//   trailers: option<fields>
```

golem-common/src/virtual_exports/http_incoming_handler.rs
```diff
@@ -0,0 +1,524 @@
...
+
+        for lv in list_values {
+            let byte_value = extract!(lv, Value::U8(inner), *inner, "not a byte")?;
+            result.push(byte_value);
+        }
+
+        Ok(HttpBodyContent(Bytes::from(result)))
+    }
+
+    pub fn to_value(self) -> Value {
```
vigoo, line 357: Instead of these `to_value` calls consider implementing the `IntoValue` type class (from wasm-rpc), that also requires you to specify the type, but it will allow us to use these in more scenarios (for example better integrating it `PublicOplogEntry` etc)

golem-worker-executor-base/src/invocation.rs
```diff
@@ -407,6 +425,123 @@ async fn invoke<Ctx: WorkerCtx>(
...
+                Ok(r) => r.expect_err("if the receiver has an error, the task must have failed"),
+                Err(_e) => anyhow!("failed joining wasm task"),
+            };
+            Err(e)?
+        }
+    };
+
+    let mut store_context = store.as_context_mut();
+
+    let current_fuel_level = store_context.get_fuel().unwrap_or(0);
```
vigoo, line 512: These  post-invocation steps are now duplicated, I guess, maybe worth extracting them to a function

golem-worker-executor-base/src/workerctx.rs
```diff
@@ -73,6 +74,7 @@ pub trait WorkerCtx:
     + Send
     + Sync
     + Sized
+    + DurableWorkerCtxView<Self>
```
vigoo, line 77: This should not be necessary - and breaks the (now unused) purpose of this trait that it is independent of `DurableWorkerCtx` (that's just one implementation - originally we had a non-durable one too).

We already have `as_wasi_http_view` and `as_wasi_view` below, isn't that enough?

golem-worker-executor-base/tests/wasi.rs
```diff
@@ -1849,3 +1849,47 @@ async fn ip_address_resolve(
     check!(result1.len() > 0);
     check!(result2.len() > 0);
 }
+
+#[test]
+#[tracing::instrument]
+async fn wasi_incoming_request_handler_compat(
```
vigoo, line 1855: better test coverage please! :) (headers, body, etc)

golem-worker-executor-base/src/workerctx.rs
afsalthaj: Also, `Ctx` decides whether worker context handles things related to durability.  So this implies the above comment from @vigoo should be made work if it's not working now, in my view - which  is simply `.as_wasi_view` where ever it is required.

golem-worker-executor-base/tests/wasi.rs
mschuwalow: Yes, this is coming :+1:

golem-worker-executor-base/src/workerctx.rs
mschuwalow: adressed

golem-common/src/virtual_exports/http_incoming_handler.rs
mschuwalow: Annoyance here is that only one half of the types here need to go to values (the other half is being converted from values), while we need to have the type for all of them.

This would suggest that we need a counterpart of IntoValue that can go from values (FromValue?) and then we would likely want to split the type portions out from both.

I wanted to avoid doing this for now and just implemented the needed functionality as plain functions. We can always implement IntoValue if we end up needing it for a specific type.

golem-worker-executor-base/tests/wasi.rs
mschuwalow: done

golem-worker-executor-base/src/invocation.rs
```diff
@@ -485,9 +485,18 @@ async fn invoke_http_handler<Ctx: WorkerCtx>(
         }
     };
 
-    let res_or_error = match receiver.await {
+    let out = receiver.await;
+
+    tracing::warn!("received response {:?}", out);
```
vigoo, line 490: Let's remove these debug warn!s (there is another one below too)

vigoo (approved):
Found two debug logs that should be removed, otherwise good to go!

golem-worker-executor-base/src/invocation.rs
mschuwalow: oh yes, sorry

PR #1253, opened 2025-01-22
Add http-handler binding type

### LINKED ISSUE #1186: WASI HTTP binding type for Worker Gateway
After completing #1184 and #1185, add support for the new invocation type as a **new binding type** in the worker gateway.

This ticket covers the following features

- Execution of the new binding type, forwarding the HTTP request as-is, with streaming request/response bodies, to the previously added new gRPC endpoint of worker executor
- Ability to define endpoints using this new binding type in the API Definition format
- Ability to define endpoints using this new binding type when using OpenAPI import
- Have an integration test showing that it work

mschuwalow:
resolves #1186


afsalthaj (approved):

golem-cli/tests/api_deployment_http_handler.rs
```diff
@@ -0,0 +1,132 @@
...
+        draft: true
+        routes:
+        - method: Post
+          path: "/test"
+          binding:
+            bindingType: http-handler
+            componentId:
+              componentId: '{component_id}'
+              version: 0
+            workerName: 'let name: string = "{worker_name}"; name'
```
afsalthaj, line 129: I believe there is nothing you could do with the response in the response mapping if bindingType is http-handler? Or can we ?

mschuwalow: Yeah, I response mapping is currently not supported in this pr.
Will bring it up during tomorrows daily :+1:

afsalthaj: So, it is confirmed this behaviour is correct.

PR #1278, opened 2025-01-29
Final durability related refactorings

vigoo:
The useful parts of https://github.com/golemcloud/golem/pull/1240 that we decided to not finish:

- Test dsl improvements (store component builder api)
- Changed how wit dependencies are fetched - this allows a single `wit-deps` based fetch and then creating separate subsets for various purposes. In the original PR this was required to create the `wit/deps` directory of `durable-wasi`, but even without that I think it is better than the previous, custom `build.rs` based solution.
- Some changes in the feature-controlled parts of `golem-common` that were used to minimize the dependencies used by `durable-wasi`, but it makes sense to have them anyway

Resolves https://github.com/golemcloud/golem/issues/1225


Makefile.toml
```diff
@@ -46,13 +47,35 @@ alias = "dev-flow"
...
+# WIT DEPENDENCIES
+[tasks.wit]
+description = "Fetches the WIT dependencies based on wit/deps.toml"
+dependencies = ["wit-host", "wit-wasm-rpc"]
+
+[tasks.wit-wasm-rpc]
+dependencies = ["wit-host"]
+script_runner = "@duckscript"
+script = """
+rm -r wasm-rpc/wit/deps
```
noise64, line 66: do we really want to delete and re-get the deps on basically every command? (i know that wit-deps uses a cache, and probably not a huge performance problem, still feels wasteful)

vigoo: This is not "regetting" them just copying them from the project root's `wit/deps`. We can try to make a better script for it but
- it needs to be working on all platforms - that's why i used this 🦆 thing
- it needs to remove things that are no longer there - i already ran into that multiple times as i was trying to figure out the proper name for the new wit package

noise64: by "regetting" i meant wit-host also runs on everything now, and yes, i understand that it would need a "task checker", will think about it (ironically golem-cli app could do it :D)

vigoo: The `wit-dep update` is designed to be runnable every time. That's why it keeps a lock file.

vigoo: (Based on https://github.com/bytecodealliance/wit-deps/issues/202 being closed)

noise64 (approved):

PR #1291, opened 2025-02-04
Worker revert feature and executor refactorings, WIP

### LINKED ISSUE #1180: Implement the undo API in worker executor
The "undo" feature is a new worker executor function exposed through it's gRPC API, having two possible modes:

- undo the worker to a specific oplog index
- convenience feature: undo the last N invocations

The "last N invocations" can be implemented on top of the "undo to the given oplog index" feature by reading back from the end of the oplog looking for the nth `ExportedFunctionInvoked` entry.

On the execution level, the undo feature will take advantage of the already existing "deleted regions" support. Similarly how the `Jump` entry is implemented, when we undo something in the worker, that will mark the "undone" region as deleted. 

In details, implementation of this ticket should consist of the following steps:

- Define the new gRPC API
- Add a new oplog entry `Undo` (it should to be different from `Jump` both to be able to differentiate them, and also because we may have to treat them differently when calculating some part of the worker status)
- In the implementation, use `get_or_create_suspended` to get the `Worker` and check its status. Undo should only work on `Idle` or `Failed` or `Exited` workers - we don't allow undo when it is executing something (running/suspended/interrupted). 
- If the request specifies "last n invocation"s, use the worker's `Oplog` to translate that into a specific oplog index
- Write the oplog entry and initiate a restart (by doing a special interrupt - I think the InterruptKind::Jump can be reused here)
- Modify `calculate_last_known_status` to take the new oplog entry into account when calculating the deleted regions, and also when calculating `WorkerStatus` (as we have to "get back" . NOTE: I think the proper implementation is to change it that it first calculates the final deleted regions, and then in all the other "folds" it takes the deleted regions into account - right now this is not the case and it can be considered a bug.
- Add the new undo API into the test DSL
- Write worker executor tests that tries out this feature (both versions)

This is not split further into smaller steps, as they only make sense and only testable together.

### LINKED ISSUE #1181: Expose the worker undo feature via REST and CLI
After #1180, expose the new functionality in the following places:

- Worker service REST API
- Worker service gRPC API
- CLI

Write an integration test and a CLI test (can be very similar to the worker executor test created in #1180) to prove it works.

vigoo:
Resolves #1180 
Resolves #1181 

WIP


golem-worker-executor-base/src/durable_host/mod.rs
```diff
@@ -312,19 +312,22 @@ impl<Ctx: WorkerCtx> DurableWorkerCtx<Ctx> {
...
-            .clone();
+            .clone()
+    }
 
-        let mut deleted_regions = self.state.replay_state.deleted_regions().await;
+    pub async fn update_worker_status(&self, f: impl FnOnce(&mut WorkerStatusRecord)) {
+        let mut status = self.get_worker_status_record();
+
+        let mut skipped_regions = self.state.replay_state.skipped_regions().await;
         let (pending_updates, extra_deleted_regions) = self.public_state.worker().pending_updates();
```
mschuwalow, line 327:
```suggestion
        let (pending_updates, extra_skipped_regions) = self.public_state.worker().pending_updates();
```
If I understand the new naming correctly

golem-worker-executor-base/src/durable_host/mod.rs
```diff
@@ -1096,7 +1109,10 @@ impl<Ctx: WorkerCtx> UpdateManagement for DurableWorkerCtx<Ctx> {
                 timestamp,
                 target_version,
                 details: details.clone(),
-            })
+            });
+            if status.skipped_regions.is_overridden() {
```
mschuwalow, line 1113: Can we have a comment explaining what this does, it's very not clear to me.

golem-worker-executor-base/src/durable_host/replay_state.rs
```diff
@@ -41,7 +41,7 @@ pub struct ReplayState {
 
 #[derive(Clone)]
 struct InternalReplayState {
-    pub deleted_regions: DeletedRegions,
+    pub skipped_regions: DeletedRegions,
     pub next_deleted_region: Option<OplogRegion>,
```
mschuwalow, line 45:
```suggestion
    pub next_skipped_region: Option<OplogRegion>,
```

golem-worker-executor-base/src/durable_host/replay_state.rs
```diff
@@ -85,19 +85,19 @@ impl ReplayState {
...
     }
 
-    pub async fn deleted_regions(&self) -> DeletedRegions {
+    pub async fn skipped_regions(&self) -> DeletedRegions {
         let internal = self.internal.read().await;
-        internal.deleted_regions.clone()
+        internal.skipped_regions.clone()
     }
 
     pub async fn add_deleted_region(&mut self, region: OplogRegion) {
```
mschuwalow, line 93:
```suggestion
    pub async fn add_skipped_region(&mut self, region: OplogRegion) {
```

golem-worker-executor-base/src/durable_host/replay_state.rs
```diff
@@ -85,19 +85,19 @@ impl ReplayState {
...
         let mut internal = self.internal.write().await;
-        internal.deleted_regions.add(region);
+        internal.skipped_regions.add(region);
     }
 
     pub async fn is_in_deleted_region(&self, oplog_index: OplogIndex) -> bool {
```
mschuwalow, line 98:
```suggestion
    pub async fn is_in_skipped_region(&self, oplog_index: OplogIndex) -> bool {
```

golem-worker-executor-base/src/worker/mod.rs
```diff
@@ -873,6 +899,105 @@ impl<Ctx: WorkerCtx> Worker<Ctx> {
...
+                        "Could not find {} invocations to revert",
+                        target.number_of_invocations
+                    )))
+                }
+            }
+        }
+    }
+
+    /// Starting from the end of the oplog, find the Nth ExportedFunctionInvoked entry's index.
+    async fn find_nth_invocation_from_end(&self, n: usize) -> Option<OplogIndex> {
```
mschuwalow, line 931: How will should this behave if we have updates / memory increases in the oplog region that is getting deleted?
We might want to delete region belonging to invocations, but keep other regions that belong to actions that users performed.

golem-worker-executor-base/src/grpc.rs
```diff
@@ -317,39 +316,28 @@ impl<Ctx: WorkerCtx, Svcs: HasAll<Ctx> + UsesAllDeps<Ctx = Ctx> + Send + Sync +
...
-        let promise_id: common_model::PromiseId =
-            promise_id.try_into().map_err(GolemError::invalid_request)?;
         let completed = self.promise_service().complete(promise_id, data).await?;
 
-        let metadata = self
-            .worker_service()
-            .get(&owned_worker_id)
-            .await
-            .ok_or(GolemError::worker_not_found(worker_id.clone()))?;
+        let metadata = Worker::<Ctx>::get_latest_metadata(&self.services, &owned_worker_id)
```
mschuwalow, line 336: wohoo, nice cleanup :tada:

golem-common/src/model/mod.rs
```diff
@@ -586,23 +596,15 @@ impl ::bincode::Decode for WorkerStatusRecord {
...
-            extensions: Decode::decode(decoder).or_else(|err| {
-                if let DecodeError::UnexpectedEnd { .. } = &err {
-                    Ok(WorkerStatusRecordExtensions::Extension1 {
-                        active_plugins: HashSet::new(),
-                    })
-                } else {
-                    Err(err)
-                }
-            })?,
+            extensions: WorkerStatusRecord::handle_decoded_extensions(Decode::decode(decoder))?,
```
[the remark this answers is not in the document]
vigoo: Seemed like a good idea and I tried but then reverted because I could not find a nice way to do it. If the public type owns the extension fields, they have to be cloned in the encode implementation which is not very nice (this struct is serialized a lot). If the serializable enum is just a view with references to the public type, then you cannot easily implement the decode side. So for now let's leave it as is and we can try again to make it nicer next time we touch it.

golem-worker-executor-base/src/worker/mod.rs
vigoo: This feature reverts everything to a given point in the oplog. Including worker updates, pending invocations, etc. I think this is what most users would expect. We could also have some more advanced "history editing" feature where you just remove random invocations and leave the rest of the history as is. I think such a feature could go into the debugging service later.

golem-api-grpc/proto/golem/worker/v1/worker_service.proto
```diff
@@ -341,4 +343,17 @@ message DeactivatePluginResponse {
     golem.common.Empty success = 1;
     WorkerError error = 2;
   }
+}
+
+
+message RevertWorkerRequest {
+  golem.worker.WorkerId worker_id = 1;
+  golem.common.RevertWorkerTarget target = 3;
```
noise64, line 351: 2?

vigoo: 2

noise64 (approved):

PR #1292, opened 2025-02-06
Allow scheduling worker invocations

### LINKED ISSUE #1250: Scheduled worker invocation
We want workers to be able to schedule future invocations for themselves (or any other worker) using a new host API. The host API will be similar to the wasm-RPC invoke API - taking a worker ID, a function name and parameters as a list of `wit-value`s. 

As the steps are not really testable separately, this ticket covers the whole implementation of:

- Defining the new API in `golem-wit`
- Extending worker executor's `Scheduler` to be able to persist scheduled invocations to workers
- Implement the new host functions using `Scheduler`
- Worker executor tests

See the epic description for a link to the product spec.

mschuwalow:
resolves #1250 

related prs:
* https://github.com/golemcloud/golem-wit/pull/45


golem-worker-executor-base/src/durable_host/golem/v1x.rs
```diff
@@ -223,14 +226,53 @@ impl<Ctx: WorkerCtx> Host for DurableWorkerCtx<Ctx> {
 
         Ok(())
     }
+
+    async fn schedule_invocation(
```
mschuwalow, line 230: This is new

golem-worker-executor-base/src/preview2/mod.rs
```diff
@@ -50,16 +51,40 @@ impl From<golem_wasm_rpc::WitValue> for golem::rpc::types::WitValue {
...
+
+impl From<self::wasi::clocks::wall_clock::Datetime> for DateTime<Utc> {
+    fn from(value: self::wasi::clocks::wall_clock::Datetime) -> DateTime<Utc> {
+        DateTime::from_timestamp(value.seconds as i64, value.nanoseconds)
+            .expect("Received invalid datetime from wasi")
+    }
+}
+
+// reexports so that we don't have to change version numbers everywhere
+pub use self::golem::api0_2_0 as golem_api_0_2_x;
```
mschuwalow, line 89: I was a bit annoyed with having to rename this everywhere when bumping the version, this should avoid it in the future

golem-worker-executor-base/src/services/scheduler.rs
```diff
@@ -135,6 +161,10 @@ impl SchedulerServiceDefault {
         let previous_hour_key = Self::schedule_key_from_timestamp(previous_hours_since_epoch);
         let current_hour_key = Self::schedule_key_from_timestamp(hours_since_epoch);
 
+        // TODO: couple of issues with this implementation
```
mschuwalow, line 164: These were all already present in the current implementation, this is just notes for later when we do a cleanup pass. I think it's fine for now for our purposes

golem-worker-executor-base/src/services/scheduler.rs
```diff
@@ -238,19 +293,39 @@ impl SchedulerServiceDefault {
...
-                Level::INFO,
-                "scheduler",
-                worker_id = owned_worker_id.worker_id.to_string()
-            );
-            self.worker_access
-                .activate_worker(&owned_worker_id)
-                .instrument(span)
-                .await;
+            // We are completely done with the action, purge it from the queue
+            self.key_value_storage
```
mschuwalow, line 325: Moved this to the end so we are only removing scheduled actions if we actually successfully ran them.
Before we might have lost actions in the presence of failures

wasm-rpc-stubgen/src/wit_generate.rs
```diff
@@ -264,6 +264,7 @@ pub fn add_dependencies_to_stub_wit_dir(def: &StubDefinition) -> anyhow::Result<
         }
     }
 
+    // TODO: take these from ${workspace_root}/wit instead of the golem-wit crate.
```
mschuwalow, line 267: Just a cleanup note

wit/host.wit
```diff
@@ -2,8 +2,8 @@ package golem:api;
 
 world golem {
   import golem:api/host@0.2.0;
-  import golem:api/host@1.1.1;
-  import golem:api/oplog@1.1.1;
+  import golem:api/host@1.2.0;
```
mschuwalow, line 5: We have tests compiled against the 1.1.x wit. The fact that tests are still passing should be enough to prove that semver works here, but I can provide a dedicated example project if desired.

golem-worker-executor-base/src/durable_host/golem/v1x.rs
```diff
@@ -61,72 +64,72 @@ impl<Ctx: WorkerCtx> HostGetWorkers for DurableWorkerCtx<Ctx> {
...
     }
 
     async fn get_oplog_index(&mut self) -> anyhow::Result<OplogIndex> {
-        golem::api0_2_0::host::Host::get_oplog_index(self).await
+        golem_api_0_2_x::host::Host::get_oplog_index(self).await
     }
 
     async fn set_oplog_index(&mut self, oplog_idx: OplogIndex) -> anyhow::Result<()> {
-        golem::api0_2_0::host::Host::set_oplog_index(self, oplog_idx).await
+        golem_api_0_2_x::host::Host::set_oplog_index(self, oplog_idx).await
```
afsalthaj, line 106: What was the need of the x? For the runtime to pick any 0.2.x version? 
We already changed it to 1.x right? So why is this part changing?

golem-worker-executor-base/src/preview2/mod.rs
afsalthaj: 👍

golem-worker-executor-base/src/services/scheduler.rs
```diff
@@ -71,6 +79,24 @@ impl<Ctx: WorkerCtx> SchedulerWorkerAccess for Arc<dyn WorkerActivator<Ctx> + Se
...
+        function_input: Vec<Value>,
+    ) -> Result<(), GolemError> {
+        let worker = self
+            .get_or_create_running(&owned_worker_id, None, None, None, None)
+            .await?;
+
+        worker
+            .invoke(idempotency_key, full_function_name, function_input)
+            .await?;
+
```
afsalthaj, line 97: Given the name is`activate_and_invoke`, what is getting "activated" here?

golem-worker-executor-base/src/services/scheduler.rs
```diff
@@ -135,6 +161,10 @@ impl SchedulerServiceDefault {
         let previous_hour_key = Self::schedule_key_from_timestamp(previous_hours_since_epoch);
         let current_hour_key = Self::schedule_key_from_timestamp(hours_since_epoch);
 
+        // TODO: couple of issues with this implementation
+        // 1: We only query scheduled actions for the current hour - 1. If we are unavailable for longer than that actions will not be run.
+        // 2: We use the timestamp of the scheduled action as a unique key. If we have 2 actions scheduled for the same point in time one will be silently discarded.
```
afsalthaj, line 166: I am not understanding. So can schedule by calling a host function. Now `n` number of workers running may schedule an invocation probably at the same time. Some of them happen to have exactly the same timestamp. And those will be discarded. Please help me understand, why there is a need of unique keys for scheduled actions?

golem-worker-executor-base/src/services/scheduler.rs
```diff
@@ -172,22 +202,47 @@ impl SchedulerServiceDefault {
...
+
+                    let result = self
+                        .promise_service
+                        .complete(promise_id.clone(), vec![])
+                        .await;
+
+                    // TODO: We probably need more error handling here as not completing a promise that is expected to complete can lead to deadlocks.
+                    match result {
+                        Ok(_) => {
+                            // activate worker so it starts processing the newly completed promises
```
afsalthaj, line 223: I didn't understand this part either. How is this related to scheduled invocation? (Couldn't spot in PR review)

golem-worker-executor-base/src/services/scheduler.rs
mschuwalow: The problem is if for some reason we have multiple actions scheduled for the same point in time. We will overwrite those in the kv store -> only keeping one.

golem-worker-executor-base/src/services/scheduler.rs
mschuwalow: Unrelated to scheduled invocation, just general notes regarding the current schedule implementation

golem-worker-executor-base/src/services/scheduler.rs
vigoo: @afsalthaj it activates the worker to be able to invoke it. this terminology is used in many places in the executor, no need to change here.

golem-worker-executor-base/src/services/scheduler.rs
```diff
@@ -238,19 +293,39 @@ impl SchedulerServiceDefault {
...
                     }
                 }
+                ScheduledAction::Invoke {
+                    owned_worker_id,
+                    idempotency_key,
+                    full_function_name,
+                    function_input,
+                } => {
+                    // TODO: We probably need more error handling here and retry the action when we fail to enqueue the invocation.
+                    // We don't really care that it completes here, but it needs to be persisted in the invocation queue.
```
vigoo, line 303: See my comment at activate_and_invoke

golem-worker-executor-base/src/services/scheduler.rs
```diff
@@ -71,6 +79,24 @@ impl<Ctx: WorkerCtx> SchedulerWorkerAccess for Arc<dyn WorkerActivator<Ctx> + Se
...
+
+    async fn activate_and_invoke(
+        &self,
+        owned_worker_id: OwnedWorkerId,
+        idempotency_key: IdempotencyKey,
+        full_function_name: String,
+        function_input: Vec<Value>,
+    ) -> Result<(), GolemError> {
+        let worker = self
+            .get_or_create_running(&owned_worker_id, None, None, None, None)
```
vigoo, line 91: This (async) blocks until the worker resumes running (including replaying the oplog and potentionally continuing an ongoing invocation which can run for an arbitrary time), and only _after_ you are calling invoke which enqueues the new invocation.

I think what we need to do instead is to "get or create suspended" - which just initializes the `Worker` but does not start it, and invoke on that, and finally activate the worker to make sure the worker is running if it was suspended. (if I remember corerctly worker activator had something that does that in a background task to not block the scheduler)

wasm-rpc/Cargo.toml
```diff
@@ -15,13 +15,14 @@ crate-type = ["cdylib", "rlib"]
 harness = false
 
 [dependencies]
-wit-bindgen-rt = { version = "=0.36.0", features = ["bitflags"] }
+wit-bindgen-rt = { version = "=0.39.0", features = ["bitflags"] }
```
vigoo, line 18: Let's keep using the same version that is in our current recommended (= depended in the stubgen and mentioned in our docs) cargo-component, 0.20, which depends on wit-bindgen 0.36

wit/deps.toml
```diff
@@ -1 +1 @@
-all = "https://github.com/golemcloud/golem-wit/archive/v1.1.7.tar.gz"
+all = "https://github.com/golemcloud/golem-wit/archive/044dba05b1f883a821cc767e151456dddb3f455f.tar.gz"
```
vigoo, line 1: Don't forget to change back to a tag after the golem-wit PR is merged

vigoo (changes requested):
Looks good, added  few comments

golem-worker-executor-base/src/services/scheduler.rs
mschuwalow: WorkerActivator#activate_worker is just
```rust
    async fn activate_worker(&self, owned_worker_id: &OwnedWorkerId) {
        let metadata = self.all.worker_service().get(owned_worker_id).await;
        match metadata {
            Some(_) => {
                if let Err(err) = Worker::get_or_create_running(
                    &self.all,
                    owned_worker_id,
                    None,
                    None,
                    None,
                    None,
                )
                .await
                {
                    error!("Failed to activate worker: {err}")
                }
            }
            None => {
                error!("WorkerActivator::activate_worker: worker not found")
            }
        }
    }
```
where Worker::get_or_create_running is
```rust
    pub async fn get_or_create_running<T>(
        deps: &T,
        owned_worker_id: &OwnedWorkerId,
        worker_args: Option<Vec<String>>,
        worker_env: Option<Vec<(String, String)>>,
        component_version: Option<u64>,
        parent: Option<WorkerId>,
    ) -> Result<Arc<Self>, GolemError>
    where
        T: HasAll<Ctx> + Send + Sync + Clone + 'static,
    {
        let worker = Self::get_or_create_suspended(
            deps,
            owned_worker_id,
            worker_args,
            worker_env,
            component_version,
            parent,
        )
        .await?;
        Self::start_if_needed(worker.clone()).await?;
        Ok(worker)
    }
```

and Worker::start_if_needed:
```rust
    pub async fn start_if_needed(this: Arc<Worker<Ctx>>) -> Result<bool, GolemError> {
        Self::start_if_needed_internal(this, 0).await
    }

    async fn start_if_needed_internal(
        this: Arc<Worker<Ctx>>,
        oom_retry_count: u64,
    ) -> Result<bool, GolemError> {
        let mut instance = this.instance.lock().await;
        if instance.is_unloaded() {
            this.mark_as_loading();
            *instance = WorkerInstance::WaitingForPermit(WaitingWorker::new(
                this.clone(),
                this.memory_requirement().await?,
                oom_retry_count,
            ));
            Ok(true)
        } else {
            debug!("Worker is already running or waiting for permit");
            Ok(false)
        }
    }

```
and WaitingWorker::new is the part that actually starts the worker, which happens in a dedicated task and does not block the caller.
```rust
struct WaitingWorker {
    handle: Option<JoinHandle<()>>,
}

impl WaitingWorker {
    pub fn new<Ctx: WorkerCtx>(
        parent: Arc<Worker<Ctx>>,
        memory_requirement: u64,
        oom_retry_count: u64,
    ) -> Self {
        let span = span!(
            Level::INFO,
            "waiting-for-permits",
            worker_id = parent.owned_worker_id.worker_id.to_string(),
        );
        let handle = tokio::task::spawn(
            async move {
                let permit = parent.active_workers().acquire(memory_requirement).await;
                Worker::start_with_permit(parent, permit, oom_retry_count).await;
            }
            .instrument(span),
        );
        WaitingWorker {
            handle: Some(handle),
        }
    }
}
```
---
So I think the current code is already not blocking the caller. Adjusted it anyway as you suggested, as I agree that 'create_suspended > invoke > activate' makes the semantics of what we are doing much more clear :+1: 


I will try to write a test that proves that, but no good idea how to set it up yet.

golem-worker-executor-base/src/services/worker_activator.rs
```diff
@@ -29,6 +30,15 @@ pub trait WorkerActivator<Ctx: WorkerCtx> {
     /// Makes sure an already existing worker is active in a background task. Returns immediately
     async fn activate_worker(&self, owned_worker_id: &OwnedWorkerId);
 
+    /// Enqueue an invocation to a worker.
+    async fn enqueue_invocation(
```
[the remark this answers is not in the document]
mschuwalow: I wanted to move the logic out of the SchedulerWorkerAccess into a more central place. This way the SchedulerWorkerAccess is just a thin wrapper around the WorkerActivator instead of implementing logic itself.

Will revert and put it back into the scheduler.

vigoo (approved):

PR #1295, opened 2025-02-07
Allow configuring compiler with the types for certain global variables

### LINKED ISSUE #1289: Infer all request path variables in API  gateway as String
```
"foo-${request.path.user-id}" <- should be inferred as string

```

afsalthaj:
Fixes #1289 

golem-rib module doesn't know anything about the semantics about `request.path` or `request.*`. It doesn't care.
However if someone writes the following code, it knows , they need to pass from (somewhere) outside - a "request", and it has to be a record that has `user-id` and `number` that are string and u64 respectively.

```rust
  let res: string = request.path.user-id;
  "worker-name-${res}"

```

However, this strongly type requirements is in fact counter intuitive. Anything under a request path is anyway `string`. This is done by being able to configure the compiler to consider certain variable (in fact global variable) to be of a particular type.

```rust
   "worker-name-${request.path.user-id}"
```

Note that I did allow some "flexibility" such as `request.path.user-id` to be both `string` and  some other type such as `u32`  (user explicitly specified). These flexibilities are taken off. It's giving more pain and instability within Rib. 

In short `request.path.user-id` can only be `string` unless you explicitly parse/cast it as u32 and use it elsewhere. In other words, you  cannot parse `request.path.user-id` to a function that accepts `u32` and expect it to automatically work (obviously, just like many languages).


golem-worker-service-base/src/gateway_rib_compiler/mod.rs
```diff
@@ -25,10 +25,15 @@ pub struct DefaultWorkerServiceRibCompiler;
 
 impl WorkerServiceRibCompiler for DefaultWorkerServiceRibCompiler {
     fn compile(rib: &Expr, export_metadata: &[AnalysedExport]) -> Result<CompilerOutput, String> {
-        rib::compile_with_limited_globals(
+        rib::compile_with_restricted_global_variables(
             rib,
             &export_metadata.to_vec(),
             Some(vec!["request".to_string()]),
+            Some(TypeDefault {
```
mschuwalow, line 32: Would it be possible to specify the full type of request instead here? I think this would also be useful as we might want to do something similar for other globals such as worker that the scripts have access to.
I.e. instead of treating it as a fallback to the type inference it is used as another input to the type inference (which might also allow less typing in users scripts I guess)

I think there are a couple of interesting cases:
1. exact path type can be extracted from the binding input, but probably not necessary,
2. exact headers are not known here, but we know that their name should be strings and their values are unknown (as we are parsing them as json currenly). Would something like map<x, y> make sense in InferredType to handle these two cases?
3. The body can be an arbitrary value. Would it work if we specify this as InferredType::Unknown

afsalthaj: automatic coercion to be avoided is the next step

afsalthaj: We can specify list of `TypeDefault`, that anything under `headers` is also string.
They are not necessarily strings by the way. I wouldn't do anything with the body as well. Not sure why do we need that kind of flexibility.


Example:

vec![

 TypeDefault {
    variable_id: "request"
    path: vec!["path"].
    InferredType::Str
 } ,
 TypeDefault {
    variable_id: "request"
    path: vec!["headers"].
    InferredType::Str
 } , 
]

afsalthaj: > instead of treating it as a fallback to the type inference for specific paths (if I understand the code correctly) it is used as another input to the type inference (which might also allow less typing in users scripts I guess)

It's in fact not a fallback. It is another input to the type inference that allows less typing in users scripts. However, Rib won't trust such outside injection as is.  Note that in any language, users explicitly convert it to `Int` or `Str` or otherwise its a compilation error

golem-rib/src/inferred_type/unification.rs
```diff
@@ -573,18 +574,22 @@ pub fn unify_with_required(
...
+                } else if inferred_type_left.is_number() && inferred_type_right.is_number() {
                     Ok(InferredType::AllOf(vec![
-                        inferred_type1.clone(),
-                        inferred_type2.clone(),
+                        inferred_type_left.clone(),
+                        inferred_type_right.clone(),
                     ]))
+                } else if inferred_type_left.is_string() && inferred_type_right.is_number() {
+                    Ok(inferred_type_right.clone())
+                } else if inferred_type_left.is_number() && inferred_type_right.is_string() {
```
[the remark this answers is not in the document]
afsalthaj: automatic coercion to be avoided is the next step

golem-worker-service-base/src/gateway_rib_compiler/mod.rs
afsalthaj: > or other globals such as worker that the scripts have access to.

Yes we can provide this too already. My logic handles that.

```
vec![

 TypeDefault {
    variable_id: "request"
    path: vec!["path"].
    InferredType::Str
 } ,
 TypeDefault {
    variable_id: "worker"
    path: vec!["name"].
    InferredType::Str
 } , 
]
```

golem-rib/src/inferred_type/unification.rs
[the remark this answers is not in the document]
afsalthaj: Squashed

golem-worker-service-base/src/gateway_rib_compiler/mod.rs
afsalthaj: This approach makes things safely flexible in many ways. For example, if you are just specifying worker name using Rib (although that will disappear soon), then you can allow to infer everything as a string (and not just request.path_

````
request.body.foo
```
which should be the worker-name and we don't need to things like

```
let x: string = request.body.foo;
x
```

afsalthaj: >exact path type can be extracted from the binding input, but probably not necessary,

Also I didn't understand. 

>exact headers are not known here, but we know that their name should be strings and their values are unknown (as we are parsing them as json currenly). Would something like map<x, y> make sense in InferredType to handle these two cases?

Hmm, not fully sure what is this. Because we could just anything under headers is a string in this PR. We don't need maps. I think these explanations are given in comments within code in this PR.

>The body can be an arbitrary value. Would it work if we specify this as InferredType::Unknown

We don't care about request.body.

mschuwalow: Makes sense, thank you for the explanation.

I was originally confused why you went with specifying types for specific paths of a global instead of typing the entire global. But now I agree that this is enough for our usecase.

afsalthaj: Awesome. Thanks for the quick reviews

afsalthaj: Plese note, I renamed some structs if you reviewing it again. Hope you won't get confused

jdegoes (approved):

PR #1303, opened 2025-02-09
Fixes #1302

### LINKED ISSUE #1302: API name with special characters causes it to be inaccessibe in Golem Cloud Console
**Description:**

Creating an API with a name containing special characters in the Golem Cloud Console leads to the API becoming inaccessible. Specifically, I have an API with the name "testingthisaswe;;", Opening the API Page for editing, or deleting leads to the following page

Image

**Steps to Reproduce:**

1. Log in to the Golem Cloud Console.
2. Navigate to the section for creating new APIs. (This is not a problem for Components as it uses ComponentId; not a Problem with API deployment as they do not have individual pages)
3. Create a new API and enter "test;;" (or another name containing special characters like ";", ":", "/", etc.) as the API name.
4. Save the new API.
5. Attempt to access or view the details of the newly created API in the Golem Cloud Console.

**Additional Information**

* Should be able to delete this using golem-cloud-cli, Yet to test this.
* This should also be a problem for Workers, though default workername is generated alphanumerical, you can set custom WorkerName with special characters
* Not sure if the problem only occurs if the special character is set at the end of name or anywhere in the name. 
* I am avoiding testing this further, figuring out which special characters and if the problem only happens if it is at the end of the name, since I cannot access it for deleting.
* Fix is easy, Restrict names to only have alpahnumerical, - , _

itsparser:
Resolves #1302


afsalthaj:
Could you please make the build green?

vigoo:
Hi! As the feedback loop is a bit slow because you are a first contributor and I have to click on an approve button here for each attempt :) I just wanted to mention make sure you:
- merge with latest `main` (we are merging in several changes per  dy)
- run `cargo make build` to make sure everything compiles
- run `cargo make fix` to make sure all clippy rules are passing and formatting 
- make sure you run the tests of whatever area you touched (in this case the worker service tests)

there are some flaky worker executor tests that may randomly fail, I'll restart your CI run when that happens.

itsparser:
> Hi! As the feedback loop is a bit slow because you are a first contributor and I have to click on an approve button here for each attempt :) I just wanted to mention make sure you:
> 
> * merge with latest `main` (we are merging in several changes per  dy)
> * run `cargo make build` to make sure everything compiles
> * run `cargo make fix` to make sure all clippy rules are passing and formatting
> * make sure you run the tests of whatever area you touched (in this case the worker service tests)
> 
> there are some flaky worker executor tests that may randomly fail, I'll restart your CI run when that happens.

Hi @vigoo , Thanks for helping me out, I have executed all the command which you have mention, but will do it one more time in same order in which you have specified and will get back to you

itsparser:
@vigoo you have missed two more command which run the test case 

`cargo make tests` 
`cargo make integration-tests`

since I am facing error in integration testing GitHub-action 
will debug this today and update here

itsparser:
@vigoo I have tested it locally can you approve that it now

afsalthaj (approved):

PR #1305, opened 2025-02-09
Fixes #1301

### LINKED ISSUE #1301: API deployment requires api-definition-id by default
**Description:**

Minor Bug

Docs for API Deployment API say - If api-definition-id is not set, it lists all API deployments. If api-definition-id is set, returns a single API deployment. Just below it says : api-definition-id | string | Yes [Yes, here implies it is not optional]

api-definition-id is required for GET Request at `localhost:9881/v1/api/deployments`; Direct GET Request leads to "failed to parse parameter `api-definition-id`: Type "string" expects an input value."

Similarly `golem-cli api-deployment list` leads to the following error

```
error: the following required arguments were not provided:
  --id <ID>

Usage: golem-cli api-deployment list --id <ID>

For more information, try '--help'.
```

**Fix** 

 * A Simple change to the docs to edit the line `If api-definition-id is not set, it lists all API deployments. If api-definition-id is set, returns a single API deployment`
 * A better fix would be to make api-definition-id optional, and it listing all API deployment, keeping in line with `golem-cli component list`, `golem-cli api list` and `golem-cli plugin list`

itsparser:
Resolves  #1301

### DIFF
--- a/golem-cli/src/clients/api_deployment.rs
+++ b/golem-cli/src/clients/api_deployment.rs
@@ -28,7 +28,7 @@ pub trait ApiDeploymentClient {
     ) -> Result<ApiDeployment, GolemError>;
     async fn list(
         &self,
-        api_definition_id: &ApiDefinitionId,
+        api_definition_id: Option<ApiDefinitionId>,
         project: &Self::ProjectContext,
     ) -> Result<Vec<ApiDeployment>, GolemError>;
     async fn get(&self, site: &str) -> Result<ApiDeployment, GolemError>;
--- a/golem-cli/src/oss/clients/api_deployment.rs
+++ b/golem-cli/src/oss/clients/api_deployment.rs
@@ -68,12 +68,17 @@ impl<C: golem_client::api::ApiDeploymentClient + Sync + Send> ApiDeploymentClien
 
     async fn list(
         &self,
-        api_definition_id: &ApiDefinitionId,
+        api_definition_id: Option<ApiDefinitionId>,
         _project: &Self::ProjectContext,
     ) -> Result<Vec<ApiDeployment>, GolemError> {
-        info!("List api deployments with definition {api_definition_id}");
+        info!("List api deployments with definition");
 
-        let deployments = self.client.list_deployments(&api_definition_id.0).await?;
+        let api_definition_id = api_definition_id.map(|n| n.0);
+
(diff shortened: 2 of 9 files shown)


golem-cli/src/clients/api_deployment.rs
```diff
@@ -28,7 +28,7 @@ pub trait ApiDeploymentClient {
     ) -> Result<ApiDeployment, GolemError>;
     async fn list(
         &self,
-        api_definition_id: &ApiDefinitionId,
+        api_definition_id: Option<&ApiDefinitionId>,
```
afsalthaj, line 31: 👍

golem-cli/src/oss/clients/api_deployment.rs
```diff
@@ -68,12 +68,17 @@ impl<C: golem_client::api::ApiDeploymentClient + Sync + Send> ApiDeploymentClien
 
     async fn list(
         &self,
-        api_definition_id: &ApiDefinitionId,
+        api_definition_id: Option<&ApiDefinitionId>,
         _project: &Self::ProjectContext,
     ) -> Result<Vec<ApiDeployment>, GolemError> {
-        info!("List api deployments with definition {api_definition_id}");
+        info!("List api deployments with definition");
```
afsalthaj, line 74: This log now sounds a bit odd.

afsalthaj:
I think the change makes sense to me. However, could you please try making the build green to gain some confidence before we  click an approval?

Note: There may be some flaky tests, and in that case, re-run failed jobs

golem-cli/src/oss/clients/api_deployment.rs
itsparser: okay will update the log

itsparser: will update the log as following

```
info!(
            "List api deployments {}",
            api_definition_id.map_or("without definition id".to_string(), |x| format!(
                "with definition ID {}",
                x
            ))
        );
 ```

itsparser:
@afsalthaj Have update the log to make more readability and also have run the check  locally to make sure the GitHub action will pass thought

itsparser:
Converting this to draft to fix the test case for once and all

afsalthaj (approved):

PR #1308, opened 2025-02-09
Allow cancelling a scheduled invocation

mschuwalow:
followup to #1292 

related prs:
* https://github.com/golemcloud/golem-wit/pull/48


golem-worker-executor-base/src/durable_host/wasm_rpc/mod.rs
```diff
@@ -426,40 +430,89 @@ impl<Ctx: WorkerCtx> HostWasmRpc for DurableWorkerCtx<Ctx> {
...
 
-        // The scheduler uses the (deterministic) datetime to deduplicate scheduled actions.
-        // The worker executor uses the idempotency key to deduplicate invocation.
-        // With both together we don't need any extra idempotency handling here.
-        self.state
-            .scheduler_service
-            .schedule(datetime.into(), action)
-            .await;
+        let cancellation_token = CancellationTokenEntry {
+            schedule_id: bincode::encode_to_vec(schedule_id, bincode::config::standard()).unwrap(),
```
vigoo, line 511: should not we use the `serialize` function from golem-common? (which prefixes the bincode binary with a serialization version number, to allow us switch serialization formats in the future)

vigoo (commented):
One small question otherwise good

golem-worker-executor-base/src/durable_host/wasm_rpc/mod.rs
mschuwalow: Makes sense, will update when back

mschuwalow: done

vigoo (approved):

PR #1312, opened 2025-02-10
Implemented cancelling pending invocations

### LINKED ISSUE #453: Ability to cancel a queued invocation by its invocation key
Each invocation already has an associated _invocation key_ but for fire-and-forget invocations it is not exposed to the user. If we return the enqueued invocation's ID then we can use the same ID for cancelling an invocation.

It can only be cancelled if it did not start running yet, and it can be done competely via the oplog (and in memory the invocation queue):
- check if the invocation is still pending
- if yes, write an entry that it is not needed
- remove invocations from the pending queue in case this entry

vigoo:
Resolves #453

### DIFF
--- a/golem-worker-executor-base/src/grpc.rs
+++ b/golem-worker-executor-base/src/grpc.rs
@@ -21,7 +21,8 @@ use golem_api_grpc::proto::golem::common::ResourceLimits as GrpcResourceLimits;
 use golem_api_grpc::proto::golem::worker::{Cursor, ResourceMetadata, UpdateMode};
 use golem_api_grpc::proto::golem::workerexecutor::v1::worker_executor_server::WorkerExecutor;
 use golem_api_grpc::proto::golem::workerexecutor::v1::{
-    ActivatePluginRequest, ActivatePluginResponse, ConnectWorkerRequest, DeactivatePluginRequest,
+    ActivatePluginRequest, ActivatePluginResponse, CancelInvocationRequest,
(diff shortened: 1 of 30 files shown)


golem-worker-executor-base/src/grpc.rs
```diff
@@ -471,6 +472,56 @@ impl<Ctx: WorkerCtx, Svcs: HasAll<Ctx> + UsesAllDeps<Ctx = Ctx> + Send + Sync +
...
+                    )
+                    .await?;
+                    worker.cancel_invocation(idempotency_key).await?;
+                    Ok(())
+                } else if metadata
+                    .last_known_status
+                    .invocation_results
+                    .contains_key(&idempotency_key)
+                {
+                    Err(GolemError::invalid_request("Invocation already completed"))
```
noise64, line 516: is cancelling already completed an error? shouldn't it e.g. returned as bool?

noise64 (approved):
Added comment on whether to signal "already completed" in the success channel instead of an error

golem-worker-executor-base/src/grpc.rs
vigoo: Yeah probably that's a better API as the user has no control in whether it has been already completed or not.
Should we change it?

noise64: I think we should change it now, as this can be quite frustrating when comes up and somebody wants to apply same logic based on it.

vigoo: done

PR #1313, opened 2025-02-10
Allow inline type annotations

### LINKED ISSUE #1290: Support Inline Type Annotations in Rib
"foo-${request.path.user-id: i32}" <- inline type annotations should be possible

afsalthaj:
Fixes #1290 

* Remove panics in parser (except in call.rs)
* Add type check module for binary math operations, allowing better error messages.
* Remove redundant error messages (it was confusing due to error msg accumulation) 
* Fix bugs in type-spec (configuring compiler with specific types for global variables) - found this as part of this work
* Being able to inline types for expressions
`foo.bar.baz` will fail to compile while `foo.bar.baz: u32` will succeed.
* Make sure inline type annotations are always consistent. We supported `1u32` before, but with this PR `1: u32` will also work keeping it consistent with other expressions as given above (keeping backward compatibility) 
* Some of the above improvements came in while writing tests for inline-type-annotation feature
* Make sure to handle all `result` possibilities in type annotation as given here: https://component-model.bytecodealliance.org/design/wit.html#results


golem-rib/src/interpreter/interpreter_stack_value.rs
```diff
@@ -205,7 +205,7 @@ mod internal {
     #[cfg(feature = "json_in_errors")]
     pub fn unable_to_complete_math_operation(left: &ValueAndType, right: &ValueAndType) -> String {
         format!(
-            "Unable to complete math operation for {}, {}",
+            "Unable to complete math operation for operands {}, {}",
```
afsalthaj, line 208: Usually this code shouldn't be hitting if the compiler was strong enough. And this is fixed in this PR.

vigoo (approved):

PR #1321, opened 2025-02-12
Type-free invocation API

### LINKED ISSUE #1122: Type free worker invocation
Currently type free invocation is half baked into golem-cli where it first tries to automatically convert to a `type_annotated_value`.

In this case, it fries to deserialize it as a `TypeAnnotatedValueJson` and then tries to convert to a real `TypeAnnotatedValue` using `typ` information. 

If this attempt fails, then it tries to fetch the worker metadata and use that information to invoke.

The intention of this ticket is to make sure such a logic is ported back to worker-service such that clients don't need repeat this logic

vigoo:
Resolves #1122 

# Type-free invocation REST API 
Previously we required explicit **type information** in the JSON input parameters on the invocation endpoints beside the JSON value representation. Without the type information it is not possible to convert the JSON value representation into the typed `Value` (and friends) representations - because in the JSON format we use field and case names to be "human friendly", but these names have to be mapped into indices when converting to the internal representation. This type information is available as part of the *component metadata*, but in the worker service when the invoke request arrives, there is no way to know what exact version (the component id itself is known from the worker id) the targeted worker is running when it processes the invocation. 

Originally when we faced this issue we first wanted to cache things in the worker service (query the worker for it's current component version, then query the metadata for that version, cache everything) but this is still incorrect because it is not guaranteed that the currently returned component version will be still true when the worker executes the invocation - there is an invocation queue involved which can also contain update attempts (which can either succeed or fail).

For the above reason we required the user to _know_ (assume) what component version they are targeting so worker service does not have to query or cache anything.

This PR changes this in the following way:
- If there is type information in the JSON (like before) everything works as before
- But if there is no type information just value, the whole JSON is sent to the worker executor using a new internal invoke API. 
- The worker executor prepares the worker for enqueuing the invocation which involves calculating its latest metadata. Here before we enqueue the invocation we already see it's actual invocation queue, and we can _assume_ that all the enqueued update attempts will succeed, and use that version to get the component metadata to interpret the JSON value.

Note that it is still not guaranteed that all those update attempts succeed and in that case it is possible that there is a mismatch in the current component version and the version used for decoding the parameters when the actual invocation happens. But this is no different from deferring this choice to the user; somebody somewhere needs to make an assumption because the invocation will happen in the future.

## REST API change
As seen in the OpenAPI yaml, the only change in the REST API is that the invocation parameters are now using a different data type, with the only difference that the type field is optional.

# CLI change
The CLI previously handled this in itself by checking if the provided JSON has type information and if not, it queried the current component version from the worker, etc. This code has been removed and it now uses this new feature of the invocation API.

# Worker executor refactoring
The `grpc` module has been "promoted to a directory" so the grpc api implementation is now in `grpc/mod.rs`. The invocation related helper traits and implementations have been extracted to a sub module, and have been refactored to remove some redundancy.

# Questions

## Why not defer decoding the JSON when the actual invocation happens?

First of all, we definitely don’t want to remove the “normal” way of invoking things, where the values are Values (or protobuf Vals or whatever - the things directly convertible to what wasmtime wants) because in many cases, for example wasm-rpc, that means a lot less conversions and encodings.
So if we would want to decode the json invocation values when the invocation starts it would mean that everything that deals with invocation parameters needs to handle two kind of invocation parameters (value and json), and this would mean for example all the oplog entries like “pending invocation”, “exported function invocation start”. So my first reason is that it would add extra complexity to all these places.

Second reason, even if we would not mind that complexity, it would make things hard for example oplog query. In oplog query we present the user “well typed values” (ValueAndType values) of everything that happened, including the invocation parameters. That’s the only way it’s actually usable on a UI. We cannot decode the json when we produce the oplog query because of the same reasons and it once again would even leak out through the oplog query API and each invocation related entry would contain either typed values or json strings, it would affect how you can search in these, etc.

Third reason I just felt (and still feel) it wrong - not from a code or complexity point of view, but regarding the responsibilities of worker executor. This human readable json format is something that in my opinion should live above worker executor, and it’s just one of infinite representations how you can describe these component model values. There’s also wave, for example, etc. It does not feel right to me to thread this as a special case through all these core components

# History
- https://github.com/golemcloud/golem/pull/730


wasm-rpc/src/json/mod.rs
```diff
@@ -39,6 +39,77 @@ struct TypeAnnotatedValueJson {
...
 
+/// A representation that optionally pairs type definition with a JSON represented value.
+///
+/// It can only be converted to any of the typed value representations if the type information
+/// is present (or provided externally).
+///
+/// The JSON format is backward compatible with `TypeAnnotatedValueJson`.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct OptionallyTypeAnnotatedValueJson {
+    typ: Option<AnalysedType>,
```
mschuwalow, line 50: let's call this 'type' in the json representation instead of 'typ'

golem-worker-service-base/src/service/worker/invocation_parameters.rs
```diff
@@ -0,0 +1,45 @@
...
+pub enum InvocationParameters {
+    TypedProtoVals(Vec<TypeAnnotatedValue>),
+    RawJsonStrings(Vec<String>),
+}
+
+impl InvocationParameters {
+    pub fn from_optionally_type_annotated_value_jsons(
+        values: Vec<OptionallyTypeAnnotatedValueJson>,
+    ) -> Result<Self, Vec<String>> {
+        let all_have_types = values.iter().all(|v| v.has_type());
```
mschuwalow, line 27: Probably mixing typed and untyped values should be a 400. Otherwise people might expect we respect them in some way if they provide types for only part of the values

mschuwalow (approved):

wasm-rpc/src/json/mod.rs
vigoo: I don't want to break the existing API (this new type is "the old" just making the `typ` field optional)

PR #1341, opened 2025-02-19
Code clean up: Named parameters in Expr nodes

### LINKED ISSUE #1125: Source Locations in Rib - Part 1: Use named fields in Expr tree
Ensure that RibExprTree avoids anonymous fields in its enum variants and uses named fields instead for better clarity and maintainability.

afsalthaj:
Resolves #1125 

This is more or less a mechanical compiler driven changes (although certain parts I had to manually change a bit)

### Struct-like Syntax for Expr tree than tuple like.

* Expr(a, b, c) has been changed to Expr { a, b, c } for better readability and maintainability.
* This ensures consistency in parameter names, making pattern matching clearer, as well as it uncovered 1 issue explained below.

### Cleanup & Refactoring

Mostly naming and additional methods to construct Expr neatly. It's a derived need of the above work. 

 Certain function naming had to be revised. Example:  I created bugs when changing the inferred type of an `Expr` that was formed using simple methods (where the `InferredType` is always set to None). This was due to the ambiguity between `add_inferred_type` and `override_inferred_type`. These are changed to `merge_inferred_type` and  `with_inferred_type` respectively. 

Another example is `fn identifier` was taking a string that it internally forms a global variable id. This was confusing (again created bugs during fixing the compilation errors in tests). 

Probably these fixes is the reason for some of the extra lines of commit in this PR. Other than that, this is just a minor mechanical change in Rib.

#### Issue uncovered
The serde round trip tests started failing after the refactor because we were/are storing `condition` of `if-else` expression as `lhs` and `lhs` as `condition`. Since reading from protobuf values also read this incorrectly (:)) we never had a bug. I have put a comment in the code, as we cannot change the grpc values as it kills backward compatibility.


vigoo:
Is this just switching from the "tuple-like" enum syntax to the "struct-like" one or are there other changes too?

afsalthaj:
Yes, unfortunately it's only a mechanical code clean up, but this is the least amount of code change that was required. this is why it was kept as a separate task

afsalthaj:
>one or are there other changes too?

There are no other changes, other than introducing proper constructor methods, which ended up having conflicts with existing some methods, which _had_ to be renamed.

vigoo (approved):

PR #1364, opened 2025-02-24
Invocation context 1

### LINKED ISSUE #1332: Invocation context in worker executor

vigoo:
Resolves #1332 

This is the first step of introducing support for tracing and invocation context into Golem.

The PR consists of three major features:

- Defines types to describe an invocation context as a stack of spans with string (but extensible) key-value attributes. The invocation context is associated with a running worker's state, and can be manipulated. It also gets propagated through invocations using the existing invocation context parameter of the gRPC APIs. Initial attributes are set containing information about the inbound HTTP request (method, path, etc) and about invocations (function name and idempotency key).
- Ability to specify the root span of an invocation from a Rib script for invocations done through API Gateway
- Support for the W3C Trace Context headers on the inbound edge of a Golem invocation. This is supported both in the API Gateway and on the direct invocation REST API. If these headers are present, the trace id and parent span id are taken from them as well as the trace state (key value pairs defined in the Trace Context spec). The trace state is not used directly in Golem but it is propagated with the tracing context so once the outgoing http support is implemented (in one of the next PRs), it is going to be forwarded downstream as the W3C trace header spec requires.

Based on this, next PRs will
- Add the ability for workers to define custom spans and add custom attributes
- Auto-define spans for RPC and outgoing HTTP requests
- Properly integrate this into the oplog 
- Add end-to-end integration test for it

Golem itself is just propagating the invocation context but not directly using this to report traces to an external tracing service. When all the follow-up changes are done it will be possible to create oplog processor plugins that can reconstruct all the span stacks for a worker and report them for example as OpenTelemetry traces.

In addition to this, the PR also contains a refactoring of the core worker invocation loop, which became unmaintainable complex during the changes coming in the past half year.


golem-common/src/model/mod.rs
```diff
@@ -858,14 +861,20 @@ impl IntoValue for WorkerStatus {
 
 #[derive(Clone, Debug, PartialEq, Encode, Decode)]
 pub enum WorkerInvocation {
-    ExportedFunction {
+    ExportedFunctionV1 {
```
mschuwalow, line 864: If this is done for oplog serialization purposes, I would propose to have a separate data type for that with mapping between the two.
If possible, I think it would make our lives much easier to not have multiple versions in domain types.

golem-worker-executor-base/src/worker/invocation_loop.rs
```diff
@@ -0,0 +1,885 @@
+// Copyright 2024-2025 Golem Cloud
```
mschuwalow, line 1: much nicer :heart:

golem-api-grpc/proto/golem/worker/invocation_context.proto
```diff
@@ -2,10 +2,41 @@ syntax = "proto3";
...
+message InvocationSpan {
+  oneof span {
+    LocalInvocationSpan local = 1;
+    ExternalParentSpan external_parent = 2;
+  }
+}
+
+message LocalInvocationSpan {
+  uint64 span_id = 1;
+  google.protobuf.Timestamp start = 2;
```
mschuwalow, line 30: do we need an end_time here?

golem-api-grpc/proto/golem/worker/invocation_context.proto
```diff
@@ -2,10 +2,41 @@ syntax = "proto3";
...
 message InvocationContext {
   golem.worker.WorkerId parent = 1;
   repeated string args = 3;
   map<string, string> env = 4;
+  optional TracingInvocationContext tracing = 5;
+}
+
+message TracingInvocationContext {
+  string trace_id = 1;               // hexadecimal representation of a 128-bit trace ID
+  repeated InvocationSpan spans = 2; // stack of spans, first item is the current, last one is the root
```
mschuwalow, line 17: What are the cases in which we would begin an invocation with more than one local stack?

I'm thinking we could have this be closer to the opentelemetry standard and just have ids for parents on each levels instead of having this stack of spans https://opentelemetry.io/docs/concepts/signals/traces/.

golem-common/src/model/mod.rs
vigoo: With bincode we must introduce a new constructor to keep compatibility. 
I would like to eventually switch to my binary serializer for these internal formats which is focusing on supporting more data evolution cases but haven't had time to fully finish and test it yet.

golem-api-grpc/proto/golem/worker/invocation_context.proto
vigoo: Turned out that I'm never keeping finished spans in memory so no. Also in the protocol, even when multiple spans are ongoing, on invocation boundaries only the "current stack" is transmitted, which consist of open spans.

golem-api-grpc/proto/golem/worker/invocation_context.proto
vigoo: See the corresponding notion doc where we decided that we want the workers to have access to the whole context stack. If we would just send ids, then we would still be able to export spans to opentelemetry but we would not be able to query information from the workers.

golem-service-base/Cargo.toml
```diff
@@ -28,6 +28,7 @@ aws-config = { workspace = true }
 aws-sdk-s3 = { workspace = true }
 bigdecimal = { workspace = true }
 bincode = { workspace = true }
+bitflags = "2.8.0"
```
afsalthaj, line 31: should this be in workspace?

golem-worker-executor-base/src/durable_host/replay_state.rs
```diff
@@ -262,6 +264,7 @@ impl ReplayState {
                             function_name.to_string(),
                             request,
                             idempotency_key.clone(),
+                            InvocationContextStack::fresh(), // TODO: read from oplog
```
afsalthaj, line 267: is this TODO to do ?

afsalthaj: or not covered in this PR?

golem-worker-executor-base/src/worker/invocation_loop.rs
```diff
@@ -0,0 +1,878 @@
...
+
+/// Outcome of processing a single command within the inner invocation loop
+enum CommandOutcome {
+    /// Break from both the inner and outer loops, there is no way to retry anything
+    BreakOuterLoop,
+    /// Break from the inner loop, setting the retry decision for the outer loop
+    BreakInnerLoop(RetryDecision),
+    /// Continue processing in the inner loop
+    Continue,
+}
```
afsalthaj, line 878: 👍 a needed refactor! looks good now

golem-worker-service-base/src/gateway_execution/gateway_http_input_executor.rs
```diff
@@ -312,10 +369,54 @@ impl<Namespace: Clone> DefaultGatewayInputExecutor<Namespace> {
                 .map(|value| IdempotencyKey::new(value.to_string()))
         };
 
+        let invocation_context = if let Some(invocation_context_compiled) =
```
afsalthaj, line 372: looks good

golem-worker-executor-base/src/durable_host/replay_state.rs
vigoo: It is coming in a next one. See the notion page with the roadmap.

golem-common/src/model/mod.rs
mschuwalow: my point is that we are using this type in other places / our domain.
I would propose having the extra constructors for bincode on a seperate type that is getting converted from/to our domain type when we need to serialize it.
This way the extra cases will not be visible to other code working with it (as they should not be using them directly if I'm understanding this correctly)

afsalthaj (approved):

golem-api-grpc/proto/golem/worker/invocation_context.proto
mschuwalow: Well they would only be able to query the current span, which I think is inline with how other opentelemetry setups work.

But makes sense if we want to support getting the full stack. In that case, should there ever be a remote parent? As users cannot access that one from within the component (at least not without some new host api)

golem-api-grpc/proto/golem/worker/invocation_context.proto
mschuwalow: I meant for eventual oplog processors. I think often you will need to report endtime to your telemetry backends. Or are plugins supposed to infer the end time by looking for the matching invocation-end entry?

As another question, how are we going to do attributes on the current span if we only persist the invocations? Or are these https://www.notion.so/golemcloud/Invocation-Context-implementation-19ea4cb355ec80a3bc69eefcc21926c4?pvs=4#19ea4cb355ec800bbe5af693dd9c74ea coming as dedicated oplog entries in another pr?

golem-service-base/Cargo.toml
vigoo: not used in any other project

mschuwalow (approved):
assuming the stuff discussed in comments is coming in followup prs as new oplog entries.

It would be nice to hide the bincode versioning from the rest of the code though, if you can fit that one in.

PR #1372, opened 2025-02-25
Scala.js wit binding generator completed

### LINKED ISSUE #1263: Full Scala.js support
There is already a proof of concept attempt to support developing Golem components using Scala by compiling to Scala.js first and then use the jco/componentize-js toolchain to package the result into a WASM component.

- https://github.com/golemcloud/golem-scalajs-wit-bindgen contains a PoC binding generator from WIT to Scala.js 
- https://github.com/golemcloud/sbt-wasm-component is an sbt plugin that wraps the binding generator and provides a macro for generating WIT from Scala

The current binding generator does not properly map all WIT features to Scala.js (does not support imports, resources, and incorrectly maps some other types).

The goal of this ticket is to have a real scalajs binding generator, built on top of `wit-bindgen` (and ready to merge upstream if they want to accept it) that supports everything that WIT and Golem supports, with tests, and integrate it into a new version of `sbt-wasm-component`. 

Improvement of the "scala to wit" macro is out of scope but the existing version should continue to work at least as good as was.

By completing this ticket, it should be possible to use any feature of Golem from Scala(.js) and to start building helper libraries to wrap low-level WIT APIs such as wasi-http into idiomatic Scala APIs.

vigoo:
Resolves #1263 
/claim #1263


jdegoes:
@vigoo The best code I have seen in decades! 👀 🎉 🕺 😆

PR #1380, opened 2025-02-27
Dynamic stubgen support for scheduled invocations

### LINKED ISSUE #1309: Add scheduled invocation to the RPC stubs
Using the new scheduled invocation host function directly from user code is inconvenient, so we want to generate scheduled invocation variants of the RPC functions in our wasm-rpc implementation. 

We already generate two functions for each exported function in our genrerated stub resources:
- one for blocking invocation
- one for async invocation (and if it is a function with no return value, it is just enqueuing a pending invocation and returns)

We can add a third one that does scheduled invocation (or 4 if we want it with and without cancellation - to be decided). 

Although we move to *stubless RPC* we did not fully deprecate the generated stubs yet, so this ticket should cover the following areas:

- Adding the scheduled invocation function calls into the generated WIT (in wasm-rpc-stubgen)
- Implement them in the generated Rust stub code (in wasm-rpc-stubgen) similar to how the existing invocation wrappers are implemented
- Implement the same thing in the dynamic linker in worker-executor-base
- Have worker executor tests both linked to stub and "stubless" 

It is probably easier to first implement the generated stub version with a test and add support in the linker for the stubless mode after.

mschuwalow:
resolves #1309


vigoo (approved):
Is the new test component different than the previous new one (`scheduled-component`)? If the only difference is how app manifest is configured to do static vs dynamic stub linking, I don't think we need to duplicate it, just copy both the composed and the non-composed wasm from the target. If they are more different then ignore this :)

PR #1381, opened 2025-02-28
Implement ranges and fix all left recursions

### LINKED ISSUE #1171: Implement Ranges in Rib
# Steps to follow

## Define the syntax, and write parsers for the following

```rust
0..10;

```

This corresponds to `Range`

which iterates from 0 until 10 , and an optional support for inclusion

```rust

0..=10

```

This corresponds to `RangeInclusive`.

We can also support for infinite variables with open-ended support

```rust

let range = 0..

```

This corresponds to `RangeFrom`

We support full range using 

```rust
..
```

and corresponds to `RangeFull`

Optionally, we can support `RangeTo` and `RangeToInclusive` 

```rust
..2;

..=2;

```

Ranges shouldn't be considered as a mere syntax sugar for list, as ranges can be lazily evaluated saving memory and computation. 
Ranges can be used in for comprehensions (except `RangeTo`), and can be used `select_index` in `list` as an example
For this reason, we are not going to convert this to list

##  Map it to the following `Expr`

Range can be considered as an expr, making it easier to support anything which Expr is able to do.

```rust

let x = foo();

let y = bar(); 

let range = x..=y;

// lazily evaluated
for i in range {
   if i < 5 {
        baz(i + 5);
   }
}

```

or

```
 let x: list<u32> = [1, 2, 3, 4,  5, 6]
 
 let y: list<u32> = x[2..5]
 // [3, 4, 5]
 
```

and therefore, the most natural next step is to add these terms in Expr language

```rust

enum Expr {
   ..... 
   Range(Range)
}

enum Idx {
   Number(usize),
   Range(Range)
}

enum Range {
   Range {
       from: Expr
        to: Expr
   }
   RangeInclusive {
     from: Expr
     to: Expr
   }
   
   RangeFrom {
      start: Expr
   }
   
   RangeTo {
     end: Expr
   }
   
   RangeToInclusive {
     end: Expr
   }
   
   RangeFull
}


```

## Handle Range in each phases of type inference

See this code

```rust
 pub fn infer_types(
        &mut self,
        function_type_registry: &FunctionTypeRegistry,
    ) -> Result<(), Vec<String>> {
        self.infer_types_initial_phase(function_type_registry)?;
        self.infer_call_arguments_type(function_type_registry)
            .map_err(|x| vec![x])?;
        type_inference::type_inference_fix_point(Self::inference_scan, self)
            .map_err(|x| vec![x])?;

        self.check_types(function_type_registry)
            .map_err(|x| vec![x])?;
        self.unify_types()?;
        Ok(())
    }

``` 

We need to handle range in each of these phases, and each of the sub phases within them. Example: initial phase has 5 sub phases in it. The inference fix point is defined for `infer_scan` and this has the most number of phases such `type_push_down` and `type_pull_up`.

Please go through each one of them and see how ranges can be compiled. Return `compilation` errors in any of these phases only if we cannot continue processing phases, otherwise leave it to type checker. Example : `let x = [a, b, c, d]; let range = x..10; ` is wrong. While it sounds simple, it is almost difficult to infer "early" that  this expr is wrong. 

### Handle Ranges in type checker

Type checker was a late addition and this may not capture all the errors, making it difficult for unification. However, any new additions into Rib should have a corresponding "handling" in type checker phase. As we go we will add more checks in type check phase, which is the primary entry point to better error messages. 

## Write tests for parser, type inference phases, and compilation phases for Ranges

Write basic tests for ranges and it should have both positive and negative tests.

## Write Interpreter for Ranges
Interpreter converts Rib Expr to Rib Byte Code which is a collection of `IR`. 
We have already handled iterative approach for handling list, and to a great extent, we believe this can be reused for ranges for lazily evaluating them, especially these `IR`s

```rust
 CreateSink(AnalysedType),
 AdvanceIterator,
 PushToSink, 
 SinkToList,
```

Interpretation of range (especially in the context of list comprehension) should mostly be the above IRs. During the real implementation, we will see if anything else would come in. 

This type of lazy evaluation is already done for list comprehension. 

With tests, ensure that a full range is not computed in usecases where it doesn't need to be computed. Example:

```rust

// lazily evaluated
for i in range {
   if i < 5 {
        baz(i + 5);
   }
}

```

afsalthaj:
Resolves #1171 and remaining left calls.

### High level
* The feature was to implement `Ranges` in Rib, but introduction of a new grammer exposed hidden problems in the existing grammar of Rib - with some existing tests with complex rib expressions stack overflowed. This is the reason for why this PR is _more than ranges implementation_.

* Wasn't it done before? A stage of removing Left-calls was done previously, but it was just not enough. Most of the times, the `combine` parser fails not only just because of recursion, but the complexity of the grammar itself. So instead of cherry picking case by case, I just further made the grammer as safe and flexible to get a reliable experience of parsing.  

* Note that the existing tests (of more than 550) never failed until we introduced `Range`.   The stack safety reasons was hard to spot. Example: the stack safety error occurs for a complex `wasm record` while the real reason was due to rest of the parsers. So with this I decided to not have anything complex as part of the simple parser grammer in the already existing (supposed to be) left-recursion-safe grammer.  You can try the stack safety problem yourself, by introducing a `position` parser on the simple_parser within the main branch, and you will see existing tests stack overflowing.

### [Details](https://github.com/golemcloud/golem/blob/implement_ranges/golem-rib/src/parser/rib_expr.rs) 

* We ported `select-field`, `select-index`, `numbers` that can have fractions and exponent component, ranges , and even `worker-function-invoke` all into the `rib-rest` component of left recursion safe grammar. This is done with careful tweaks and compromises in favor of stack safety.  All have a `.` delimiter before parsing `many(rest)`.

* For example: instead of failing early at `123.foo.56: f64` to be parsed as a number we consume everything to be combined with the simple parser output on the left side, to identify whether it is a valid number or not and return proper error message.  The grammer is now more flexible, that you can select an index from a `rib-expr` and not just from specific `identifier` or `sequence`. This helps with easiness in maintaining and understanding the parser, while being stack safe, and leave it to the compiler to fail for invalid code.

* The work improved the robustness of the parser by a lot. The grammer of each individual expressions have become stricter (example: `-` can occur in between a number only after the `e` or `E` in the exponent part), while the overall grammer becomes more flexible (that you can select a field from any rib-expr than just a record or identifier, similar to index selection mentioned above) but fails with a proper message at compile time, with no more random stack overflows error. We now have `position` for both simple-expr and rib-rest and unlike main branch, this PR doesn't fail.

## [Introducing dynamic index selection (not a literal number) in existing `SelectIndex`](https://github.com/golemcloud/golem/blob/implement_ranges/golem-rib/src/expr.rs#L59)
This work was important, becaues range selection on a sequence is desugarded to a for comprehension and it requires selecting index by expression over selecting by a literal number. You can imagine the subsequent changes that is required for type inference, type pull up phase (pulling up inner type to outside depends on the type of index (number or a range), and type push down phases (pushing down the overall type of a selection expression depends on the type of index which can be a range or number). 

## [Automatic type inference of range and numbers](https://github.com/golemcloud/golem/blob/implement_ranges/golem-rib/src/type_inference/index_selection_type_binding.rs)

`list[1..2]` automatically bind the type of `U64` to 1 and 2. This goes for simple index too which is `list[1]` where 1 will be of type `U64` unless user specify it using a type annotation. This auto inference is for better developer experience.

This sounds like an obvious behaviour, but that is achieved with a separate compilation phase which comes after the phase of binding the type annotations to the corresponding expressions.

## Range Implementation

Adding Range functionality fell right into the places in existing Rib code - add extra nodes, and handle each compilation phase, and add the changes into interpreter. Nothing had to be rewritten/refactored to make ranges work. :)  

```
1..10
```

```
1..=10
```

```
1..
```

As mentioned by John, it is ambiguous or impossible to parse `..`, `..x` and therefore they aren't supported.

### [Lazy Ranges ](https://github.com/golemcloud/golem/blob/implement_ranges/golem-rib/src/interpreter/rib_interpreter.rs#L2633)

Ranges are lazy description (just like rust) that you can return this value

```rust
 let x = 1..;
 x
```

Nothing will happen, it just returns the range information as a `record`.

### [Iterate Range](https://github.com/golemcloud/golem/blob/implement_ranges/golem-rib/src/interpreter/rib_interpreter.rs#L2782)

```rust
for i in 1..10 {
  yield i;
}

```

Not  that `1..10` and is never pre-computed (obviously). This was done with the help of existing iterator level functionalities at the interpreter side. We were able to reuse.

### Disallow iterating through infinite range

```rust
for i in 1.. {
  yield i
}
```

This is disallowed in Rib, with an error message at the interpreter level "make sure range is finite to avoid infinite computation". Disallowing at compiler level could be done , but looks like more difficult due to rest of the details which is not worth the time now.

### [Allow using infinite ranges within select-ranges](https://github.com/golemcloud/golem/blob/implement_ranges/golem-rib/src/interpreter/rib_interpreter.rs#L2585)

With range selection (explained) we do allow `my-list[1..]`, that it doesn't go into infinite computation as the computation is bound by the size of the list `my-list`. 

This is done with the help of proper desugar of range select expressions [to list-comprehensions. ](https://github.com/golemcloud/golem/blob/implement_ranges/golem-rib/src/compiler/desugar.rs#L19) This required new internal instructions and expression such as `length`. 

### [Range Selection](https://github.com/golemcloud/golem/blob/implement_ranges/golem-rib/src/interpreter/rib_interpreter.rs#L2536)

```rust
let x: list<u64> = [1, 2, 3]

x[1..2]

```

### Iterate with a range selection

```rust
let result = my-worker.checkout();

for i in result[2..5] {
   yield i
}

```

is infact two for-comprehensions after desugaring. 

```rust
let result = my-worker.checkout();

for i in result[2..] {
   yield i
}

```

PS: type-inference tests and interpreter-tests modules are flattened because it was very hard to identify which tests stack overflowed and which tests aren't.  So everything is in 1 module tests and no internal nesting - wasn't worth it.


jdegoes:
@afsalthaj Big improvements and I love the detailed PR writeup!

golem-rib/src/parser/rib_expr.rs
```diff
@@ -12,180 +12,1187 @@
...
+                function_name,
+                Some(type_parameter2),
+                vec![
+                    Expr::identifier_global("foo", None),
+                    Expr::identifier_global("bar", None),
+                    Expr::identifier_global("baz", None),
+                ],
+            ),
+        ]);
+        assert_eq!(expr, expected);
```
afsalthaj, line 1196: 90% of these tests are existing tests, as select_field.rs and select_index.rs files are deleted. No existing tests are deleted.

vigoo (approved):

PR #1383, opened 2025-02-28
Various fixes while investigating ifs errors

mschuwalow:
* add docker-compose setup using localstack, making it easier to reproduce s3 issues
* better logging for errors during upload to s3
* get rid of duplicated blobstorage apis in component service and reuse shared blobstorage
* remove now unused test-component due to move of golem-cli


golem-service-base/src/storage/blob/s3.rs
```diff
@@ -254,7 +253,11 @@ impl S3BlobStorage {
             SdkError::ConstructionFailure(_) => "Construction failure".to_string(),
             SdkError::TimeoutError(_) => "Timeout".to_string(),
             SdkError::DispatchFailure(inner) => {
-                format!("Dispatch failure: {}", inner.as_connector_error().unwrap())
+                // normal display of the error does not expose enough useful information
```
noise64, line 256: Please fix all error mapping to have logs. Our open issue in prod is about contruction failure.

vigoo:
The code changes look good. Other questions:
- Why did you remove the `file-server` test component?
- I understand you wanted to test with localstack and created a new docker compose setup for that. But I think this way (where is it put etc) it just even more increases confusion in this repo:
  - now we have two docker compose files in the root - `docker-compose-postgres.yaml` and `docker-compose-sqlite.yaml` that in my opinion nobody should ever use :) because they try to build the images from scratch and you can just do a `cargo make run` and run the services locally, if you are developing them.
  - We also have two docker compose files in the `docker-examples` directory that are using our published docker images. I also think those are obsolete for "local testing" because we have the single-executable golem - but makes sense to keep them as examples of how to run the containerized versions, in case someone wants to setup their own infra.
  - And now this new one in a third place in the root.

So in my opinion it would be nice to clean this docker situation up a bit

mschuwalow:
https://github.com/golemcloud/golem/pull/1383#issuecomment-2693469518

> Why did you remove the file-server test component?

This one is unused now as it's only used in the cli tests. Worker executor tests do not use it.

> I understand you wanted to test with localstack and created a new docker compose setup for that. But I think this way (where is it put etc) it just even more increases confusion in this repo:
    now we have two docker compose files in the root - docker-compose-postgres.yaml and docker-compose-sqlite.yaml that in my opinion nobody should ever use :) because they try to build the images from scratch and you can just do a cargo make run and run the services locally, if you are developing them.
    We also have two docker compose files in the docker-examples directory that are using our published docker images. I also think those are obsolete for "local testing" because we have the single-executable golem - but makes sense to keep them as examples of how to run the containerized versions, in case someone wants to setup their own infra.
    And now this new one in a third place in the root.

Makes sense. Are you fine with me moving everything to docker-examples?

vigoo (approved):

PR #1388, opened 2025-03-07
Add app & library plugins

### LINKED ISSUE #1322: Simple Component transformer plugin type
We have component transformer plugins but they require a 3rd party web server running, performing the transformation. A very common use case is to link the uploaded component with another component - we should support this built-in, by storing a WASM binary in blob storage and performing the component composition in the component service when the plugin is applied.

mschuwalow:
resolves #1322


golem-component-service-base/src/api/dto.rs
```diff
@@ -0,0 +1,210 @@
+use crate::model::plugin as local_plugin_model;
+use crate::model::plugin::PluginWasmFileReference;
+use golem_common::model::plugin as common_plugin_model;
+use golem_common::model::plugin::{PluginOwner, PluginScope};
+use golem_service_base::poem::TempFileUpload;
+use poem_openapi::Multipart;
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, poem_openapi::Union)]
```
mschuwalow, line 10: These are moved here and separated for two reasons:

* we now have different types for creation between the rest and grpc api. There is now a domain type that unifies both that they convert to in the api layer.
* We have a different representation for plugins in the rest and grpc api as I don't want to expose blob storage paths through the rest api -- they have no meaning for external users.

golem-component-service/src/api/plugin.rs
```diff
@@ -103,13 +105,71 @@ impl PluginApi {
         record.result(response)
     }
 
+    /// Registers a new library plugin
+    #[oai(
+        path = "/library-plugins/",
```
mschuwalow, line 110: New endpoints for the two new plugin types as we need to stream the wasms

golem-service-base/src/replayable_stream.rs
```diff
@@ -0,0 +1,236 @@
...
+use bytes::Bytes;
+use futures::future::BoxFuture;
+use futures::stream::BoxStream;
+use futures::Stream;
+use futures::StreamExt;
+use std::convert::Infallible;
+use std::fmt::{Debug, Display};
+use std::future::Future;
+
+pub trait ReplayableStream: Send + Sync {
```
mschuwalow, line 24: Cleaned up a bit and extracted to a more central location

golem-service-base/src/service/plugin_wasm_files.rs
```diff
@@ -0,0 +1,92 @@
...
+use crate::storage::blob::{BlobStorage, BlobStorageNamespace};
+use bytes::Bytes;
+use golem_common::model::plugin::PluginWasmFileKey;
+use golem_common::model::AccountId;
+use tracing::debug;
+
+const LIBRARY_PLUGIN_STORE_LABEL: &str = "library_plugins";
+
+/// Service for storing plugin wasm files.
+pub struct PluginWasmFilesService {
```
mschuwalow, line 27: In principle this could be separated for library and app plugins.

Alternatively, as the this is using content hashing, we could also merge it with initial component files and store everything here, but that might make debugging harder.

Want me to go for either of those options?

openapi/golem-service.yaml
```diff
@@ -3388,6 +3388,170 @@ paths:
             application/json; charset=utf-8:
               schema:
                 $ref: '#/components/schemas/ErrorBody'
+  /v1/library-plugins:
```
mschuwalow, line 3391: An alternative api would be to have a merged ComponentTransformer plugin type and have a nested type to encode the direction of the composition.

I feel this way will be easier to explain in docs / more intuitive for users though

golem-client/build.rs
```diff
@@ -3,41 +3,37 @@ use std::ffi::OsString;
 use std::path::{Path, PathBuf};
```
mschuwalow, line 3: This should now be rerun whenever the golem-service.yaml in the root changes

golem-service-base/src/service/plugin_wasm_files.rs
vigoo: I think it's ok like this

vigoo (approved):

PR #1390, opened 2025-03-07
Cleanup docker tests, simplify db tests and reduce flakiness

mschuwalow:


afsalthaj:
would you mind how this change reduced flakiness?

It looks it was intentionally added tests to use multiple DB instances in parallel, to test if it is working properly. Aren't they?
With these changes it is reduced to only one DB instance. Am I right? So flakiness is reduced here by reduced test coverage?

mschuwalow:
Couple of points:

* Increased connection pool size for tests

* Docker{Postgres,MySql}Rdbs attempt to start many containers in parallel, failing in ci quite consistently https://github.com/golemcloud/golem/actions/runs/13705034106/job/38391598184. Doing the startups sequentially makes it more stable, but increases runtime of the tests.

* https://github.com/golemcloud/golem/blob/main/golem-worker-executor-base/src/services/rdbms/tests.rs#L2619
These guys spawn tests in parallel, whichc is already handled as part of the test-framework / runner. Having two levels of parallelism is unnecessary and also makes that test more likely to timeout (as the test framework cannot see what is going on in there), especially when running with a low thread count.

* Spawning multiple posgres / mysql instances makes the test runs much heavier and our test code more complex. I don't see a good reason for doing it. Overall performance is going to be decreased as a single instance can easily handle whichever load we are producing during the tests, more is just going to make things slower.
Testing being connected to multiple database instances is fine, but if we want to have that test we should still do that against a single db instance using multiple independent pools / aliases for the same instance.

* Running these as independent statements instead of as part of a transaction is flaky, as  sometimes changes are not yet visible https://github.com/golemcloud/golem/blob/main/golem-worker-executor-base/src/services/rdbms/tests.rs#L2619.
I actually do not know exactly what caues this. Will look into it and try to articulate what is going on here. I'm not sure what postgres guarantees regarding stale reads across connections

golem-test-framework/src/components/component_service/docker.rs
```diff
@@ -43,7 +42,6 @@ pub struct DockerComponentService {
 }
 
 impl DockerComponentService {
-    const NAME: &'static str = "golem_component_service";
```
mschuwalow, line 46: As much as possible I changed these to work with dynamic names, so there is less risk of name conflicts

golem-test-framework/src/components/component_service/docker.rs
```diff
@@ -52,62 +50,59 @@ impl DockerComponentService {
         component_compilation_service: Option<(&str, u16)>,
         rdb: Arc<dyn Rdb + Send + Sync + 'static>,
         verbosity: Level,
-        keep_container: bool,
```
mschuwalow, line 55: This is not needed now. Containers are automatically removed when dropped and start / stop / restart were refactored

golem-test-framework/src/components/rdb/docker_postgres.rs
```diff
@@ -40,57 +32,36 @@ impl DockerPostgresRdb {
...
+        info!("Starting Postgres container");
+
         let database = Self::DEFAULT_DATABASE;
         let password = Self::DEFAULT_PASSWORD;
         let username = Self::DEFAULT_USERNAME;
+        let port = Self::DEFAULT_PORT;
 
-        let name = "golem_postgres";
-        let image = testcontainers_modules::postgres::Postgres::default()
-            .with_tag("12")
```
mschuwalow, line 53: 12 is end of lived since November of last year and the caches in blacksmith seem to not include it, making the pulls very slow

PR #1397, opened 2025-03-09
rdbms parallel test with using multiple databases

justcoon:
related to: #1390

### DIFF
--- a/golem-test-framework/src/components/rdb/docker_mysql.rs
+++ b/golem-test-framework/src/components/rdb/docker_mysql.rs
@@ -28,7 +28,7 @@ pub struct DockerMysqlRdb {
 
 impl DockerMysqlRdb {
     const DEFAULT_PORT: u16 = 3306;
-    const DEFAULT_USERNAME: &'static str = "mysql";
+    const DEFAULT_USERNAME: &'static str = "root";
     const DEFAULT_PASSWORD: &'static str = "mysql";
     const DEFAULT_DATABASE: &'static str = "mysql";
 
@@ -42,8 +42,7 @@ impl DockerMysqlRdb {
 
         let container = testcontainers_modules::mysql::Mysql::default()
             .with_tag("8")
-            .with_env_var("MYSQL_PASSWORD", password)
-            .with_env_var("MYSQL_USER", username)
+            .with_env_var("MYSQL_ROOT_PASSWORD", password)
             .with_env_var("MYSQL_DATABASE", database)
             .with_network(NETWORK)
             .start()
@@ -77,6 +76,15 @@ impl DockerMysqlRdb {
 
     pub fn public_connection_string(&self) -> String {
         self.info.public_connection_string()
+    }
+
+    pub fn public_connection_string_to_db(&self, db_name: &str) -> String {
+        let db_info = MysqlInfo {
+            database_name: db_name.to_string(),
+            ..self.info.clone()
+        };
+
+        db_info.public_connection_string()
     }
 
     pub fn private_connection_string(&self) -> String {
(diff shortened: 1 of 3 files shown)


golem-test-framework/src/components/rdb/docker_mysql.rs
```diff
@@ -79,6 +78,15 @@ impl DockerMysqlRdb {
         self.info.public_connection_string()
     }
 
+    pub fn public_connection_string_to_db(&self, db_name: &str) -> String {
```
mschuwalow, line 81: This is just Self#public_connection_string

golem-worker-executor-base/src/services/rdbms/tests.rs
```diff
@@ -2846,6 +2847,147 @@ fn test_rdbms_pool_key_masked_address() {
     check!(key.masked_address() == "postgres://user:*****@localhost:5432?abc=xyz&secret=*****");
 }
 
+#[test]
+async fn mysql_par_test(mysql: &DockerMysqlRdb, rdbms_service: &RdbmsServiceDefault) {
+    let db_address = mysql.public_connection_string();
+    let rdbms = rdbms_service.mysql();
+    let mut db_addresses = create_test_databases(rdbms.clone(), &db_address, 3, |db_name| {
```
mschuwalow, line 2854: As this is using the same db address, the pool will not create independent connections.
Instead, it will reuse the same connection for all 3 of the tests.

What exactly do you want to test here?
* That multiple workers can be connected to the same database? If yes, this is already implicitly tested through the various tests in this module getting executed concurrently. I would not say a dedicated test for this is unneccessary.

* That multiple workers can be connecto to multiple databases? If yes, we need to have different connection strings for them to avoid sharing in the pool. The way I would propose this is to start the same container using multiple aliases, i.e.
`docker create --network-alias db_1 --network-alias db_2 ...` and run tests against those.

mschuwalow (commented):
Thanks for doing the followup. I originally removed the test in my pr to stabilize / simplify our tests. But if we can have it without too much overhead, let's absolutely do it.

If you are looking at this anyway. There is one more issue with the test suite I was struggling to fix:
test_containers has a [hardcoded timeout](https://github.com/testcontainers/testcontainers-rs/blob/83e46894f80713fe3b257f40694a9d89f4d45469/testcontainers/src/core/client/bollard_client.rs#L8
) of 2 minutes when starting the container For some reason postgres is sometimes [timing out / failing](https://github.com/golemcloud/golem/actions/runs/13747754881/job/38483348387?pr=1388#step:13:1317) this one.

I don't think this happened before the rdms / blacksmith prs, but I don't see what caused it.

golem-worker-executor-base/src/services/rdbms/tests.rs
justcoon: probably variable names are not best,

but those are db connections urls  

```
"mysql://root:mysql@localhost:55000/test_db_0", "mysql://root:mysql@localhost:55000/test_db_1", "mysql://root:mysql@localhost:55000/test_db_2", "mysql://root:mysql@localhost:55000/mysql"
```
pools are created for url, which may have connection to specific underlying database, with specific user, or additional custom attributes

test are aiming to test if `RdbmsServiceDefault` working properly with multiple pools accessing by multiple workers

golem-test-framework/src/components/rdb/docker_mysql.rs
justcoon: not sure what you mean by that, 
it is not same `public_connection_string` (which is connecting to default database),
this is helper function which allowing to create connection string to different underlying database

mschuwalow: ah you are right, missed that part

mschuwalow (approved):

golem-worker-executor-base/src/services/rdbms/tests.rs
mschuwalow: missed the db_name part in the connection strings.

This looks nice then, lgtm

PR #1403, opened 2025-03-11
Check oplog processor plugins implement the proper interface

### LINKED ISSUE #1261: Check if an oplog processor plugin implements the oplog processor interface when defining it

mschuwalow:
resolves #1261

### DIFF
--- a/golem-component-service-base/src/service/component.rs
+++ b/golem-component-service-base/src/service/component.rs
@@ -63,6 +63,7 @@ use std::vec;
 use tap::TapFallible;
 use tempfile::NamedTempFile;
 use tokio::io::BufReader;
+use tokio::sync::RwLock;
 use tokio_stream::Stream;
 use tokio_util::compat::FuturesAsyncReadCompatExt;
 use tokio_util::io::ReaderStream;
@@ -309,7 +310,7 @@ pub fn create_new_versioned_component_id(component_id: &ComponentId) -> Versione
 }
 
 #[async_trait]
-pub trait ComponentService<Owner: ComponentOwner>: Debug {
+pub trait ComponentService<Owner: ComponentOwner>: Debug + Send + Sync {
     async fn create(
         &self,
         component_id: &ComponentId,
@@ -449,11 +450,11 @@ pub trait ComponentService<Owner: ComponentOwner>: Debug {
 }
 
 pub struct ComponentServiceDefault<Owner: ComponentOwner, Scope: PluginScope> {
-    component_repo: Arc<dyn ComponentRepo<Owner> + Sync + Send>,
-    object_store: Arc<dyn ComponentObjectStore + Sync + Send>,
-    component_compilation: Arc<dyn ComponentCompilationService + Sync + Send>,
+    component_repo: Arc<dyn ComponentRepo<Owner>>,
+    object_store: Arc<dyn ComponentObjectStore>,
+    component_compilation: Arc<dyn ComponentCompilationService>,
     initial_component_files_service: Arc<InitialComponentFilesService>,
-    plugin_service: Arc<dyn PluginService<Owner::PluginOwner, Scope> + Sync + Send>,
+    plugin_service: Arc<dyn PluginService<Owner::PluginOwner, Scope>>,
     plugin_wasm_files_service: Arc<PluginWasmFilesService>,
 }
 
@@ -465,12 +466,12 @@ impl<Owner: ComponentOwner, Scope: PluginScope> Debug for ComponentServiceDefaul
 
 impl<Owner: ComponentOwner, Scope: PluginScope> ComponentServiceDefault<Owner, Scope> {
     pub fn new(
-        component_repo: Arc<dyn ComponentRepo<Owner> + Sync + Send>,
-        object_store: Arc<dyn ComponentObjectStore + Sync + Send>,
-        component_compilation: Arc<dyn ComponentCompilationService + Sync + Send>,
+        component_repo: Arc<dyn ComponentRepo<Owner>>,
+        object_store: Arc<dyn ComponentObjectStore>,
+        component_compilation: Arc<dyn ComponentCompilationService>,
         initial_component_files_service: Arc<InitialComponentFilesService>,
-        plugin_service: Arc<dyn PluginService<Owner::PluginOwner, Scope> + Sync + Send>,
-        library_plugin_files_service: Arc<PluginWasmFilesService>,
+        plugin_service: Arc<dyn PluginService<Owner::PluginOwner, Scope>>,
+        plugin_wasm_files_service: Arc<PluginWasmFilesService>,
     ) -> Self {
         ComponentServiceDefault {
             component_repo,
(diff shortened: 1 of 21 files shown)


golem-component-service-base/src/service/plugin.rs
```diff
@@ -208,28 +229,79 @@ impl<Owner: PluginOwner, Scope: PluginScope> PluginServiceDefault<Owner, Scope>
             }
         }
     }
+
+    async fn check_oplog_processor_plugin(
```
mschuwalow, line 233: This is the main thing implementing the ticket description

golem-component-service-base/src/service/plugin_installation.rs
```diff
@@ -0,0 +1,547 @@
...
+use golem_service_base::repo::plugin_installation::PluginInstallationRecord;
+use golem_service_base::service::plugin_wasm_files::PluginWasmFilesService;
+use reqwest::multipart::{Form, Part};
+use tracing::info;
+use tracing::info_span;
+use tracing_futures::Instrument;
+use wac_graph::CompositionGraph;
+
+#[derive(Debug)]
+pub struct PluginInstallationService<Owner: ComponentOwner, Scope> {
```
mschuwalow, line 43: We have a cyclical wiring issue here (component service used plugin service and for the new function the plugin service would use the component service). There were 3 options for fixing:

* Put everything into a single trait
* Use some kind of Arc<Mutex>> structure to support the cyclical wiring and replace the implemenation at the end.
* Make the graph non-cyclical

I went for option 3 by extracting plugin installation related operations and adjusting the component service to accept a function for transforming the plugins. This function will end up getting bound to this guy, tying the know

golem-component-service-base/src/service/component.rs
```diff
@@ -320,10 +328,40 @@ pub trait ComponentService<Owner: ComponentOwner>: Debug {
         installed_plugins: Vec<PluginInstallation>,
         dynamic_linking: HashMap<String, DynamicLinkedInstance>,
         owner: &Owner,
-    ) -> Result<Component<Owner>, ComponentError>;
+        apply_transformations: impl AsyncFnOnce(
```
mschuwalow, line 331: This guy is new and is used to break the cycle with the plugin service.
Converted the trait to a struct to be able to use AsyncFn here

golem-component-service-base/src/service/plugin_installation.rs
```diff
@@ -0,0 +1,547 @@
...
+    component_repo: Arc<dyn ComponentRepo<Owner> + Sync + Send>,
+    object_store: Arc<dyn ComponentObjectStore + Sync + Send>,
+    plugin_service: Arc<dyn PluginService<Owner::PluginOwner, Scope> + Sync + Send>,
```
mschuwalow, line 46: This guy could be converted to a struct as well, I will go for it before merging

golem-test-framework/src/components/component_compilation_service/docker.rs
```diff
@@ -39,13 +39,6 @@ impl DockerComponentCompilationService {
     pub async fn new(
```
mschuwalow, line 39: unrelated cleanups here

golem-component-service-base/src/service/plugin_installation.rs
vigoo: I solved a similar problem in the worker executor with the 2nd option. (`LazyWorkerActivator`). 

Having to pass an arbitrary function to things like `ComponentService::create` everywhere (which _has_ to be the same call to the plugin installation service, by the way, but nothing enforces this) is very much "taste" of what good code should be. This leaks the specifics of what it means to create a component by having to know it in each usage point. 

It would be a bit nicer if it is not a function parameter but a specific trait for applying transformations, for what you can then look for its implementations and find that the plugin installation service it is. That still breaks the cycle but makes it much readable in my opinion.

I also feel that doing it the 2nd way would have been much less work, you would not have been forced to delete the service traits (which we debated about on Slack and did not get into a consensus) and it would have been similar to what we have already in the executor.

golem-test-framework/src/components/component_compilation_service/docker.rs
vigoo: Why did you change this / what was the two kind of constructors used before?

golem-test-framework/src/components/worker_executor/docker.rs
```diff
@@ -48,25 +48,6 @@ impl DockerWorkerExecutor {
         worker_service: Arc<dyn WorkerService + Send + Sync + 'static>,
         verbosity: Level,
         shared_client: bool,
-    ) -> Self {
```
noise64, line 51: did you check how cloud breaks without these? AFAIR these were introduced for that

golem-component-service-base/src/service/plugin_installation.rs
noise64: agree with @vigoo, we should not use a function for something that has only one implementation and for something that should not be customizable. all of this seems to be a quest for justifying using AsyncFn-s when we can just keep using patterns as they are, or solve it by the usual "extract another modul", in this case e.g. a service.

golem-test-framework/src/components/component_compilation_service/docker.rs
mschuwalow: Just a leftover from another pr. They used to be used to set a default for a specific kind of env vars, but I got rid of it entirely as there is only a single implementation across both oss and cloud https://github.com/golemcloud/golem/pull/1390/files#diff-defcdcc5cbd9c47aee747607c42d64036dea791683cfa67c00448bb1b9aea0d5L51

Just forgot to delete these now redundant constructors

golem-component-service-base/src/service/plugin_installation.rs
mschuwalow: I can go for the LazyWorkerActivator approach if you prefer.

Regarding having to pass that function everywhere. I can also introduce a service sitting on top of the component and plugin_installation_service that just wires this call. I don't think we have that many usesite, so I didn't go for it to lower overall amount of code.

Want me to do either of those?

vigoo: I would prefer the following, which I think is the least amount of code change and also keeping the interfaces clean:
- Undo all the unnecessary changes to these services (deletion of the traits, split up)
- Add the component-service dependency to the plugin service as an `Arc<Mutex<Option` and set it at the end of wiring code.

Worrying about the heavyness of that mutex does not make sense in this situation because adding/updating a component is not on the hot path at all, and it involves steps like storing multiple megabytes of data on S3, parsing the WASM and even linking it, calling a series of external component transformer services, etc. Having an extra lock and dereferencing does not change anything. It could be an `RwLock` to make it nicer and _if_ it would be on the hot path it could be some unsafe delayed-initialized field but it is not so let's not do that.

mschuwalow: Ok, will update

mschuwalow: done

vigoo (approved):

Issue #1103, opened 2024-12-02
Clean up tasks: Rib


afsalthaj:
- [x] A bit more user friendly error message for block expressions such as "the last line in a block cannot end with `;` . Currently `expected ; at position x` is slightly polluted with more messages.
- [x] A few re-namings (review code)
- [x]  More steps in type check module to not leak wrong code into unification. There was about 2000+ lines of code addition during the type-check module addition and that covers a lot. However, it is still not good enough! 
- [x]  Make sure math operations have flexibility with types. The current type inference is very strict. `let z: u64 = 1u8+ 2u32`. Currently type inference fails if a variable is found to have `AllOf(U64, U8, U32)`, which is the case in the above example - given our standard type inference machinery. Probably the best way to overcome this is to consider resolve initial value's type with yield type during push down phase. It is during push down the initial value type is pushed down to the yield type. If yield type is OneOf(u64, u32) but initial value is u8, then we cast down yield type to the initial value. If number overflows then that's runtime exception obviously.
- [x]  Add more tests into first class worker support, and migrate any remaining tests to first-class-worker-support

afsalthaj:
<img width="870" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM1ZEQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--bbf931592b2df77176f2137b32834f4fc58d49d3/image.png" alt="image.png" />


Above issues were resolved. The remaining will be separate tickets

PR #1404, opened 2025-03-12
Improve/Cleanup Rib: Improve type inference for binary math op

afsalthaj:
Resolves part of #1103 
Binary math operations were not inferrred properly. This PR fixes it


afsalthaj:
Just realised I need to handle div. Will fix it in a while

afsalthaj:
This will also help pushing down the types properly - just like in Rust allowing

PR #1405, opened 2025-03-12
Better test coverage for plugins

### LINKED ISSUE #1108: Better test coverage for plugins

mschuwalow:
resolves #1108


golem-component-service-base/src/service/transformer_plugin_caller.rs
```diff
@@ -0,0 +1,168 @@
...
+
+impl SafeDisplay for TransformationFailedReason {
+    fn to_safe_string(&self) -> String {
+        self.to_string()
+    }
+}
+
+#[derive(Debug)]
+pub struct TransformerPluginCallerDefault {
+    client: reqwest::Client,
```
mschuwalow, line 62: The client maintains an internal connection pool per host. Not recreating it per plugin call should resolve this todo:
https://github.com/golemcloud/golem/pull/1405/files#diff-f4ae9ef1c03598ea0766776182ed99fa41f5b79f88a02fcfd50e4e804f03f3c0L924

golem-test-framework/src/components/component_service/mod.rs
```diff
@@ -85,20 +86,17 @@ pub enum PluginServiceClient {
 }
 
 #[async_trait]
-pub trait ComponentService {
-    fn client_protocol(&self) -> GolemClientProtocol;
+trait ComponentServiceInternal {
```
mschuwalow, line 89: These are only used to implement the defaults in the main trait, so there is no reason to expose them to other parts of the code

golem-test-framework/src/components/component_service/mod.rs
```diff
@@ -85,20 +86,17 @@ pub enum PluginServiceClient {
...
     fn component_client(&self) -> ComponentServiceClient;
     fn plugin_client(&self) -> PluginServiceClient;
+    fn plugin_wasm_files_service(&self) -> Arc<PluginWasmFilesService>;
+}
 
+#[allow(private_bounds)]
```
mschuwalow, line 95: This is just here because I didn't make ComponentServiceInternal public.
This is perfectly fine, but as the name of ComponentServiceInternal is not exported, nobody outside this crate can implement the trait. I think this is desirable behaviour, but if we want to allow outside implementations we just need to export the internal trait

test-components/app-and-library/Makefile.toml
```diff
@@ -14,5 +14,5 @@ dependencies = ["build"]
 condition = { env_set = ["CARGO_MAKE_CRATE_CURRENT_WORKSPACE_MEMBER"] }
 script_runner = "@duckscript"
 script = """
-cp ../target/wasm32-wasip1/release/app_and_library_plugin_${CARGO_MAKE_CRATE_CURRENT_WORKSPACE_MEMBER}.wasm ../../app_and_library_plugin_${CARGO_MAKE_CRATE_CURRENT_WORKSPACE_MEMBER}.wasm
+cp ../target/wasm32-wasip1/release/app_and_library_${CARGO_MAKE_CRATE_CURRENT_WORKSPACE_MEMBER}.wasm ../../app_and_library_${CARGO_MAKE_CRATE_CURRENT_WORKSPACE_MEMBER}.wasm
```
mschuwalow, line 17: Just some renaming as I'm not only using it for plugin tests now.

golem-component-service-base/src/service/component.rs
```diff
@@ -920,94 +898,10 @@ impl<Owner: ComponentOwner, Scope: PluginScope> ComponentServiceDefault<Owner, S
...
-            Ok(body.to_vec())
-        } else {
-            Err(ComponentError::TransformationFailed(
-                TransformationFailedReason::HttpStatus(response.status()),
-            ))
-        }
+        self.transformer_plugin_caller
+            .call_remote_transformer_plugin(component, data, url, parameters)
+            .await
+            .map_err(ComponentError::TransformationFailed)
```
mschuwalow, line 904: Decided to not go with 502 here, as it would be ambiguous with alb errors. Added a check that at least asserts that we have an error message exposing the downstream failure.

golem-test-framework/src/components/component_service/mod.rs
mschuwalow, line 95: This is only here because I didn't make ComponentServiceInternal.
This is perfectly fine, but means that nobody outside the crate can implement this trait.
If we want to allow outside implementations, I can make it public.

vigoo (approved):

PR #1407, opened 2025-03-12
Improve wall duration of ci run

mschuwalow:


mschuwalow:
Some job names changed, to we'll need to update the required checks if we merge this

vigoo (approved):

PR #1408, opened 2025-03-12
Support for the download file API in the component

0xvasanth:
- Added API support to the download file from component
    - `{{baseUrl}}/v1/components/:component_id/versions/:version/file-contents/:file_path - "/Epifi.pdf"`
- Fixed the Name of the `InitialComponetFile` to `ComponentFile` since that make more sense for the struct

### DIFF
--- a/golem-service-base/src/service/initial_component_files.rs
+++ b/golem-service-base/src/service/initial_component_files.rs
@@ -21,7 +21,7 @@ use futures::stream::BoxStream;
 use golem_common::model::{AccountId, InitialComponentFileKey};
 use tracing::debug;
 
-const COMPONENT_FILES_LABEL: &str = "initial_component_files";
+const INITIAL_COMPONENT_FILES_LABEL: &str = "initial_component_files";
 
 /// Service for storing initial component files.
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


golem-component-service-base/src/service/component.rs
```diff
@@ -1343,6 +1367,52 @@ impl<Owner: ComponentOwner, Scope: PluginScope> ComponentService<Owner>
...
+                .get(&owner.account_id(), &file.key)
+                .await
+                .map_err(|e| {
+                    ComponentError::initial_component_file_upload_error(
+                        "Failed to get component file",
+                        e,
+                    )
+                })?
+                .ok_or(ComponentError::FailedToDownloadFile)?;
+            let mapped_stream = stream.map_err(|e| {
```
afsalthaj, line 1406: Very minor. I would just inline this with the above rather than calling it a mapped_stream. It's just mapping error.

vigoo:
I would not change the names - this feature is called "initial file system", as they are the **initial** files a worker gets when it starts up. This is how we always referred to it in our docs and release notes etc.

golem-component-service/src/api/component.rs
```diff
@@ -508,6 +508,43 @@ impl ComponentApi {
         record.result(response)
     }
 
+    /// Download file in a Component
+    #[oai(
+        path = "/:component_id/versions/:version/file-contents/:file_key",
```
mschuwalow, line 513:
```suggestion
        path = "/:component_id/versions/:version/file-contents/:file_path",
```

mschuwalow (approved):

PR #1409, opened 2025-03-13
Better type inference

afsalthaj:
As part of #1103 

All of the changes are geared to improving DX of Rib and bug fixes except for 1 simple renaming of a variable (inferred_type_stack is replaced with inferred_expr_stack)  in type_pull_up.rs file, which is done, because of it was confusing. The same file has some important changes too, so still worth reviewing.

## Better type inference

Before this PR the following code doesn't work unless those numbers are type annotated as `u64`. Infact, there is a subtle behaviour , that you can annotate the operands to be `u8`,  and annotate `z` to be `u64`, while the worker function `foo` takes a `u64`. This won't happen again.

```rust
let worker = instance("my-worker");
let z = 1 + 2;
worker.foo(z)

```
**With this PR , the above code works,** because it knows what the type of `z` is and propagates all the way to the operands. This time, inspired from languages like Rust, we make the types of operands strict. In the above code `foo` takes a u64, and hence the operands of `z` is also u64 - quite simple on a high level.

Before this PR, push down for math operations were disallowed for binary ops, since math operations were allowed to work on different types of operands. This confuses the users. Better to have a proper compilation error, than surprises. 

What this also means is, you will get a compilation error, if you do `let z = 1: u8 + 2`, because now `1` has conflicting types. This is the same behaviour in Rust too. I think that's the correct way. Obviously we don't have a proper "cast" method that a developer of Rib can use. But that doesn't mean we have to compromise the type safety of Rib internally for the upcoming release. We already have `type annotation` in Rib that unblocks user to a great extent for any strict compilation error.

## Easier type inference for result types

Before this PR the following code fails asking users to put a type to those numbers (200, the two 400s). @vigoo  when using Rib, asked this question as well.

```
  let input = { request : { path : { user : "jak" } }, y : "baz" };
  let result = my-worker-function(input);
   
   
   match result {
     ok(result) => { body: result, status: 200 },
     err(result) => { status: 400, body: 400 }
 }
 ```

With this PR, it just works.

For those are looking for details: This happened obviously because strict typing - i.e, which `ValueAndType` should the interpreter create at runtime when someone says `200` within  a wasm record. Well, we could pick one out of `u64, u32 etc` but it produced more unreliable behaviour. 

But this is also annoying to the users, because it's hard to justify even if there is a technical reason. So to improve DX, i have added a new [compilation phase ](https://github.com/golemcloud/golem/blob/better_literal_inference/golem-rib/src/type_inference/infer_orphan_literals.rs) inferring these orphan literals. Firstly, the phase is really safe, that it doesn't hurt any other phase. Meaning, if you remove this , all that's going to happen is users having to type annotate. This phase should improve the DX with Rib atleast to some extent. Similar to Rust picking `i32` for any numbers that are not typed, Rib picks a type based on the sign and fractional digit count of the numbers - so its' not i32, but one of u64, s64 or F64.

## Migration of any remaining gateway tests to use first worker support

Migrated any tests that were using legacy API definitions.This is an [example](https://github.com/golemcloud/golem/pull/1409/commits/f3696b79732135177574505d317d59b1ab5c2354) commit. 

This uncovered another subtle bug in compilation (which essentially result in user having to type annotate which can be annoying) and that's fixed.

###Bug Fixes
Please see this commit as a[ bug fix](https://github.com/golemcloud/golem/pull/1409/commits/96f87a193feb5d5c643c8e34866629c10e6083f6), where `request.path.user-id` and `request.body.user-id` was given together, it failed to compile due to global variable type binding phase being wrong. Note that, it was an existing test that failed when migrated to first class worker support.  
It's a hot fix - the minimal reliable change required to fix the bug.

### CODE REVIEW

File: golem-rib/src/type_inference/type_push_down.rs [hunk 1]
```diff
@@ -213,6 +213,20 @@ pub fn push_types_down(expr: &mut Expr) -> Result<(), RibCompilationError> {
...
+                inferred_type,
+                ..
+            } => {
+                lhs.add_infer_type_mut(inferred_type.clone());
+                rhs.add_infer_type_mut(inferred_type.clone());
+                queue.push_back(lhs);
+                queue.push_back(rhs);
+            }
+
+            // Note that binary math operations cannot be pushed down. There is no way to properly
```
afsalthaj, line 228: This is wrong. Fixing it.

golem-rib/src/type_inference/infer_orphan_literals.rs
```diff
@@ -0,0 +1,145 @@
...
+// This is more of an optional stage to help with a better
+// DX for Rib users, while not affecting the type inference reliability.
+// A standalone expression is something that doesn't influence push down or pull up phases, or
+// doesn't require further scanning but may require a final pull up.
+// This is not a risk because there is no way it can influence
+// the type of the rest of the rib program as they are standing alone.
+// Mostly this simply helps with the inference of the return value of a Rib script.
+// If there is any possible mistake in assigning types, type_checker phase
+// or unification phase will capture it.
+// However, this phase may not be perfectly assigning a reasonable type to all types of literals in the program
```
afsalthaj, line 13: Only some parts of the comment makes sense now

golem-rib/src/type_inference/mod.rs
```diff
@@ -259,13 +262,6 @@ mod type_inference_tests {
 
     #[test]
     fn test_inference_inline_type_annotation_7() {
-        // ok(1) isn't automatically inferred since no further hints are given
-        let mut invalid_rib_expr = Expr::from_text(r#"ok(1)"#).unwrap();
-
-        let result = invalid_rib_expr.infer_types(&FunctionTypeRegistry::empty(), &vec![]);
-
-        assert!(result.is_err());
```
afsalthaj, line 267: These will get inferred with this change. They are not compilation errors anymore.

PR #1411, opened 2025-03-13
Various flaky test fixes

mschuwalow:


golem-api-grpc/src/lib.rs
```diff
@@ -218,23 +218,9 @@ pub mod proto {
...
     #[cfg(test)]
     mod tests {
-        use test_r::test;
-
         use crate::proto::golem;
         use prost::Message;
-        use std::str::FromStr;
-
-        #[test]
-        fn test_uuid() {
```
mschuwalow, line 228: No assertion here

golem-common/src/model/mod.rs
```diff
@@ -2410,7 +2410,7 @@ mod tests {
             let start = SystemTime::now();
             let worker_id = target_worker_id.into_worker_id(&shard_ids, SHARD_COUNT);
             let end = SystemTime::now();
-            println!(
```
mschuwalow, line 2413: cleaned up our logs a bit and got rid of all printlns

golem-component-service-base/tests/tests.rs
```diff
@@ -26,9 +26,10 @@ pub struct Tracing;
 
 impl Tracing {
     pub fn init() -> Self {
-        init_tracing_with_default_debug_env_filter(&TracingConfig::test_pretty_without_time(
-            "component-service-base-tests",
-        ));
+        init_tracing_with_default_debug_env_filter(
+            &TracingConfig::test_pretty_without_time("component-service-base-tests")
+                .with_env_overrides(),
```
mschuwalow, line 31: I want to be able to override test logs to go to a file for detailed debugging

golem-test-framework/src/components/worker_executor_cluster/docker.rs
```diff
@@ -65,14 +65,17 @@ impl DockerWorkerExecutorCluster {
...
+            let worker_executor_join = tokio::spawn(
+                Self::make_worker_executor(
+                    redis.clone(),
+                    component_service.clone(),
+                    shard_manager.clone(),
+                    worker_service.clone(),
+                    verbosity,
+                    shared_client,
+                )
+                .in_current_span(),
```
mschuwalow, line 77: better propagation of spans

golem-worker-executor-base/src/lib.rs
```diff
@@ -160,7 +160,7 @@ pub trait Bootstrap<Ctx: WorkerCtx> {
             .in_current_span(),
         );
 
-        Ok(())
+        Ok(grpc_port)
```
mschuwalow, line 163: so tests work without static ports

golem-worker-executor-base/tests/api.rs
```diff
@@ -1745,26 +1765,25 @@ async fn long_running_poll_loop_works_as_expected(
...
-            format!("0.0.0.0:{}", host_http_port)
-                .parse::<SocketAddr>()
-                .unwrap(),
-        )
-        .await
-        .unwrap();
-        axum::serve(listener, route).await.unwrap();
-    });
+
+    let listener = tokio::net::TcpListener::bind("0.0.0.0:0").await.unwrap();
```
mschuwalow, line 1769: Changed all of these to work without fixed ports. Given the number of tests we have, we had to keep quite a large range of ports free

golem-worker-executor-base/tests/api.rs
```diff
@@ -1887,32 +1907,42 @@ async fn long_running_poll_loop_interrupting_and_resuming_by_second_invocation(
...
         .await;
 
+    let mut rx = executor.capture_output_with_termination(&worker_id).await;
+
     {
         let mut response = response.lock().unwrap();
         *response = "first".to_string();
     }
 
-    fiber.await.unwrap();
```
mschuwalow, line 1913: This is completed when the second invocation is enqueued. If we had bad luck the first invocation was not completed at that time, causing the test to timeout. 

Looking for the lock message of the first poll loop being completed should be more reliable and work without sleeps

golem-worker-executor-base/tests/api.rs
```diff
@@ -2156,29 +2183,51 @@ async fn long_running_poll_loop_connection_can_be_restored_after_resume(
...
+        }
+    }
+
+    // change the response. Next loop will be the last
     {
         let mut response = response.lock().unwrap();
         *response = "first".to_string();
     }
 
-    let start = Instant::now();
```
mschuwalow, line 2164: If we had bad luck the interleaving was like this, causing the test to timeout

log("Calling the poll endpoint\n") *> connect *> change_response *> log("Received first\n") *> log("Poll loop finished\n")

golem-worker-executor-base/src/services/rdbms/sqlx_common.rs
```diff
@@ -712,7 +712,7 @@ where
         let mut tx_conn = self.tx_connection.0.lock().await;
 
         if tx_conn.open {
-            DB::TransactionManager::start_rollback(&mut tx_conn.connection);
+            let _ = DB::TransactionManager::rollback(&mut tx_conn.connection).await;
```
mschuwalow, line 715: This was the reason for the flakiness in the sql tests I reported a few days ago.

What happens here is the rollback is started asynchronously on the connection, but it is getting returned to the pool.
What can happen then is that somebody else picks up the connection from the pool and starts running work on it. The rollback is then interleaved with the new work, causing various weird failures.

Waiting for the the rollback to complete when dropping the connection should fix this.

@justcoon am I missing something with this one?

golem-api-grpc/src/lib.rs
afsalthaj: Should we put assertions or delete this? :)

golem-common/src/serialization.rs
```diff
@@ -119,7 +118,7 @@ mod tests {
     #[test]
     pub fn roundtrip_component_id() {
         let example = Some(ComponentId::new_v4());
-        println!("example: {example:?}");
+        info!("example: {example:?}");
```
afsalthaj, line 121: Should we have it at all? If there is a bug, someone can put `info`? Or we have a standard that some complex tests have logs?

golem-worker-executor-base/src/services/rdbms/sqlx_common.rs
vigoo: nice catch

golem-worker-executor-base/tests/api.rs
vigoo: Ah this whole test must have been written _before_ we had persistent invocation queues, based on the "// Invoke blocks until the invocation starts" comment. Invoke no longer does this - previously it was blocking on a mutex until the execution actually started.

I've reviewed these tests many times and my eyes always skipped this comment

vigoo (approved):
Very nice, thanks!

PR #1418, opened 2025-03-14
Include openapi yaml in client crate

mschuwalow:


vigoo (approved):

vigoo (changes requested):
I think this will break publishing, because it refers to something outside of the crate root. That's why for example in wasm-rpc the WITs included are copied (duplicated) by a cargo make step

mschuwalow:
This is the one within the crate.
Include bytes takes a relative path https://doc.rust-lang.org/std/macro.include_bytes.html

Issue #1415, opened 2025-03-14
Extend complex rib test cases, fix scoping  and add necessary features


afsalthaj:
- [x] Variable scoping - a small revisit- we never had variable scope bugs and therefore I never had to look at this part of the code. However, with more new features such as list comprehension, its good to tackle this properly with added tests under variable_scope module - handled in #1424 
- [ ]  Extend complex_rib_test. Tests take time. Currently there are 500 required tests which covers a lot. But good to have a test coverage setup
- [ ] Make sure list flatmap is enabled

PR #1424, opened 2025-03-16
Fix any ambiguous variable related bugs and other necessary improvements for 1.2

afsalthaj:
Trying to ensure any problems with ambiguous identifiers, fix bugs if any around identifier inference.

This is a fix that partly solves https://github.com/golemcloud/golem/issues/1415 , which is an issue that got separated from the already closed issue: https://github.com/golemcloud/golem/issues/1103 (which already handled quite a lot)

This is not a clean up PR, rather making sure code such as the following (which has many _probably_ ambiguous varaibles) works, which didn't work before.

```rust
               let x: u64 = 1;
               let x = x;

               let result1 = match some(x + 1:u64) {
                  some(x) => match some(x + 1:u64) {
                     some(x) => x,
                     none => x
                  },
                  none => x
               };

               let z: option<u64> = none;

               let result2 = match z {
                  some(x) => x,
                  none => match some(x + 1:u64) {
                     some(x) => x,
                     none => x
                  }
               };

               { result1: result1, result2: result2 }

```

One of the key algorithms floating around is top down and bottom up traversals of Expr. Obviously the way it is done now, is to write loops the way it is required for different phases, and sometimes even a combination of visitors and queues to work with immutable exprs. With this PR, first of all we are making it structured. The reason for this is, being able to enqueue mutable references to all nodes in a tree (something that rust doesn't like inherently - hence unsafe code, but collated to just 1 module with a struct [ExprVisitor](https://github.com/golemcloud/golem/blob/fix_scoping/golem-rib/src/type_inference/expr_visitor.rs#L10)  fixed scoping issues that fully enqueued mutable queues keep the order of expressions better / correct. THis will also reduce huge clones for immutable expressions (the use of immutable expressions came in to avoid recursion in phases like type_pull_up and there by avoid stack unsafety, but with this PR, we reduce clones too, along with a better and maintainble algorithm across the board

## TBD later (as this will be code refactoring and not a bug fix)
https://github.com/golemcloud/golem/issues/1428 

`ExprVisitor` that deals with enqueueing all expressions with a stack and queue inside may not be always required, and can affect compile time. The existing efficient code to be brought into the ExprVisitor side of things later and make sure every phases use the functions within ExprVisitor. But that's not the concern of this PR.


afsalthaj:
Any more coding in this PR, falls into "clean up/refactor" which I don't want to do prior to a release. Therefore raised a separate ticket #1428 - which deals with more of code clean up, something that doesn't hurt if not done now

vigoo (approved):

PR #1434, opened 2025-03-19
Use latest component version to retrieve files if there is no worker

mschuwalow:
Slightly different than what we discussed, but I believe this way it's fully consistent with other binding types

### DIFF
--- a/golem-worker-service-base/src/gateway_execution/file_server_binding_handler.rs
+++ b/golem-worker-service-base/src/gateway_execution/file_server_binding_handler.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use crate::empty_worker_metadata;
-use crate::gateway_execution::WorkerDetail;
+use crate::gateway_execution::WorkerDetails;
 use crate::getter::{get_response_headers_or_default, get_status_code};
 use crate::service::component::{ComponentService, ComponentServiceError};
 use crate::service::worker::{WorkerService, WorkerServiceError};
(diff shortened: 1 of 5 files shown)


golem-worker-service-base/src/gateway_execution/mod.rs
```diff
@@ -49,19 +48,19 @@ pub struct GatewayResolvedWorkerRequest<Namespace> {
 }
 
 #[derive(Debug, Clone, PartialEq)]
-pub struct WorkerDetail {
-    pub component_id: VersionedComponentId,
+pub struct WorkerDetails {
+    pub component_id: ComponentId,
```
mschuwalow, line 52: Removed the version here to make it clearer that it's not used at this level

golem-worker-service-base/src/gateway_execution/file_server_binding_handler.rs
```diff
@@ -173,6 +174,49 @@ impl DefaultFileServerBindingHandler {
...
+        worker_detail: &WorkerDetails,
+    ) -> Result<Component, FileServerBindingError> {
+        // Two cases, we either have an existing worker or not (either not configured or not existing).
+        // If there is no worker we need use the lastest component version, if there is none we need to use the exact component version
+        // the worker is using. Not doing that would make the blob_storage optimization for read-only files visible to users.
+
+        let component_version = if let Some(worker_id) = worker_detail.worker_id() {
+            let worker_metadata = self
+                .worker_service
+                .get_metadata(&worker_id, empty_worker_metadata())
```
afsalthaj, line 189: So what happens if the worker has gone down by this time, and started using the latest component version?

afsalthaj: I know this approach is taken in other places. Just came to my mind, but sorry that I merged before getting an answer

PR #1436, opened 2025-03-19
Default site for OSS localhost API deployment

afsalthaj:


mschuwalow:
I think it would be better to handle this in the cli rather than on the server side (and probably this default only makes sense while on an oss context, not in cloud). Like this defaults about where the cli is starting golem leak here.

afsalthaj:
cloud has a separate @

afsalthaj:
@mschuwalow  you are right

PR #1437, opened 2025-03-20
Better error messages, and allow trailing commas for pattern match

afsalthaj:
I request you to fully read this description before jumping into the code.

We had a parse error that emits a lot of errors when there was a trailing comma in a pattern match after a match arm. Why does it even talk `tuple` when you are already committed to a pattern match for instance. The answer is given later down below.

```rust
error: API Definition Service - Error: 400 Bad Request, Parse error at line: 10, column: 1
Unexpected `}`
Expected none, letter, digit, `_`, `-`, (, [, {, for, reduce, match, let, if, whitespaces, `{`, true, false, some, ok, err, letter or digit, `:`, `/`, `(`, `"`, !, `[` or whitespace
Unable to parse custom constructor name
Invalid syntax for tuple type
.... 
```

This PR is raised to fix a few things in and around this. Before I explain why this happened, let me explain a few things in this PR

* Firstly we already had test assertions on common parse error messages from long time back. These tests already include cases of invalid syntax for pattern match expressions such as [this](https://github.com/golemcloud/golem/blob/fixes_2/golem-rib/src/parser/errors.rs#L285), [this](https://github.com/golemcloud/golem/blob/fixes_2/golem-rib/src/parser/errors.rs#L304), [this](https://github.com/golemcloud/golem/blob/fixes_2/golem-rib/src/parser/errors.rs#L323) etc.  But this didn't include the case with trailing commas.  I will get to that.

*  Firstly, I also made sure trailing commas are supported. I do that every time when I write a pattern match in any language (scala, rust etc). So there is no need to add any more test around common mistakes when writing pattern match. So it will not happen again.

### But, why did such error accumulation occur even without backtracking? Is there a fix?

* This can be better understood if you reproduce the same behaviour with `if-else` condition where you fully skipped a rib expression after `else`. i.e,  `if 1 > 2 then 1 else;`  . Now this will behave exactly the same as pattern match case with a trailing comma (before this PR). It emits all possibilities that can come after `else`. 

* `combine` collects all the initial token of valid Rib, that is valid after `else` in the error message. In the case of pattern match, it emits all possibilities of initial token for a match constructor which includes rib-expressions itself.  Example:`ok(x) => ...`. This was the reason for accumulation. **_With this PR, this is fixed too._**

### Final important cleanups
*  In this PR, we also remove white space related parse errors. Again, please find these changes in tests that assert on error messages.

* We also remove these random/inconsistent errors such as "Invalid ok expression". Example `ok(x` is invalid ok expression, but developers already know this is a `result` type. There is no need to _explain_ Rib in a syntax error message. Users already know its a `result` type, and now that we have source locations too to back this up. A good Rib editor can also show the snippet of code that's syntatically invalid with the source locations. Also, such error messages doesn't exist in standard languages too. So we are keeping it simple. Meaning just say `expected `)``

### For future

I did look at skipping white spaces as a whole, and pre-tokenised stream is in our radar from day 1 after we moved from our own tokenizer-lexer-parser to combine's concise way of doing things. 

I tried it along with this PR, just in case if its easy, but I bumped into horrible compilation errors, along with losing real source locations. End of the day, that's a code clean up, and doesn't really count to releasing golem 1.2 fixing critical issues.


vigoo (approved):

afsalthaj:
Merged from the other PR

PR #1439, opened 2025-03-20
Distinguish query and path params

afsalthaj:


golem-worker-service-base/src/gateway_execution/request.rs
```diff
@@ -121,29 +121,28 @@ impl RichRequest {
...
         }
 
-        let header_value = Value::Object(header_records);
-
         let mut basic = serde_json::Map::from_iter(vec![
-            ("path".to_string(), merged_request_path_and_query),
-            ("headers".to_string(), header_value),
+            ("path".to_string(), Value::Object(path_key_values)),
+            ("query".to_string(), Value::Object(query_key_values)),
+            ("headers".to_string(), Value::Object(header_key_values)),
```
noise64, line 145: Should this also be non plural?

noise64 (approved):

golem-worker-service-base/src/gateway_execution/request.rs
afsalthaj: Probably not the right thing to change now. It has been `headers` from day 1. `request.headers.foo`.
We can take a look after asking John in another PR, that will need fixes in other tests

jdegoes: I think `headers` is correct as it's a field that contains _all headers_.

PR #1441, opened 2025-03-20
Readable error messages when BadRequest from API Gateway

afsalthaj:
The existing error messages as part of BadRequest from API Gateway are extremely bad. It doesn't provide much value add rather than a confusing and verbose error message back to the user.  So when there is a BadRequest, I have decided to only emit the "expectation (i.e, types)" that is readable, along with status-code 400.  

#### Before:

```rust
Invalid input: Input request doesn't match the requirements for rib expression to execute: Key 'query' not found in json_map.
Requirements. Record(TypeRecord { fields: [NameTypePair { name: "path", typ: Record(TypeRecord { fields: [NameTypePair
 { name: "user-id", typ: Str(TypeStr) }] }) }, NameTypePair { name: "query", typ: Record(TypeRecord { fields: [NameTypePair
  { name: "after", typ: U64(TypeU64) }, NameTypePair { name: "limit", typ: U8(TypeU8) }] }) }] })
```

#### After:

```rust
Invalid http request. Required types in request: 
  record { body: record { bar_key: string, foo_key: u32}, path: record{user-id: string} }
```

I would like to improve this error message, only after solving _static_ validation of query and path parameters, which may take slightly more time. Let's not wait until then in my opinion and get this going as an incremental improvement.


golem-worker-service-base/src/gateway_execution/gateway_http_input_executor.rs
```diff
@@ -655,21 +649,28 @@ impl<Namespace: Send + Sync + Clone + 'static> GatewayHttpInputExecutor
...
-        ).map_err(|err| format!("Input {key} doesn't match the requirements for rib expression to execute: {}. Requirements. {:?}", err.join(", "), analysed_type))?;
+        let input_value = input.get(key).ok_or(GatewayHttpError::BadRequest(format!(
+            "Missing `{}` in http request",
+            key
+        )))?;
+
+        let parsed_value = TypeAnnotatedValue::parse_with_type(input_value, analysed_type)
+            .map_err(|_| {
+                GatewayHttpError::BadRequest(format!(
+                    "Invalid http request. Required types in request: {}",
```
vigoo, line 664: I think it would be better to say "Available types in requires:"

PR #1442, opened 2025-03-20
Remove constraints on API deployment delete

afsalthaj:
- [x] Implement Removal of constraints
- [x] Prove the behaviour using integration tests of golem
- [x] Single API deployment -> Update Component (fail 👎 ) -> Delete API deployment -> Update Component (success 👍 )
- [x]  Multiple API deployments -> Update Component (fail 👎 ) -> Delete 1st API deployment -> Update Component (fail 👎 ) -> Delete 2nd API deployment -> Update Component (success 👍 )
- [x] Multiple API deployments of which second API deployment doesn't use any functions -> Update Component(fail 👎 ) -> Delete 1st API deployment -> Update Component (success 👍 )
- [x] Fix failing integration tests (it looks like there was a bug within the tests. see `golem-component-service-base/tests/all/service/mod.rs`.


golem-worker-service-base/src/service/gateway/api_deployment.rs
```diff
@@ -624,13 +658,19 @@ where
...
             self.deployment_repo
                 .delete(existing_deployment_records.clone())
                 .await?;
 
             self.set_undeployed_as_draft(existing_deployment_records)
                 .await?;
 
+            self.remove_component_constraints(existing_api_definitions, auth_ctx)
+                .await?;
+
```
afsalthaj, line 673: There are many instances like this - what happens if x runs and y fails. It's hard to squash everything into repo level transactions.

But that's a different topic. Not a concern of this PR.

mschuwalow: Let's leave a comment in the code for this so it's easier to find all the places with broken semantics later when we want to fix

mschuwalow (approved):

golem-worker-service-base/src/service/gateway/api_deployment.rs
afsalthaj: Sorry I missed this comment. 

These race conditions and transaction guarantee problems cannot be done just with Repository and any service layers. It's an architectural revisit. 

Not saying the current state is too bad, and I am pretty sure all of us the whole team knows it and aware of it - and there fore these localised comments will only confuse us - ending up thinking - only these commented places have issues and our solution becomes too specific and localised (wasting time solving it in a suboptimal way)

PR #1444, opened 2025-03-21
Use component name instead of id in api definitions

mschuwalow:
TLDR:

Before:
```
          id: 0b6d9cd8-f373-4e29-8a5a-548e61b868a5
          version: 0.0.1
          draft: true
          routes:
          - method: post
            path: /good/syntax
            binding:
              componentId:
                componentId: 0b6d9cd8-f373-4e29-8a5a-548e61b868a5
                version: 0
              response: |
                  { status: 200, body: "x" }
          - method: GET
            path: /bad/syntax
            binding:
              componentId:
                componentId: 0b6d9cd8-f373-4e29-8a5a-548e61b868a5
                version: 0
              response: |
                  { status: 200, body: "x" }
```

After:
```
          id: 0b6d9cd8-f373-4e29-8a5a-548e61b868a5
          version: 0.0.1
          draft: true
          routes:
          - method: post
            path: /good/syntax
            binding:
              component:
                name: test-component
                version: 0
              response: |
                  { status: 200, body: "x" }
          - method: GET
            path: /bad/syntax
            binding:
              component:
                name: test-component
                version: 0
              response: |
                  { status: 200, body: "x" }
```


afsalthaj:
Please merge this after #1442

golem-common/src/cache.rs
```diff
@@ -53,11 +52,10 @@ pub struct Cache<K, PV, V, E> {
     name: &'static str,
 }
 
-#[async_trait]
 pub trait SimpleCache<K, V, E> {
-    async fn get_or_insert_simple<F>(&self, key: &K, f: F) -> Result<V, E>
+    fn get_or_insert_simple<F>(&self, key: &K, f: F) -> impl Future<Output = Result<V, E>>
```
mschuwalow, line 56: Need this because I want to work with non-'static futures

golem-worker-service-base/src/gateway_api_definition/http/api_definition.rs
```diff
@@ -219,6 +217,38 @@ impl TryFrom<grpc_apidefinition::ApiDefinition>
     }
 }
 
+impl TryFrom<HttpApiDefinition> for grpc_apidefinition::ApiDefinition {
```
mschuwalow, line 220: These were moved next to the types they belong to. I think some leftover from some cleanup

golem-worker-service-base/src/service/gateway/api_definition.rs
```diff
@@ -173,15 +194,118 @@ pub trait ApiDefinitionService<AuthCtx, Namespace> {
...
+            component_by_id_cache_size: 1024,
+        }
+    }
+}
+
+type ComponentByNameCache = Cache<ComponentId, (), Option<Component>, String>;
+
+type ComponentByIdCache = Cache<ComponentName, (), Option<Component>, String>;
+
+// TODO: cache mappings
```
mschuwalow, line 222:
```suggestion
```

golem-worker-service-base/src/service/gateway/mod.rs
```diff
@@ -12,8 +12,27 @@
...
+use golem_common::model::ComponentId;
+use golem_service_base::model::ComponentName;
 pub mod api_definition;
 pub mod api_definition_validator;
 pub mod api_deployment;
 pub mod http_api_definition_validator;
 pub mod security_scheme;
+
+#[async_trait]
+pub trait ConversionContext: Send + Sync {
```
mschuwalow, line 25: This is used to translate between our internal and public representation of components in the api definitions.

golem-worker-service-base/src/service/gateway/api_definition.rs
```diff
@@ -138,13 +145,27 @@ pub trait ApiDefinitionService<AuthCtx, Namespace> {
         auth_ctx: &AuthCtx,
     ) -> ApiResult<CompiledHttpApiDefinition<Namespace>>;
 
+    async fn create_with_oas(
```
mschuwalow, line 148: These were moved here as they are duplicated between the grpc and http apis

afsalthaj:
@mschuwalow  Should there be version too? I guess we can avoid version and just make use of the latest version or if worker is already running get the one from it's metadata? I am not able to remember the reason for version too. There was a good reason once, but now I can't remember. Or probably @vigoo  can remind me.

golem-common/src/cache.rs
afsalthaj: I see, need to look at your PR more where exactly we forced to make this change. It's interesting such a change was required.

openapi/golem-service.yaml
```diff
@@ -4515,13 +4515,24 @@ components:
...
+      - name
+      - version
     GatewayBindingData:
       type: object
       properties:
         bindingType:
           $ref: '#/components/schemas/GatewayBindingType'
-        componentId:
-          $ref: '#/components/schemas/VersionedComponentId'
+        component:
```
afsalthaj, line 4534: Probably we need to keep things backward compatible. We can't break it at 1.2, because 1.1 users - almost all - during different hackathons used API definitions.

golem-worker-service-base/src/api/api_definition.rs
```diff
@@ -489,7 +525,7 @@ impl From<SecuritySchemeReferenceData> for SecuritySchemeReference {
 #[serde(rename_all = "camelCase")]
 #[oai(rename_all = "camelCase")]
 pub struct GatewayBindingResponseData {
-    pub component_id: Option<VersionedComponentId>, // Optional to keep it backward compatible
+    pub component: Option<GatewayBindingComponent>, // Optional to keep it backward compatible
```
afsalthaj, line 528: So, is this backward compatible?

openapi/golem-service.yaml
```diff
@@ -4515,13 +4515,24 @@ components:
       enum:
       - directory
       - file
+    GatewayBindingComponent:
+      type: object
+      properties:
+        name:
+          type: string
+        version:
```
afsalthaj, line 4523: I think version may be avoided? I think in the long run, even if there is some amount of reasoning to keep the version technically, its just annoying users specifying it. 

Workers can pick the latest version of component _at any point_ for reasons! So there isn't a s strong reason to keep the version anymore. There is type check, but that can be done using latest component version available. There will be corner cases - for example - workers are already running with slightly order version even if there is a new component version. However, let's not allow users to piggyback on older types when defining API definition for the first time. It's much safer, easier to understand and document.

openapi/golem-service.yaml
vigoo: We are totally breaking backward compatibility with the other changes we decided to do, so it does not matter

golem-common/src/cache.rs
mschuwalow: This guy will create a non-static future when used https://github.com/golemcloud/golem/pull/1444/files#diff-0b676bf48730736dc951c862a53c431e12ae2610bfbba134476906a6c4dbeddaR222 (which is fine, as we know exactly for how long it is needed).

Making it work with static futures is possible, but would mean cloning the Arc, introducing an Arc for the caches and introducing (and propagating) a Clone constraint for AuthCtx. I think this way is more elegant

openapi/golem-service.yaml
vigoo: I think it makes sense to not require component version in the yaml you upload. This will help in integration with the app manifest too - there it's lifecycle will be more tightly integrated with the other things so having to manually update these versions seems to be problematic. Also it is logical that you the api description assumes the "actual" version - even if the actual version in the app's repository may not be the actual version that is deployed in golem. 
So I think it's good to make it optional, but I think we can also keep it as an optional thing you can specify to force using some kind of specific version.

We definitely need to still have a specific component version as soon as the definition was typechecked. That will be the version we know it was working for, and the one that we use in the constraint so we cannot break the deployed API with a breaking change. 

So (although I'm not fully familiar with all the internal details of our api definition code) I think we could keep things as they are internally, but resolve a missing component version into the current latest one at the edge where the definition gets uploaded.

mschuwalow: Sounds good. Will make the versions optional in a followup pr if you are fine with that

afsalthaj (approved):

PR #1447, opened 2025-03-24
Fix query params and avoid redundant lookups in hot path

afsalthaj:
The changes are small, but these are some fundamental changes in my view - and very important. 

* Look up query or path or headers only if needed by the Rib script. Not for every http request. Infact, look up from `poem::Request` for specific header keys that are _required_ and transform/parse them specifically, instead of collecting everything together.
* Make sure if a number is passed in path or query, but rib expects a string (which by default expects string for these global paths except body), it should work. I think this might be a regression and the lack of tests in this space need to be fixed
* Better error messages. Such as instead of `invalid request. expected : record { headers: {user-id: string} }`, with this PR, it will be `missing header: user-id`.  Bring back more details into the error message as well  "invalid http request body \n Invalid value for key `foo`: expected number, found string \n expected body: ...". 
* More precise approach rather than blindy using`wasm_rpc::TypeAnnotatedValue::parse_with_type` after collecting all fields into HashMap and then forming a ValueAndType etc in hot path. This in a hot path is not good. I think we only did some compiler driven refactoring in this space of the code after August 2023 where we quickly glued everything together - more of as a POC.  This is gone now, for the most part. We use wasm_rpc mappings only for request body. Here too, I guess I am missing `ValueAndType::parse_with_type` than `TypeAnnotatedValue::parse_with_type` but I haven't gone much deep into it, nor it is a concern of this PR. But I believe any reduction of loops and hoops in the hot path is an advantage. We will see.
* Try to make use of `RichRequest.inner` especially headers, as poem has efficient lookups
* Make sure both `request.header.x` as well as `request.headers.x` work. This has always been confusing to me and I don't want to keep going back to docs to see which one is correct

I have focussed on these fundamental requirements, and hasn't done any extra refactoring as part of this work!

### DIFF
--- a/golem-worker-service-base/src/gateway_execution/gateway_http_input_executor.rs
+++ b/golem-worker-service-base/src/gateway_execution/gateway_http_input_executor.rs
@@ -16,7 +16,7 @@ use super::auth_call_back_binding_handler::AuthorisationSuccess;
 use super::file_server_binding_handler::FileServerBindingSuccess;
 use super::http_handler_binding_handler::{HttpHandlerBindingHandler, HttpHandlerBindingResult};
 use super::request::{
-    authority_from_request, split_resolved_route_entry, RequestHeaderValues, RequestPathValues,
+    authority_from_request, split_resolved_route_entry,
     RichRequest, SplitResolvedRouteEntryResult,
 };
 use super::to_response::GatewayHttpResult;
@@ -55,7 +55,6 @@ use golem_wasm_rpc::{IntoValue, IntoValueAndType, ValueAndType};
 use http::StatusCode;
 use poem::Body;
 use rib::{RibInput, RibInputTypeInfo, RibResult, TypeName};
-use serde_json::Value;
 use std::collections::HashMap;
 use std::str::FromStr;
 use std::sync::Arc;
@@ -103,21 +102,9 @@ impl<Namespace: Clone> DefaultGatewayInputExecutor<Namespace> {
         request: &mut RichRequest,
         binding: &WorkerBindingCompiled,
     ) -> GatewayHttpResult<RibResult> {
-        let mut rib_input: serde_json::Map<String, Value> = serde_json::Map::new();
-
-        // phase 1. we only have the request details available
-        {
-            let request_value = request
-                .as_json_with_body()
-                .await
-                .map_err(GatewayHttpError::BadRequest)?;
-            rib_input.insert("request".to_string(), request_value);
-        }
-
         let worker_detail = self
             .get_worker_details(
                 request,
-                &rib_input,
                 &binding.worker_name_compiled,
                 &binding.idempotency_key_compiled,
                 &binding.component_id,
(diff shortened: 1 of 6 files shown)


golem-worker-service-base/src/gateway_rib_compiler/mod.rs
```diff
@@ -45,6 +45,11 @@ impl WorkerServiceRibCompiler for DefaultWorkerServiceRibCompiler {
                     path: Path::from_elems(vec!["headers"]),
                     inferred_type: InferredType::Str,
                 },
+                GlobalVariableTypeSpec {
+                    variable_id: VariableId::global("request".to_string()),
+                    path: Path::from_elems(vec!["header"]),
+                    inferred_type: InferredType::Str,
+                },
```
afsalthaj, line 52: I keep bumping into this problem. It's very hard to say which one is more intuitive. 
`request.path` or `request.query` or `request.body` don't have this singular plural confusion - `request.header.auth-bearer` or `request.headers.auth-bearer`. May be `header` is more correct. But I am pretty sure users will need to keep coming back to docs to re-assert themselves.

golem-worker-service-base/src/gateway_execution/request.rs
```diff
@@ -67,38 +75,36 @@ impl RichRequest {
...
 
-    fn request_query_values(&self) -> Result<RequestQueryValues, String> {
+    pub fn request_query_values(&self) -> Result<RequestQueryValues, String> {
         RequestQueryValues::from(&self.query_params(), &self.query_info)
             .map_err(|e| format!("Failed to extract query values, missing: [{}]", e.join(",")))
     }
 
-    fn request_header_values(&self) -> Result<RequestHeaderValues, String> {
+    pub fn request_header_values(&self) -> Result<RequestHeaderValues, String> {
         RequestHeaderValues::from(self.underlying.headers())
```
afsalthaj, line 102: I think I will remove these RequestHeaderValues and RequestQueryValues, and simply return a HashMap similar to path values. There is no need to wrap and unwrap in this hot path even if its negligible. Not much typsafety is attained here.

golem-worker-service-base/src/gateway_execution/gateway_http_input_executor.rs
```diff
@@ -647,36 +601,178 @@ impl<Namespace: Send + Sync + Clone + 'static> GatewayHttpInputExecutor
 }
 
 async fn resolve_rib_input(
-    input: &serde_json::Map<String, Value>,
+    rich_request: &mut RichRequest,
```
mschuwalow, line 604: One thing I don't really like here is that we mutate the RichRequest (due to potentially moving the body).
Would it be possible to refactor this code so we move the request here? (So that it can not be used twice by accident)

afsalthaj: It's just like any server backend I guess - that once we consumes body. At the same time we don't need to access body if its not being used too. One thing that we could do is to populate body within RichRequest, once we access the body.

PR #1449, opened 2025-03-24
Optional component version in api definition

mschuwalow:


mschuwalow:
Note that the grpc api was not adjusted.
There are some open questions I have there and the api is currently not really user facing. I will adjust it once we had an opportunity to talk about it.

afsalthaj (approved):

PR #1450, opened 2025-03-25
Make sure request body can be used in multiple rib scripts (if any) within one request

afsalthaj:
* The code still ensures that we don't consume request body if unused in any rib script. So this is _NOT_ early consumption and reuse across
* If API definition has multiple rib scripts exists that needs to look up request body (ending up resulting in consuming request body from poem::Request), it should still behave lazy (i.e, only if its used) and yet reuse, avoiding potentially "empty" body errors.

### DIFF
--- a/golem-worker-service-base/src/gateway_execution/request.rs
+++ b/golem-worker-service-base/src/gateway_execution/request.rs
@@ -33,6 +33,7 @@ pub struct RichRequest {
     pub path_param_extractors: Vec<PathParamExtractor>,
     pub query_info: Vec<QueryInfo>,
     pub auth_data: Option<Value>,
+    pub cached_request_body: Value
 }
 
 impl RichRequest {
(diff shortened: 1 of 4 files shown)


golem-worker-service-base/src/gateway_execution/request.rs
```diff
@@ -29,13 +29,25 @@ const COOKIE_HEADER_NAMES: [&str; 2] = ["cookie", "Cookie"];
 /// Thin wrapper around a poem::Request that is used to evaluate all binding types when coming from an http gateway.
 pub struct RichRequest {
     pub underlying: poem::Request,
```
afsalthaj, line 31: I tried this private too, but too many compilation errors. Not the purpose of this PR

vigoo:
It was very hard to me to first understand what this is trying to solve, especially because of comments like "in 99% this is not happening". I almost wrote here that I don't have any idea what this is about when I realized that you must be talking about that there multiple rib scripts associated now with a single endpoint, the one that generates the actual response, and another that can produce an invocation context. But anyway, just writing this here for further reference to make it easier to understand what this is all about.

vigoo (approved):

PR #1451, opened 2025-03-25
Attempt to fix a race condition when interrupting workers for update

vigoo:
This is a simple fix for the issue reported about `unwrap()` panicking when trying to automatically update a worker.

The fix is just to simply ignore this error - because the only way it can fail is if the receiver of this channel is already being dropped, which means the run loop has quit anyway; this is basically the same effect we want from the interruption (stop the running worker so we can update it) so we can safely ignore it.


noise64:
is it worth to add a log for it? ro not that useful?

noise64 (approved):

vigoo:
I don't think it adds any useful knowledge

PR #1457, opened 2025-03-29
New component analysis implementation based on wit-parser

### LINKED ISSUE #1458: Migrate to wit-parser and store the whole “binary wit” in component metadata

vigoo:
Resolves #1458 

Switches to `wit-parser` as the source of truth for extracting a component's exports. Also serializes the "WIT binary representation" and the root package's name and version into the component metadata. In addition to that, we also add the root package name and version as separate fields to the `component_versions` table for future use.

Marked some old (C# and Swift) tests ignored which are not officially supported languages anymore (and they are not recompiled with latest toolchains for a while)


golem-component-service-base/src/repo/component.rs
```diff
@@ -1072,7 +1080,9 @@ impl<Owner: ComponentOwner> ComponentRepo<Owner> for DbComponentRepo<sqlx::Postg
                     cv.component_type AS component_type,
                     cv.available AS available,
                     cv.object_store_key AS object_store_key,
-                    cv.transformed_object_store_key AS transformed_object_store_key
+                    cv.transformed_object_store_key AS transformed_object_store_key,
+                cv.root_package_name AS root_package_name,
```
noise64, line 1084: indent

PR #1471, opened 2025-03-30
grpc integration

### LINKED ISSUE #1201: gRPC and OpenAPI Integration
This specification defines a new feature whereby Golem workers will be able to interact with gRPC and OpenAPI services in a type-safe way, similar to how Golem workers interact with other workers in a type-safe way.

This feature will greatly expand the number of applications that can be built and deployed on Golem and simplify usability of existing Golem applications currently written using lower-level libraries, such as HTTP client libraries.

## Background

In Golem's current worker-to-worker communication, *WIT* interfaces describe the public interface of components, with `wasm-rpc` handling the transformation of these interfaces to support the addressing of specific workers. For external service communication, we need similar but stateless transformations from Protobuf and OpenAPI specifications to WIT.

The current process for worker-to-worker communication is as follows:

1. The user is writing a component `A`, which depends on a component `B`. The user adds component `B` as a dependency of component `A`, by editing the Golem application manifest file (a YAML file read by `golem-cli`).
2. The public interface of component `B` is described by WIT (WASM Interface Types).
3. `golem-cli`, using the `wasm-rpc` project, can generate a transformed WIT file that is the stateful version of the original WIT exported by component `B`. This stateful aspect allows developers to target specific workers, which is necessary for worker-to-worker communication--but which is not generally necessary for gRPC or OpenAPI, because these protocols are stateless.
4. `wasm-rpc` generates and compiles stubs for RPC, which implement the transformed (stateful) WIT, and require Golem host functions for performing the actual RPC. Note that this step is [in the process of being simplified](https://github.com/golemcloud/golem/pull/1199/files), and shortly, the stubs will be dynamically added on the server side, in the `worker-executor`, using a custom linker, and this model of dynamic stub generation **must** be used for both gRPC and OpenAPI support.
5. The user writes component `A`, invoking the transformed WIT, and linking against the stubs, to eliminate the requirement on the transformed WIT. This step is disappearing as worker-to-worker transitions to the server-side, dynamically generated stubs; it will not be necessary for gRPC or OpenAPI support.
6. The `worker-executor` provides the Golem host functions necessary for RPC.

The process for supporting gRPC and OpenAPI will proceed in a similar fashion:

1. The user is writing a component `A`, which depends on gRPC defined by Protobuf `B` and OpenAPI API defined by schema `C`. The user adds a reference to `B` as one dependency of type `grpc`, and a reference to `C` as another dependency of type `openapi`, by editing the Golem application manifest file.
2. `golem-cli`, using an extended and enhanced `wasm-rpc` , generates a WIT file for `B`, corresponding to an idiomatic encoding of the gRPC services into WIT; and another WIT file for `C`, corresponding to an idiomatic encoding of the OpenAPI API into WIT.
3. The user writes component `A`, invoking the generated WIT for `B` and `C` as necessary to access the functionality of API `B` and API `C`.
4. The `worker-executor`, when executing instances of component `A`, dynamically stubs the WIT interfaces and links them to the instances, allowing instances to interact with the gRPC service and the OpenAPI API.

In supporting both gRPC and OpenAPI, there are several major challenges:

1. For an arbitrary gRPC or OpenAPI spec, programmatically generating an equivalent WIT that is "WIT idiomatic", and which does not look or feel like it was programmatically generated. This is much more challenging for OpenAPI schemas than it is for Protobuf.
2. [Dynamically adding stubs](https://github.com/golemcloud/golem/pull/1199/files) in the worker-executor corresponding to the generated WIT. These stubs will be implemented in Rust and programmatically execute gRPC and OpenAPI invocations using appropriate metadata.
3. Ensuring durable execution of the API calls at the level of the individual calls, using the same mechanisms that the `worker-executor` is already using to make WASI calls durable (essentially, branching off record/play back mode and reading from / writing to the oplog).

Beyond these major challenges, there are several other essential features of the implementation to consider:

- Capturing the protobuf and OpenAPI schemas during component creation and update. These will be captured by `golem-cli`, the command-line interface for Golem.
- Modifying the Golem application manifest file (YAML) parser and schema to accept new types of dependencies, including the protobuf files and OpenAPI files. Currently, the only type of dependency supported by the parser and schema is `wasm-rpc`, for describing worker-to-worker communication.
- Modifying the component creation and update REST APIs to accept information on the new types of dependencies.
- Storing the protobufs and OpenAPI files during component creation and update; or more precisely, storing a structured representation that has enough information for the `worker-executor` to generate and add the dynamic stubs.
- Accessing the protobuf and OpenAPI structured information from inside the `worker-executor`, so when a worker is launched, they have what they need to dynamically add the gRPC / HTTP stubs.

## Input Formats

-   Protocol Buffers v3 (proto3) with gRPC service definitions
-   OpenAPI 3.0.x YAML/JSON specifications

### Package Translation

-   Must generate WIT package name and version
-   For gRPC: Use proto package name
-   For OpenAPI: Use sanitized info.title
-   Version must come from:
    -   gRPC: Configuration input
    -   OpenAPI: info.version

### Type Mappings

Protobuf and OpenAPI types must be mapped into their most precise WIT types. A sketch of a few possible mappings for Protobuf is shown below:

```wit
oneof         -> variant
message       -> record
string        -> string 
int32         -> s32 
int64         -> s64 
uint32        -> u32 
uint64        -> u64 
float         -> float32 
double        -> float64 
bool/boolean  -> bool 
repeated T    -> list<T> 
optional T    -> option<T>
```

Note that OpenAPI schemas may embed JSON schemas, which can contain patterns. To the extent possible and reasonable, patterns should be converted into validations that occur in the stubs, to ensure that where possible, local and highly descriptive errors are produced--rather than relying on a remote service to produce a useful error in response to some kind of schema validation failure.

### Error Handling

The following sketch of an error type could inform design of a generalized error type. The actual error type used in the implementation must be at least as capable and as expressive as the provided one.

```wit
variant error {  
  unauthorized { message: string }, 
  not-found { resource: string, id: string }, 
  validation-error { fields: list<string> }, 
  rate-limited { retry-after: u32 }, 
  server-error { message: string }, 
}
```

### Authentication Types

The following sketches of authentication types could inform design of generalized authentication types. The actual authentication types used in the implementation must be at least as capable and as expressive as these sketches:

```wit
record bearer-auth {
  token: string, scheme: string, 
}
record basic-auth {
  username: string, password: string, 
}
record api-key-auth {
  key: string, 
}
 ```

### Message Translation

-   Each message type must become a WIT record
-   Names must be converted to kebab-case
-   Must preserve all fields and their relationships
-   Nested messages must become separate records

### Service Translation

-   Each service must become a WIT interface
-   Each RPC method must become a WIT function
-   Must return result<response-type, error> for WIT-idiomatic error handling

Example:

```wit
message  GetUserRequest  {   string user_id =  1; }
```

Must become:

```wit
record get-user-request {  user-id: string, }
```

The numerical order of fields in the protobuf MUST correspond to the linear order of fields in WIT.

### Schema Translation

-   Each components.schema must become a WIT record
-   Required fields must be non-optional
-   Must preserve all relationships between types

### Inline Type Translation

-   All anonymous/inline schema definitions must be converted to named WIT types
-   Names could be synthesized systematically using the following rules (or similar):
    1.  For request bodies: `{path}-{method}-request-body`
    2.  For response bodies: `{path}-{method}-response-body`
    3.  For array items: `{parent-type}-item`
    4.  For nested objects: `{parent-type}-{field-name}`
    5.  For parameters: `{path}-{method}-params`

Example:

```yaml
paths:
  /users: 
    post: 
      requestBody: 
        content: 
          application/json: 
            schema: 
              type: object
              properties: 
                name: 
                  type: string 
                addresses: 
                  type: array 
                items: 
                  type: object
                  properties: 
                    street: string 
                    city: string
```

Could generate something like:

```wit
record users-post-request-body {
  name: string,
  addresses: list<users-post-request-body-addresses-item>, 
}
record users-post-request-body-addresses-item {
  street: string, 
  city: string,
}
 ```

The exact naming scheme may vary, but must be:

-   Deterministic
-   Generate valid WIT identifiers
-   Maintain clear relationship to source structure
-   Avoid name collisions
-   Create traceable mappings for worker-executor's use

### Path Translation

-   Must support paths to any number of levels deep
-   Each base resource path must generate interface(s)
-   Must group related operations logically
-   Must handle path/query parameters correctly

### Header Handling

Must specially handle these headers:

```wit
Authorization     -> auth field in request 
ETag              -> version field in response 
If-Match          -> expected-version in request 
Last-Modified     -> last-updated in response
```

All other headers must be mechanically translated to kebab-case option<string> fields.

### Resource Function Generation Rules

The following sketches of resource interfaces could inform design of generalized resource handling for REST APIs. The actual interfaces used in the implementation must be at least as capable and as expressive as these sketches:

```wit
interface resource-collection {
  list: func(params: list-params) -> result<list-response, error>; 
  create: func(params: create-params) -> result<item, error>; 
}
interface resource-item {
  get: func(id: string) -> result<item, error>; 
  update: func(id: string, params: update-params) -> result<item, error>; 
  delete: func(id: string) -> result<unit, error>; 
}
```

### Conflict Resolution

Potential algorithm for resolving name conflicts:

1.  Appending type suffix (_record, _params, _result, etc.)
2.  If still conflicting, error out requiring manual resolution

### Generated Names

-   Must be valid WIT identifiers
-   Must use kebab-case
-   Should not exceed 64 characters
-   Must prefix reserved words with %

## Dynamic Stub Requirements

Dynamic stubs must be added to the `worker-executor` for all of the gRPC and OpenAPI dependencies. These stubs must, of course, match the type signatures of the generated WITs. 

Durability must be ensured using the same mechanism that is used to provide durability for WASI inside the `worker-executor` (namely, direct interaction with the mode, whether record or playback, and the worker oplog).

## Testing Requirements

Automated tests are required at the following levels:

- **Unit**. Unit testing should ensure correctness of each part of the system developed.
- **Integration**. Integration should ensure that when the components are composed (e.g. `golem-cli` with `wasm-rpc`), they function together as specified.
- **System**. End-to-end system tests should verify the correctness of the entire system, all the way from `golem-cli` to the `worker-executor`.

In particular, there must exist automated tests against REAL gRPC and OpenAPI APIs, which verify both the transformed WIT that is generated, and also which actually invoke the APIs from within a test worker, to verify that the dynamically added stubs in the `worker-executor` are working correctly. Durability tests must be added to verify durability of any gRPC or HTTP interacts that happen from within the generated stubs.

The solution will be tested against the following resources:

OpenAPI:
 - https://github.com/cloudflare/api-schemas
 - https://github.com/openai/openai-openapi
 - https://github.com/github/rest-api-description

gRPC:

- https://grpcb.in/
- https://cloud.google.com/text-to-speech/docs/reference/rpc

We have several additional OpenAPI and gRPC resources we will be "surprise testing" against to ensure sufficient scope and quality.

## Acceptance

We will accept the first solution which, in our sole opinion and estimation:

1. Demonstrates deep understanding of the purpose of this feature, even, if necessary, extending or changing parts of the specification in order to achieve the end goal in a way which is compatible with how worker-to-worker communication currently works, which is highly friendly to developers, and which does not sacrifice type safety, performance, or other characteristics important for production adoption of the feature.
2. Demonstrates high-quality, best practices in the Rust implementation and associated tests. Code should be modular, well-organized, strongly typed, free of duplication, and demonstrate the best of what Rust has to offer; with consistent conventions as utilized by the best parts of the existing code base.
3. Demonstrates a systematic and well-thought out approach to testing, having ample unit tests, integration tests, and system tests, which collectively verify end-to-end usage scenarios, including interacting with real-world gRPC and OpenAPI APIs from within Golem workers, in a type-safe way.
4. Has been updated with latest changes coming from our head branch.
5. Is repeatably passing the CI build.
6. Is sufficiently well-documented to enable other developers to build off the solution, and end-developers to use the solution in their own Golem applications.

webbdays:
/fixes #1201 (grpc part only) 

Grpc integration :

- wit generation

   - new crate called wit-carpenter that handles translation from grpc, openapi to wit and also to generate some metadata as req.
   - from_grpc function can used to generate wit, filedescriptor set from protobuf files
   - from_openapi still in WIP 

- dynamic linking changes
  - using existing wasm_rpc mechanism but with a introduction of a concept called remote
  - remote is some server/service/component/api whose function/endpoint we are calling via rpc mechanism.
  - so, i have categorized remote into grpc, openapi, golem-worker (default)
  - created some functions to convert to and from json. ( used TypeAnnotatedValue enum as an intermediate)
  - used proto_reflect lib (not using server side reflection, but via fds) and generated dynamic message using some metadata like file descriptor set we get for each remote.
  - GrpcCLient to send req and receive. (used tonic)
  
- golem cli changes
  - added "golem-cli wit generate [options]" command to generate wit from grpc and openapi
  - changes while building client for component to gen and store remote specific metadata. example for grpc we need to send filedescriptorset and package name
  - manifest parsing changes like addition of remote filed in dependency definition object (if not provided defualt to golem-worker)
  
- example rpc-grpc example
  - user can use customCommand in golem manifest files to auto generate wit form protobuf root on build every time.
  - currently user has to impl dummy methods for each Guests. (looking to auto generate template from wit).
  

Pending
- Still have some corner cases that I have to cover like.

    -  Enums.
    -  Empty message to a empty record.

- Currently when bindings are generated it is erroring with empty record.

    - Security review of where to send the auth token and how. Currently as dummy I have put in metadata a bearer token.
- Generating boiler component from a wit (user generates wit from a command I have added to golem cli)

Tested against 
grpc://grpcb.in::9000 withoit tls and
9001 with tls.

Thanks for reading.


test-components/rpc-grpc/components-rust/rpc-grpc-api-caller/golem.yaml
```diff
@@ -0,0 +1,19 @@
...
+    template: rust
+
+# Example for adding dependencies for Worker to Worker communication:
+# See https://learn.golem.cloud/docs/app-manifest#fields_dependencies for more information
+#
+dependencies:
+ rpc-grpc:api-caller:
+ - target: rpc-grpc:grpcbin
+   type: wasm-rpc
+   remote: grpc
```
jdegoes, line 19: This looks wrong. For a grpc dependency, it should look like this:

```
type: grpc
proto:
  file: ../../example/service.proto
```

jdegoes: Other protobuf-specific options would be nested under `proto`, for example, a target package name (although this could be inferred from the protobuf file).

webbdays: I was thinking on like remotes
dynamic rpc with different remotes like grpc, openapi, golem-worker (it uses wasmrpc)

webbdays: wit is generated manually or via customCommand.
when building cleint via build_client for dynamic wasmrpc in cli
i check for remote and generate metadata and store in application context and it used used when sending dynamic linking object. 
if it is grpc i look at grpc folder inside the component like wise other openapi remotes.

golem-worker-executor-base/src/durable_host/dynamic_linking/wasm_rpc.rs
```diff
@@ -788,14 +787,25 @@ async fn grpc_call<Ctx: WorkerCtx + HostWasmRpc + ResourceStore + Send>(
                 .to_case(Case::Pascal)
         );
 
-        let service_descriptor = descriptor_pool
-            .get_service_by_name(&service_full_name)
-            .unwrap();
+        let Some(service_descriptor) = descriptor_pool
```
jdegoes, line 790: You have used `wasm_rpc.rs` for your gRPC code. But this is poor code organization.

The `wasm_rpc.rs` module should ONLY contain dynamic linking code for wasm_rpc protocol, which is NOT the gRPC protocol.

You should make a new `grpc.rs`, which contains all the linking code for gRPC, and you should use that at a higher level in an appropriate way.

webbdays: sure will separate the logic into grpc.rs

test-components/rpc-grpc/components-rust/rpc-grpc-api-caller/golem.yaml
jdegoes: You build support for this on wasm-rpc but wasm-rpc is only for wasm-rpc, not for gRPC. It's a different dependency type and needs to be reflected like that in the manifest file.

Now obviously you have to do many of the same steps as for wasm-rpc, but that's an implementation detail solved with code reuse / abstraction and should not affect the user-visible surface area of the feature. From a user's perspective, wasm-rpc is one dependency type, grpc is another dependency type, each requiring different things.

A grpc dependency requires a file that contains the proto.

jdegoes:
@webbdays It looks like you tried to do OpenAPI too, though that's not finished. The bar would be much higher for gRPC + OpenAPI, so even if you want to do OpenAPI too, I would recommend getting through gRPC.

Which means: removing OpenAPI stuff from this PR, doing the organization mentioned above, and the improvements to DX and compatibility with our existing manifest architecture.

Please ping me when you're ready for this PR to have another (closer) look! 🙏

webbdays:
Yes, first I started with openapi some time back.
Sure will focus on grpc for now. Grpc streams on server side implemented.
Stuck with some error Will push it soon once fixed and will work on
remaining cli part.

Thanks.

PR #1474, opened 2025-04-01
Rib Repl

afsalthaj:
### Rib Repl

Allows to interactively work with golem workers using Rib language

#### Note
There are infinite nuances when building a REPL inspecting various parts of Rib compiler progressively, with a progressive interpreter. This work is a manifestation of a small experiment done a few days back to a decently working version.

#### An optional embedded executor
While Repl is a `lib` that can be used by `golem-cli` (which further uses app manifest to deal with component dependency management), the developers of golem (and may be users if they really want to) can clone golem project and do the following

```bash
cd golem-rib-repl
cargo run --features embedded
```

Otherwise, the only for users to use Repl will be through `golem-cli` when we integrate this with app manifest

When enabling `embedded` feature, it pull in various internal dependencies of golem such as `golem-service-base` etc.
As such the `lib` of `Repl` is depending only on `wasm-ast`, `wasm-rpc` and obviously `golem-rib`.

#### Repl features
There will be something or the other that we can improve forever, but it is managing to resolve a lot of things:
It is better to show it as a demo video, to see where we are with this

https://github.com/user-attachments/assets/be60334d-b4e0-4472-a317-525acb8285b0


golem-rib-repl/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+wasmtime-wasi-http = { workspace = true }
+
+rustyline = "15"
+termion = "1.5.6"
+tokio = { version = "1", features = ["full"] }
+dirs = "4.0.0"
+async-trait = "0.1.87"
+anyhow = "1.0.97"
+uuid = { version = "1.15.1", features = ["v4"] }
+serde_json = "1.0.140"
```
afsalthaj, line 43: Please review cargo.toml later

noise64 (approved):

Issue #1464, opened 2025-03-30
Update to the latest JS tooling


vigoo:
First try the latest version of everything (componentize-js + jco) by hand.

If it does not work correctly, try to create a patched version (reapplying our previous workarounds, etc)

Then:
- Update the test components
- Update the golem-ts library
- Update the the templates in golem-cli
- Update the docs

Note that part of this is also dropping our custom TS binding generator and use whatever they created since then (see more info in the upstream PR we created)

PR #1475, opened 2025-04-01
Update js/ts test-components

mschuwalow:
The ts-rpc component is also a preview of the new ts project structure.

part of https://github.com/golemcloud/golem/issues/1464


test-components/ts-rpc/common-ts/scripts/jco-wrapper
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env node
```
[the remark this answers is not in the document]
mschuwalow: And actually uncessary with they way it's now invoked in the package.json.
Will drop

test-components/js-3/package.json
```diff
@@ -1,10 +1,10 @@
 {
   "type": "module",
   "scripts": {
-    "build": "jco componentize -w wit -o out/component.wasm main.js"
+    "build": "componentize-js main.js -w wit -o out/component.wasm main.js"
   },
   "devDependencies": {
-    "@golemcloud/componentize-js": "0.10.5-golem.3",
-    "@golemcloud/jco": "1.4.4-golem.1"
+    "@bytecodealliance/componentize-js": "^0.18.0",
```
noise64, line 7: i would keep these pinned

test-components/ts-rpc/components-ts/caller/golem.yaml
```diff
@@ -6,13 +6,7 @@
...
-    customCommands:
-      copy:
-      - command: cp dist/caller-ts.wasm ../..
-      - command: cp ../dist/caller-composed-ts.wasm ../..
 dependencies:
   caller:
   - target: counter
-    type: static-wasm-rpc
\ No newline at end of file
+    type: wasm-rpc
```
noise64, line 12: all of these were using static rpc and custom commands deliberately: to be able to test both static and non-static rpc with 1 build. we might have to change this a bit (e.g. because of the new meta data step), but this way it will break the static rpc tests.
cc @vigoo

test-components/js-4/out/.gitkeep
noise64, line 1: where are these coming from?

noise64: ah, ok, so they are only for non-manifested ones?

test-components/ts-rpc/components-ts/caller/golem.yaml
mschuwalow: Makes sense. I missed the copying step for the composed version of the components.
There is a different problem with this, will explain during standup

vigoo (approved):

PR #1487, opened 2025-04-04
bug fixes and clean ups

afsalthaj:
* Make sure no conflicting types pass the compile (a follow up to [fixing the bug](https://github.com/golemcloud/golem/pull/1483) reported by @justcoon), where he mentioned that Rib did allow some wrong types to compile (and later failed during interpretation).
* A few more clean ups in error messages, and add more negative tests
* Cleanup and improvements in RibRepl
    * Expose APIs that app manifest can dynamically load component dependency to rib repl session (probably - we will come to know more when we really integrate)
    * Make sure a REPL can be run in a configurable way in the integration test context
    * Add better docs


golem-rib/src/inferred_type/unification.rs
```diff
@@ -721,10 +721,6 @@ pub fn unify_with_required(
                         inferred_type_left.clone(),
                         inferred_type_right.clone(),
                     ]))
-                } else if inferred_type_left.is_string() && inferred_type_right.is_number() {
```
noise64, line 724: 👍

afsalthaj: I am happy with "foo" if you give me 1 and "foo" :D 
Python inspired :P

noise64 (approved):

PR #1505, opened 2025-04-09
Bugfix: Fix variable conflicts when metadata has enums and variants

### LINKED ISSUE #1491: Rib: Enable variable declarations (let bindings) to reuse names already defined as functions, enums, or variants
**Description:**

Below is API for `golem:todolist/api.{add}` 

```rib
response: |-
  let title: string = request.body.title;
  let deadline: option<string> = request.body.deadline;
  let result = golem:todolist/api.{add}({title: title, priority: high, deadline: deadline});
  {status: 200: u64, body: "Success"}
worker-name: |-
  "badworker-todo-list-2"
```

The compilation error is: happens even though deadline option<string>

```
error: API Definition Service - Error: 400 Bad Request, Rib compilation error: error in the following rib found at line 3, column 39
`{title: title, priority: high(), deadline: deadline()}`
found within:
`golem:todolist/api.{add}({title: title, priority: high(), deadline: deadline()})`
cause: type mismatch at path: `deadline`. expected option<string>
invalid argument to the function `golem:todolist/api.{add}`
```

This error happens in the complier side, in case deadline exists, but is not needed like below, responsemappinginput has deadline and it is nullable 
```
let deadline: option<string> = request.body.deadline;
let result = golem:todolist/api.{add}({title: title, priority: high, deadline: none});
```

afsalthaj:
Fix #1491 

Let's say we have a metadata with the following enum and variant

```
enum foo {
  x,
  y,
  z
}

```

then the following should work  (sounds obvious, but it never did)

```rust
let x = 1;
let y = 2;
add(x, y)
```

Given `process` takes the enum `foo`,

```rust
let x = x;
let y = x;
process-enum-fn(x, y)

```

This scoping was handled before. However during enum resolution phase (this is when compiler assigns the type of `enum` to identifier with the same `enum` name),  the  actual `variable-id` was not taken into account - that it only compared names. So this is only a 1 liner fix.  

In the above case the variable-id of the identifier `x` (within the add function) will be a local identifier, and hence during enum resolution, it can be skipped.

This is the same for variants with no arguments too

### DIFF
--- a/golem-rib/src/type_inference/enum_resolution.rs
+++ b/golem-rib/src/type_inference/enum_resolution.rs
@@ -42,7 +42,7 @@ mod internal {
                 type_annotation,
             } = expr
             {
-                if enum_cases.cases.contains(&variable_id.name()) {
+                if enum_cases.cases.contains(&variable_id.name()) && !variable_id.is_local() {
                     *expr = Expr::Call {
                         call_type: CallType::EnumConstructor(variable_id.name()),
                         generic_type_parameter: None,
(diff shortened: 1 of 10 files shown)


golem-rib/src/type_inference/enum_resolution.rs
```diff
@@ -71,14 +71,17 @@ mod internal {
...
-                    // Retrieve the possible no-arg variant from the registry
-                    let key = RegistryKey::FunctionName(variable_id.name().clone());
-                    if let Some(RegistryValue::Value(AnalysedType::Enum(typed_enum))) =
-                        function_type_registry.types.get(&key)
-                    {
-                        enum_cases.push(variable_id.name());
-                        *inferred_type = inferred_type
-                            .merge(AnalysedType::Enum(typed_enum.clone()).clone().into());
+                    // If variable is local, it takes priority over being a global enum
+                    if !variable_id.is_local() {
```
afsalthaj, line 75: This is the fix for enum - if the variable id is not local, then we go ahead and assign the type

vigoo (approved):

PR #1047, opened 2024-11-01
Make string interpolation flexible

afsalthaj:
In response to https://discord.com/channels/1134448700572319785/1134448701260169249/1300769907755061268

**_and tracked as part of https://github.com/golemcloud/golem/issues/1035 which is currently in progress_**

Previously I  tested the html str responded from worker, however, not inlining HTML within Rib. Things failed when inlining within a string interpolation, as indicated by discord user.

#### Rib
```scala
let style = "font-family: Arial, sans-serif; display: flex; align-items: center; justify-content: center; height: 100vh; background: linear-gradient(135deg, #1e3c72, #2a5298); color: #ffffff; text-align: center; padding: 20px;";

let user: u64 = request.path.user-id;

let body = "
  <!DOCTYPE html>
     <html>
       <head>
      </head>
      <body style='${style}'>
         <div>
           <h1>Welcome to Golem Cloud!</h1>
           <p style='font-size: 1.5em;'>Hello, User ID: ${user}</p>
           <p style='color: #dcdcdc;'>Explore your cart contents and more on this personalized page.</p>
       </div>
    </body>
  </html>"

{Content-Type: "text/html", status: 200u64, body: body}

```

#### Result

image

NOTE: Having a complex HTML within Rib within API Definition JSON is really hard. Everything has to be 1 line. Or else use golem-cloud console Rib editor.

https://github.com/afsalthaj/golem-worker-gateway-examples/blob/main/html/api-definition.json

Issue #1462, opened 2025-03-30
Disallow string interpolation for anything other than literals


vigoo:
Disallow string interpolation for anything other than literals (strings, numbers, etc.), since concatenation of “records” and “resources” and “module instances” doesn’t make sense.

PR #1508, opened 2025-04-09
allow string concatenation with any type

afsalthaj:
Fixes #1047  #1462 

In short, with this PR the following works (previously it will throw run time error if you try to interpolate a record)

```rust
  let x = "foo";
  let y = "bar";
  let z = {foo: "baz"};
  let n: u32 = 42;
  let result = "${x}-${y}-${z}-${n}";
  result
```

will result in the following, where each value is a wasm-wave string.

```
"foo-bar-{foo: baz}-42"
```

The ticket 1462 says to allow only literals, but here _literals_ mainly intend to mean - anything that can be _safely_ interpolated. In that case, `enums`  or primitives (number, string, char) and interestingly developers may want to stringify a no-arg variant term too (whose value can be sometimes available only at runtime disallowing a compile time check as to whether or not a particular variant value is amongst the ones with no-args). In details, this approach is suboptimal, and therefore the above approach which is both intuitive and hence user friendly.

### DIFF
--- a/golem-rib/src/interpreter/rib_interpreter.rs
+++ b/golem-rib/src/interpreter/rib_interpreter.rs
@@ -1938,7 +1938,7 @@ mod tests {
 
         assert_eq!(
             result.get_val().unwrap().value,
-            Value::String("foo-bar-{foo: baz}-42".to_string())
+            Value::String("foo-bar-{foo: \"baz\"}-42".to_string())
         );
     }
 
--- a/golem-rib/src/interpreter/rib_interpreter.rs
+++ b/golem-rib/src/interpreter/rib_interpreter.rs
@@ -1400,16 +1400,23 @@ mod internal {
         interpreter_stack: &mut InterpreterStack,
         arg_size: usize,
     ) -> Result<(), String> {
-        let literals = interpreter_stack.try_pop_n_literals(arg_size)?;
+        let value_and_types = interpreter_stack.try_pop_n_val(arg_size)?;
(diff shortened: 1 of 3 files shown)


jdegoes (approved):

vigoo:
I agree it should work with enums and zero-argument variants, but I'm not sure about allowing it to everything like the record in the provided example. I can't imagine when I would like to embed a WAVE representation of a record in a string in Rib and have the feeling that by allowing it we just make these errors runtime - the user will only realize this is not what he wanted when the Rib script generates weird strings.

afsalthaj:
@vigoo  Hmmm. 

If we forget about variants (i.e, it is almost impossible to allow just zero arg variants), we can have a proper implementation of just allowing enums and primitives - and get away from `ToString` approach for any type. Variants with no-args may sound obvious, but in detail its just impossible. 

```rust
let x = request.body.variant-value;
"${x}-foo"
```

where x can be any of the following as an example

```json
{ "none": null }
{ "any": null }
{ "restricted": ["one", "two", "three"] }
```


If we have requirements such variants should also be able to stringified, the neat solution is to do something as above.

The thing is, no language has that restriction in a string concatenation except for libraries that come up safe string interpolation, that each elem should have an instance of SafeDisplay etc.

jdegoes:
@vigoo I think it will be quite useful for logging values. 

Rib doesn't have a log function but that will be useful at some point soon.

PR #1511, opened 2025-04-10
clean ups in API definition (for CORS and security) - to make DX integration easier

afsalthaj:
This is in-progress (this is not code refactor, but main intention is to make peripherals easier to understand and thereby make DX integration easier)

Code evaluation and simplifying certain fields to make DX integration easier
* Fixing comments about the fields in gateway binding fields. I will be adding more documentation here.
* Removing unused (unused after https://github.com/golemcloud/golem/pull/1444) on CORS in the request front-end data. 
* Removing redundant `cors` field anywhere else too (such as within the binding, and outside the binding, there was CORS - probably making it closer to OAS - but this is not needed in native API definition. This can be removed.
* (Yet to be done) Removing higher level security field to match Open API Specific format, where securities are listed globally, and every route should be picking only 1 among them. I think this is confusing and less user friendly. If user is importing OAS, they might need to do this (or they might have this already), but the native API definition doesn't need to follow it exactly. Export to OAS will still be lossless with this data.

TODO:
- [x] Make sure cors tests are expanded with request with and without origin.

### DIFF
--- a/golem-worker-service-base/src/api/api_definition.rs
+++ b/golem-worker-service-base/src/api/api_definition.rs
@@ -230,7 +230,6 @@ pub struct RouteRequestData {
     pub method: MethodPattern,
     pub path: String,
     pub binding: GatewayBindingData,
-    pub cors: Option<HttpCors>,
     pub security: Option<String>,
 }
 
@@ -251,7 +250,6 @@ impl RouteRequestData {
             path,
             binding,
             security,
-            cors: self.cors,
         })
     }
 }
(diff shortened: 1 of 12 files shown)


afsalthaj:
While CORS is mainly checked against `OPTIONS` (preflight of browsers), it is also mandatory to check origin for all request (all methods) and not rely just on browsers decision using preflight response. This was missing.  I added it.

Also we need extra checks for a preflight request. These are also added.

golem-worker-service-base/src/api/api_definition.rs
```diff
@@ -479,7 +458,7 @@ impl From<HttpMiddlewares> for MiddlewareData {
 
         for i in value.0.iter() {
             match i {
-                HttpMiddleware::AddCorsHeaders(cors0) => cors = Some(cors0.clone()),
+                HttpMiddleware::Cors(cors0) => cors = Some(cors0.clone()),
```
afsalthaj, line 461: CORS is not just sending preflight response or  adding response headers to other requests, although most browsers acknowledge it and disallow further requests (after preflight)
So renamed it to CORS, that it has things to do during `middleware_in` and `middleware_out`. The impl is fully inspired from `poem` (further down below)

golem-worker-service-base/src/gateway_api_definition/http/api_definition.rs
```diff
@@ -82,29 +82,31 @@ impl HttpApiDefinition {
...
-            for security_scheme_reference in security_schemes {
-                let security_scheme = security_scheme_service
-                    .get(
-                        &security_scheme_reference.security_scheme_identifier,
-                        namespace,
-                    )
-                    .await
-                    .map_err(ApiDefinitionError::SecuritySchemeError)?;
+        let security_schemes_in_definition = request
+            .routes
```
afsalthaj, line 86: The outer level security was not helping users before, and acted as something _to keep it as close to open-api-spec_. But I have decided to remove it, and we can still generate lossless open-api spec. Only route level security is needed (and that's all we documented too)

noise64 (approved):

PR #1525, opened 2025-04-14
Static validation of query and path params, and important improvements including a bug fix

### LINKED ISSUE #1463: Static validation of query and path parameters than at runtime for Rib
Static validation of query and path parameters than at runtime for Rib. It will reduce the difficulty in debugging when user accidentally uses request.path.* in Rib, and passed a request.query parameter.

afsalthaj:
Fixes #1463 

* Add static validations for query and path look ups. Currently an API definition with route such as `/foo/{userid|` and a rib script `requedt.query.user-id` (which can never succeed in runtime) can be successfully registered. And so is the case when it is `/foo?{user-id}` and `request.path.user-id`. With this PR, thee are now validated, and will throw validation errors during API definition registration.

See `golem-worker-service-base/gateway_api_definition/api_definition.rs` to see the specific changes.

Following important clean ups also applied this time.

As a summary, much more details in `RibCompilationError` and introduction `RibRuntimeError`, along with handling panics that may happen in the rib runtime, with reduced number of clones that was taking place in api-gateway

Details:

* fix a bug where types where not unified when complex types where inlined as argument to instance
* Spotted a lot of avoidable clones introduced recently, and these are reduced/removed as much as possible in the api gateway hot path. The one that existed for a long time was the `clone` on rib-byte-code and `rib-input` which is avoided almost 100 percent. The `WorkerDetail` struct was cloned a couple of times, and these are avoided too.
* `RibInterpreter` returns a `Result<ValueAndType, RibRuntimeError>` instead of `Result<ValueAndType, String>` where RibRuntimeError is a fine grained structured error message (within the time limits), which also ensure to keep the structure of invocation of error (which is WorkerServiceError that has GolemError). It is upto the client of interpreter to use the structure or not. Currently there is a reasonable `Display` for rib interpreter error which is used in Repl and gateway. 
* the math operations were not handling overflow of operation, and it was neither handled divide by zero. These are handled.
*  There is a RibCompilationError which is much more structured - it can be RibParseError, actual compilation-error due to type mismatch, internal byte-code generation error etc.


golem-rib/src/interpreter/rib_runtime_error.rs
```diff
@@ -0,0 +1,234 @@
...
+    },
+    IndexOutOfBound {
+        index: usize,
+        size: usize,
+    },
+    InfiniteComputation {
+        message: String,
+    },
+    InputNotFound(String),
+    InvariantViolation(InvariantViolation),
```
afsalthaj, line 31: Hopefully in future, we can confidently unwrap for violation of the the invariant.
Also in future, we can add or remove more errors into this enum - to get a clear idea of which one can be user exposed or not.

For now, its just better than what was before

golem-worker-service-base/src/gateway_api_definition/http/api_definition.rs
```diff
@@ -839,6 +882,203 @@ impl CompiledRoute {
...
+                        }
+                    })
+                    .collect::<Vec<_>>(),
+
+                _ => vec![],
+            }
+        } else {
+            vec![]
+        }
+    }
```
afsalthaj, line 1081: These are just moved to a single module.

Cargo.lock
```diff
@@ -3645,6 +3645,7 @@ dependencies = [
 name = "golem-worker-service-base"
 version = "0.0.0"
 dependencies = [
+ "anyhow",
```
noise64, line 3648: is this still needed after reverting the anyhow usage?

golem-rib-repl/src/rib_repl.rs
```diff
@@ -233,6 +232,21 @@ impl RibRepl {
     }
 }
 
+#[derive(Debug)]
+pub enum RibExecutionError {
+    RibCompilationError(RibCompilationError),
+    RibRuntimeError(RibRuntimeError),
+}
+
+impl Display for RibExecutionError {
```
noise64, line 241: let's add an Error instance too

golem-rib/src/compiler/byte_code.rs
```diff
@@ -14,14 +14,53 @@
...
+    AnalysedTypeConversionError(String),
+    PatternMatchDesugarError,
+    RangeSelectionDesugarError(String),
+    UnexpectedTypeError {
+        expected: TypeHint,
+        actual: TypeHint,
+    },
+}
+
+impl Display for RibByteCodeGenerationError {
```
noise64, line 39: error instance?

golem-rib/src/interpreter/rib_runtime_error.rs
```diff
@@ -0,0 +1,234 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum InvariantViolation {
+    InsufficientStackItems(usize),
+    InternalCorruptedState(String),
+    InstructionJumpError(InstructionId),
+}
+
+impl Display for RibRuntimeError {
```
noise64, line 176: error impl?

golem-worker-service-base/Cargo.toml
```diff
@@ -76,6 +76,7 @@ tracing-subscriber = { workspace = true }
 url = { workspace = true }
 urlencoding = { workspace = true }
 uuid = { workspace = true }
+anyhow = "1.0.97"
```
noise64, line 79: is this needed still? if so, please add to workspace cargo

afsalthaj: No i will clean up the dependencies.

afsalthaj: Excuse this accidental addition. I just disabled this auto update of cargo.toml in IDE when I simply add an import. 
Anyway the dependency is removed :)

noise64 (approved):

PR #1527, opened 2025-04-15
 Allow reusing plugin (name,version) pairs 

### LINKED ISSUE #1300: Recreating a Plugin with same name and version after deleting fails.
**Description:**

After deleting a plugin, attempting to create a new plugin with the *exact same name and version* fails consistently. Workaround is to restart the Golem backend server. I have verified this issue using both the `golem-cli` and direct POST requests to the backend's plugin endpoint.

**Steps to Reproduce:**

1.  Create a plugin with a specific name and version using either `golem-cli plugin register` or a POST request to the backend's plugin endpoint.
2.  Verify that the plugin is successfully created and listed by direct GET request or `golem-cli plugin list`. 
3.  Delete the plugin and verify that the plugin is successfully deleted. (direct GET request)
5.  Attempt to create a new plugin with the *same exact name and version* as the previously deleted plugin, using either `golem-cli plugin register` or a POST request to the backend's plugin endpoint. 

**Workaround:**

*   Restarting the Golem backend server resolves the issue. After a restart, the plugin can be created successfully. (In previous version, It did not need a restart of the server, just waiting a few minutes would resolve the issue (Maybe).)

**Environment:**

*   Golem Backend Server version 1.1.16
*   `golem-cli` version 1.1.16

**Additional Information:**

*   This bug is probably due to Golem backend not properly cleaning up or releasing some internal state related to the deleted plugin.

mschuwalow:
resolves #1300 

Introduces a synthetic plugin_id primary key in the plugin table and uses that when referencing plugins in installations.
This id is completely hidden from users and doesn't show up in apis at all.

The (name, version) unique index is now partial, allowing to reusing the pair if the previous value was unregistered / deleted.
There is a new internal (grpc-only) api get_plugin_by_id that is used to resolve plugins even when they are unregistered.


openapi/golem-service.yaml
```diff
@@ -4544,19 +4544,6 @@ components:
           type: string
         invocationContext:
           type: string
-        allowOrigin:
```
mschuwalow, line 4547: This a leftover from a different pr. @afsalthaj was this intentional?

After this pr I'll open a small followup adding an up-to-date check for the openapi.yaml to our ci

integration-tests/tests/api/plugins.rs
```diff
@@ -566,3 +560,85 @@ async fn app_plugin(deps: &EnvBasedTestDependencies, _tracing: &Tracing) {
 
     assert_eq!(response, Ok(vec![Value::U64(2)]))
 }
+
+/// Test that a plugin can be recreated after deleting it
+#[test]
+async fn recreate_plugin(deps: &EnvBasedTestDependencies, _tracing: &Tracing) {
```
mschuwalow, line 566: These two assert the now fixed issues in the ticket

golem-test-framework/src/components/component_service/mod.rs
```diff
@@ -90,6 +92,82 @@ pub trait ComponentServiceInternal: Send + Sync {
...
+            installed_plugins: stream::iter(component.installed_plugins)
+                .then(async |install| {
+                    let (plugin_name, plugin_version) = match install.plugin {
+                        ReferencedPlugin::Registered(inner) => {
+                            (inner.plugin_name, inner.plugin_version)
+                        }
+                        _ => Err(anyhow!("reference to unregistered plugin"))?,
+                    };
+
+                    // TODO: This should error if this is none, but currently the http api cannot resolve the id, but the types force us to provide something here.
```
mschuwalow, line 156: This is an ugly side effect of us abstracting over the http and grpc api in the test framework (we have a number of these already).
For now the tests are happy, but we should give it some though how we want to improve the test framework when it comes to this.

golem-component-service-base/src/api/mapper.rs
```diff
@@ -0,0 +1,90 @@
...
+
+use crate::api::dto;
+use crate::model as domain;
+use crate::service::plugin::{PluginError, PluginService};
+use futures::{stream, StreamExt, TryStreamExt};
+use golem_common::model::component::ComponentOwner;
+use golem_common::model::plugin::{PluginInstallation, PluginScope};
+
+#[async_trait::async_trait]
+pub trait ApiMapper<Owner: ComponentOwner>: Send + Sync {
```
mschuwalow, line 25: This thing translates between our internal domain and the public api we want to expose to users.

openapi/golem-service.yaml
```diff
@@ -7482,6 +7468,50 @@ components:
             $ref: '#/components/schemas/ProducerField'
       required:
       - fields
+    ReferencedPlugin:
```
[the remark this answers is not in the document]
mschuwalow: Replaced with a simple flag to keep backwards compat in the api

golem-component-service-base/src/service/transformer_plugin_caller.rs
```diff
@@ -162,3 +166,29 @@ impl<Owner: ComponentOwner> TransformerPluginCaller<Owner> for TransformerPlugin
         Ok(body.to_vec())
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SerializableComponent {
```
vigoo, line 172: Why did you introduce this "clone" of the `Component` type in `golem_service_base` here? It was already serializable

golem-test-framework/src/model.rs
```diff
@@ -0,0 +1,42 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use golem_common::model::plugin::{DefaultPluginScope, PluginTypeSpecificDefinition};
+
+#[derive(Debug, Clone)]
+pub struct PluginDefinitionCreation {
```
vigoo, line 18: Why do we need a test-framework specific type for this?

vigoo (commented):
Looks good, I just two minor concerns about introducing "duplicated" types

golem-test-framework/src/model.rs
mschuwalow: This serves as an abstraction over the grpc and rest api creation of plugins.

The functions using this used to use the domain type directly, but this no longer works as there is now an extra id / conversion process happening https://github.com/golemcloud/golem/pull/1527/files/d53bd412a1c31aa46e938680dfdb5ee78ae74819#diff-f6c708d1157caac855950eb82eb62277e56663669630375743f5c720497f7da8L1411.

Instead of trying to generalize these types to be usable in both contexts I decided it would be clearer to specialize them. What we have now is:

```
(this thing) -----> rest dto type ----> component service domain type
             \                       /
              \---> grpc dto type - /
```

golem-component-service-base/src/service/transformer_plugin_caller.rs
mschuwalow: The name is not great tbh.

The purpose of this thing is to be a projection of the component that is exposed to transformers.
The internal Component type in golem_service_base (which is not / doesn't need to be serializable anymore btw) includes plugin (installations) via their id, which is something that I wanted to hide from the api for now.
So this thing is a projection of the component that doesn't include this information.
If we do include it, we need to decide whether:
* we expose the plugin_id (which is otherwise not reachable from the api)
* we expose plugin_name + version (which requires a conversion here and we need to figure out how to encode "dangling" plugins)
* some other strategy

I decided to delay making this decision for now and just not expose this information until it is needed.

openapi/golem-service.yaml
afsalthaj: I hope you resolved this doubt yourself recently :)

golem-component-service/db/migration/postgres/011__plugin_id.sql
```diff
@@ -0,0 +1,30 @@
...
+
+UPDATE component_plugin_installation
+    SET plugin_id = p.id
+    FROM plugins AS p
+    WHERE p.name = plugin_name AND p.version = plugin_version;
+
+ALTER TABLE component_plugin_installation ALTER COLUMN plugin_id SET NOT NULL;
+
+ALTER TABLE component_plugin_installation
+    DROP CONSTRAINT IF EXISTS component_plugin_installation_plugin_name_plugin_version_fkey,
```
mschuwalow, line 24: these names match what postgres should pick by default, but after deploying we should double check that these where indeed deleted (made if IF EXISTS so the deployment as a whole should not fail)

vigoo (approved):

PR #1528, opened 2025-04-15
Custom durability related fixes

vigoo:
Fixes custom durability payload serialization

Changes the behavior of `PersistenceLevel::PersistNothing` mode. 
Previously side-effects in these sections were always re-executed during replay, and their results were not persisted.
With this PR we always persist the results (for observability reasons) and also guarantee that side-effects do not get "randomly" reexecuted during replay mode. Instead the guest code is responsible for not executing the persist-nothing region during replay, for example by wrapping the whole thing using the durability host API. Not doing so fails the worker with divergence errors and this is expected.

Corresponding golem-wit PR: https://github.com/golemcloud/golem-wit/pull/56


golem-api-grpc/proto/golem/worker/public_oplog.proto
```diff
@@ -280,3 +286,9 @@ message OplogEntryWithIndex {
   uint64 oplog_index = 1;
   OplogEntry entry = 2;
 }
+
+enum PersistenceLevel {
+    PERSIST_NOTHING = 0;
+    PERSIST_REMOTE_SIDE_EFFECTS = 1;
+    SMART = 2;
```
noise64, line 293: 🧠

golem-test-framework/src/dsl/mod.rs
```diff
@@ -1352,11 +1354,11 @@ impl<T: TestDependencies + Send + Sync> TestDsl for T {
     }
 
     async fn check_oplog_is_queryable(&self, worker_id: &WorkerId) -> crate::Result<()> {
-        let _oplog = TestDsl::get_oplog(self, worker_id, OplogIndex::INITIAL).await?;
+        let oplog = TestDsl::get_oplog(self, worker_id, OplogIndex::INITIAL).await?;
 
-        // for (idx, entry) in oplog.iter().enumerate() {
-        //     debug!("#{}: {entry:#?}", idx + 1);
-        // }
+        for (idx, entry) in oplog.iter().enumerate() {
```
noise64, line 1359: do you want to keep this?

noise64 (approved):

golem-test-framework/src/dsl/mod.rs
vigoo: Yes because now it has a nice output that does not take infinite space :)

golem-worker-executor-base/tests/durability.rs
```diff
@@ -0,0 +1,115 @@
...
+use crate::{LastUniqueId, Tracing, WorkerExecutorTestDependencies};
+use golem_test_framework::dsl::TestDslUnsafe;
+
+inherit_test_dep!(WorkerExecutorTestDependencies);
+inherit_test_dep!(LastUniqueId);
+inherit_test_dep!(Tracing);
+
+#[test]
+#[tracing::instrument]
+async fn custom_durability_1(
```
mschuwalow, line 36: Does it make sense to add two more tests for failure cases (also as documentation for how this behaves):

* worker fails after `persist_infallible` is called
* worker fails after the side effect is performed (the http callback in this case) but before `persist_infallible` is called

golem-test-framework/src/dsl/mod.rs
noise64: Would it worth putting the loop inside enabled(debug)?

golem-worker-executor-base/tests/durability.rs
vigoo: There is an existing test in `transactions` that now demonstrates this (more or less) so I think we are good for now. I'm going to use this new behavior in `golem-llm` and add new test cases here for whatever else I find.

golem-test-framework/src/dsl/mod.rs
vigoo: Why? :) 
This is a very specific check executed in the end of executor tests only, not affecting anything else

noise64: sorrry, missed that this is test code

PR #1534, opened 2025-04-16
Make service start functions "Send"

noise64:
Required for implementing "auto local server start" in golem-cli

### DIFF
--- a/golem-service-base/src/db/sqlite.rs
+++ b/golem-service-base/src/db/sqlite.rs
@@ -358,7 +358,8 @@ pub async fn migrate(
     info!("DB migration: sqlite://{}", config.database);
     let mut conn = SqliteConnection::connect_with(&config.connect_options()).await?;
     let migrator = sqlx::migrate::Migrator::new(migrations).await?;
-    migrator.run(&mut conn).await?;
+    // See: https://github.com/launchbadge/sqlx/issues/954, Send required for golem-cli
+    futures::executor::block_on(migrator.run(&mut conn))?;
     let _ = conn.close().await;
     Ok(())
 }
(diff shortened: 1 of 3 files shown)


mschuwalow (approved):

golem-service-base/src/db/sqlite.rs
```diff
@@ -358,7 +358,8 @@ pub async fn migrate(
     info!("DB migration: sqlite://{}", config.database);
     let mut conn = SqliteConnection::connect_with(&config.connect_options()).await?;
     let migrator = sqlx::migrate::Migrator::new(migrations).await?;
-    migrator.run(&mut conn).await?;
+    // See: https://github.com/launchbadge/sqlx/issues/954, Send required for golem-cli
+    futures::executor::block_on(migrator.run(&mut conn))?;
```
mschuwalow, line 362: Not a fan of blocking the async runtime thread, but as this is only done during startup probably fine

golem-worker-executor-base/src/lib.rs
```diff
@@ -100,7 +100,7 @@ const VERSION: &str = golem_version!();
 pub struct RunDetails {
     pub http_port: u16,
     pub grpc_port: u16,
-    pub epoch_thread: std::thread::JoinHandle<()>,
+    pub epoch_thread: std::sync::Mutex<Option<std::thread::JoinHandle<()>>>,
```
mschuwalow, line 103: Can we move the mutex + Option into the cli?
I don't like making the types less precise here as we will always have a joinhandle.

golem-service-base/src/db/sqlite.rs
noise64: I hope nobody is fan of that :) see the linked issue, reasoning is what you wrote

golem-worker-executor-base/src/lib.rs
noise64: - we cannot
- we expose async service creation methods, and i would say, it is preferable to make them returning Send futures in general, which makes them usuable in more situations
- the type is not less precise at all, to actually _join_ a join handle jo have to consume it, which means you have to take it out from the mutex, i would argue that even without the mutex it should have been an option, so one can join it without destroying the whole RunDetails

PR #1547, opened 2025-04-22
Component search (multiple names with each one having specific details)

### LINKED ISSUE #1469: Batch lookup for returning a set of component metadata based on names
with optionally returning a specific or latest version (compared to returning all);

afsalthaj:
Fixes #1469 

The PR is implemented based on the features required by @noise64  and made the search a POST. This is what we and noise thought so.

### DIFF
--- a/golem-component-service/src/api/component.rs
+++ b/golem-component-service/src/api/component.rs
@@ -24,8 +24,7 @@ use golem_common::model::ComponentFilePathWithPermissionsList;
 use golem_common::model::{ComponentId, ComponentType, Empty, PluginInstallationId};
 use golem_common::recorded_http_api_request;
 use golem_component_service_base::model::{
-    ComponentSearchParameters, DynamicLinking, InitialComponentFilesArchiveAndPermissions,
-    UpdatePayload,
+    ComponentSearch, DynamicLinking, InitialComponentFilesArchiveAndPermissions, UpdatePayload,
 };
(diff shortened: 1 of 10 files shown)


golem-component-service/src/api/component.rs
```diff
@@ -421,6 +422,29 @@ impl ComponentApi {
         record.result(response)
     }
 
+    #[oai(path = "/search", method = "post", operation_id = "search_components")]
+    async fn get_components_batch_post(
+        &self,
+        components_search: Json<Vec<ComponentSearchParameters>>,
```
mschuwalow, line 428: I would suggest not using a raw array as the top level json type. That will make it very difficult to change this endpoint in a backward compatible way (e.g. adding pagination).
Better would be to have an object as the top level (even when it only has a single array field for now)

mschuwalow (approved):

golem-component-service/src/api/component.rs
afsalthaj: Makes sense

afsalthaj: Resovled

afsalthaj:
Added sqlite and postgres tests for the query. I missed it.

Just realised service tests works with just sqlite repo. And repo tests deals with both postgres and sqlite. This is ok, but 
I wish it was just just service tests that works with both postgres and sqlite repo as dependencies. May be there is more reasoning to it. And merging this PR.

PR #1560, opened 2025-04-25
Propagate TypeOrigin in Rib

### LINKED ISSUE #1523: Propagate TypeOrigin in InferredType tree in Rib
We don't have precise compile time errors, resulting in most of the errors being "ambiguous types inferred" instead of "expecting x, but found y". This became more evident in REPL.

### Solution
Specifying `TypeOrigin` in InferredType allows clean and precise compilation errors, and allows removing duplicate error messages.
This is now much possible given that we have `SourceSpan` in Expr tree with the recent releases.

It's much better if we are able to squeeze this as early as possible especially we managed to finish Golem REPL, and we don't want complex compilation error messages in REPL

```rust
#[derive(Debug, Clone, Ord, PartialOrd)]
enum TypeOrigin {
    DeclaredAt(SourceSpan),
    Default,
    RequiredByCall(SourceSpan),
    Other
}

#[derive(Debug, Clone, Ord, PartialOrd)]
pub struct InferredTypeWithOrigin {
    pub inferred_type: Box<InferredType>,
    pub origin: TypeOrigin,
}


#[derive(Debug, Clone, Ord, PartialOrd)]
pub enum InferredType {
    Bool,
    S8,
    U8,
    S16,
    U16,
    S32,
    U32,
    S64,
    U64,
    F32,
    F64,
    Chr,
    Str,
    List(Box<WithOrigin>),
    Tuple(Vec<WithOrigin>),
    Record(Vec<(String, WithOrigin)>),
    Flags(Vec<String>),
    Enum(Vec<String>),
    Option(Box<WithOrigin>),
    Result {
        ok: Option<Box<WithOrigin>>,
        error: Option<Box<WithOrigin>>,
    },
    Variant(Vec<(String, Option<WithOrigin>)>),
    Resource {
        resource_id: u64,
        resource_mode: u8,
    },
    Range {
        from: Box<WithOrigin>,
        to: Option<Box<WithOrigin>>,
    },
    Instance {
        instance_type: Box<InstanceType>,
    },
    OneOf(Vec<WithOrigin>),
    AllOf(Vec<WithOrigin>),
    Unknown,
    // Because function result can be a vector of types
    Sequence(Vec<WithOrigin>),
}

```

And we pass `InferredTypeWithOrigin`  all the way.

Now what this allows is as follows

Let's say we get a conflicting error message such as `InferredType::AllOf({u32, required_by_call("foo"),  {u64, declared_at(line_36)}))`

In the above case, we have a clear way of saying "expected u32, but declared as u64 at line number 36".

This also allows us to pick and prioritise types

afsalthaj:
Fixes #1523 

Progress:

* Better default type handling
* Improved compilation error messages (this will be ongoing - can't do everything in a single go)

# What's done?

I request you to try out the following Rust code in Rustrover, with inline type-hints enabled

Copy the code and change `none => x` under `bar` to none => 1. Notice how Rust infers types: with none => x, both foo and bar are u64; with none => 1, they fall back to i32. This PR makes Rib behave the same.

```rust

    let x = 1;
    let x = x;

    let foo = match Some(x + 1) {
        Some(x) => x,
        none => x
    };

    let z: Option<u64> = None;

    let bar = match z {
        Some(x) => x,
        none => x
    };

    foo;
    bar;


```

The example is slightly convoluted to highlight nuances like how x + 1 forces 1 to become u64, showing that Rib, like Rust, **avoids unsafe runtime casts.**

# From a REPL perspective

Let's see how this works out in REPL

```rust
>>> let x = 1
>>> x
1
>>> call_worker_function_that_takes_u64(x)
"success"
```

With this PR the above snippet works. It sounds obvious ofcourse but never worke. Now under the hood, there is a switch of types happening - as shown below

```rust
>>> let x = 1
>>> x
1
>>>:type x
s32
>>> call_worker_function_that_takes_u64(x)
>>>:type x
u64
```

In the existing release version the compiler ask you to annotate the very first line before proceeding - not needed anymore.

# Default types and global variables

`request.path.user` if not annotated is a `string` (a long time feature now)

```rust
let input = request.path.user;
foo(input)
```

But what if `foo` takes a `u64`. Current rib fails wit  "expected - u64, found - string". If the type was annotated as u64 this would work. Or if the feature is disabled (configurable), then Rib would have compiled, but gateway always enables this feature.

With this PR,  user can still ignore to annotate a type for `request.path.user` and will be correctly inferred as u64.

# Compilation Errors

For simple scenarios, such as if you pass u64 instead of a string to a function, you are already getting precise error messages, such a the following

```rust
error in the following rib found at line 4, column 22
`"afsal"`
found within:
`foo("afsal")`
cause: type mismatch. expected u32, found string
invalid argument to the function `foo`

```

or

```rust
error in the following rib found at line 4, column 50
`product`
found within:
`golem:it/api.{cart(user_id).add-item}(product)`
cause: type mismatch at path: `quantity`. expected u32
invalid argument to the function `golem:it/api.{cart(user_id).add-item}`
```

This was done some time ago, by introducing type-checker module before unification. But still there are a few scenarios that type checker may find it hard to identify, which one is expected and which one is actual. 

This PR opens the door to making the errors more precise. For example, the following:

```rust
 let x: option<u64> = some(1);
          match x {
            some(x) => x,
            none => "none"
          }
```

```
 error in the following rib found at line 4, column 24
            `x`
            cause: type mismatch. expected string, found u64
            expected string based on pattern match branch at line 5 column 21
```

However, there was a limit to which I could improve altogether. With type-origin propagation, we get more details to _decide_ things, but those details are not yet enough to get a 100 percent clean compilation errors. We need to jump 1 more hurdle which is https://github.com/golemcloud/golem/issues/1567 to get there, which may take time.


golem-rib/src/interpreter/rib_interpreter.rs
```diff
@@ -1697,8 +1697,8 @@ mod tests {
     #[test]
     async fn test_interpreter_variable_scope_0() {
         let rib_expr = r#"
-               let x: u64 = 1;
-               let y = x + 2u64;
```
afsalthaj, line 1701: No need to type annotate anymore!

vigoo (approved):

PR #1561, opened 2025-04-25
Fix component grpc serialization

mschuwalow:


mschuwalow:
Alternatively we can expose project_id on the ComponentOwner trait.
I would prefer to avoid exposing it more in oss, but no strong feelings.

noise64 (approved):

golem-component-service-base/src/model/component.rs
```diff
@@ -117,10 +117,8 @@ impl<Owner: ComponentOwner> From<Component<Owner>> for golem_service_base::model
     }
 }
 
-impl<Owner: ComponentOwner> From<Component<Owner>>
-    for golem_api_grpc::proto::golem::component::Component
-{
-    fn from(value: Component<Owner>) -> Self {
+impl From<Component<DefaultComponentOwner>> for golem_api_grpc::proto::golem::component::Component {
+    fn from(value: Component<DefaultComponentOwner>) -> Self {
```
afsalthaj, line 121: What was the issue? One is a generic type and the other is concrete

PR #1564, opened 2025-04-28
Fixes for golem cli api def integration

noise64:
- return invocation_context for GET api definitions (ro wroker and file server types), so it can be diffed in golem-cli
- wrap all calls of Component::form_binary with spawn_blocking, so golem executable won't hang on compilation service


afsalthaj (approved):

golem-component-compilation-service/src/service/compile_worker.rs
```diff
@@ -179,12 +180,20 @@ impl CompileWorker {
             .await?;
 
             let start = Instant::now();
-            let component = Component::from_binary(&engine, &bytes).map_err(|e| {
-                CompilationError::CompileFailure(format!(
-                    "Failed to compile component {:?}: {}",
-                    component_with_version, e
-                ))
-            })?;
+            let component = spawn_blocking({
```
afsalthaj, line 183: 👍

PR #1568, opened 2025-04-29
Exposed plugin installation management as a batch operation

vigoo:
Required for https://github.com/golemcloud/golem-cli/issues/105

### DIFF
--- a/golem-component-service-base/src/repo/component.rs
+++ b/golem-component-service-base/src/repo/component.rs
@@ -34,6 +34,7 @@ use golem_service_base::repo::plugin_installation::{
 };
 use golem_service_base::repo::RepoError;
 use sqlx::{Postgres, QueryBuilder, Row, Sqlite};
+use std::collections::{HashMap, HashSet};
 use std::fmt::{Debug, Formatter};
 use std::marker::PhantomData;
 use std::result::Result;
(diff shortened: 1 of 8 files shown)


golem-component-service-base/src/repo/component.rs
```diff
@@ -1958,98 +1929,15 @@ impl<Owner: ComponentOwner> ComponentRepo<Owner>
...
         &self,
         owner: &<<Owner as ComponentOwner>::PluginOwner as PluginOwner>::Row,
         component_id: &Uuid,
-        plugin_installation_id: &Uuid,
+        actions: &[PluginInstallationRepoAction<Owner>],
     ) -> Result<u64, RepoError> {
         let mut transaction = self
             .db_pool
-            .with_rw("component", "uninstall_plugin")
+            .with_rw("component", "apply_plugin_installation_changes")
```
noise64, line 1940: apart from using a transaction, i think this should hold a lock (e.g. SELECT FOR UPDATE) representing the component, otherwise parallel calls will result in mixed applications.

vigoo: This PR is doing the update exactly the same way as before - just applying multiple changes at once (not saying it was not bad before)

noise64: it was :)

noise64: i'm fine with postponig the fix, but do not be surprised that this can misbehave until reviewing these

noise64 (approved):

afsalthaj (approved):

PR #1573, opened 2025-04-30
Followup on type-origin 

afsalthaj:


afsalthaj:
I just need to fix the following test

```

Failed tests:
 - rib::interpreter::rib_interpreter::tests::test_interpreter_pattern_match_on_tuple_with_wild_pattern (called `Result::unwrap()` on an `Err` value: RibTypeError(error in the following rib found at line 0, column 0
`record.request`
cause: type mismatch. expected record
cannot select path from record.request since it is not a record type. Found: conflicting types: string, record{path: record{user: string}}, record{user: string}

))

```

vigoo (approved):

PR #1574, opened 2025-05-01
Add more invalid record types into the list of invalid rib syntax test cases, and handle spaced path patterns in gateway

afsalthaj:


golem-rib/src/parser/rib_expr.rs
```diff
@@ -169,7 +169,7 @@ where
     >,
     Input::Position: GetSourcePosition,
 {
-    choice((attempt(flag()), attempt(record())))
+    choice((attempt(flag()), record()))
```
afsalthaj, line 172: We can have custom parser (and that too with involves real parsing in between forgetting source locations from the rest - combine allows it) written to avoid the attempt on flag and get precise error messages for `flag` too. But I don't think its worth it. The PR simply gives importance over a frequently used expression which is `record` than `flags` which by itself is a simple expression with less possibilities of syntax errors.

afsalthaj: Fixed

noise64 (approved):

PR #1579, opened 2025-05-09
Upgrade poem

afsalthaj:


jdegoes (approved):

afsalthaj:
Don't want to change the openapi spec (looks like that's changing with latest poem, which is interesting)
I raised this PR, because a couple of bugs that came through `darling` which poem depends came up when I made certain crates from golem (repl, wasm-ast, wasm-rpc etc) in other projects, and this PR did solve it

vigoo:
Upgraded in https://github.com/golemcloud/golem/pull/1646

PR #1569, opened 2025-04-29
Api Definition by Importing OpenAPI yaml in integration-test

Nanashi-lab:
closes #1559 

completely fixes the issue in #1559 

cause - 
`golem-worker-service/src/grpcapi/api_definition`

```rust
            create_api_definition_request::ApiDefinition::Openapi(definition) => {
                let converted = OpenApiHttpApiDefinition(
                    serde_json::from_str(&definition).map_err(|_| bad_request("Invalid JSON"))?,
                );
```                

fix - 
Now accepts both JSON and YAML, added two more tests (works in both http-mode and grpc)
deps.done file is created while doing `cargo make fix` (This wasnt the case, a week+ ago)

PR #1647, opened 2025-05-20
Change the license

vigoo:


Nanashi-lab:
Both new test from PR #1569 
create_openapi_yaml_definition, create_openapi_json_definition both fail with 

```rust
called `Result::unwrap()` on an `Err` value: Item(Error400(Messages(ErrorsBody { errors: ["Invalid openapi api definition: Parse error at line: 6, column: 5\nUnexpected `{`\nExpected letter\n"] })))
```

I just noticed, I will do a quick fix for this.

PR #1654, opened 2025-05-25
ci: split code-checks and unit-tests

laz-001:
* the execution time (of the ci run) on github free-tier runners is quite high, splitting the code-checks from the unit-tests halves the time.

this simplifies the run of actions in own forks.

* Added a  workflow_dispatch to simplify running/rerunning from ui/cli.

This is in essence a strict refactoring step, no functional change was introduced to the test-suites.

(i could not validate this on my fork due to the custom blacksmith runners)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,6 +6,8 @@ on:
     branches:
       - main
   pull_request:
+  # Enables to start/restart the workflow from the GitHub UI/CLI
+  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


laz-001:
be aware that contributing here is difficult, because of your custom runners.

To get ci feedback, one must use the own fork, and your current ci run has limits when using free-tier github ci runners.

"this [PR] simplifies the run of actions in own forks." - and does not hurt your custom-runners setup.

I can super-easy workaround this, but took the time to create a PR so others benefit from it.

All good, will focus more on the main task at hand.

vigoo:
> and does not hurt your custom-runners setup.

we are using blacksmith executors because they are faster, but it is just running the same github actions flow that you've modified here

PR #1656, opened 2025-05-26
provide makefile windows overrides for ci

laz-001:
`cargo make` will use the windows overrides when on a windows ci runner.

## Compatibility

All original tasks for non-windows remain unchanged.

(There is possibly a more elegant way to avoid this code-duplication. But this here works, as-is, and leaves non-windows 100% as-is)

this is an non-breaking change, it can be merged to main without any side-effects.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -455,12 +455,34 @@ mkdir -pv ../data
 ./golem-worker-service --dump-openapi-yaml > ../golem-worker-service.yaml
 '''
 
+[tasks.generate-worker-service-openapi.windows]
+description = "Generates openapi spec for worker service (Windows override)"
+cwd = "./target"
+script_runner = "pwsh"
+script_extension = "ps1"
+script = '''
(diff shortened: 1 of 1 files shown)


Makefile.toml
```diff
@@ -455,6 +455,17 @@ mkdir -pv ../data
 ./golem-worker-service --dump-openapi-yaml > ../golem-worker-service.yaml
 '''
 
+[tasks.generate-worker-service-openapi.windows]
```
vigoo, line 458: For these simple ones I would prefer to have only one version, but change that to use `duckscript` (simple platform independent scripting language supported  by cargo-make). We already used in other places for example in https://github.com/golemcloud/golem-cli/blob/main/Makefile.toml

It may not be powerful enough for check-configs (I am not sure), in that case it is ok to have a separate powershell version, but in general having a single version of these will make it easier to maintain

laz-001: Fully agree here, will give `@duckscript` a try.

PR #1660, opened 2025-05-27
Automatic updates of a worker see a consistent view of the filesystem during replay

### LINKED ISSUE #1629: Automatic updates of a worker should not cause it to see new files in replay mode
Currently workers see the new state of files in replay mode when the component is updated.
Instead workers should always see the same content of files as it was originally in live mode.

Whenever we go through a worker update when replaying, we need to adjust the state of the filesystem in accordance with the update. The rules for updating are as follows:

ro -> ro: overwrite file
ro -> rw: overwrite file
ro -> missing: delete file
rw -> ro: fail the update (or maybe overwrite and warn the user?)
rw -> rw: leave the file alone
rw -> missing: leave the file alone
missing -> ro: create the file
missing -> rw: create the file

mschuwalow:
resolves #1629

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ To work on **Golem** you need to install (via package installers or manually) th
 - [Protobuf](https://github.com/protocolbuffers/protobuf#protobuf-compiler-installation)
   - The [prost crate](https://crates.io/crates/prost) requires `protoc`
   - Requires **version 28** or later
-- For integration tests  
+- For integration tests
   - [redis](https://redis.io/downloads/)
   - [docker](https://www.docker.com)
 - To be able to run all services with `cargo-make run` with a merged log view:
@@ -142,13 +142,25 @@ There are two ways now to run Golem locally:
 
 ### Using cargo make run
 
-By running
+To run the project locally using `cargo make run` you'll need to configure the `GITHUB_CLIENT_ID` and `GITHUB_CLIENT_SECRET`.
+To do so you'll need to create a new GitHub OAuth application.
 
-```shell
-cargo make run
-```
+1. Go to github.com
+2. Navigate to `Settings > Developer settings > OAuth Apps`
+3. Create a new OAuth App. Use the following settings:
 
-all the services are going to be built and started as individual native processes, configured to communicate with each other. Beside that, an `nginx` process is going to be started to unify the various processes' HTTP APIs, and `lnav` is started to merge their logs.
+    Authorization callback URL: `http://localhost`
+4. Generate a new client secret.
+5. Export the following environment variables:
+
+    GITHUB_CLIENT_ID: Client ID of the OAuth application created in step (3)
+
+    GITHUB_CLIENT_SECRET: Client secret created in step (4)
+
+Ensure that `lnav`, `nginx` and `redis` commands are available in your PATH.
+You can now start golem-cloud locally using `cargo make run`.
+
+After it is started you can configure the cli to use the local instance using `golem profile new --component-url http://localhost:9881/  cloud cloud-local`.
 
 ### Using the Single Golem Executable
 
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -735,13 +735,20 @@ dependencies = ["build-release-override-linux-arm64"]
 description = "Runs all the services locally"
 dependencies = ["build"]
 
-condition = { fail_message = "Requires lnav, nginx and redis on path. Install them with your package manager." }
-condition_script = ["nginx -v", "lnav --version", "redis-server --version"]
+condition = { env_set = [ "GITHUB_CLIENT_ID", "GITHUB_CLIENT_SECRET"], fail_message = "Requires lnav, nginx and redis on path. Install them with your package manager. GITHUB_CLIENT_ID and GITHUB_CLIENT_SECRET env vars must be set." }
+
+condition_script = [
+    "nginx -v",
+    "lnav --version",
+    "redis-server --version",
+    ""
+]
 
 script = '''
 mkdir -pv data logs
 
-redis-server --port 6380 --save "" --appendonly no &> logs/redis.log &
+# start redis
+redis-server --port 6380 --appendonly yes --dir data/redis &> logs/redis.log &
 redis_pid=$!
 
 export RUST_BACKTRACE=1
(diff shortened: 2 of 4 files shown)


golem-common/src/model/mod.rs
```diff
@@ -585,18 +585,10 @@ pub struct WorkerStatusRecord {
...
-        active_plugins: HashSet<PluginInstallationId>,
-    },
-    Extension2 {
-        active_plugins: HashSet<PluginInstallationId>,
-        deleted_regions: DeletedRegions,
-    },
+    pub active_plugins: HashSet<PluginInstallationId>,
+    pub deleted_regions: DeletedRegions,
+    /// The component version at the starting point of the replay. Will be 0 if only automatic updates were used or the version of the latest snapshot based update
+    pub component_version_for_replay: ComponentVersion,
```
mschuwalow, line 591: This is the starting point of the replay and needs to be consistent with the deleted / skipped regions

golem-common/src/model/mod.rs
```diff
@@ -585,18 +585,10 @@ pub struct WorkerStatusRecord {
     pub total_linear_memory_size: u64,
     pub owned_resources: HashMap<WorkerResourceId, WorkerResourceDescription>,
     pub oplog_idx: OplogIndex,
-    pub extensions: WorkerStatusRecordExtensions,
-}
-
-#[derive(Clone, Debug, PartialEq, Encode, Decode)]
-pub enum WorkerStatusRecordExtensions {
```
mschuwalow, line 592: As this can be recomputed from the oplog I decided to go that way (as I need that anyway for component_version_for_replay)

golem-worker-executor-base/src/durable_host/mod.rs
```diff
@@ -2476,47 +2572,207 @@ impl<Ctx: WorkerCtx> WasiHttpView for DurableWorkerCtxWasiHttpView<'_, Ctx> {
...
+
+    let futures_phase_2 = files.iter().map(|file| {
+        let path = root.join(PathBuf::from(file.path.to_rel_string()));
+        let file = file.clone();
+        let permissions = file.permissions;
+        let file_loader = file_loader.clone();
+
+        let existing = current_state.get(&path);
+
+        async move {
```
mschuwalow, line 2676: The rules here are a bit arbitrary, but I tried to be as predictable as possible

golem-worker-executor-base/src/services/worker.rs
```diff
@@ -236,8 +242,26 @@ impl WorkerService for DefaultWorkerService {
...
                     });
 
-                if let Some(status) = status_value {
-                    details.last_known_status = status;
+                match status_value {
+                    Some(Ok(status)) => {
+                        details.last_known_status = status;
+                    }
+                    // We had a status, but it was written in a previous format and is not longer valid -> recompute
+                    Some(Err(_)) => {
```
mschuwalow, line 250: This error here is only for deserializtion -> we are dealing with an old version.

golem-worker-executor-base/tests/wasi.rs
```diff
@@ -616,6 +616,232 @@ async fn file_write_read(
     );
 }
 
+#[test]
+#[tracing::instrument]
+async fn file_updates(
```
mschuwalow, line 621: Testing all the various updates + restores here

golem-common/src/model/mod.rs
mschuwalow: Not quite, there are 3 versions that are relevant.

1: The "latest" component version which is what external users expect and is obtained after completely replaying the oplog.
2: The "initial" component version that the worker was originally created with.
3: The component version that is used to start the replay. This is the version that needs to be used for the earliest non-deleted region of the oplog. This is the thing I'm storing in the metadata.

These three will always be 2 <= 3 <= 1.

Consider the following scenarios:
1: Worker created with component version 0, automatically updated to version 1 -> the version for replay is 0, as we are always replaying from the beginning.
2: Worker created with component version 0, manually updated to version 1 -> the version for replay is 1, as everything before the update will be skipped.
3: Worker created with component version 0, manually updated to version 1, automatically updated to version 2 -> the version for replay is 1, as we are skipping the (0,1) region but need to replay the (1,2) region.

This is further complicated by reverts, etc. so this starting point is not possible to figure out without looking at the full oplog, so I think storing it in the metadata makes sense.

afsalthaj: Ok, thanks Maxim.

But I still have a doubt skimming through code :) May be you can help fix this doubt.

Currently the create_instance function has the logic of popping out the pending updates information, which has the new component version, to create WorkerCtx. The replay is done as part of recovering instance state, which is always after creating the instance (InvocationLoop - outer)

```
 let component_version = worker_metadata
            .last_known_status
            .pending_updates
            .front()
```

This implies, the worker context's  consist of the component metadata corresponding to the updated component's version even during the replay.

So I am a bit confused when we say

> Worker created with component version 0, automatically updated to version 1 -> the version for replay is 0, as we are always replaying from the beginning.

afsalthaj: I kind of get the first points though, there can be difference between the real "latest" and the one used in "replay"

vigoo:
I'm going to review this fully a bit later

golem-common/src/model/oplog.rs
```diff
@@ -780,8 +780,6 @@ impl OplogEntry {
                 | OplogEntry::Exited { .. }
                 | OplogEntry::PendingWorkerInvocation { .. }
                 | OplogEntry::PendingUpdate { .. }
-                | OplogEntry::SuccessfulUpdate { .. }
```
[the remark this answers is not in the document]
mschuwalow: What do you mean by non-deterministic here?

golem-worker-executor-base/src/durable_host/mod.rs
```diff
@@ -180,8 +179,21 @@ impl<Ctx: WorkerCtx> DurableWorkerCtx<Ctx> {
...
         );
 
-        let (file_use_tokens, read_only_paths) = prepare_filesystem(
-            file_loader,
+        debug!(
+            "Worker {} starting replay from component version {}",
+            owned_worker_id.worker_id, worker_config.component_version_for_replay
+        );
+
+        let component_metadata = component_service
```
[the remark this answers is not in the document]
mschuwalow: It's no longer passed as parameter. We used to pass it, but after this pr no component metadata is passed, just the version to start at.

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -703,44 +747,63 @@ impl<Ctx: WorkerCtx + DurableWorkerCtxView<Ctx>> DurableWorkerCtx<Ctx> {
                                 ),
                             )
                             .await;
-                        RetryDecision::None
+
+                            debug!("Finalizing automatic update to version {target_version}");
+                        }
+                        _ => {
+                            panic!("Expected snapshot-based update description")
```
vigoo: Wrong panic message

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -703,44 +747,63 @@ impl<Ctx: WorkerCtx + DurableWorkerCtxView<Ctx>> DurableWorkerCtx<Ctx> {
...
+        (*self.state.read_only_paths.write().unwrap()) = compute_read_only_paths(&current_files);
+
+        self.state.component_metadata = new_metadata;
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum ProcessReplayEventsResult {
```
vigoo: This does not seem to be used anywhere?

vigoo (approved):

PR #1665, opened 2025-05-28
Merge OSS and cloud - 1

### LINKED ISSUE #1618: Merge worker service cloud+oss
In this step we merge `golem-worker-service-base`, `golem-worker-service` and `cloud-worker-service` crates into a single one, as well as their REST and gRPC APIs. 

- We no longer going to have a separate worker service client generated, this has to be reflected in CLI
- Authentication related code (parametrizations, context, etc) need to be cleaned up

Merging the component service crates is tracked in a separate ticket, but they should end up using the same cleaned up constructs for abstracting cloud authentication.

### LINKED ISSUE #1619: Merge component service oss+cloud
In this step we take the `golem-component-service-base`, `golem-component-service` and `cloud-component-service` crates and merge into one crate. All the previously existing behaviors (like requiring auth tokens or not) should remain available through configuration. There should be a single gRPC and REST API remaining.

- After this, we no longer have separate clients generated from OpenAPI, this has to be reflected in the CLI
- Auth related code (abstractions, context objects, etc) need to be cleaned up and refactored. This should be done in sync with merging the worker service. (I gave a bigger estimate for that ticket assuming that is done first but the order is not that important)

mschuwalow:
resolves #1618
resolves #1619

Mostly mechanical changes here. More interesting refactors will be done in smaller followup prs

* Removed all oss services and replaced them with their cloud counterparts
* Merge the grpc apis
* Removed the oss client. Only a single client exists now
* Removed oss only types (but left the polymorphism intact for now. This I will do in the next pr)
* Adjust the test-framework to use cloud services and the "cloud service"
* Fix a number of small bugs that I encountered while getting the tests working


afsalthaj (approved):

afsalthaj:
Let @vigoo  take a look as well before merge

vigoo (approved):

Issue #1666, opened 2025-05-29
Make sure atmost 1 value in the stack after rib interpretation


afsalthaj:
TL;DR: There needs to be an assertion that makes sure there is only 1 value in the stack atmost after interpretation

Currently AssignVar is setting up variable value, however not loading it's result which is a Unit to back to stack in a stack based interpreter - which should be the intuitive approach given all other instructions do so.

5 of 650+ tests are failing when this was experimented.

This ticket is not about making it work, but finding out the proper reason for why this was the case.

afsalthaj:
Tried to reproduce it and I don't think there is any issue after the big cleanups and improvements https://github.com/golemcloud/golem/releases/tag/v1.3.0-dev.11

PR #1667, opened 2025-05-29
Fix runtime (or REPL) errors when invoking worker function in a loop

### LINKED ISSUE #1585: Bug: worker calls in a range throwing runtime errors (probably only in REPL, as there are tests in golem-rib)
```
>>> for i in range { cart.add-item({product-id: "foo", name: "foo", price: 42, quantity: 42}); yield "success"; }
[runtime error] internal error: InternalCorruptedState("failed to obtain an iterator from the stack")
>>>
```

May be this is only in REPL environment

Image

afsalthaj:
Fixes #1585 

### Before

image

### After

image

* In short, simplified REPL script interpretation, instead of REPL overly controlling the stack, env etc of the rib interpreter.
* Exposed InstructionId for REPL to hook on to and cache the results instead of diffing byte-code etc which originally worked, but comes with difficulty to understand when there is a bug.
* While caching results of instructions, REPL makes sure to not use cache for repeated evaluation of the same instruction in live mode
* There was a bit of issue in rib's interpreter too, that's fixed - see rib_intrepreter.rs


golem-rib-repl/src/rib_repl.rs
```diff
@@ -355,22 +350,49 @@ impl ReplRibFunctionInvoke {
...
-                component_id,
-                component_name,
-                worker_name.map(|x| x.0),
-                function_name.0.as_str(),
-                args.0,
-            )
-            .await
-            .map_err(|e| e.into())
+        let component_id = self.repl_state.dependency().component_id;
+        let component_name = &self.repl_state.dependency().component_name;
```
afsalthaj, line 359: These will be changed a bit when REPL supports multiple functions soon

golem-rib/src/interpreter/rib_interpreter.rs
```diff
@@ -252,67 +250,71 @@ impl Interpreter {
...
         }
 
-        let stack_value = stack
-            .pop()
-            .unwrap_or_else(|| RibInterpreterStackValue::Unit);
-
-        let rib_result =
-            RibResult::from_rib_interpreter_stack_value(&stack_value).ok_or_else(no_result)?;
+        match byte_code_cursor.last() {
+            Some(RibIR::AssignVar(_)) => Ok(RibResult::Unit),
```
afsalthaj, line 307: An instruction that currently doesn't push anything to stack. This is tracked in a separate issue, probably this is fine for now.

https://github.com/golemcloud/golem/issues/1666

As reading, there are various ways to ensure stack containts atmost 1 value at the end, by using instructions like `Drop`, but not a concern of this PR

afsalthaj:
Please give some more time, just found an instance where same instruction-id can actually be called again due to jumps in the stack. Need to handle that!

jdegoes (approved):

PR #1724, opened 2025-05-31
Confirm request path, query and header values can be numbers while type requirement is string

### LINKED ISSUE #1589: [Possible Bug/Already Fixed] Verify path parameter can be numbers when rib is expecting a string
There is already a test that make sure numbers are accepted in request path parameters even if the rib script is expecting a string

https://github.com/golemcloud/golem/blob/main/golem-worker-service-base/tests/api_gateway_end_to_end_tests.rs#L645

However, 1 user in the golem hackathon experienced some issue.

afsalthaj:
Fixes #1589 

Works as expected, and I am not sure during hackathon developer one of the users said that's not the case

Expanded an existing test case for this behaviour


afsalthaj:
Merging this as this is only 3 lines addition which is expanding a test

PR #1727, opened 2025-06-01
Command support in REPL

### LINKED ISSUE #1606: Support of commands in rib-repl and implement a few in golem-cli
## Commands

- `:load`. Loads a component by name or id, optionally with a version range.
- `:clear`. Clears the history of the REPL and starts fresh, with any auto-loaded components re-loaded.
- `:unload`. Unloads a component by name or id, optionally with a version range.
- `:exports` . Lists all exports of all loaded components.
- `:type <name>`. Looks up type information associated with a function and pretty prints the result in WIT format.
- `:help`. Displays information on available commands and their usages and contains a link to a Rib reference.

`clear`, `exports` (questionable, but rib repl can have exports that are more in line with how we call methods in Rib), `type` and `help` can be done in rib-repl.

rib-repl decides all the supported commands. There is no need of any additional commands from outside.

`load`, `unload`can be implemented by the clients, i.e, golem-cli.

The tasks are completed only when golem-cli implement all commands

afsalthaj:
Fix #1606 

Commands in REPL. Implemented in a way golem-cli (or wasmtime) should be able to extend or override command set (as discussed during retreat) 

There is an integration with `Clap` but note that golem_rib_repl's own `Command` trait do not enforcing users to use `Clap` for parsing input.

### Adding new commands (and overriding)
```rust
pub struct TypeInfo;

impl Command for TypeInfo {
    type Input = Expr;
    type Output = AnalysedType;
    type InputParseError = RibCompilationError;
    type ExecutionError = RibCompilationError;
   
   // implement functions
}

let mut registry = CommandRegistry::default();

// push needs an instance of UntypedCommand, and every Command already has an instance of UntypedCommand
registry.push(MyCustomCommand); 

let repl = RibRepl::bootstrap(RibReplConfig { registry });

repl.run()

```

### Overriding built-in commands from golem-cli (if needed)

* There is already `CommandRegistry::built_in()`. `type-info` is part of built_in. 

* These commands can be overriden by golem-cli (or wasmtime) if needed. Example: if `golem-cli` decides to override exports, it can create its own command and register and as far as the name matches golem-cli's export implementation will take precedence. 

* Command name is the kebab case of the command type name. Just decided to do this, after I gave same name for two commands.  

### Currently supported commands

* Commands are auto completed, and all options are available once user hits `:` and keep clicking tab

Examples:

### Exports
image

### TypeInfo
image

### Clear
image

### Command Error
If `clap::Parser` is used to implement the `parse` function in `Command` above, help and error documents are automatically. It is upto the client what to use as a parser. Given golem-rib is going to be used in other runtimes soon, its better not to enforce clients who want to add new commands to depend on `Clap` as such.

The inbuilt commands of golem-rib-repl are using clap for parse and documentations

image

### Command Help

image

### Other fix 

* Resources and instances could never be the return value before. This resulted in just compiler errors often. Now we made it flexible by tweaking a bit of interpreter for the benefit of users. 
It's a trade off after thinking and _trying out a lot_ from a user perspective.

Just used record type for now - no other choice

image

instead of 

image

Load and Unload commands can be supported as part of supporting multiple components _properly_. It will create more problems to expose these functionalities now, especially from golem-cli

### REPL API changes

It breaks compatibility but it's a simple change. Add `None` for command-registry when creating `RibReplConfig`


golem-rib-repl/src/rib_context.rs
```diff
@@ -0,0 +1,36 @@
+use crate::{RawRibScript, ReplPrinter};
+use rib::RibCompiler;
+
+// A projection of internal repl_state that could be useful
+// for advanced customisation of REPL commands.
+pub struct ReplContext<'a> {
+    printer: &'a dyn ReplPrinter,
+    rib_script: &'a RawRibScript,
+    compiler: &'a RibCompiler,
```
afsalthaj: Example: compiler is always retrieved from internal state (because a compiler of Rib can change if there is a dynamic load of a new component) and made it available for any clients who want to build a new component, but they will be restricted from making any change or mutate it.

golem-rib-repl/src/rib_execution_error.rs
```diff
@@ -0,0 +1,21 @@
+use rib::{RibCompilationError, RibRuntimeError};
+use std::fmt::{Display, Formatter};
+
+#[derive(Debug)]
+pub enum RibExecutionError {
+    RibCompilationError(RibCompilationError),
+    RibRuntimeError(RibRuntimeError),
+    Custom(String),
```
afsalthaj, line 8: TODO;  this can be more precise, just wanted to see if the design works

golem-rib-repl/src/rib_repl.rs
```diff
@@ -144,57 +154,80 @@ impl RibRepl {
...
-            // History is never used for any progressive compilation or interpretation
-            let _ = self.editor.add_history_entry(rib);
-            let _ = self.editor.save_history(&self.history_file_path);
+                let repl_context =
+                    ReplContext::new(self.printer.as_ref(), &script, &rib_compiler);
 
-            match compile_rib_script(&self.current_rib_program(), &self.repl_state) {
-                Ok(compiler_output) => {
-                    let helper = self.editor.helper_mut().unwrap();
+                executor.run(args.as_str(), &repl_context);
```
afsalthaj: args here is as much as possible a string. For the most part a REPL command has a single argument however, that may not be case for clients like golem-cli or wasmtime. But they can rely on Clap etc for their arg parsing to form a structured input. The current design in this PR allows this with the new trait `Command` that we expose (which is fully typesafe), which internally talks about a real typed `Input` that's retrieved.

afsalthaj: When they register their Command that works with varying input and output, it already has an instance of `UntypedCommand` and hence Repl will end up having a collection of `UntypedCommand`

golem-rib-repl/src/command/builtin.rs
```diff
@@ -0,0 +1,150 @@
...
+use crossterm::cursor::MoveTo;
+use crossterm::{
+    execute,
+    terminal::{Clear as TermClear, ClearType},
+};
+use golem_wasm_ast::analysis::AnalysedType;
+use rib::{CompilerOutput, Expr, FunctionDictionary, RibCompilationError};
+use std::io::stdout;
+
+#[derive(Clone)]
```
afsalthaj: Remove this need

jdegoes (approved):

PR #1734, opened 2025-06-03
Removed multiple result value support

### LINKED ISSUE #1373: Remove support for multiple return values
Support for multiple return values have been removed from everywhere in the ecosystem (wasmtime, wit-parser, etc). We still have support it in many levels, we could try to remove it as well to simplify the code if we can do it without breaking backward compatibility.

Otherwise we can schedule this for Golem 2.0

vigoo:
Resolves #1373 

Also changed the internal representation (including in the serialized oplog) of values and types to `ValueAndType` (a much more "user friendly" representation) from `TypeAnnotatedValue` (a protobuf-generated type).

Previously on the worker executor level invocation results were represented by a `Vec<Val>` that corresponds to the wasmtime result. This was further encoded into a single `TypeAnnotatedValue` where anonymous result values were organized into a tuple, and named ones into a record. As in practice there are no multiple return values, this ended up with the following possible representations:

- 1-tuple containing a single non-unit result value
- empty record representing the unit result value

After this change we represent the result value as `Option<ValueAndType>` or `Option<TypeAnnotatedValue>` and only convert it to an empty-record representation where needed

There are no real reason to keep `TypeAnnotatedValue` in the codebase but this PR only eliminates it from the inner parts of worker executor. A follow-up issue has been created.

### DIFF
--- a/golem-worker-executor/src/model/public_oplog/wit.rs
+++ b/golem-worker-executor/src/model/public_oplog/wit.rs
@@ -29,7 +29,7 @@ use golem_common::model::public_oplog::{
     SuccessfulUpdateParameters, TimestampParameter, WriteRemoteBatchedParameters,
 };
 use golem_common::model::Timestamp;
-use golem_wasm_rpc::{NodeBuilder, WitValue, WitValueBuilderExtensions};
+use golem_wasm_rpc::WitValue;
 
 impl From<PublicOplogEntry> for oplog::OplogEntry {
     fn from(value: PublicOplogEntry) -> Self {
@@ -101,9 +101,7 @@ impl From<PublicOplogEntry> for oplog::OplogEntry {
                 consumed_fuel,
             }) => Self::ExportedFunctionCompleted(oplog::ExportedFunctionCompletedParameters {
                 timestamp: timestamp.into(),
-                response: response
-                    .map(WitValue::from)
-                    .unwrap_or(WitValue::builder().record().finish()), // NOTE: representing unit result value as empty record
+                response: response.map(WitValue::from),
                 consumed_fuel,
             }),
             PublicOplogEntry::Suspend(TimestampParameter { timestamp }) => {


golem-rib/src/compiler/worker_functions_in_rib.rs
```diff
@@ -69,7 +69,7 @@ impl WorkerFunctionsInRib {
 pub struct WorkerFunctionType {
     pub function_key: RegistryKey,
     pub parameter_types: Vec<AnalysedType>,
-    pub return_types: Vec<AnalysedType>,
+    pub return_type: Option<AnalysedType>,
```
afsalthaj, line 72: 👍

golem-rib/src/type_inference/mod.rs
```diff
@@ -2648,10 +2648,7 @@ mod tests {
             vec![AnalysedExport::Function(AnalysedFunction {
                 name: function_name.to_string(),
                 parameters: analysed_function_parameters,
-                results: vec![AnalysedFunctionResult {
-                    name: None,
-                    typ: output,
-                }],
+                result: Some(AnalysedFunctionResult { typ: output }),
```
afsalthaj, line 2651: 👍

golem-rib-repl/src/repl_printer.rs
```diff
@@ -215,6 +215,12 @@ fn format_type_list(types: &[InferredType]) -> String {
     }
 }
 
+fn format_return_type(typ: &Option<InferredType>) -> String {
+    typ.as_ref()
+        .map(|t| wasm_wave::wasm::DisplayType(&AnalysedType::try_from(t).unwrap()).to_string())
+        .unwrap_or_else(|| "()".to_string())
+}
```
afsalthaj, line 222: 👍

afsalthaj (approved):

afsalthaj:
https://github.com/golemcloud/golem/pull/1737 will really help 

In the PR description,  for the following comments:

>and only convert it to an empty-record representation where needed (for example for the Rib evaluator).

This is no more needed with the above PR, and stays consistent and easy to understand what's going on. This makes rib interpreter more reliable as well. rib_interpreter is also a compoent that benefits from removing multi-return support allowing it to avoid _unreliable_ logic, such as if the value-and-type is of certain shape, then it is `Unit`.

afsalthaj (approved):

PR #1737, opened 2025-06-04
make rib interpreter aware that the worker function invoke can return optional value handling `Unit` functions

afsalthaj:
Raised against no-multi-return branch

This PR doesn't affect any other part of code other than golem-rib, repl, the tests in worker-service-base and will help continuing the original PR faster if this gets merged

- [x] golem-rib-interpreter small logical changes to properly handle optional value-and-type from function invocation
- [x] golem-worker-service-base api-gateway tests end to end is working (with fixes)
- [x] golem-rib tests are working with fixes in tests
- [x] golem-rib-repl tests integration tests working // an extra logic to be removed if the invoke_and_await_typed in test framework returns Option<AnalysedType>. If tests are not changed, keep it as is 

The integration tests that are failing is already failing in https://github.com/golemcloud/golem/tree/no-multi-return and hence not touching them for now

### DIFF
--- a/golem-rib-repl/src/invoke.rs
+++ b/golem-rib-repl/src/invoke.rs
@@ -31,7 +31,7 @@ pub trait WorkerFunctionInvoke {
         worker_name: Option<String>,
         function_name: &str,
         args: Vec<ValueAndType>,
-    ) -> anyhow::Result<ValueAndType>;
+    ) -> anyhow::Result<Option<ValueAndType>>;
 }
 
 // Note: Currently, the Rib interpreter supports only one component, so the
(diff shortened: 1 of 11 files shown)


golem-rib-repl/src/invoke.rs
```diff
@@ -48,7 +48,7 @@ impl ReplRibFunctionInvoke {
         Self { repl_state }
     }
 
-    fn get_cached_result(&self, instruction_id: &InstructionId) -> Option<ValueAndType> {
+    fn get_cached_result(&self, instruction_id: &InstructionId) -> Option<Option<ValueAndType>> {
```
afsalthaj, line 51: Could clean this up but this logic works 

One option is the cache result can be present or absent. Second option is the actual function  result

golem-worker-service-base/src/gateway_rib_interpreter/mod.rs
```diff
@@ -178,17 +177,10 @@ impl<Namespace: Clone + Send + Sync + 'static> RibFunctionInvoke
...
-                    Value::Record(vec![]),
-                    analysed_type::record(vec![]),
-                ))
-            }
-        }
+        let tav_opt = executor.execute(worker_request).await.map(|v| v.result)?;
+
+        tav_opt
+            .map(|tav| ValueAndType::try_from(tav).map_err(|x| x.into()))
+            .transpose()
```
afsalthaj, line 184: This is the core change in worker-gateway. Instead of unwrapping the option and creating another structure, we simply propoagate the optional value, for rib_interpreter to do the right thing.

Issue #1514, opened 2025-04-10
Implement Exactly-Once Semantics for golem:rdbms Support


jdegoes:
# Exactly-Once Semantics for Database Transactions

## Problem

Golem provides fault-tolerant execution of WASM components by recording all host function interactions in a durable operation log (oplog). In the case of infrastructure failure (e.g., process crash, hardware fault), this allows component execution to resume precisely from the point of interruption.

However, when a component executes a database transaction, and a failure occurs **after the database has made the transaction durable, but before Golem records the outcome in the oplog**, the transaction may be **re-executed** during replay. This breaks exactly-once semantics and may result in duplicate effects (e.g., double inserts, duplicate charges) if the database operation is not idempotent.

This issue outlines the mechanism for achieving exactly-once semantics for database transactions under two execution models:

1. **Databases with native transaction introspection**
2. **Databases without such support**

## Class 1: Databases with Native Transaction Status Introspection (e.g., Postgres)

Some databases (like modern Postgres versions) allow retrieving a unique identifier for the current transaction and querying its commit status after a crash.

**Approach:**

- At the start of a transaction, Golem retrieves a transaction identifier from the database and logs it to the oplog before executing any component logic.
- On commit or rollback, Golem logs an `end_transaction` marker in the oplog indicating the outcome.
- During replay, if a `begin_transaction` is encountered, Golem **scans forward in the oplog**:
  - If a corresponding `end_transaction` is present: no recovery action is needed.
  - If missing: Golem queries the database using the stored transaction ID.
    - If the database confirms the transaction was committed: do not replay.
    - If the transaction was aborted or never completed: replay.
    - If the database cannot determine the status: fail the instance.

This strategy avoids schema changes and imposes minimal performance overhead. It should be used wherever supported.

## Class 2: Databases Without Transaction Introspection (e.g., MySQL)

Databases that do not expose transaction status must rely on a fallback mechanism involving durable, internal metadata.

**Approach:**

- When a transaction is about to begin, Golem checks for the existence of a special internal table (`golem_transactions`). If the table does not exist, Golem attempts to create it.
  - If creation fails due to insufficient permissions, the instance is failed immediately with a descriptive error.
- Golem generates a unique transaction identifier (UUID) and:
  - Writes it to the internal table with a status of `incomplete`
  - Logs a `begin_transaction` entry to the oplog using the UUID
- The component transaction is executed.
[...]

PR #1741, opened 2025-06-04
Exactly-Once Semantics for golem:rdbms

justcoon:
fixes: #1514
/claim #1514 

new oplog entries:
1.  `BeginRemoteTransaction` - Begins a transaction operation  - entry is added to oplog after transaction begin
2. `PreCommitRemoteTransaction` - Pre-Commit of the transaction, indicating that the transaction will be committed -  entry is added to oplog before transaction commit is executed 
3. `PreRollbackRemoteTransaction` - Pre-Rollback of the transaction, indicating that the transaction will be rolled back -  entry is added to oplog before transaction rollback is executed 
4. `CommittedRemoteTransaction` - Committed transaction operation, indicating that the transaction was committed -  entry is added to oplog after successful transaction commit 
5. `RolledBackRemoteTransaction` -  Rolled back transaction operation, indicating that the transaction was rolled back - entry is added to oplog after successful transaction rollback 

`BeginRemoteTransaction` is first oplog entry related to DB transaction
`PreCommitRemoteTransaction` and `CommittedRemoteTransaction` entries will be around DB commit
`PreRollbackRemoteTransaction` and `RolledBackRemoteTransaction` entries will be around DB rollback
`CommittedRemoteTransaction` or `RolledBackRemoteTransaction` oplog entry is last oplog entry related to transaction

TODO:
- [ ] recovery test improvements - initial implementation using `TestOplog` where append of oplog entries failing based on patterns in worker name. [Oplog](https://github.com/golemcloud/golem/blob/main/golem-worker-executor-base/src/services/oplog/mod.rs#L170-L193) functions do not have `Result`  as return type, in case of error, it just ends with panic, ([PrimaryOplog](https://github.com/golemcloud/golem/blob/main/golem-worker-executor-base/src/services/oplog/primary.rs#L468-L490)), in this PR new oplog functions were added, where add of entry have in response `Result type` - discuss Oplog function with `Result`  as return type, find better way how to control when specific action should fail
- [x] golem-wit updates
- [x] code cleanup - remove println ...

oplog snippet 

```json
[
    {
        "type": "BeginRemoteTransaction",
        "timestamp": "2025-06-04T17:51:36.626Z",
        "transactionId": "740"
    },
    {
        "type": "ImportedFunctionInvoked",
        "timestamp": "2025-06-04T17:51:36.638Z",
        "functionName": "rdbms::postgres::db-transaction::execute",
        "request": { ...
        },
        "response": {...
        },
        "wrappedFunctionType": {
            "type": "WriteRemoteTransaction",
            "index": 16
        }
    },
    {
        "type": "Error",
        "timestamp": "2025-06-04T17:51:36.662Z",
        "error": "error while executing at wasm backtrace:\n    0:  0x3fc08 - wit-component:shim!indirect-golem:rdbms/postgres@0.0.1-[method]db-transaction.commit\n    1:  0x2a6a1 - <unknown>!<wasm function 231>\n    2:  0x2af54 - <unknown>!<wasm function 238>: Runtime error: worker rdbms-service-postgres-763b5c94-2128-4fca-b130-682ffaa8ec10-Fail1OnPreCommitRemoteTransaction failed on PreCommitRemoteTransaction 1 times"
    },
    {
        "type": "Jump",
        "timestamp": "2025-06-04T17:51:36.842Z",
        "jump": {
            "start": 16,
            "end": 21
        }
    },
    {
        "type": "BeginRemoteTransaction",
        "timestamp": "2025-06-04T17:51:36.848Z",
        "transactionId": "744"
    },
    {
        "type": "ImportedFunctionInvoked",
        "timestamp": "2025-06-04T17:51:36.857Z",
        "functionName": "rdbms::postgres::db-transaction::execute",
        "request": {...
        },
        "response": {...
        },
        "wrappedFunctionType": {
            "type": "WriteRemoteTransaction",
            "index": 22
        }
    },
    {
        "type": "PreCommitRemoteTransaction",
        "timestamp": "2025-06-04T17:51:36.868Z",
        "beginIndex": 22
    },
    {
        "type": "CommitedRemoteTransaction",
        "timestamp": "2025-06-04T17:51:36.888Z",
        "beginIndex": 22    }
]
```
where transaction failed on commit, on appending `PreCommitRemoteTransaction` entry to oplog, which means transaction was not commited, and was restarted


cloud-debugging-service/src/debug_session.rs
```diff
@@ -396,6 +399,21 @@ fn get_oplog_entry_from_public_oplog_entry(
...
+        PublicOplogEntry::PreCommitRemoteTransaction(_) => {
+            Err("Cannot override an oplog with a pre commit remote transaction".to_string())?
+        }
+        PublicOplogEntry::CommittedRemoteTransaction(_) => {
+            Err("Cannot override an oplog with a committed remote transaction".to_string())?
+        }
+        PublicOplogEntry::PreRollbackRemoteTransaction(_) => {
+            Err("Cannot override an oplog with a pre rollback remote transaction".to_string())?
+        }
+        PublicOplogEntry::RolledBackRemoteTransaction(_) => {
```
afsalthaj, line 414: I might find the answer as I go the PR. But what's the guarantee of having `RolledBackRemoteTransaction` for every successful rollback?

golem-common/src/model/public_oplog/mod.rs
```diff
@@ -677,6 +709,16 @@ pub enum PublicOplogEntry {
...
     /// Change the current persistence level
     ChangePersistenceLevel(ChangePersistenceLevelParameters),
+    /// Begins a transaction operation
+    BeginRemoteTransaction(BeginRemoteTransactionParameters),
+    /// Pre-Commit of the transaction, indicating that the transaction will be committed
+    PreCommitRemoteTransaction(RemoteTransactionParameters),
+    /// Pre-Rollback of the transaction, indicating that the transaction will be rolled back
+    PreRollbackRemoteTransaction(RemoteTransactionParameters),
+    /// Committed transaction operation, indicating that the transaction was committed
+    CommittedRemoteTransaction(RemoteTransactionParameters),
```
afsalthaj, line 722: If `PreCommitRemoteTransaction` exists, and if `CommittedRemoteTransaction` doesn't exist, we introspect the transaction status from db (for postgres to be precise). Am I right?

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2434,6 +2435,243 @@ impl PrivateDurableWorkerState {
...
+                        .lookup_oplog_entry_with_condition(
+                            begin_index,
+                            OplogEntry::is_end_remote_transaction,
+                            OplogEntry::no_concurrent_side_effect,
+                        )
+                        .await;
+
+                    // println!(
+                    //     "begin_transaction_function tx_id {tx_id}, pre_index {pre_index} end_entry {} jumped {jumped}",
+                    //     end_entry.clone().map(|(i, _)| i.to_string()).unwrap_or("None".to_string())
```
afsalthaj: Remove these

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2434,6 +2435,243 @@ impl PrivateDurableWorkerState {
...
+                    //     // );
+                    //     restart = !rolled_back;
+                    // }
+                } else {
+                    restart = true;
+                }
+
+                if restart {
+                    // We need to jump to the end of the oplog
+                    self.replay_state.switch_to_live().await;
```
afsalthaj, line 2537: Wasn't this already live at this point?

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2434,6 +2435,243 @@ impl PrivateDurableWorkerState {
...
+                            OplogEntry::no_concurrent_side_effect,
+                        )
+                        .await;
+
+                    // println!(
+                    //     "begin_transaction_function tx_id {tx_id}, pre_index {pre_index} end_entry {} jumped {jumped}",
+                    //     end_entry.clone().map(|(i, _)| i.to_string()).unwrap_or("None".to_string())
+                    // );
+                    if end_entry.is_none()
+                        && pre_entry.is_pre_commit_remote_transaction(begin_index)
```
afsalthaj: I think I would personally restructure this code in a small way as follows

You are already doing a pre-entry check  before which internally checks  if its a is_pre_commit_remote_transaction. So instead of checking pre_commit and pre_rollback together, I would just just check each of them and handle the logic as separate functions. And obviously avoids this second check, but most improtantly more readable.

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2434,6 +2435,243 @@ impl PrivateDurableWorkerState {
...
+                let tx_id = try_match!(
+                    begin_entry,
+                    OplogEntry::BeginRemoteTransaction {
+                        timestamp: _,
+                        transaction_id,
+                    }
+                )
+                .map_err(|_| GolemError::runtime("Unexpected oplog entry"))?;
+
+                let (tx_id, tx) = handler.create_replay(&tx_id).await?;
```
afsalthaj: Hmm. Interesting. Isn't `handler.create_replay` always ` RdbmsTransactionState::Closed(RdbmsTransactionId::new(transaction_id)),` ?

golem-worker-executor/src/durable_host/mod.rs
afsalthaj: Ok, no. I get it.

golem-common/src/model/public_oplog/mod.rs
afsalthaj: Ok got it I guess

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2434,6 +2435,243 @@ impl PrivateDurableWorkerState {
...
+                        .await;
+
+                    // println!(
+                    //     "begin_transaction_function tx_id {tx_id}, pre_index {pre_index} end_entry {} jumped {jumped}",
+                    //     end_entry.clone().map(|(i, _)| i.to_string()).unwrap_or("None".to_string())
+                    // );
+                    if end_entry.is_none()
+                        && pre_entry.is_pre_commit_remote_transaction(begin_index)
+                    {
+                        let committed = handler.is_committed(&tx_id).await?;
```
afsalthaj: Ok, if there is no end log  in oplog corresponding to a pre-log exist, we re verify the status of the real transaction - and - that makes sense obviously. 

But is this `is_committed` always reliable, given your postgres implementation for `is_committed` is relying  `txid_status`  to see if its committed. 

For old transactions, is there a chance `txid_status` can given false negative results? Given transaction wrap arounds or auto-vacuuming for that table can takes place and given its very internal postgres may nor provide a guarantee here? 

I did try to read a bit about this while reviewing - and ended up in things like `relfrozenxid` along with `txid_status`.  But I don't want to go deeper, pretty sure you might have done the research as well.

In short, I just want to know, do we know or have a guess about for how much recent transactions, this logic is reliable  - especially for postgres.?

Let us also ask @noise64  about a third opinion here, if I am over worrying here.

golem-worker-executor/src/durable_host/mod.rs
justcoon: for `golem:rdbms` yes, but I tried to do this impl. in generic way, in theory it can be used for something else in future

golem-worker-executor/src/durable_host/mod.rs
justcoon: If worker crash (after DB transaction commit, but before Committed oplog entry add), and It is restarted immediately,  I think there is low probability that related transaction id will be unknown or reused, in case of postgres

it can be issue if worker will not be immediately restarted, or started again for a long time.

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2434,6 +2435,232 @@ impl PrivateDurableWorkerState {
...
+                        // if the remote transaction was not committed, we need to restart
+                        restart = !committed;
+                    }
+                    // NOTE: rollback is not checked, because by default transaction is rolled back
+                    // it is not optimal to restart transaction which will end with rollback
+                } else {
+                    restart = true;
+                }
+
+                if restart {
```
afsalthaj: May be minor:

So would you mind having a `should_restart` and `on_restart` function.?

`should_restart` could be even more precise talking about pre-commit-remote-transaction.

i.e, is there no pre-commit-remote-transaction, just `restart` (true)
if there is pre-commit-remote-transaction, then check if there is corresponding end transaction.
If there is no end transaction, check if `handler.is_committed()` and `!result-of-that`

`on_restart` switch to live mode, and then ensure to delete the region so that next replay can completey forget about whatever happened, and then get a new transaction-id and start with `BeginRemoteTransaction`.


I am usually not that adamant about splitting functions. But I think in this case, it calls for such granularity for better readability. Secondly I re-explained just to ensure this is what you tried to do.

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2434,6 +2435,232 @@ impl PrivateDurableWorkerState {
...
+
+                    Ok((begin_index, tx))
+                } else {
+                    Ok((begin_index, tx))
+                }
+            }
+        } else {
+            Err(GolemError::runtime("Unexpected durable function type").into())
+        }
+    }
```
afsalthaj, line 2715: This comment is not specific about this line but generally about the approach.

Generally 1 question that I have is, the need of these new oplog entires. I don't think its too much required.

* In the live mode, you are anyways getting a transaction id, and that is being encoded within `BeginRemoteTransaction`. 

* So if there is no end remote transaction, you can still check with `handler.is_committed()` which internally uses postgres's internal tables etc. 

* After introspection and found that if the transaction hasn't happened,  you can switch to live, and making sure to delete the regions and start everything again with a new transaction-id. 

* In this case I am not even talking about pre-commit-transaction or committed-transaction, and rollbacks. Implies, no particular use of these new entires in this particular logic, and yet we achieved the same outcome of exactly once semantics

I am not quite convinced of these new entires, or may be I am yet to get the precise reason of these new logs, such as pre-commit and post-commit and pre-rollback and post-rollback.

If you could help me convince, or confirm if my concern is reasonable.

PS: I generally understand some advantage in having as much events to be tracked as possible.

justcoon: > So if there is no end remote transaction, you can still check with handler.is_committed() (using postgres's internal tables etc). After introspection and found that if the transaction hasn't happened, you can switch to live, and making sure to delete the regions and start everything again with a new transaction-id as new oplog entries

I think this is valid point, and based on `handler.is_committed()`, (which using transaction status), it is possible to change it in way that `PreCommit` and `PreRollback` oplog entries will not be needed

however at the other side, with `PreCommit` and `PreRollback` entries, it is more clear in oplog, what is happening there

afsalthaj: @justcoon  Along with the new entires of pre-commit, and pre-rollback and corresponding end entries, I am also doubtful of `WriteRemoteTransaction` because that also be dealt using `WriteRemoteBatch` which we already have.

justcoon: Hi,  `WriteRemoteTransaction` and `WriteRemoteBatch`, are kinda very similar considering attribute/field in structure, 

but I think, what is important is naming here, with  `WriteRemoteTransaction`  it is clearly  saying that specific functions are invoked in scope of transaction, (see oplog snippet in PR description)

justcoon: I thought about the need to have `PreCommit` and `PreRollback` entries, and think they will be needed considering postgres transaction status, it would be problem to distinguish if transaction was rolled back in case when user code executed rollback or because execution fail

if there is `BeginTransaction` oplog entry, and missing end
1. if implementation do NOT have  `PreCommit` and `PreRollback` entries, oplog replay will use get transaction status function, to get status in case if [postgres txid_status](https://pgpedia.info/t/txid_status.html), it will return aborted, but at this point we do not know if it was aborted/rolled back because user code, or some other failure
2. if implementation have `PreCommit` and `PreRollback` entries (like current one), if there are missing Pre entries, we know clearly that it failed/crashed before user executed commit/rollback

golem-worker-executor/src/durable_host/mod.rs
justcoon: I am not sure if this making sense to separate to smaller functions, as in general 
`replay_state.lookup_oplog_entry_with_condition` reading oplog entries, and it is not optimal to read it twice
in general most of functionality in `begin_transaction_function` for rollback and commit is common, just conditions for should restart are little bit different

golem-worker-executor/src/durable_host/mod.rs
afsalthaj: Sure. If there is a missing pre-roll-back entry and the status is aborted, then the transaction is a definite  fail and not by the user.

Could you explain was the exact reason for pre-commit and committed transaction?

justcoon: > Could you explain was the exact reason for pre-commit and committed transaction?
In general, considering current implementation, and that those transaction oplog entries are used for rdbms only

`PreCommit` entry is not needed as based on `handler.is_committed()` we can check if transaction was committed (if `Commit` entry is missing)

but as `PreRollback` entry is needed (as was mentioned above), I think is better to have also `PreCommit`

with `PreCommit` and `PreRollback` entries, it is more clear in oplog, what is happening there

afsalthaj: Ok. It looks like you are getting my questions and agreeing with it, and is aware of your implementation. I will ask @vigoo whenever he is free and hope he sees these message threads.. He had concerns around these new oplog entries too. Thanks for your responses for big reviews :)

afsalthaj:
@vigoo , whenever you are free, let's get this PR reviewed since it has been hanging around for a while.
https://github.com/golemcloud/golem/pull/1741#discussion_r2203246254

golem-worker-executor/src/services/oplog/mod.rs
```diff
@@ -168,6 +168,11 @@ pub trait Oplog: Any + Debug + Send + Sync {
     /// Adds a single entry to the oplog (possibly buffered)
     async fn add(&self, entry: OplogEntry);
 
+    async fn add_safe(&self, entry: OplogEntry) -> Result<(), String> {
```
vigoo, line 169: This (and `add_and_commit_safe` below neither) is not adding any safety over the `add` which panics by design in case it fails to write to the oplog.

All the oplog operations are "panicking" on purpose, because we do not want leaking failures of persisting the worker's state to user facing error channels. Instead, for example if an executor loses connection to Redis and fails to write to it, we want it to crash as soon as possible so another executor can pick up and recover the workers affected.

So please remove these new methods.

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2479,6 +2480,21 @@ impl PrivateDurableWorkerState {
         }
     }
 
+    async fn add_deleted_region(&mut self, index: OplogIndex) {
```
vigoo: We have both "skipped regions" (for skipping previous attempts of running a transaction for example) and "deleted regions" (used by the worker revert). This is adding a skipped region, so let's rename it to `add_skipped_region` if we keep it.

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2479,6 +2480,21 @@ impl PrivateDurableWorkerState {
...
+        if self.is_live() {
+            let deleted_region = OplogRegion {
+                start: index.next(), // need to keep the BeginAtomicRegion entry
```
vigoo: This comment makes only sense in the context where it was copied from. So I think `index.next()` with the comment should be moved back to the place where it was extracted from, and this helper method should just take the start index as is.

golem-worker-executor/src/durable_host/mod.rs
vigoo: After reviewing it more, this is not really just adding a skipped region but:
- checking if we are in live mode
- if we are NOT (this is incorrect in the code), adding a JUMP to skip the remaining of the oplog and get into live mode

So it should be called something like `add_jump_to_skip_remaining_oplog`.

But also the `self.replay_state.switch_to_live().await;` call is strongly related to this operation - basically it always have to happen before this. 

So I think it would be best to just undo the extraction of this method

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2479,6 +2480,21 @@ impl PrivateDurableWorkerState {
         }
     }
 
+    async fn add_deleted_region(&mut self, index: OplogIndex) {
+        if self.is_live() {
```
vigoo: The code below only makes sense in replay mode:
- create a skipped region for the **non-replayed part of the oplog**
- and adding a jump entry to the end of the oplog

It is true though that where this extracted method is called, there is a call before to `replay_state.switch_to_live` which already moved the replay pointer to the end of the oplog so we are in live mode.

But it is very misleading to check this here and do nothing if it's false. (As you can see even I thought in the previous comment that it should be an `is_replay` check). 

So my conclusion here is to just not have this extracted method

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2580,6 +2587,178 @@ impl PrivateDurableWorkerState {
         }
     }
 
+    pub async fn begin_transaction_function<Tx, Err>(
```
vigoo, line 2599: This whole functionality should be exposed as part of the `Durability` API (for the internals, and later also through the durability host functions). I'm going to write a suggested bigger change below that helps achieving this.

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2580,6 +2587,178 @@ impl PrivateDurableWorkerState {
...
+            let pre_entry = self
+                .replay_state
+                .lookup_oplog_entry_with_condition(
+                    begin_index,
+                    OplogEntry::is_pre_remote_transaction,
+                    OplogEntry::no_concurrent_side_effect,
+                )
+                .await;
+
+            let jumped = self
```
vigoo: This part is not how it should work and is incorrect in many ways. 
- The`get_oplog_entry` gets the next oplog entry and verifies it is `BeginRemoteTransaction`. While getting it, it automatically skips the skipped regions.
- `begin_index.next()` is the next oplog entry which is not necessarily have anything to do with the transaction, it can be many things.
- then the code is checking if that one is in a skipped region here and if yes, then gets one more `BeginRemoteTransaction`. Even if it would be the way to do it, one problem with this would be that it only does it once, although transactions can be retried multiple times. 

But instead, there are two choices:
- if the transaction ID remains the same when retrying, then simply not including the `BeginRemoteTransaction` in the skipped region, and then just continue as is.
- I suspect the reason for this code is that a new transaction ID is generated when a transaction is restarted. If that's true, then I think this can be done correctly by _including_ the the `BeginRemoteTransaction` in the skipped region. That way the first `get_oplog_entry` always gets the latest one and there is no need to know about skpped regions in this code at all (that is constrained to `ReplayState` and that handles it correctly and recursively)

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2580,6 +2587,178 @@ impl PrivateDurableWorkerState {
...
+                // we need to check if the remote transaction was committed
+                if end_entry.is_none() && pre_entry.is_pre_commit_remote_transaction(begin_index) {
+                    // if the remote transaction was not committed, we need to restart
+                    should_restart = !handler.is_committed(&tx_id).await?;
+                }
+                // NOTE: rollback is not checked, because by default transaction is rolled back
+                // it is not optimal to restart transaction which will end with rollback
+            } else {
+                // if pre entry is not found, we need to restart
+                should_restart = true;
```
vigoo, line 2675: When getting `pre_entry` the `no_concurrent_side_effect` condition is used. (Also for `end_entry` above). Restarting a transaction must not interfere with other side-effects, and having this condition here is part of achieving that. 
But what it does is that if there WAS a concurrent side_effect with the transaction, it is going to return `None` and we get into this branch of execution.
Here we decide that we should restart, even if `pre_entry` was `None` because there was interfering side-effects, not because there was no pre_entry. This is the opposite of what we need to do because we cannot restart it! Restarting it means we also drop the results of the already performed other side effects, and rerun them, which is against the idempotence requirements.
The only thing we can do in this case is to fail. This is what the streaming HTTP response durability layer is doing too.
And this whole behavior must be controlled by `self.allow_idempotence` - if that is set to `true` we don't have to fail and we can rerun the other side effects too.

Check how `begin_function` implements this exactly.

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -3092,3 +3271,17 @@ macro_rules! get_oplog_entry {
         }
     };
 }
+
+#[async_trait]
+pub trait RemoteTransactionHandler<Tx, Err>
+where
+    Err: From<WorkerExecutorError>,
+{
+    async fn create_new(&self) -> Result<(String, Tx), Err>;
```
vigoo: Let's have a `TransactionId` newtype instead of using `String` and `&str` in this API

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -3092,3 +3271,17 @@ macro_rules! get_oplog_entry {
         }
     };
 }
+
+#[async_trait]
+pub trait RemoteTransactionHandler<Tx, Err>
```
vigoo, line 3303: There is a logic above with a comment
```
                // NOTE: rollback is not checked, because by default transaction is rolled back
                // it is not optimal to restart transaction which will end with rollback
```

I think this decision should also be coming from the `RemoteTransactionHandler`, not hardcoded there.

golem-worker-executor/src/services/oplog/mod.rs
justcoon: I added this function to be able to do required tests scenarios, till now, I was not able to find better solution to test oplog entry add failure after successful DB request (commit, rollback)

as you mentioned, in current design, in case of oplog failure, it panic, which leads to situation that whole executor fail 

but this look like also mean, that temporary write problem related to one worker will affect also other workers which running in same executor

with `add_safe` (which is used in this PR) in case of oplog failure, worker was restarted and continued where it ended before (see oplog snippet in PR description)

if we consider panic approach, is there some regular way how to test such scenarios?
because if I do same with panic approach whole test case fail

thanks for help/recommendation

golem-worker-executor/src/durable_host/mod.rs
justcoon: I removed extracted function, as was recommended

golem-worker-executor/src/durable_host/mod.rs
justcoon: I removed extracted function, as was recommended

golem-worker-executor/src/durable_host/mod.rs
justcoon: thanks, will take a look on that, 
yes, new transaction ID is generated for new restarted transaction, that is reason why there is  new `BeginRemoteTransaction` entry

golem-worker-executor/src/durable_host/mod.rs
justcoon: I removed extracted function, as was recommended

golem-worker-executor/src/durable_host/mod.rs
justcoon: `TransactionId` added

golem-worker-executor/src/durable_host/mod.rs
justcoon: changed based on recommendation

> I suspect the reason for this code is that a new transaction ID is generated when a transaction is restarted. If that's true, then I think this can be done correctly by including the the BeginRemoteTransaction in the skipped region. That way the first get_oplog_entry always gets the latest one and there is no need to know about skpped regions in this code at all (that is constrained to ReplayState and that handles it correctly and recursively)

golem-worker-executor/src/durable_host/mod.rs
justcoon: thanks, `RemoteTransactionHandler.is_rolled_back` is used now for this case

golem-worker-executor/src/durable_host/mod.rs
justcoon: I changed implementation, it is using `self.allow_idempotence` now

justcoon:
Hi @vigoo,

Regarding your [proposal](https://github.com/golemcloud/golem/pull/1741#pullrequestreview-3082150947) about using begin/end remote write:

I had also considered that approach, but ultimately decided to add new entries instead. However, I'm not opposed to the idea.

One of the main differences between `begin_remote_write` and `begin_remote_transaction` functions is the presence of a `RemoteTransactionHandler`. This handler manages both the creation/beginning of remote transactions and provides functions to check whether a remote transaction has been committed or rolled back. These functions help resolve situations where an end entry is missing, either by continuing the execution of worker functions (which should add the end entry) or by triggering a transaction restart. In the case of RDBMS, restarting a transaction means creating a new one.

If we want to expose this functionality through host functions, we would need to add an additional WIT interface for `RemoteTransactionHandler` that would be implemented by the worker. The worker executor could then invoke it as needed, similar to how load/save snapshot operations work.

Because that, I think it is better to have specific oplog entries for transaction, we can decide if we want to expose that as host functions or not

Let me know if I am missing something or if you have any other suggestions,

Thank you very much

vigoo:
From your TODO list, don't worry about backward compatibility. We decided to break many things in the 1.3 release so it won't be compatible with any older worker anyway.

vigoo:
> If we want to expose this functionality through host functions, we would need to add an additional WIT interface for RemoteTransactionHandler that would be implemented by the worker. The worker executor could then invoke it as needed, similar to how load/save snapshot operations work.

That kind of callback interface is not really possible with these WIT interfaces (I mean what you write is technically probably possible but very inconvenient). Instead the version of this exposed as host function should be refactored similarly to what I did with `Durability` (which, if you remember, originally also was a higher order function somewhere, taking what to run as a parameter) in a way that it just returns data ("what to do") and the control is on the caller side. Then of course the logic can be still hidden in a library that implements it with the host functions always the same.

I don't want this to be implemented before this PR is merged though.

Because we are going to break compatibility at least once more in 1.4 according to the current plans, I am giving this another review now with the intent to have it in 1.3 as we planned. (And modify it later if needed)

golem-worker-executor/src/services/oplog/mod.rs
vigoo: > but this look like also mean, that temporary write problem related to one worker will affect also other workers which running in same executor

Yes but because they are durable, they are just going to resume in another (or restarted) node. So it just affects performance in theory.

I think we don't have a good way to test such panics now. So let's leave it as is in the first version.

PR #1743, opened 2025-06-06
Support multiple components

### LINKED ISSUE #1608: Support for multiple components in golem-rib and golem-rib-repl
How to distinguish functions across components is already implemented in golem-rib using type-parameters (if there are conflicts)

Rib and RibRepl should allow multiple components , and so that golem-cli can forget about selecting a specific component when firing `golem-repl`.

Currently RibRepl expose extensible functionality on how to load "one" single component, and it should be changed to loading multiple components, such that, golem-cli can load whatever components available in the context and load it up.

https://github.com/golemcloud/golem/issues/1606

afsalthaj:
Fix #1608 

- [ ]  auto-completes with all function-names in REPL in the context of multiple components (being able to choose a method
- [ ]  Adding more tests (currently only ensuring major tests in golem-rib working)
- [ ]  Better type inference: - only small changes required.  Ex: ambiguous interface, but choosing a method makes it precise. 
- [ ]  88 tests (out of 650) remaining but mostly done - testing things from REPL directly and making changes accordingly
- [ ]  A few minor grpc changes
- [ ]  some copies and clones, just to get the logic working


afsalthaj:
Status: All tests in rib passes
image

golem-rib/regression_tests/lib.rs
```diff
@@ -15,108 +15,108 @@ use rib::{
 #[test]
 async fn test_rib_regression() {
     let expr = r#"
-
+              let worker = instance();
```
afsalthaj, line 18: This is now mandatory. If a component / worker needs to be used in rib script, there needs to be an instance.

jdegoes (approved):

PR #1753, opened 2025-06-11
Merge golem-component-service-base and golem-component-service

### LINKED ISSUE #1732: Merge component-service cloud and base

mschuwalow:
resolves #1732


golem-component-service/src/authed/component.rs
```diff
@@ -0,0 +1,540 @@
...
+use std::sync::Arc;
+use std::vec;
+
+pub struct AuthedComponentService {
+    base_component_service: Arc<dyn ComponentService>,
+    auth_service: Arc<dyn BaseAuthService>,
+    project_service: Arc<dyn ProjectService>,
+}
+
+impl AuthedComponentService {
```
mschuwalow, line 48: Getting rid of this separate layer of services with auth handling is not really possible without making everything more complex. Problem is that there are a lot of interdependencies between the services, so keeping this separate layer and doing auth at the edge makes things simples.

golem-component-service/src/authed/component.rs
```diff
@@ -0,0 +1,540 @@
...
+use golem_service_base::clients::auth::BaseAuthService;
+use golem_service_base::clients::project::ProjectService;
+use golem_service_base::model::ComponentName;
+use std::collections::HashMap;
+use std::fmt::{Debug, Formatter};
+use std::sync::Arc;
+use std::vec;
+
+pub struct AuthedComponentService {
+    base_component_service: Arc<dyn ComponentService>,
```
vigoo: This is a bit confusing that there is a `base_component_service` which is `ComponentService` and an `auth_service` which is a `BaseAuthService` :)

golem-common/src/model/plugin.rs
```diff
@@ -282,45 +279,6 @@ impl PluginScope {
             Self::Project(scope) => &scope.project_id == project_id,
         }
     }
-
-    pub async fn accessible_scopes(
```
vigoo, line 286: Why isn't this necessary anymore?

mschuwalow: This is inline in the AuthedPluginService

PR #1755, opened 2025-06-11
Ensure stateful instance

afsalthaj:
The state was kept only when a worker name was given to `instance` function before. With this PR, it does even without it. I thought it was a 5 mins task, and indeed it was simple in `golem-rib` but there were nuances in REPL. Anyway, it works now!

Added integration tests that test direct Rib invocation and also test this behaviour with components including resource. 

Now, with this, `instance` function is autocompleted all the way to `instance()` instead of stopping at `instance` and waiting for arguments. Implies, in REPL user hardly needs to give a worker name  most of the time -  a minor DX improvement 

image

image


afsalthaj:
There is 1 thing to be fixed. Will fix it soon.

golem-rib/src/interpreter/rib_interpreter.rs
```diff
@@ -450,86 +473,199 @@ mod internal {
...
+
+                            _ => bail_corrupted_state!(
+                                "expected a field named 'to' to be of type U64, but it was not"
+                            ),
+                        }
+                    }
+
+                    _ => bail_corrupted_state!(
+                        "expected a field named 'from' to be of type S8, U8, S16, U16, S32, U32, S64, U64, but it was not"
+                    ),
```
afsalthaj, line 570: A few fixes in the range when adding more tests.

PR #1762, opened 2025-06-16
Allow disabling oauth2 logins in the cloud service

### LINKED ISSUE #1758: Allow disabling signups / github integration

mschuwalow:
resolves #1758


cloud-service/src/login/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+use golem_service_base::db::Pool;
+use std::sync::Arc;
+
+pub struct LoginSystemEnabled {
+    pub login_service: Arc<dyn LoginService>,
+    pub oauth2_service: Arc<dyn OAuth2Service>,
+}
+
+pub enum LoginSystem {
+    Enabled(LoginSystemEnabled),
```
mschuwalow, line 41: All the oauth2 / login stuff is gated behind this.

Optimally we would not even migrate the tables if this is disabled, but it's not worth the effort to implement currently imo

cloud-service/src/login/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+pub use self::service::LoginError;
+pub use self::service::LoginService;
+use crate::config::LoginConfig;
+use crate::service::account::AccountService;
+use crate::service::token::TokenService;
+use golem_service_base::db::Pool;
+use std::sync::Arc;
+
+pub struct LoginSystemEnabled {
+    pub login_service: Arc<dyn LoginService>,
```
mschuwalow, line 36: These should probably be reorganized and their error hierarchies simplified / merged, but I'm leaving as is for now.

noise64: you mean e.g. having a unified a LoginService?

noise64: Apart from this naming (system) the PR looks good, but i understand that would mean renaming and / or merging the current login and oauth2.
But that seems somthing we can do a follow-up later, getting CLI working seems more important.

noise64 (approved):

PR #1764, opened 2025-06-17
Remove wasmtime dep from client

mschuwalow:


golem-rib/src/compiler/ir.rs
```diff
@@ -195,6 +169,32 @@ mod protobuf {
...
+            match value {
+                WorkerNamePresence::Present => {
+                    golem_api_grpc::proto::golem::rib::WorkerNamePresence::Present
+                }
+                WorkerNamePresence::Absent => {
+                    golem_api_grpc::proto::golem::rib::WorkerNamePresence::Absent
+                }
+            }
+        }
+    }
```
afsalthaj, line 196: I think you can remove this structure both from proto and code. I forgot to may be remove it. If you could try, that will be good since you are anyways touching these parts

golem-client/Cargo.toml
```diff
@@ -21,7 +21,7 @@ golem-common = { path = "../golem-common", version = "=0.0.0", default-features
 golem-wasm-ast = { path = "../wasm-ast", version = "=0.0.0", default-features = false, features = [
     "analysis",
 ] }
-golem-wasm-rpc = { path = "../wasm-rpc", version = "=0.0.0", default-features = false }
+golem-wasm-rpc = { path = "../wasm-rpc", version = "=0.0.0", features = ["stub"], default-features = false }
```
vigoo, line 24: This is not correct, the `stub` feature replaces the bindings to the `cargo-component` generated one to be used in generated stub components while `golem-client` is to be used from any non-wasm crates. It maybe accidentally working but that feature flag is not for that.

golem-api-grpc/Cargo.toml
```diff
@@ -14,7 +14,7 @@ harness = false
 
 [dependencies]
 golem-wasm-ast = { path = "../wasm-ast", version = "=0.0.0", default-features = false, features = ["protobuf"] }
-golem-wasm-rpc = { path = "../wasm-rpc", version = "=0.0.0", default-features = false, features = ["host-bindings", "protobuf"] }
```
vigoo, line 17: why is it requiring host bindings now :o

mschuwalow:
This is not ready for review, just something I was playing around with.

Will close

PR #1766, opened 2025-06-18
Final clean ups to 1608

afsalthaj:
I wanted to do this soon, before going to take up tasks that are unrelated to Rib

* Reduce the need of Expr clones in various compilation phases by keeping `source_span` instead of `Expr` by itself in error spans. Error structures owned the Expr and that was the base reason for clones. With this change, it led to an internal error called RibTypeErrorInternal which gets converted to `RibTypeError' only at the end. The former holds just the source span, and the latter holds the expression that is invalid. 
* Making sure the instance type is narrowed down at every loop in the fix point such that finally it has only 1 component dependency.
* Avoid parent_expr information in error messages in favor of code maintainability, as parent expression is (seen to) provide less value to users. Users have line number and column number, along with the wrong (non compiling) expression and that's enough.
* Avoid expr clone in type pull up phase (hasn't changed much speed :) , yet thought of avoiding it
* regression tests is back to original speed of 5s, althought it is not because of removing clones, but a few small changes elsewhere.


golem-rib/src/compiler/byte_code.rs
```diff
@@ -1629,29 +1629,6 @@ mod compiler_tests {
             );
         }
 
-        #[test]
-        fn test_invalid_arg_size_variants() {
-            let metadata = internal::metadata_with_variants();
-
-            let expr = r#"
-               let regiser_user_action = register-user(1, "foo");
-               let result = golem:it/api.{foo}(regiser_user_action);
```
afsalthaj, line 1638: Because we don't allow this anymore

PR #1771, opened 2025-06-19
Rib comments

### LINKED ISSUE #1611: Support WIT style comments in Rib
- [ ] Update grammer
- [ ] Create parser for comments
- [ ] Add tests

afsalthaj:
Fix #1611 

Tests are based on the possibilities I tried in WIT file. 

Although their [documentation](https://component-model.bytecodealliance.org/design/wit.html#documentation) talks about `///` and `/**`, `//` and `/*...` are also supported. 

The WIT IDE plugin may show error  for things like `  add: func(value: /*foo*/ u64);`, however it still works and is supported in Rib

### DIFF
--- a/golem-rib/src/parser/comment.rs
+++ b/golem-rib/src/parser/comment.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use combine::parser::char::{char, spaces, string};
+use combine::parser::char::{char, newline, spaces, string};
 use combine::{
     any, attempt, choice, eof, many, none_of, not_followed_by, optional, parser, Parser, Stream,
 };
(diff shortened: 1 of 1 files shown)


golem-rib/src/parser/pattern_match.rs
```diff
@@ -18,6 +18,7 @@ use combine::{attempt, not_followed_by, sep_end_by1, ParseError, Parser};
 use match_arm::*;
 
 use crate::expr::Expr;
+use crate::parser::comment::comments;
```
afsalthaj, line 21: for pattern matches, not every token is recursively going back to full rib parser, which would have taken care of comments automatically, and hence we support comments in pattern match expressions explicitly.

golem-rib/src/parser/comment.rs
```diff
@@ -0,0 +1,314 @@
...
+        .map(|(_, _, _, _): ((), (), (), ())| ())
+}
+fn line_comment<Input>() -> impl Parser<Input, Output = ()>
+where
+    Input: Stream<Token = char>,
+{
+    (
+        spaces().with(choice!(attempt(string("///")), attempt(string("//"))).map(|_| ())),
+        many(none_of(vec!['\n']).map(|_| ())),
+        choice((string("\r\n").map(|_| ()), string("\n").map(|_| ()), eof())),
```
afsalthaj: use newline() method

noise64: is this still TODO?

afsalthaj: I will update it before merge. Right now not able to push a commit.

noise64 (approved):

jdegoes (approved):

PR #1779, opened 2025-06-25
test-framework: Support testing with multiple accounts

mschuwalow:
* all individual components in the test-framework have no knowledge of "current" accounts anymore
* TestDsl is no longer implemented directly implemented for TestDependencies, but you rather select whether you want to use admin or a normal (ad-hoc created) using `.admin()` and `.user()`. These have access to the account_id and token, so you can script things not supported by the TestDsl using the raw clients using these

In practice the hierarchy is now something like this for the test-framework:

raw clients (grpc/http) < specific services like ComponentService (abstracts over / hides the underlying client) < TestDsl (hides / abstracts users and authentication)


afsalthaj:
LGTM 

(except for certain patterns such as `executor.into_admin()` - and i would consider it as a minor thing).

PR #1786, opened 2025-06-30
Fix handling of failed update attempts to non-existing versions

### LINKED ISSUE #1522: Updating a worker to a non-existing component version makes it non-responsive to further invocations

vigoo:
Resolves #1522


golem-worker-executor/src/worker/mod.rs
```diff
@@ -1572,15 +1572,47 @@ impl RunningWorker {
...
                     &account_id,
                     &component_id,
                     component_version,
                 )
-                .await?
+                .await
+            {
+                Ok((component, component_metadata)) => Ok((component, component_metadata)),
+                Err(error) => {
+                    if component_version != worker_metadata.last_known_status.component_version {
```
mschuwalow, line 1587: Does it make sense to inspect the error here and look for an explicit not_found?

Imo for infrastructure errors it would be better to not fail the update and retry it whenever the component service comes back up. Plus the following get (for the current version) will fail in the case of infrastructure errors anyway, and we need to handle it.

mschuwalow (approved):
minor comment, but also fine as is

golem-worker-executor/src/worker/mod.rs
vigoo: I wanted that first but then found that out here isn't a specific error case for this and didn't want to change GolemError as I know you were also looking into that.

PR #1793, opened 2025-07-02
Component version env var during update

### LINKED ISSUE #1630: Ensure that environment variables work correctly when updating workers
@afsalthaj mentioned that he saw some weird behaviour when updating workers, specifically regarding the component version environment variable.

mschuwalow:
partially resolves #1630 

Fix updating of the env var for snapshot based updates.
For automatic updates env-var updates don't work at all, so there is no way to support it currently.


golem-worker-executor/src/worker/invocation.rs
```diff
@@ -48,6 +48,8 @@ pub async fn invoke_observed_and_traced<Ctx: WorkerCtx>(
     let mut store = store.as_context_mut();
     let was_live_before = store.data().is_live();
 
+    debug!("Beginning invocation {full_function_name}");
```
vigoo, line 51: This can be deleted

mschuwalow: It's actually really useful for me to more easily read the logs during tests runs ;)
So I was planning to keep it if you are fine with it

PR #1803, opened 2025-07-08
Expose worker error logs in worker service http api errors

mschuwalow:
Also moves GolemError to golem-service-base and reuses it to reduce some duplication


golem-common/src/model/oplog.rs
```diff
@@ -968,32 +968,38 @@ pub enum DurableFunctionType {
 }
 
 /// Describes the error that occurred in the worker
-#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode)]
-pub enum WorkerError {
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Encode, Decode)]
+pub enum WorkerTrapCause {
```
mschuwalow, line 972: Renamed this to avoid confusion with worker service errors

golem-service-base/src/error/worker_executor.rs
```diff
@@ -876,3 +910,44 @@ impl Display for WorkerOutOfMemory {
...
+            InterruptKind::Restart => write!(f, "Simulated crash via the Golem API"),
+            InterruptKind::Suspend => write!(f, "Suspended"),
+            InterruptKind::Jump => write!(f, "Jumping back in time"),
+        }
+    }
+}
+
+impl Error for InterruptKind {}
+
+#[cfg(feature = "worker-executor")]
```
mschuwalow, line 922: just to avoid pulling in wasmtime when only compiling a single service in ci that doesn't need it.

This will cause rebuilds when switching between the two (workspace build and single service build), so there is an argument for just always building it.

(just making the conversions functions in the worker-executor is a bit inconvenient currently because of how our Durability struct is defined)

golem-service-base/src/error/worker_executor.rs
```diff
@@ -12,20 +12,22 @@
...
+use golem_wasm_rpc_derive::IntoValue;
+use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
+use std::error::Error;
+use std::fmt::{Display, Formatter};
 use tonic::Status;
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Encode, Decode)]
-pub enum GolemError {
+pub enum WorkerExecutorError {
```
mschuwalow, line 30: Just to have slightly less confusing naming

golem-worker-service/src/api/common.rs
```diff
@@ -43,16 +45,21 @@ pub enum ApiTags {
...
+    /// Error log of the worker
+    pub stderr: String,
 }
 
-#[derive(Object, Debug, Clone)]
-pub struct MessageBody {
-    message: String,
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, poem_openapi::Object)]
+#[oai(rename_all = "camelCase")]
+pub struct ErrorBodyWithOptionalWorkerError {
```
mschuwalow, line 50: This is the main result.
For example an error response from a worker failure might now look like this:
```json
{
  "error": "Invocation failed",
  "worker_error": {
    "cause": "<wasm error that cause the worker to fail>",
    "stderr": "<stderr from the failed invocation>"
  }
}
```

giving the cli more flexibility in how to display it without having to parse the strings

golem-debugging-service/src/from_value.rs
```diff
@@ -657,13 +706,34 @@ impl FromValue for GolemError {
...
 
-                        Ok(GolemError::Unknown { details })
+                        Ok(WorkerExecutorError::Unknown { details })
                     }
 
                     _ => Err("Failed to get GolemError. Not a Record".to_string()),
                 },
-                (22, None) => Ok(GolemError::ShardingNotReady),
-                (23, Some(error)) => match error.deref() {
+                (22, None) => Ok(WorkerExecutorError::ShardingNotReady),
```
mschuwalow, line 714: tbh, I'm not quite sure how these are being used.

Will dig into it some other time as it seems quite brittle

vigoo (approved):

PR #1841, opened 2025-07-14
Change worker's owner to be project

### LINKED ISSUE #1604: Change storage ownership to be project based
Currently the `BlobStorageNamespace`/`KeyValueStorageNamespace` and `IndexedStorageNamespace` types are grouping data per account. We want to change this to be per project, because projects can be transferred between accounts.

If we want to preserve component and worker data while doing this we have to generate migration scripts renaming all the existing data. (We do not need to work on a downtime-free migration as golem-cloud is still in developer preview mode).

vigoo:
Requires https://github.com/golemcloud/golem-wit/pull/65

Resolves #1604 

**THIS IS A BREAKING CHANGE** in many ways, breaking the blob and KV store, the oplog and the APIs.

- Worker related data is now namespaced by project and not account
- Worker executor works with project as the worker owner and not account
- Plugin definitions and resource limits are resolved to be a worker's _owner project's owner account_ not the account created the worker.


golem-worker-executor/src/model/public_oplog/mod.rs
```diff
@@ -351,60 +321,7 @@ impl PublicOplogEntryOps for PublicOplogEntry {
                         function_name,
                         request,
                         response,
-                        wrapped_function_type: wrapped_function_type.into(),
-                    },
-                ))
-            }
-            OplogEntry::ExportedFunctionInvokedV1 {
```
mschuwalow, line 358: :tada:

golem-worker-executor/src/services/projects.rs
```diff
@@ -0,0 +1,304 @@
...
+    async fn resolve_project(
+        &self,
+        account_id: &AccountId,
+        project_name: &str,
+    ) -> Result<Option<ProjectId>, WorkerExecutorError>;
+}
+
+pub fn configured(
+    config: &ProjectServiceConfig,
+    cache_config: &ComponentCacheConfig,
```
mschuwalow: Let's make a dedicated cache config for this instead of reusing the component cache config.
Can probably be part of ProjectServiceConfig::Grpc

golem-worker-executor/src/services/projects.rs
```diff
@@ -0,0 +1,304 @@
...
+    project_id: ProjectId,
+}
+
+#[async_trait]
+impl ProjectService for ProjectServiceDisabled {
+    async fn get_project_owner(
+        &self,
+        _project_id: &ProjectId,
+    ) -> Result<AccountId, WorkerExecutorError> {
+        Ok(self.account_id.clone())
```
mschuwalow: Let's assert here that the account id is as expected, just to detect misconfigurations earlier instead of continuing with possibly wrong data

golem-worker-executor/src/services/projects.rs
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    async fn resolve_project(
```
mschuwalow, line 301: same here, let's check the account_id and make the expected project name part of the configuration (and check that as well)

vigoo:
> Looks good to me.
> 
> One small question, why did you decide to use the project_id as the owner instead of the component_id (as we can always resolve component_id -> project_id -> account_id if needed)

Hm actually the component ID is already part of the worker ID so if we would do that, we could basically just drop the `OwnedWorkerId` thing. I did not think of this because we already had this concept of `WorkerId` + `AccountId` and changing that to be project based looked like the simplest/fastest refactoring step. 

So basically we can decide to 
- doing it this way - pass the project id around and get it from the upper layers so we don't need to query and cache it whenever we need it for something
- further simplify the inside of the executor by dropping this (sometimes we still need account id though to track who created a worker) but resolve (and cache) the component owner whenever needed

I think it may be a good idea to do the second but I would rather do it as a separate second step, just as a simplification refactoring, primarily because of time constraints. What do you think?

mschuwalow:
> > Looks good to me.
> > One small question, why did you decide to use the project_id as the owner instead of the component_id (as we can always resolve component_id -> project_id -> account_id if needed)
> 
> Hm actually the component ID is already part of the worker ID so if we would do that, we could basically just drop the `OwnedWorkerId` thing. I did not think of this because we already had this concept of `WorkerId` + `AccountId` and changing that to be project based looked like the simplest/fastest refactoring step.
> 
> So basically we can decide to
> 
>     * doing it this way - pass the project id around and get it from the upper layers so we don't need to query and cache it whenever we need it for something
> 
>     * further simplify the inside of the executor by dropping this (sometimes we still need account id though to track who created a worker) but resolve (and cache) the component owner whenever needed
> 
> 
> I think it may be a good idea to do the second but I would rather do it as a separate second step, just as a simplification refactoring, primarily because of time constraints. What do you think?

sounds good to me :+1:

PR #1842, opened 2025-07-15
Support wasi:config/store in worker executor

### LINKED ISSUE #1227: Implement wasi-runtime-config in worket executor

mschuwalow:
resolves #1227


vigoo:
I was thinking about the high level approach, and wanted to point out one thing that I'm undecided whether it is a problem or not.
If we do this in the "usual" way of treating the `get-config(key)` function as a durable function, it means every time the worker accesses a configuration key, we have a corresponding oplog entry with information stored about both the key and the value. 
With your method there is no oplog entry about the configuration access at all, but I guess the config map is part of the initial create entry, and successful update entries (did not check the code in detail yet but that's my understanding). 

Both has the same functionality and depending on how frequent the config access is, having them durable may have a performance impact, but on the other hand from an observability point of view the per-call durability seems to be nicer, especially if a larger config map is associated with the worker - with the second approach we don't have any information about which particular keys the code is accessing and when.

This of course can also be seen as an advantage, because with the second approach adding/removing get-config calls is not causing any divergence while with the first approach it is.

golem-worker-executor/src/durable_host/config.rs
```diff
@@ -0,0 +1,41 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::DurableWorkerCtx;
+use crate::preview2::wasi::config::store::{Error, Host};
+use crate::workerctx::WorkerCtx;
+
+/// `wasi:config/store` implementation
+impl<Ctx: WorkerCtx> Host for DurableWorkerCtx<Ctx> {
```
vigoo: Small detail, but all the other host implementations are in "module directories" so to be consistent I would be more happy to see this in `config/mod.rs`

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -200,6 +201,12 @@ impl<Ctx: WorkerCtx> DurableWorkerCtx<Ctx> {
         )
         .await?;
 
+        // TODO: pass config vars from component metadata
```
vigoo, line 204: What about this TODO?

vigoo: I see, it is coming as a next step.

vigoo (approved):

PR #1844, opened 2025-07-16
Minor: Worker files - List file as part of list_directory

### LINKED ISSUE #1761: Worker file Endpoints
There are currently two endpoints for interacting with worker files. 

1) get_files - which can only get details of a specific folder, not of specific file or just list all files

example - 
`http://localhost:9881/v1/components/<component_uri>/workers/<worker_name>/files/%2Fdata%2Ftest.png`
`Bad file type: /data/test.png`

example -
`http://localhost:9881/v1/components/<component_uri>/workers/<worker_name>/files/%2Fdata`
This works and outputs all files under data, with modified, size and permission

Make it so that it displays details of the files if file is inputed

2) get_file_content -> works great (the name the download file should be the same as file in worker, currently it is / is converted _ and full path is used)

Nanashi-lab:
/closes #1761 

Changes
for function list_directory -> Rather than outputing NotaDirectory Error, this now Display File(FileSystemNode)
If file is requested at the endpoint, it displays the the meta details of the file rather than displaying Bad file error

This is not the most needed change, especially with new CLI command making it easier to browse files.
I did this first and then the Cli, so. Good to learn atleast. 
I update the integration test to test file node as well. 

The Cli changes are independent, and will work without this.

### DIFF
--- a/golem-api-grpc/proto/golem/worker/v1/worker_service.proto
+++ b/golem-api-grpc/proto/golem/worker/v1/worker_service.proto
@@ -302,7 +302,7 @@ message GetFileSystemNodeRequest {
 message GetFileSystemNodeResponse {
   oneof result {
     ListFileSystemNodeResponse success = 1;
-    WorkerError error = 3;
+    WorkerError error = 2;
   }
 }
 


golem-api-grpc/proto/golem/workerexecutor/v1/worker_executor.proto
```diff
@@ -400,7 +400,6 @@ message ListDirectoryResponse {
   oneof result {
     ListDirectorySuccessResponse success = 1;
     golem.common.Empty not_found = 2;
-    golem.common.Empty not_a_directory = 3;
```
mschuwalow, line 403: Please introduce a dedicated new oneof case when it's a file and include the metadata there. It's fine to break backwards compat of the grpc api here.

Nanashi-lab: Is it necessary ? 

ListDirectorySuccessResponse, -> Outputs repeated golem.worker.FileSystemNode 
Since file outputs FileSystemNode, it as part of ListDirectorySuccessResponse, it is still repeated golem.worker.FileSystemNode, with just one element. 

So for directory -> vec[of all FileSystemNode under it]
for file -> vec[Single FileSystemNode] 
 
I think this is a better way, than adding a new oneof which lists a single filesystemnode

Nanashi-lab: If the grpc must match the enum inside, I can remove file(FileStstemNode) and make it part of successful call in vec[FileSystemNode]

mschuwalow: Please just make it a new message.

The reason not to reuse the ListDirectorySuccessResponse is that it includes a list of FileSystemNode, while we only have one here.
We also want the client to be able to easily tell whether it tried to list a file or a directory. Consider how it works with the `ls` command:
```
ls foo
> foo
```
Do I have a file foo here or a folder foo with a file foo in it? It's better to avoid such problems if we have the option.

The reason for making it a message instead of inlining is:
* Repeated is not supported inside of oneof afaik
* It's inconsistent with our other existing apis
* It makes it more difficult to evolve the api in the future

---

While you are doing that, please also update the name of the rpc method (ListDirectory) and the messages (ListDirectoryRequest, ListDirectoryResponse, ListDirectorySuccessResponse) as they no longer match the semantics of the method.
I would propose something like GetFileSystemNode, where the result is File { self_node: FileSystemNode } | Directory { nodes: repeated FileSystemNode } | DoesNotExist | Error.

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2017,16 +2017,39 @@ impl<Ctx: WorkerCtx + DurableWorkerCtxView<Ctx>> FileSystemReading for DurableWo
...
-        {
-            let metadata = tokio::fs::metadata(&target).await.map_err(|e| {
-                WorkerExecutorError::FileSystemError {
-                    path: path.to_string(),
-                    reason: format!("Failed to get metadata: {e}"),
-                }
-            })?;
-            if !metadata.is_dir() {
-                return Ok(ListDirectoryResult::NotADirectory);
+        let metadata = tokio::fs::metadata(&target).await.map_err(|e| {
```
mschuwalow, line 2020: The name of the enclosing method (list_directory) is no longer accurate.
Please rename it to something that matches the new semantics

Nanashi-lab:
@mschuwalow I am still not done, there is a small bug in how file gets passed through worker-executor and then worker-service

golem-api-grpc/proto/golem/worker/v1/worker_service.proto
```diff
@@ -294,21 +294,22 @@ message SearchOplogSuccessResponse {
...
   golem.worker.TargetWorkerId worker_id = 1;
   string path = 2;
 }
 
-message ListDirectoryResponse {
+message GetFileSystemNodeResponse {
   oneof result {
-    ListDirectorySuccessResponse success = 1;
-    WorkerError error = 2;
+    ListFileSystemNodeResponse success = 1;
```
Nanashi-lab, line 304: I have left this flatten, repeated FileSystemNode. This is for service, user facing, worker executor has seperated File and Directory. By this time, the structure is already converted into vec<componentfilestsytemnode>, and this is passed to both http endpoint and grpc service. Trying to pass whole structure with both File() and Directory(), requires unnecessary code in both http endpoint and 
service, as executor grpc needs to be converted into service grpc. 

forwarding.rs -> flattens the values

golem-worker-executor/src/workerctx/mod.rs
```diff
@@ -443,10 +442,10 @@ pub trait PublicWorkerIo {
 #[async_trait]
 pub trait FileSystemReading {
     // List the contents of a directory. Will return an error if the path is not a directory.
```
mschuwalow:
```suggestion

```

Nanashi-lab: I missed this one. Thanks again.

Nanashi-lab:
Test pass locally, tested it

PR #1867, opened 2025-07-25
Forward stdout/stderr when debugging workers

### LINKED ISSUE #1845: Debugging service stdout/err/log streaming
We want the debugger clients to see stdout/err/log entries emitted during debugging sessions.

This requires two major changes:

- Worker executor needs to be configurable in a way that it emits these entries during recovery too. During normal worker recovery (not a debugging session) we don't want the users to get what has been previously sent out before the recovery.
- These log events need to be merged into the debugging service's connection stream

#### Worker executor configuration
See `DurableWorkerCtx::emit_log_event`. Here the current logic is:
- if we are in replay mode, never emit the event
- if we are in live mode, as the persisted log events are treated as "hint" entries (to not cause divergence), it is possible that some log entries would be double emitted after replay. this is handled by remembering the hashes of the log entries seen during replay.
- 

The (either per-worker or per app?) configuration should be accessed in this `emit_log_event` and it should make it always emit the events when turned on.

#### Merging the events into the connection
Check how `WorkerEventStream` is implemented - that's the gRPC worker connection stream in the worker executor. Similar to that, using the `WorkerEventReceiver`, the events should be mergeable into the debugger connection.

### LINKED ISSUE #1865: Span not found crash with parallel http requests
Making two long-running HTTP requests in parallel can lead to a `Runtime error: Span db08258923c4d83d not found` getting the worker to a stuck state.

mschuwalow:
resolves #1845


golem-debugging-service/test-components/shopping-cart/src/lib.rs
```diff
@@ -18,21 +16,6 @@ struct State {
     items: Vec<ProductItem>,
 }
 
-fn reserve_inventory() -> Result<(), &'static str> {
```
mschuwalow, line 21: nondeterminism was messing with being able to assert on the exact structure of logs emitted

vigoo (approved):

PR #1887, opened 2025-08-07
Build with panic=abort

### LINKED ISSUE #1886: Panic in the worker invocation loop leaves the worker in a corrupted state
Goal of this ticket is to fix the following scenario:

* For some reason the worker invocation loop panics (in this case a stdout issue during replay, but I guess we have more of these):
```
/home/mschuwalow/Downloads/logs/worker-executor.9094.log         ┌2025-08-06T14:06:49.586248+0200  INFO api_request{api="assign_shards" api_type="grpc"}:waiting-for-permits{worker_id="509d4a81-ca58-4bc5-8493-4853e85b858a/f19b641a-813d-4456-adb1-43295e835478"}:invocation-loop{worker_id="509d4a81-ca58-4bc5-8493-4853e85│
/home/mschuwalow/Downloads/logs/worker-executor.9094.log         │2025-08-06T14:06:49.669275+0200 ERROR api_request{api="assign_shards" api_type="grpc"}:waiting-for-permits{worker_id="509d4a81-ca58-4bc5-8493-4853e85b858a/f19b641a-813d-4456-adb1-43295e835478"}:invocation-loop{worker_id="509d4a81-ca58-4bc5-8493-4853e85│
/home/mschuwalow/Downloads/logs/worker-executor.9094.log         │    0: 0x7fc0c1 - <unknown>!<wasm function 13371>                                                                                                                                                                                                           │
/home/mschuwalow/Downloads/logs/worker-executor.9094.log         │    1: 0x2421b8 - <unknown>!<wasm function 5185>                                                                                                                                                                                                            │
/home/mschuwalow/Downloads/logs/worker-executor.9094.log         │    2: 0x80af21 - <unknown>!post_test:example-exports/test-example-api#fetch-json                                                                                                                                                                           │
/home/mschuwalow/Downloads/logs/worker-executor.9094.log         │2025-08-06T14:06:50.058260+0200 ERROR api_request{api="assign_shards" api_type="grpc"}:waiting-for-permits{worker_id="509d4a81-ca58-4bc5-8493-4853e85b858a/f19b641a-813d-4456-adb1-43295e835478"}:invocation-loop{worker_id="509d4a81-ca58-4bc5-8493-4853e85│
/home/mschuwalow/Downloads/logs/worker-executor.9094.log         │failed printing to stdout: Resource temporarily unavailable (os error 35) panic_backtrace=   0: std::backtrace_rs::backtrace::libunwind::trace                                                                                                              │
/home/mschuwalow/Downloads/logs/worker-executor.9094.log         │             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/../../backtrace/src/backtrace/libunwind.rs:117:9
```
* The panic in the invocation loop causes the entire spawned invocation loop in the worker executor (golem-worker-executor/src/worker/invocation_loop) to panic/exit without notifying the parent RunningWorker instance about it
* This results in a corrupted state of the worker: the invocation queue is not running / receiver for the notification queue is dropped / ... while the parent worker instance thinks that everything is fine. Trying to interact with the worker will start failing now in various "weird" ways. For example trying to invoke the worker will panic as we unwrap a senderror due to the dropped receiver side of the notification queue

mschuwalow:
resolves #1886

### DIFF
--- a/wasm-rpc/Cargo.toml
+++ b/wasm-rpc/Cargo.toml
@@ -13,7 +13,6 @@ include = []
 
 [lib]
 path = "src/lib.rs"
-crate-type = ["cdylib", "rlib"]
 harness = false
 
 [dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -298,3 +298,9 @@ redis-protocol = { git = "https://github.com/golemcloud/redis-protocol.rs.git", 
 wasmtime = { git = "https://github.com/golemcloud/wasmtime.git", branch = "golem-wasmtime-v33.0.0" }
 wasmtime-wasi = { git = "https://github.com/golemcloud/wasmtime.git", branch = "golem-wasmtime-v33.0.0" }
 wasmtime-wasi-http = { git = "https://github.com/golemcloud/wasmtime.git", branch = "golem-wasmtime-v33.0.0" }
+
+[profile.dev]
+panic = "abort"
+
+[profile.release]
+panic = "abort"
--- a/golem-worker-service/src/gateway_api_definition/http/api_oas_convert.rs
+++ b/golem-worker-service/src/gateway_api_definition/http/api_oas_convert.rs
@@ -717,7 +717,7 @@ struct ExtractedBindingData<'a> {
     cors_preflight: Option<&'a HttpCors>,
 }
 
-fn extract_binding_data(binding: &GatewayBindingCompiled) -> ExtractedBindingData {
+fn extract_binding_data(binding: &GatewayBindingCompiled) -> ExtractedBindingData<'_> {
     match binding {
         GatewayBindingCompiled::Worker(w) => ExtractedBindingData {
             component_id: Some(&w.component_id),


wasm-rpc/Cargo.toml
```diff
@@ -13,7 +13,6 @@ include = []
 
 [lib]
 path = "src/lib.rs"
-crate-type = ["cdylib", "rlib"]
```
mschuwalow, line 16: This is needed due to https://github.com/rust-lang/cargo/issues/6313#issuecomment-480014371.

Do we have a need for the cdylib currently? If yes, I will check whether I can find a workaround

vigoo: I think it's needed when it is used in guest code (like the generated rpc stubs which we still support, although probably not for long)

vigoo:
The test execution (test-r, but also the built-in one) relies on capturing panicks so I think if it is enabled like this, it is going to make our failing tests no longer properly reported. Is there a way to only enable this for the service executables?

mschuwalow:
The tests are still built with panic unwind (I don't think you can even override that), so afaik nothing needs to be done there

wasm-rpc/Cargo.toml
mschuwalow: Ok, will mark as draft while looking for a workaround then

mschuwalow: Had a look. So from what I understand cdylib is only needed when compiling to wasm for the "main" / the crate that is directly being compiled to wasm (e.g. rpc stubs, component using golem-rust, ...). Any libraries that are used can stay as rlib.

If we do want to have the cdylib the best workaround seems to be to create a new dedicated crate (i.e. golem-wasm-rpc-lib) that reexports everything from the underlying crate and only compiles as a cdylib.
But I would prefer to avoid doing this.

TLDR: I think we should be fine without the cdylib. I would like to merge and try. If issues show up, I can introduce a new dedicated crate as workaround.

vigoo: You can try it out by creating an app with two components using `static-wasm-rpc` dependency type (or one from the `test-components`) and just tell golem-cli to use the local modified wasm-rpc crate instead of the published one (there is an env var for that)

mschuwalow: Yes, I tested the non-cdylib with a test-component and it works

vigoo (approved):

PR #1888, opened 2025-08-07
debugging-service: Do not error when trying to write to oplog

mschuwalow:
Fixes issue reported by @andrueandersoncs

### DIFF
--- a/golem-worker-executor/src/model/mod.rs
+++ b/golem-worker-executor/src/model/mod.rs
@@ -249,10 +249,7 @@ impl TrapType {
                             Some(WorkerExecutorError::ValueMismatch { details }) => {
                                 TrapType::Error(WorkerError::InvalidRequest(details.clone()))
                             }
-                            _ => {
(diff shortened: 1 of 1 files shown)


golem-debugging-service/src/oplog/debug_oplog.rs
```diff
@@ -129,9 +129,8 @@ impl Oplog for DebugOplog {
         self.inner.length().await
     }
 
-    async fn upload_payload(&self, _data: &[u8]) -> Result<OplogPayload, String> {
-        // in a debugging session we don't need to upload anything
-        Err("Workers in debug cannot upload any new data to oplog".to_string())
+    async fn upload_payload(&self, data: &[u8]) -> Result<OplogPayload, String> {
```
mschuwalow, line 132: Going past the last entry in the oplog could cause the playback rpc to fail as there was a race between this and the check_interrupt mechanism for trapping the worker.

PR #1964, opened 2025-08-19
Miscalleneous improvements and fixes in Rib REPL side to interact with code first agents properly

afsalthaj:
Note: Code first agents has moved away from considering agents as wit-resource, but still good to get this PR in.

* Handle static functions in resource.  Until now, rib handled only resource constructors where it was sure what is the exact return type - i.e, same  resource type. Now it can call a static function. The resource name in the resource type is helping out in situations where the return type of a static function is a resource but wrapped in `result`. `result<agent, string>` as an example.

```rust
let worker = instance("my-worker");
 // we created an instance of cart - underlying `new` constructor
let default-cart = worker.cart("param1", "param2"); 

 // we called a static fn `create` in `cart` without creating an instance of cart, Note: We still consider `cart` to be a member of `worker`, and hence `worker.cart`
let cart-custom-result = worker.cart.create("foo"); 

match cart-custom-result {
  ok(custom-cart) => custom-cart.checkout(),
  err(_)          => default-cart.checkout()
}
```

* Minor change: Rib REPL printer changes. Resource types is not supported in wasm-wave but Rib REPL handled this for a long time. It used to check if the final result is a resource value, and if so handle it separately, otherwise piggyback on wasm-wave print. With this PR, it sort of extend this functionality to print values that may have resource type in it at any position. Example: `result<resource-type, string>`

* Minor change: With multiple testing, I bumped into the following errors in integration tests, for postgres. The culrpit was unique network id, and I had to clear a lot of my local to get it working back again, and later I bumped into it again. So I just changed to whatever it was before - i.e, to not use unique network etc. May be for CI, it doesn't matter, but it just a matter of time, that someone else bump into the same situation in their local. 

    >Failed to start Postgres container: Client(CreateNetwork(DockerResponseServerError { status_code: 400, message: "all predefined address pools have been fully subnetted" }))


vigoo (approved):
I'm not sure about the unique network thing - that was added because it helped with CI flakiness

afsalthaj:
@vigoo  The unique network thing did solve CI flakiness around worker-gateway api integration tests. However, I did fix some logical issues with those tests later on and removed that flakiness I believe. So now may be, unique network is not mandatory. Just to be cautious, I didn't fully delete the code that does this unqiue network thing, but just disabled it. I can try re-running the CI job for this PR multiple times just to be sure that I didn't mess up with CI - before merging

afsalthaj:
Reverted that change, just to ensure I am not changing any other behaviour of other tests.

The API tests actually became flaky after I re-introduced unique network, but not changing it again.

PR #1972, opened 2025-08-19
Split CLI integration tests into two jobs

vigoo:


vigoo:
This did not get any faster, will try a different split

vigoo:
Does not seem to worth it

PR #1982, opened 2025-08-21
environment shares

mschuwalow:


golem-registry-service/db/migration/postgres/001_init.sql
```diff
@@ -650,3 +650,55 @@ CREATE TABLE environment_plugin_installation_revisions
 
 CREATE INDEX environment_plugin_installation_revisions_plugin_idx
     ON environment_plugin_installation_revisions (plugin_id);
+
```
noise64, line 653: the repo tests now use the sqlite one, so please also add it to that one

golem-registry-service/db/migration/postgres/001_init.sql
```diff
@@ -650,3 +650,55 @@ CREATE TABLE environment_plugin_installation_revisions
...
+    modified_by UUID      NOT NULL,
+
+    current_revision_id BIGINT    NOT NULL,
+
+    CONSTRAINT environment_shares_pk
+        PRIMARY KEY (environment_share_id),
+    CONSTRAINT environment_shares_environments_fk
+        FOREIGN KEY (environment_id) REFERENCES environments,
+    CONSTRAINT environment_shares_accounts_fk
+        FOREIGN KEY (grantee_account_id) REFERENCES accounts (account_id)
```
noise64, line 672: foreign keys should be covered by index (environment_id is okay, because environment_shares_environment_grantee_uk is a covering index for environment_id, but grantee_account_id has no indexes)

golem-registry-service/src/repo/environment_share.rs
```diff
@@ -0,0 +1,417 @@
...
+                        "#})
+                            .bind(environment_id)
+                            .bind(revision.environment_share_id)
+                            .bind(grantee_account_id)
+                            .bind(&revision.audit.created_at)
+                            .bind(revision.audit.created_by)
+                            .bind(revision.revision_id)
+                    )
+                    .await
+                    .to_error_on_unique_violation(EnvironmentShareRepoError::ShareViolatesUniqueness)?;
```
noise64, line 274: i would simply consider this concurrent modification error, and do not make more specific ones, like RevisionForUpdateNotFound in update (which is kinda also misleading, it is not "not found", rather, not the current one). just like with the prev repos i would like to keep them ConcurrentModification errors, as the clients should behave the same: refetch their data, and then decide what to do.

it is good to have more specific errors for user provided identitiy uniqueness errors (version, name, email), but for optimistic lock errors i do not think that client can do anything meaningfully different, and depending on interleave patterns they might get one or the other.

golem-registry-service/db/migration/postgres/001_init.sql
```diff
@@ -650,3 +650,55 @@ CREATE TABLE environment_plugin_installation_revisions
...
+);
+
+CREATE UNIQUE INDEX environment_shares_environment_grantee_uk
+    ON environment_shares (environment_id, grantee_account_id)
+    WHERE deleted_at IS NULL;
+
+CREATE TABLE environment_share_revisions
```
noise64, line 685: for env component plugin installations i collapsed the revision - revision-plugin-relations tables into one. it is also works like this, but i think for these specfic versioned relations it is fine to skip the "empty" revision table

mschuwalow: I think it's nice to keep this structure as it's easier to add additional columns later on.
In the domain an environment share is very much it's own thing (considering we even have a ui view that lists all of the shares and lets people update / delete them), so I think it's worth keeping.

noise64 (approved):

PR #1985, opened 2025-08-21
Update main.wit in template-golem-agent-ts wit and update SDK with bug fixes

afsalthaj:
Only this version of SDK reasonably behave (and avoid surprises for invocations - not just type-def extractions).
Also updating  template-golem-agent-ts wit/main.wit, without which it won't work


test-components/my_agent.wasm
afsalthaj, line 1: Oops! I  will remove this

PR #1987, opened 2025-08-24
Add auth checks

### LINKED ISSUE #1901: Copy of Registry repo: HTTP API domains (sites) and certificates

mschuwalow:
resolves #1901

POC for now to show the proposed approach, will do a couple of iterations and then propagate.

High level idea is to resolve the user to an AuthCtx early on in the request lifecycle that can _very cheaply_ be used to do authorization checks.
Any read operations against the db apply basic visibility checks also return the relevant information to check auth for the retrieved resource.

### DIFF
--- a/golem-registry-service/db/migration/postgres/001_init.sql
+++ b/golem-registry-service/db/migration/postgres/001_init.sql
@@ -79,9 +79,7 @@ CREATE TABLE account_revisions
     CONSTRAINT account_revisions_accounts_fk
         FOREIGN KEY (account_id) REFERENCES accounts,
     CONSTRAINT account_revisions_plans_fk
-        FOREIGN KEY (plan_id) REFERENCES plans,
-    CONSTRAINT account_revisions_check_roles
-        CHECK ((roles & ~3) = 0) -- for 3 roles (bits 0,1),
+        FOREIGN KEY (plan_id) REFERENCES plans
 );
 
 CREATE TABLE tokens
@@ -689,7 +687,5 @@ CREATE TABLE environment_share_revisions
     CONSTRAINT environment_share_revisions_pk
         PRIMARY KEY (environment_share_id, revision_id),
     CONSTRAINT environment_share_revisions_environment_shares_fk
-        FOREIGN KEY (environment_share_id) REFERENCES environment_shares,
-    CONSTRAINT environment_share_revisions_check_roles
-        CHECK ((roles & ~7) = 0) -- for 3 roles (bits 0,1,2)
+        FOREIGN KEY (environment_share_id) REFERENCES environment_shares
 );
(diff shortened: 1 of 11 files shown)


golem-registry-service/db/migration/postgres/001_init.sql
```diff
@@ -687,24 +679,17 @@ CREATE TABLE environment_share_revisions
...
-    revision_id          BIGINT NOT NULL,
-    role                 INT    NOT NULL,
-
-    CONSTRAINT environment_share_revision_roles_pk
-        PRIMARY KEY (environment_share_id, revision_id, role),
-    CONSTRAINT environment_share_revision_roles_environment_share_revisions_fk
-        FOREIGN KEY (environment_share_id, revision_id) REFERENCES environment_share_revisions
+        FOREIGN KEY (environment_share_id) REFERENCES environment_shares,
+    CONSTRAINT environment_share_revisions_check_roles
+        CHECK ((roles & ~7) = 0) -- for 3 roles (bits 0,1,2)
```
noise64: not sure i would do a check for this, as we do double check in the repo layer anyway, and if we need to add a new role, we have to drop / recreate this.
(as in, we do not add checks for enums neither, to check their range)

golem-registry-service/src/repo/account.rs
```diff
@@ -43,11 +43,13 @@ pub trait AccountRepo: Send + Sync {
     async fn get_by_id(
         &self,
         account_id: &Uuid,
+        include_deleted: bool,
```
noise64: is this really needed, used?

golem-registry-service/src/repo/account.rs
```diff
@@ -297,50 +252,46 @@ impl AccountRepo for DbAccountRepo<PostgresPool> {
...
             .fetch_optional_as(
                 sqlx::query_as(indoc! {r#"
-                    SELECT a.created_at AS entity_created_at, ar.account_id, ar.revision_id, ar.name, ar.email, ar.plan_id, ar.created_at, ar.created_by, ar.deleted
+                    SELECT a.created_at AS entity_created_at, ar.account_id, ar.revision_id, ar.name, ar.email, ar.plan_id, ar.roles, ar.created_at, ar.created_by, ar.deleted
                     FROM accounts a
                     JOIN account_revisions ar ON ar.account_id = a.account_id AND ar.revision_id = a.current_revision_id
-                    WHERE a.email = $1 AND a.deleted_at IS NULL
+                    WHERE
+                        a.email = $1
+                        AND ($2 OR a.deleted_at IS NULL) -- check deletion
```
noise64: it feels weird for me to send comments as part of mostly simple queries

golem-registry-service/src/repo/account.rs
mschuwalow: will remove

PR #1991, opened 2025-08-25
Mocking imported AI WIT interfaces and adding imports to the TS agent base template

### LINKED ISSUE #1918: Integrate golem-ai into the TS agent template
All the Golem AI APIs must be accessible from the TS template, but by default we should not compose any of the implementations with the user's component. 

Some parts of this:

- include the d.ts of all APIs in the typescript template
- include at least "commented out" configuration for each AI library for each provider in the template's YAML, so the used ones can be enabled (as either plugins or via build-time composition - to be decided) easily
- "mock" the unused ones in the executor's dynamic linker

vigoo:
Resolves #1918

- [x] Mocks imported golem-ai WIT interfaces in the executor
- [x] Adds all golem-ai libraries to the base TS agent template (Note: WIT files are manually copied from the `golem-ai` repo for now)
- [x] Fix issues in wasm-rquickjs preventing generating the import modules
- [x] Generate d.ts definitions for the imports and include them in the agent template
- [x] Add commented-out dependencies in the agent tempalte for all library's all provider
- [x] Merge https://github.com/golemcloud/golem-agentic/pull/73 and update in this PR 
- [x] golem-ai dev release (and update links if needed)


cli/golem-templates/templates/ts/ts-app-component-default/components-ts/component-name/golem.yaml
```diff
@@ -4,4 +4,144 @@ components:
   componentname:
     template: ts
 
+# The following block contains commented-out dependencies for various Golem AI libraries.
```
noise64, line 7: currently we only have one template, but later i think we might want to move this where we replace `#golem-app-manifest-component-hints`

cli/golem-templates/templates/ts/ts-app-component-default/components-ts/component-name/golem.yaml
```diff
@@ -4,4 +4,144 @@ components:
...
+# For each area (such as LLM, Search, etc) only one of the providers can be commented out.
+# If no provider dependency is commented out, then using that AI API will be a runtime failure.
+dependencies:
+  componentname:
```
noise64, line 11: is this allowed currently? as having a dep with empty componentname? also should this be componentName?

Cargo.toml
```diff
@@ -319,7 +319,7 @@ walkdir = "2.5.0"
 warp = "0.3.7"
 wasm-encoder = "0.235.0"
 wasm-metadata = { version = "0.235.0", features = ["oci"] }
-wasm-rquickjs = "0.0.5"
+wasm-rquickjs = "0.0.9"
```
noise64, line 322: should we also bump the moonbit generator, so it won't print the debug lines?

cli/golem-templates/templates/ts/ts-app-component-default/components-ts/component-name/golem.yaml
noise64: ahh nevemind, need more coffe

Cargo.toml
vigoo: I am doing that in my next PR

cli/golem-templates/templates/ts/ts-app-component-default/components-ts/component-name/golem.yaml
vigoo: it works

PR #2004, opened 2025-08-27
Add new name filter and special support in cli

### LINKED ISSUE #1836: Agent instance discovery
- host function
- API
- CLI
- integration tests

Based on agent definition discovery, worker enumeration and worker resource tables.

### LINKED ISSUE #1912: Make the public oplog entry changes available on the WIT oplog entry types

vigoo:
As with the latest changes agents are just workers, this PR only tries to do the minimum to support agent instance listing on top of the existing features.

Agent-oriented API and CLI is coming in the next version.

So this PR:
- Adds a new string filter command `startswith` which is convenient to filter for agent types (as the name starts with the agent type name)
- Adds a new option to the `worker list` command in CLI to pass an agent type name. If passed, it looks up the component that implements it, and modifies the query to use that component, and adds an additional `startswith` filter to the query.

The same technique (find component by agent type name, and use the proper) filter can be used on the REST API and on the Host API if any user needs it before the dedicated agent-oriented APIs appear.

Resolves #1836 
Resolves #1912 (new resource name/owner fields in public oplog entries in WIT)

### DIFF
--- a/golem-worker-executor/src/model/public_oplog/wit.rs
+++ b/golem-worker-executor/src/model/public_oplog/wit.rs
@@ -210,18 +210,18 @@ impl From<PublicOplogEntry> for oplog::OplogEntry {
                 timestamp: timestamp.into(),
                 resource_id: id.0,
                 name,
-                owner
+                owner,
             }),
             PublicOplogEntry::DropResource(ResourceParameters {
                 timestamp,
                 id,
                 name,
-                owner
+                owner,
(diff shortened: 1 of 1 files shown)


vigoo:
This is failing because it requires a new version of `golem-rust` with the WIT updates. As that is time consuming, I'll wait to see if there is any other change affecting the golem host WITs, and do update everything before the first RC.

noise64 (approved):

PR #2013, opened 2025-09-01
Update sdk to get better error messages

afsalthaj:


afsalthaj:
I am merging this as it is only a version update 

image

PR #2015, opened 2025-09-01
Fix agent initialization

### LINKED ISSUE #2011: Invoke on agents not working (and init errors)
The initialization / invoke seems to be not working with main branch of golem  (latest commit: d80697101c63c2060540c3aedd665fd58238286d).

### Issue 1
I tried to invoke  and I got this in return. 

```rust
>>> let x = instance[weather-agent]("weather-agent(1)")
()
>>> x.get-weather("foo", {data: "foo", value: 42})
[runtime error] failed to invoke function my:agent/weather-agent.{get-weather}: Worker Service - Error: 500 Internal Server Error, Invalid request: expected 2, got 1 parameters

```

#### Issue 2 or an observation 

Now the second thing that I observed is, invoking it again resulted "agent never got initialized error"

```rust

>>> let x = instance[weather-agent]("weather-agent(1)")
()
>>> x.get-weather("foo", {data: "foo", value: 42})
[runtime error] failed to invoke function my:agent/weather-agent.{get-weather}: Worker Service - Error: 500 Internal Server Error, Invalid request: expected 2, got 1 parameters
>>>
>>>
>>> x.get-weather("foo", {data: "foo", value: 42})
err(custom-error({
  value: {
    nodes: [
      prim-string("Agent is not initialized")
    ]
  },
  typ: {
    nodes: [
      {
        name: none,
        owner: none,
        type: prim-string-type
      }
    ]
  }
}))
```

#### Note
Not sure how much it helps, just to make sure these are all making use of recent changes, I _INTENTIONALLY_ gave a wrong agentId and it picked up and gave the correct. So pretty sure it isn't about wrong/invalid agent-id

```rust
>>> let x = instance[weather-agent]("weather-agent(1)))))))")
()
>>> x.get-weather("foo", {data: "foo", value: 42})
[runtime error] failed to invoke function my:agent/weather-agent.{get-weather}: Worker Service - Error: 500 Internal Server Error, Invalid request: Invalid agent id: Failed to parse parameter value 1)))))))))))): trailing characters after value at 1..2
```

But it gave a 500 Internal Server error. May be it's true when Rib natively supports it.

### How to reproduce?

Create the template as usual, but with the following code, and then do `..golem repl`

```ts
import {
    BaseAgent,
    agent,
    prompt,
    description,
} from '@golemcloud/golem-ts-sdk';

interface CustomData {
    data: string;
    value: number;
}

@agent()
class AssistantAgent extends BaseAgent {
    @prompt("Ask your question")
    @description("This method allows the agent to answer your question")
    async ask(name: string): Promise<string> {
        const customData = { data: "Sample data", value: 42 };

        const remoteWeatherClient = WeatherAgent.get(1);
        const remoteWeather = await remoteWeatherClient.getWeather(name, customData);

        return (
            `Hello! I'm the assistant agent (${this.getId()}) reporting on the weather in ${name}. ` +
            `Here’s what the weather agent says: "\n${remoteWeather}\n". `
        );
    }
}

@agent()
class WeatherAgent extends BaseAgent {
    private readonly userId: number;

    constructor(userId: number) {
        super()
        this.userId = userId;
    }

    @prompt("Get weather")
    @description("Weather forecast weather for you")
    async getWeather(name: string, param2: CustomData): Promise<string> {
        return Promise.resolve(
            `Hi ${this.userId} Weather in ${name} is sunny. Params passed: ${name} ${JSON.stringify(param2)}. ` +
            `Computed by weather-agent ${this.getId()}. `
        );
    }
}


```

```rust
>>> let x = instance[weather-agent]("weather-agent(1)")
()
>>> x.get-weather("foo", {data: "foo", value: 42})
```

### FYI if it helps
1. It looks like `initialize` parameters are used to `invoke` get-weather. Hence the first mismatch. Let me confirm if that's the case.
2. These are new bugs as these invokes used to work before - and hopefully this regression is a minor fix.
2. At SDK side, there are invoke tests (which can be done only after init, and therefore init tests too): 
Tests in SDK: https://github.com/golemcloud/golem-agentic/blob/main/golem-ts-sdk/tests/invoke.test.ts
Agents impored in SDK tests: https://github.com/golemcloud/golem-agentic/blob/main/golem-ts-sdk/tests/sampleAgents.ts

vigoo:
Resolves #2011

### DIFF
--- a/golem-common/src/model/agent/mod.rs
+++ b/golem-common/src/model/agent/mod.rs
@@ -79,7 +79,7 @@ pub enum AgentError {
     InvalidMethod(String),
     InvalidType(String),
     InvalidAgentId(String),
-    CustomError(ValueAndType),
+    CustomError(#[wit_field(convert = golem_wasm_rpc::WitValue)] ValueAndType),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Encode, Decode, IntoValue)]
@@ -403,7 +403,7 @@ impl Display for NamedElementValue {
 #[cfg_attr(feature = "poem", oai(discriminator_name = "type", one_of = true))]
 #[serde(tag = "type")]
 pub enum ElementValue {
-    ComponentModel(ValueAndType),
+    ComponentModel(#[wit_field(convert = golem_wasm_rpc::WitValue)] ValueAndType),
     UnstructuredText(TextReference),
     UnstructuredBinary(BinaryReference),
 }
--- a/golem-worker-executor/src/worker/mod.rs
+++ b/golem-worker-executor/src/worker/mod.rs
@@ -1341,7 +1341,10 @@ impl<Ctx: WorkerCtx> Worker<Ctx> {
                             invocation: WorkerInvocation::ExportedFunction {
                                 idempotency_key: IdempotencyKey::fresh(),
                                 full_function_name: "golem:agent/guest.{initialize}".to_string(),
-                                function_input: vec![agent_id.parameters.clone().into_value()],
+                                function_input: vec![
+                                    agent_id.agent_type.clone().into_value(),
+                                    agent_id.parameters.clone().into_value(),
+                                ],
                                 invocation_context: invocation_context.clone(),
                             },
                         });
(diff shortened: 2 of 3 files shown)


vigoo:
There are some follow-up bugs to be fixed before merge

PR #2020, opened 2025-09-02
Small CLI improvements

vigoo:
- Hiding the Rust templates by default (only show with `--dev-mode`)
- Showing an initial version of the "rendered agent exports" instead of the WIT interface:

```
╔═
║ Started Rib REPL for component my:agent using version 0
║
║ Component name:    my:agent
║ Component ID:      c2ab5b41-48f4-4c25-956f-177dd967d15b
║ Component type:    Durable
║ Component version: 0
║ Project ID:        77b4400d-2e7a-468e-9a57-c14be53d30eb
║ Component size:    2.82 MiB
║ Created at:        2025-09-02 08:29:49.321714 UTC
║ Exports:
║   assistant-agent() agent constructor
║   assistant-agent.ask(name: string) -> return-value: string
║   weather-agent(username: string) agent constructor
║   weather-agent.getWeather(name: string, param2: record { data: string, value: s32 }) -> return-value: string
╚═
```

### DIFF
--- a/cli/golem-cli/src/command.rs
+++ b/cli/golem-cli/src/command.rs
@@ -134,6 +134,10 @@ pub struct GolemCliGlobalFlags {
     #[arg(long, global = true, display_order = 110)]
     pub show_sensitive: bool,
 
+    /// Enable experimental, development-only features
+    #[arg(long, short, global = true, display_order = 111)]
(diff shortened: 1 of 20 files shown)


afsalthaj:
Should it be "Exports" or just "Agents"

PR #2032, opened 2025-09-04
Fix template example

afsalthaj:
The interface is not really the type that should showcase "object". It should be just real object type in typescript. Maxim pointed this out when testing RC1


cli/golem-templates/templates/ts/ts-app-component-default/components-ts/component-name/src/main.ts
```diff
@@ -5,7 +5,7 @@ import {
     description,
 } from '@golemcloud/golem-ts-sdk';
 
-interface CustomData {
+type CustomData = {
```
vigoo, line 8: Is the SDK capturing properly the name of this too? (Will the AnalysedType have a name in the agent metadata)

vigoo (approved):

cli/golem-templates/templates/ts/ts-app-component-default/components-ts/component-name/src/main.ts
afsalthaj: I have to set the name in AnalysedType. It's going to be only in upcoming releases of SDK.

And yes, the SDK can see the alias name too, but someone can inline that object type without any name as well.


`function get-weather(myData: {data: string, value: string})`

So should it be `myData` that should be the name of the type or the type alias name (if an alias exist)?

afsalthaj: I guess no names if its inlined

vigoo: I don't know, hate these anonymous types :) I think there is a danger of using the same names for different things, which will most likely fail in the wrapper generation as we are generating separate WIT types for all these.

vigoo: If you are not giving a name in the metadata, it gets an automatic name (`elementX` currently with a counter X)

afsalthaj: > I don't know, hate these anonymous types :)

I undertand, but TS :)

PR #2047, opened 2025-09-06
Update SDK and fix overflow error related to plan limit

afsalthaj:
Fixing an overflow error.

```
... usual server logs &


agent()", idempotency_key: "623a8cbb-bc55-41a6-8452-ab46375a9409", function: "my:agent/assistant-agent.{ask}"
thread 'tokio-runtime-worker' panicked at cloud-service/src/service/plan_limit.rs:218:30:
attempt to subtract with overflow
fish: Job 1, '/Users/afsalthaj/projects/ribbb…' terminated by signal SIGABRT (Abort)
```

Also, I am not sure why it popped up all of a sudden

Also I was testing with  the update SDK - so I might as well have it in the same PR


afsalthaj:
@vigoo A new release of SDK is on it's way. I would make it ready-for-review then

mschuwalow (approved):

cloud-service/src/service/plan_limit.rs
```diff
@@ -215,7 +215,12 @@ impl PlanLimitService for PlanLimitServiceDefault {
     ) -> Result<ResourceLimits, PlanLimitError> {
         let plan = self.get_plan(account_id).await?;
         let fuel = self.account_fuel_repo.get(account_id).await?;
-        let available_fuel = plan.plan_data.monthly_gas_limit - fuel;
+        let available_fuel = plan
+            .plan_data
+            .monthly_gas_limit
+            .checked_sub(fuel)
```
noise64, line 221: i first thought that this should be a saturating sub, but given this is an i64, it should be manually checked for not going under 0? with this way it can go into -MAX and only then to 0

PR #2048, opened 2025-09-08
Try to increase "open file limit" on "golem server" start

noise64:
On mac with defaults (which is 256 soft limit):
```
golem server run --clean -vvv
  2025-09-08T09:16:21.692653Z DEBUG golem_cli::context: Loaded profiles, profile_name: local, manifest_profile: None
    at cli/golem-cli/src/context.rs:141

  2025-09-08T09:16:21.692681Z DEBUG golem_cli::log: set log output, output: Stdout
    at cli/golem-cli/src/log.rs:134

Selected profile: local
  2025-09-08T09:16:21.695058Z DEBUG golem::command_handler: File limit increase result: Ok(138240)
    at cli/golem/src/command_handler.rs:37
    ```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4079,6 +4079,7 @@ dependencies = [
  "prometheus 0.13.4",
  "regex",
  "reqwest 0.12.23",
+ "rlimit",
  "rustls 0.23.31",
  "serde",
  "tempfile",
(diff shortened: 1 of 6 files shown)


mschuwalow:
This should be calling this under the hood https://www.man7.org/linux/man-pages/man2/setrlimit.2.html.

This will only modify the current process if i'm understanding correctly, right?

noise64:
@mschuwalow, yes it is only for the current process, so we just try to bump the soft limit, and the hard limits and os specific things should be handled by the library.

afsalthaj (approved):

PR #2053, opened 2025-09-08
Add a websearch agent example/template

mschuwalow:


cli/golem-templates/templates/ts/ts-app-component-websearch-summary-example/components-ts/component-name/src/main.ts
```diff
@@ -12,7 +12,7 @@ import {
   send as sendToLLM
 } from 'golem:llm/llm@1.0.0'
 import { startSearch } from 'golem:web-search/web-search@1.0.0';
-import { env } from 'node:process';
+import { env } from 'process';
```
vigoo: This is definitely not good. It's called `node:process` and there is even a test for it in wasm-rquickjs

vigoo: https://github.com/golemcloud/wasm-rquickjs/blob/main/tests/runtime.rs#L928

mschuwalow: Yes, will revert this one. Was just testing some stuff while debugging the import issue :+1:

PR #2054, opened 2025-09-08
Make sure  identifiers of string literals are handled properly in rib agent construction

afsalthaj:
But I have tried to test the behaviour with extending an existing unit test case and assert on the worker name
There can be couple of other things to do in this space.


vigoo (approved):

afsalthaj:
There was another regression, just fixed.
Will merge once things are green

PR #2062, opened 2025-09-09
Better stderr logs in case of errors

mschuwalow:


mschuwalow:
There is an existing test (stderr_returned_for_failed_component) which should cover this case, but it obviously doesn't.

Trying to figure out where exactly it's going wrong. Will update the pr / open a separate on for it.

vigoo:
I think the original implementation intentionally only collected output from the last try to avoid showing the retried error multiple times. 
But if we think this is more useful I'm ok with it

mschuwalow:
This should still only show each log message once.

For the previous invocation case it looks like we are only writing them to the oplog once, so that will only emit each message once across all retry attempts.

For the current invocation case, the worker_events include the is_live parameter, which is false for deduplicated error messages (by the hash), so there we can avoid emitting logs multiple times if we only consider live ones.

vigoo (approved):

PR #2078, opened 2025-09-11
Fix desktop-app after agent changes

mschuwalow:


cli/golem-cli/src/command_handler/component/mod.rs
```diff
@@ -387,11 +387,7 @@ impl ComponentCommandHandler {
                 .await?;
         }
 
-        if component_views.is_empty() {
```
noise64, line 390: i think it should have been `component_name.is_some() && component_views.is_empty()`, in those cases it should fail (or at least that is how it is done (hopefully) in other places)

cli/golem-cli/src/command_handler/api/deployment.rs
```diff
@@ -160,29 +160,22 @@ impl ApiDeploymentCommandHandler {
...
+                    .selected_project_id_or_default(project.as_ref())
+                    .await?
+                    .0,
+                id,
+            )
+            .await
+            .map_service_error()?
+            .into_iter()
+            .map(ApiDeployment::from)
+            .collect::<Vec<_>>();
```
noise64, line 178: finally :D

cli/desktop-app/src/pages/workers/create/index.tsx
```diff
@@ -41,6 +41,22 @@ const formSchema = z.object({
 export default function CreateWorker() {
   const navigate = useNavigate();
   const { componentId, appId } = useParams();
+
+  const [agentConstructorDetails, setAgentConstructorDetails] = useState<
+    string[]
+  >([]);
+  useEffect(() => {
+    API.componentService.getComponentById(appId!, componentId!).then(res => {
+      const agentSuffix = " agent constructor";
```
andrueandersoncs, line 50: surely* this ad-hoc parser isn't the correct way to list constructors..?

mschuwalow: Nah, It's not. 

@noise64 will hopefully adust the cli output to make this better / expose agentTypes.
But for now this should be good enough.

cli/golem-cli/src/command_handler/component/mod.rs
mschuwalow: I believe this is handled in the lines above the diff

```
        if component_views.is_empty() && component_name.is_some() {
            // Retry selection (this time with not allowing "not founds")
            // so we get error messages for app component names.
            self.ctx
                .app_handler()
                .opt_select_components(
                    component_name.iter().cloned().collect(),
                    &ApplicationComponentSelectMode::CurrentDir,
                )
                .await?;
        }

```

mschuwalow: e.g.

```
❯ ../../golemcloud/golem/target/debug/golem component list foo
Selected profile: local
warn: No versions found for component foo

error: The following requested component names were not found:
  - foo

Available application components:
  - golem-qa:websearch
  ```

Issue #2096, opened 2025-09-16
Moonbit generation fails when there is a type that is union of only literal strings


afsalthaj:
* The extraction of types succeeds however, moon-bit side fails for `type UnionWithOnlyLiterals = "foo" | "bar" | "baz"`.
* This particular bug was discovered when I developed an agent that has all of the types that the SDK supports. In fact, all other types worked with moonbit generation.

Here is the agent code

```ts
// main.ts

import {BaseAgent, agent} from '@golemcloud/golem-ts-sdk';

import {UnionWithOnlyLiterals} from './model';


@agent()
class FooAgent extends BaseAgent {
    constructor(readonly input: string) {
        super();
        this.input = input;
    }
    
    async myFun(param: UnionWithOnlyLiterals): Promise<UnionWithOnlyLiterals> {
        return param;
    }
}
```

```ts
// model.ts


// The type that doesn't work
export type UnionWithOnlyLiterals = "foo" | "bar" | "baz";

// A type that's working - as an example that proves other types in this file is working
export type TaggedUnion = {tag: 'a', val: string} | {tag: 'b' , val: string};

```

The log for `golem app build` says this (see the last line)

```rust

✔ Metadata saved successfully in .metadata/generated-types.ts!
    Executing external command 'npx --no rollup -- -c' in directory /Users/afsalthaj/projects/my-agent-app/components-ts/my-agent

virtual:agent-main → ../../golem-temp/ts-dist/my-agent/main.js...
created ../../golem-temp/ts-dist/my-agent/main.js in 299ms
    Injecting JS module /Users/afsalthaj/projects/my-agent-app/components-ts/my-agent/../../golem-temp/ts-dist/my-agent/main.js into QuickJS WASM /Users/afsalthaj/
         projects/my-agent-app/components-ts/my-agent/../../wasm/golem_agent.wasm
    Generating agent wrapper for my:agent
      Extracting agent types from /Users/afsalthaj/projects/my-agent-app/components-ts/my-agent/../../golem-temp/agents/my_agent.dynamic.wasm
      Generating agent WIT interface for my:agent
      Generating agent WIT interface implementation to /Users/afsalthaj/projects/my-agent-app/components-ts/my-agent/../../golem-temp/agents/my_agent.wrapper.wasm
[...]

PR #2100, opened 2025-09-16
Add ts code first agent integration tests and fix issues

### LINKED ISSUE #1824: Copy of Rust SDK: integration test suite
We need integration tests for the macros.

afsalthaj:
Fixes #1824 

Migration of manual testing scenarios automated in golem. Serves as more of a regression test.
Every invocation goes through a Rpc call. Function returns the same input (at any point) mostly.

We can expand these tests (with arbitrary input - requires more time/thoughts on it) and more scenarios in near future. 

## Issues found and fixed through this integration test (excluding a few miscellaneous/minor ones)

1. https://github.com/golemcloud/golem/issues/2096 (already resolved by @vigoo in another PR - but not yet working, this time from SDK)

2. Union with literals not working: https://github.com/golemcloud/golem-agentic/issues/131

3. Result type not working at golem execution stage: https://github.com/golemcloud/golem-agentic/issues/134

4. Variants with boolean not working: https://github.com/golemcloud/golem-agentic/issues/138

5.  Rib Compilation error - randomly occurring (very confusing): Getting resolved by https://github.com/golemcloud/golem-agentic/pull/104

```
>>> x.fun-union-complex-type(case1("foo"))
[compilation error]
[position] 0
[expression]
[cause] cannot determine the type
[help] conflicting types: record { a: string, b: s32, c: bool }, bool
```

6. Typescript SDK: Decimal values are truncated to integers. Resolved in https://github.com/golemcloud/golem-agentic/pull/105

7. Worker service run time errors (looks like very random)

Doesn't happen after the above fixes.
```
>>>  let x = bar-agent(...)
>>> x.fun-string("foo")
[runtime error] failed to invoke function my:agent/bar-agent.{fun-string}: Worker Service - Error: 500 Internal Server Error, Invalid request: in $.tuple.[1].component-model.nodes.[0].variant-value.1 expected option, got s32
>>>
```

8. Singleton worker failure and Rib REPL experience
Resolved. See [this]( https://github.com/golemcloud/golem/pull/2100#issuecomment-3298962062) (my comment is quoted there)

 
9.  Key word conflicts such as `ok` , `err` and Rib struggling to distinguish (fixed)

Invoke works (unlike REPL) with `%` escape, which is not a valid syntax in Rib. Currently Rib is fixed to distinguish without escape pattern

>Several tokens are reserved WAVE keywords: true, false, inf, nan, some, none, ok, err. Variant or enum cases that match one of these keywords must be prefixed with %.

Note tha `%ok("foo")` doesn't compile in Rib REPL though  - so that needs to be fixed

More integration tests on multimodal-&-unstructured to be done after fixing https://github.com/golemcloud/golem-agentic/issues/128

### DIFF
--- a/cli/golem-cli/tests/app/mod.rs
+++ b/cli/golem-cli/tests/app/mod.rs
@@ -1014,12 +1014,13 @@ async fn test_ts_code_first_complex() {
     .await;
 
     // Union that has only literals
-    run_and_assert(
-        &ctx,
-        "fun-union-with-only-literals",
-        &["foo", "bar", "baz"],
-    )
-        .await;
+    run_and_assert(&ctx, "fun-union-with-only-literals", &["foo"]).await;
+
+    // Union that has only literals
+    run_and_assert(&ctx, "fun-union-with-only-literals", &["bar"]).await;
+
+    // Union that has only literals
+    run_and_assert(&ctx, "fun-union-with-only-literals", &["baz"]).await;
 
     // A number type
     run_and_assert(&ctx, "fun-number", &["42"]).await;


vigoo:
> Let's say bar-agent has zero constructors. This is more or less a singleton worker as we call it.
Now for some reason we gave some wrong values and it resulted in a failure (example: type failure from SDK). One of the solutions is to delete the worker (meaning using cli to delete the worker outside the REPL) or make sure there is some random id constructor parameter (which may not be something we always do).

> Just repeating, this comment is mainly about REPL experience (and not executor behaviour) - that when people have zero parameter constructor agents and accidentally making mistake, it's a hard rewind altogether. Not to mention, even I am just "trying" to recreate the worker using let x = bar-agent() forgetting the fact that, that's not the case. Is this an acceptable limitation for 1.3? If yes, then all good. If no, do anyone has any idea how to make this experience good from "REPL" perspective.

I don't feel like this is a limitation. First of all you probably very rarely want to create a parameterless agent. Anyway if you do, just have to understand that referencing the agent in rib is not "creating an agent" but "referencing an agent" so even if you write `let x = agent()` multiple times you always just refer to the same agent. 
This is exactly what we had before when you referenced to the same worker by name (`let x = instance("agent"); let y = instance("agent");`).

If an agent fails you can delete (`golem agent delete`) or you can revert the last invocation(s) (`golem agent revert`). Having access to these commands from the repl itself may be a nice feature but I don't feel it important at all.

afsalthaj:
@vigoo 

https://github.com/golemcloud/golem/pull/2100#issuecomment-3298962062

Sounds reasonable. Thanks.

vigoo:
I think this error:
```
 [cause]  expected variant { ok(string), err(option<string>) }
```
is quite nice and understandable, would not change it. it talks about rib, and in the rib docs we talk about variants and this is the syntax we use for showing exported agent APIs too

afsalthaj:
https://github.com/golemcloud/golem-agentic/issues

Trying to merge this integration test once the bugs in the above issue are fixed.

PR #2104, opened 2025-09-16
Fix oplog archiving behaviour under high load

### LINKED ISSUE #2089: Single executable with multilayer oplog can lead to panics
Using this reproduction https://github.com/mschuwalow/golem-oom-repro, the single executable consistently fail during replays because of ooms.

```
thread 'tokio-runtime-worker' panicked at golem-worker-executor/src/durable_host/replay_state.rs:321:60:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: core::option::unwrap_failed
   4: golem_worker_executor::durable_host::DurableWorkerCtx<Ctx>::create::{{closure}}
   5: <golem_worker_executor::workerctx::default::Context as golem_worker_executor::workerctx::WorkerCtx>::create::{{closure}}
   6: golem_worker_executor::worker::invocation_loop::InvocationLoop<Ctx>::run::{{closure}}
   7: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
```

Switching to a single layer oplog fixes it.

mschuwalow:
resolves #2089

* ensure that the oplog can still be read even if entries in the cache are getting evicted in the background
* limit size of oplog chunks (mostly relevant when a lot of load is present in the worker)

NOTE: There are still a bunch of panics in the oplog code. I didn't want to touch is as part of this, but it should be refactored to properly return results

### DIFF
--- a/golem-worker-executor/src/services/oplog/multilayer.rs
+++ b/golem-worker-executor/src/services/oplog/multilayer.rs
@@ -370,7 +370,7 @@ impl OplogService for MultiLayerOplogService {
 
         if !full_match {
             for layer in &self.lower {
-                let partial_result = layer.read(owned_worker_id, idx, remaining as u64).await;
+                let partial_result = layer.read(owned_worker_id, idx, remaining).await;
                 let full_match = match partial_result.first_key_value() {
                     None => false,
                     Some((first_idx, _)) => {


vigoo:
Those panics were intentional but of course we can review this decision.
The way I distinguished between panics and returning results:
- if it is something the user should know about or can do about, it is a result
- if it is something fatal the user has nothing to do with - like not being able to read or write the oplog - then it is a panic, with the intention that it kills the "bad executor" (that let's say lost access to redis somehow) as soon as possible, and because it runs durable workers, everything automatically continues running on other nodes (or a restarted, working one) where hopefully there is no such error

While I know this above scenario is very simplistic I also think that just returning these errors to user land (errors while writing to the oplog for example would end up as failing host functions, failing the worker itself!) is not good either

golem-worker-executor/src/services/oplog/tests.rs
```diff
@@ -730,37 +730,31 @@ async fn entries_with_large_payload(_tracing: &Tracing) {
 }
 
 #[test]
-#[flaky(10)]
```
vigoo, line 733: :tada:

PR #2111, opened 2025-09-18
Ability to return an agent-error in discover-agent-types

vigoo:
This makes it easier to show a nice error (as it does not goes through a js exception -> rust panic -> captured stderr flow)


vigoo:
The integration tests are failing because of breaking the golem:agent/guest API. But to have the new agent SDK I have to merge this.

PR #2114, opened 2025-09-18
Allow http apis for agents: pass constructor params from worker service to rib compiler.

afsalthaj:


golem-worker-service/src/gateway_rib_compiler/mod.rs
```diff
@@ -115,3 +119,40 @@ impl WorkerServiceRibCompiler for DefaultWorkerServiceRibCompiler {
         compiler.compile(rib.clone())
     }
 }
+
+fn get_constructor_args(input_schema: &DataSchema) -> Vec<AnalysedType> {
```
mschuwalow, line 123: This is duplicated with the code in the rip repl from what I can tell

afsalthaj: Yes. Would you mind fixing it? 
Or as I understand, you are already into this in one of your branch. So if you don't mind merging it to this PR (or other way around), that will be great.

The main thing I care is a few more tests ... What happens if I don't pass enough arguments to constructor? What happens if not enough parameters to method?

If you are continuing to work on this, I request you to add those tests too :)

afsalthaj: Ok I see you approved. So I will merge this, and you could do the things mentioned above as a separate PR.

PR #2115, opened 2025-09-18
worker-service: add a few more test cases for agents in rib

mschuwalow:


golem-worker-service/tests/api_gateway_end_to_end_tests.rs
```diff
@@ -154,6 +155,69 @@ async fn test_api_def_with_agent() {
...
+        get_api_def_with_worker_binding("/foo?{userid}", response_mapping).await;
+
+    let result = CompiledHttpApiDefinition::from_http_api_definition(
+        &api_specification,
+        &internal::get_agent_component_metadata(),
+        &test_namespace(),
+        &(Box::new(TestConversionContext) as Box<dyn ConversionContext>),
+    );
+
+    assert2::assert!(let Err(RouteCompilationErrors::RibError(RibCompilationError::RibTypeError(_))) = result);
```
afsalthaj: How about not passing any parameter or more than 1 parameter to get-weather? I know it must be tested already within in golem-rib, but just to be sure. It will cover construction of agents as well as calling agent methods

PR #2119, opened 2025-09-19
worker-service: integration test for agentic components

mschuwalow:
Small followup to yesterday's changes


golem-test-framework/src/dsl/mod.rs
```diff
@@ -187,6 +189,11 @@ impl<'a, DSL: TestDsl> StoreComponentBuilder<'a, DSL> {
         self
     }
 
+    pub fn agentic(mut self) -> Self {
```
vigoo: This does not have to be user controlled. There is a component metadata extraction already (except in a few tests cases that are intentionally trying to load corrupted wasms), that could be used to automatically decide this (and it's going to be always true in 1.4). So basically decide it by checking if golem:agent/guest is in the component exports.

mschuwalow: This is used in the implementation already. The reason I decided it this way for now is that the vast majority of components is not agents currently, so the work to check / analyze the component is wasted.

I would propose to make it the default and remove it when then branching is no longer needed.

Note/Question regarding the metadata extraction: Currently the metadata is extracted server-side while the agent types are extracted client side (and we trust the client). Should we move the agent type extraction into the server side as well?

mschuwalow: Actually nvm, will remove the explicit agentic() configuration.

vigoo: Let's keep the extraction on the client side for now; it's quite heavyweight, having it in the server would completely change the server's scaling requirements etc. (because it requires to create and run wasmtime instances). Which is maybe what we want to do in the future, but I don't want to do this switch now

golem-common/src/model/agent/extraction.rs
```diff
@@ -296,3 +298,20 @@ struct FunctionInfo {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use assert2::assert;
+    use std::path::PathBuf;
+    use std::str::FromStr;
+    use test_r::test;
+
+    #[test]
+    async fn can_extract_agent_types_from_component_with_dynamic_rpc() -> anyhow::Result<()> {
```
mschuwalow, line 338: This was failing before because of wasi-io-pollable

PR #2144, opened 2025-09-27
Add human in the loop template

mschuwalow:


integration-tests/tests/worker.rs
```diff
@@ -1693,3 +1693,78 @@ async fn resolve_components_from_name(deps: &EnvBasedTestDependencies, _tracing:
             ])
     );
 }
+
+#[test]
+#[tracing::instrument]
+#[timeout(120_000)]
+async fn promise(deps: &EnvBasedTestDependencies, _tracing: &Tracing) {
```
mschuwalow: Reason to move this test here is that the promise completion is going through the worker proxy, which is currently not supported in the worker executor test suite.

Optimally it would not go through the proxy if the worker is on the current instance, which would make the tests work there. But as this problem exists in other places in the code as well (e.g. worker fork), I would prefer to do it like this for now and make the proxy aware of workers on the current instance in a later change.

golem-worker-executor/src/services/promise.rs
```diff
@@ -48,32 +44,20 @@ pub trait PromiseService: Send + Sync {
     async fn delete(&self, promise_id: PromiseId);
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct DefaultPromiseService {
     key_value_storage: Arc<dyn KeyValueStorage + Send + Sync>,
-    promises: Arc<DashMap<PromiseId, PromiseState>>,
```
mschuwalow, line 54: All this logic seems to be leftovers from a previous implementation and was unused

golem-worker-executor/src/durable_host/golem/v1x.rs
```diff
@@ -133,6 +133,15 @@ impl<Ctx: WorkerCtx> Host for DurableWorkerCtx<Ctx> {
         &mut self,
         promise_id: golem_api_1_x::host::PromiseId,
     ) -> anyhow::Result<Vec<u8>> {
+        // only the agent that originally created the promise is woken up when it is completed.
+        if golem_common::base_model::WorkerId::from(promise_id.worker_id.clone())
+            != *self.worker_id()
+        {
+            return Err(anyhow!(
+                "Tried awaiting a promise not created by the current agent"
```
mschuwalow, line 145: This was already effectively unsupported before, just added an error so that we don't end up in weird states.

PR #2148, opened 2025-09-28
Fix websearch-summary-example template

vigoo:
- Fix the HTTP API name
- Moved the API def to top to be aligned with the other two templates

### DIFF
--- a/cli/golem-templates/templates/ts/ts-app-component-websearch-summary-example/components-ts/component-name/golem.yaml
+++ b/cli/golem-templates/templates/ts/ts-app-component-websearch-summary-example/components-ts/component-name/golem.yaml
@@ -1,5 +1,27 @@
 # golem-app-manifest-header
 
+httpApi:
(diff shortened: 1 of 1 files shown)


cli/golem-templates/templates/ts/ts-app-component-websearch-summary-example/components-ts/component-name/golem.yaml
```diff
@@ -1,5 +1,27 @@
...
+              let agent = research-agent();
+              let result = agent.research(request.query.topic);
+              {status: 200, body: result, headers: {ContentType: "text/plain"}}
+
+  ## Uncomment this to enable deployment of the http api
+  # deployments:
+  #   local:
+  #   - host: localhost:9006
+  #     definitions:
+  #     - componentname-api
```
mschuwalow:
```suggestion
  #     - component-name-api
```

Issue #2152, opened 2025-09-29
Fix agent ID parsing


noise64:
- use kebab-casing
- if it is easy, let's also add a "WAVE" normalize option, so whitespaces are removed

PR #2155, opened 2025-09-29
Reworked native and WIT cased naming for agent types and AgentID normalization

noise64:
Resolves: #https://github.com/golemcloud/golem/issues/2152

- added "wrapped type name" accessors to AgentType and AgentId
- agent metadata lookups can be now done by wrapped or original agent type names
- added ToWitNaming trait to convert AgentTypes or parts of to WIT naming convention
- use the above for:
 - constructor wit_args
 - CLI help
 - AgentID parsing
- switched to std::sync::Mutex for the metadata caches, given currently calculating the values are blocking (and fast) operations
- an unrealted small fix: git version detection now only consider tags starting with "v", this was needed as we now also have tags for golem-rust releases, and that can create bad version info for the server and CLI


golem-rib/src/type_inference/custom_instance_spec.rs
```diff
@@ -33,13 +34,22 @@ impl CustomInstanceSpec {
     //                    and has to be part of the ComponentDependencies
     pub fn new(
         instance_name: String,
-        parameter_types: Vec<AnalysedType>,
+        parameter_types_for_rib: Vec<AnalysedType>,
```
afsalthaj: Let's discuss on this. For any instance creation, theoretically it's just 1 set of parameter types.

noise64: as discussed, reverted

golem-common/src/model/agent/mod.rs
```diff
@@ -63,7 +64,7 @@ pub struct AgentConstructor {
 }
 
 impl AgentConstructor {
-    pub fn wit_arg_types(&self) -> Vec<AnalysedType> {
+    pub fn arg_types(&self) -> Vec<AnalysedType> {
```
vigoo: I disagree with this rename. The returned argument types are specific to the WIT wrapper - they represent how the wrapper encodes `DataSchema` as WIT types, such as using list of variants for multimodal, etc.
The "non-wit" argument types is the `input_schema` itself.

golem-common/src/model/agent/mod.rs
```diff
@@ -747,31 +743,24 @@ impl AgentId {
...
     }
 }
 
 #[async_trait]
 pub trait AgentTypeResolver {
-    async fn resolve_agent_type(&self, agent_type: &str) -> Result<AgentType, String>;
-}
-
-#[async_trait]
-impl<F> AgentTypeResolver for F
```
vigoo, line 760: Why did this implementation hurt you? :)

noise64: it was not used?

golem-common/src/model/agent/mod.rs
noise64: they are NOT having wit compatible field names, and i started using for this difference in apis wit_ vs native_, as this is "kinda wit"

golem-common/src/model/agent/mod.rs
noise64: on a more serious nore, i wanted to understand where are instances i might have to adapt, when i explored the code, so i removed to check if something went unnoticed, and given it was not used, i did not put it back, as i even forgot about it

noise64: or maybe i missed something (you are using it for debugging?)? then i will add it back

golem-common/src/model/agent/mod.rs
```diff
@@ -747,31 +743,24 @@ impl AgentId {
...
-where
-    F: for<'a> Fn(&'a str) -> Pin<Box<dyn Future<Output = Result<AgentType, String>> + 'a + Send>>
-        + Send
-        + Sync,
-{
-    async fn resolve_agent_type(&self, agent_type: &str) -> Result<AgentType, String> {
-        self(agent_type).await
-    }
+    fn resolve_wit_agent_type(&self, agent_type: &str) -> Result<AgentType, String>;
```
vigoo: This does not feel right to me. The trait has nothing that would be specific to the WIT wrapper in any way, although it's implementation on `ComponentMetadata` is. But we will need the same thing for the "real" agent type name very soon, even if at the moment it is not required anywhere and then duplicating this trait feels wrong. Adding a second method also feels wrong, but less wrong :) 
Maybe there could be an `AgentTypeName` sum type with an `Original` and `Wrapper` case and accep that here as parameter?

golem-common/src/model/agent/mod.rs
noise64: so for me WIT means (at least i think _should_ mean), that it is actually a WIT compatible types, with WIT names

golem-common/src/model/agent/wit_naming.rs
```diff
@@ -0,0 +1,244 @@
...
+    DataSchema, ElementSchema, NamedElementSchema, NamedElementSchemas,
+};
+use golem_wasm_ast::analysis::{
+    AnalysedType, NameOptionTypePair, NameTypePair, TypeEnum, TypeFlags, TypeHandle, TypeList,
+    TypeOption, TypeRecord, TypeResult, TypeTuple, TypeVariant,
+};
+
+/// ToWitNaming allows converting discovered AgentTypes to WIT and WAVE compatible naming for named
+/// elements
+pub trait ToWitNaming {
```
vigoo, line 26: Can we use this in the actual wrapper generator too?

noise64: agree, we should

golem-common/src/model/agent/mod.rs
noise64: > This does not feel right to me

The whole mixed "kinda WIT but not really WIT" does not feel right for me, and was used _wrong_ and randomly in the CLI and REPL and we kept fixing it or thinking that it works :)

So for now i'm caring more about that it works, and not about how it is expressed, but I'm open to any naming suggestions, that clearly helps making it clear the we are using the original or the "WIT and WAVE compatible" naming.

As for my current naming: i prefixed with _WIT_  for WIT compatible or with _native_ (original) naming.

TBH this whole resolver trait is a bit puzzling for me, as (currently) only used for agent id parsing, and apart from tests the only impl is the metadata (which could be stubbed easily); the naming and scope of it hinted something much more important to me, but i did not find any other usecase, everybody else goes directly to the metadata. And for this purpose, the only correct spelling for now is the WIT. So i'm really not sure why this trait should exists, and why we spend time arguing on naming it, but probably you have plans for it.

golem-common/src/model/agent/wit_naming.rs
noise64: i will add this

golem-common/src/model/agent/mod.rs
vigoo: Then this should be called `wit_arg_types` AND it should work on the "wit-ified" types. Because otherwise it does not make any sense. This representation of `DataSchema` as wit types only exists in the WIT wrapper.

golem-common/src/model/agent/mod.rs
vigoo: Maybe in the end it was not used, I definitely used it at some point in some tests. The idea was to be able to test this thing without having to produce a full component metadata.

golem-common/src/model/agent/mod.rs
noise64: That sounds good, will change to that

noise64: (It was not obvious for me until now, as the rib repl used these in the original casing, which is not true anymore)

golem-common/src/model/agent/mod.rs
noise64: there is a custom test type registry used for that currently

golem-common/src/model/agent/mod.rs
noise64: as for the future, will both agent id naming exists together? if not, then we can just drop here the wit prefix (or decide later, how this should be handled), but i see no harm in being explicit here for now, especially if we will have mixed modes. (and this naming of the interface method does not break anything AFAIU for users, so this can be done anytime)

vigoo: There is a process (going through 1.3 and 1.4) of "moving" to the higher level agent invocation from the low level wit one, and part of that is that agent types got gradually added to the component metadata, and these things will evolve more. This allows me to have more freedom doing these steps, as all that's needed to parse an agent ID is to be able to get the `AgentType` once it parsed the agent type name. I don't want to tie it to that particular cached component metadata object that has a much bigger scope. So please don't remove it.

In general, don't get me wrong, I think this PR is definitely trying to solve the confusing duality of having the wrapper and the original things at the same time. I think my problem is that it represents the two with the same `AgentType` data structure which just moves the confusion to a different level for me, as it is still the agent structure (with methods and constructor using `DataSchema` as input and output, having possible data types that are not WIT types) but the names of things is already renamed to the ones in the wrapper. And distinguishing these two kind of `AgentType`s only by the name of functions that return them. 

The original idea is that `AgentType` describes the agent level interface which is in this pr now called the "native" agent type, and the original component exports describe the low level one, and these can be resoled to each other (for example `find_agent_method_or_constructor` finds the high level one from the low level name). 

With this now we have 3 different representations and the new one is somewhere between the two real ones, and I feel like it is going to lead to even more confusion in the future especially as we evolve the core further. 

I'm not sure what the solution is, but what I would try is first of all to distinguish the two "agent types" on the type level too, and then fix the ambiguity based on what each of them are used for. (I mean in something that describes the wrapper, there cannot be `DataSchema` only `AnalysedType`, for example)

golem-common/src/model/agent/mod.rs
noise64: updated

afsalthaj (approved):

PR #2158, opened 2025-09-30
Stream logs in the REPL

### LINKED ISSUE #2150: Stream logs from invoked agents in the repl
With this it would be much nicer to use the REPL to debug things, as you can add prints to the things you play with.

I think it should work by just spawning a background streaming connection to every agent the repl every invokes, with some inbox to receive idempotency keys to watch for (so we don't stream logs from invocations from outside that repl session). Then live dump all the incoming entries so invocations that never return for some reason can also be debugged with it.

vigoo:
image

Resolves #2150

### DIFF
--- a/cli/golem-cli/src/command.rs
+++ b/cli/golem-cli/src/command.rs
@@ -561,6 +561,9 @@ pub enum GolemCliSubcommand {
         /// Optional script_file to run, when defined the repl will execute the script and exit
         #[clap(long, conflicts_with_all = ["script"])]
         script_file: Option<PathBuf>,
+        /// If true (default), the REPL will stream logs from the invoked agents. Can be controlled with the :logs command in the REPL.
+        #[clap(long)]
+        stream: Option<bool>,
     },
     /// Generate shell completion
     Completion {
--- a/cli/golem-cli/src/command_handler/rib_repl.rs
+++ b/cli/golem-cli/src/command_handler/rib_repl.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::command::shared_args::DeployArgs;
+use crate::command::shared_args::{DeployArgs, StreamArgs};
 use crate::command_handler::Handlers;
 use crate::context::Context;
 use crate::error::NonSuccessfulExit;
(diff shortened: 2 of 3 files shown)


vigoo:
I'll hide the [INVOKE] and [STREAM] lines in a next commit

vigoo:
image

cli/golem-cli/src/command.rs
```diff
@@ -561,6 +561,9 @@ pub enum GolemCliSubcommand {
         /// Optional script_file to run, when defined the repl will execute the script and exit
         #[clap(long, conflicts_with_all = ["script"])]
         script_file: Option<PathBuf>,
+        /// If true (default), the REPL will stream logs from the invoked agents. Can be controlled with the :logs command in the REPL.
+        #[clap(long)]
+        stream: Option<bool>,
```
noise64: i think all the other "bool" flags now are using `bool` as type, making them behaving only as flags without values. could we keep this consistent and change this flag to something like
```
disable_streaming: bool
no_stream: bool
```
?

vigoo: sure

cli/golem-cli/src/command_handler/rib_repl.rs
```diff
@@ -37,17 +37,22 @@ use golem_wasm_rpc::json::OptionallyValueAndTypeJson;
...
 use rib::{ComponentDependency, ComponentDependencyKey};
 use std::path::{Path, PathBuf};
+use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 use uuid::Uuid;
 
 #[derive(Clone)]
 pub struct RibReplHandler {
     ctx: Arc<Context>,
+    stream_logs: Arc<AtomicBool>,
```
noise64, line 47: so far every state was stored inside the ctx (there is even a rib repl state there), and this was the "rule" because that way handlers can be constructed in other handlers easily using the context.
for the rib repl handler this does not matter that much, but would be more consistent to put it there

noise64: i can live with this exception, i just don't want to see this pattern in other places, where it actually matters
(as in, it can make sense to have handler instance vars, opposed to be fully global ones, but so far i avoided those)

noise64 (approved):
added some minor notes

cli/golem-cli/src/command_handler/rib_repl.rs
vigoo: It felt weird to me to add something that is the local state of the REPL to the global context. It just something that has to be shared between the `WorkerFunctionInvoke` implementation and the `Logs` `Command`, and I've put it on `RibReplHandler` because that's the one implementing `WorkerFunctionInvoke`.

Issue #2147, opened 2025-09-27
Async awaitPromise host function in TS


vigoo:
Use https://github.com/golemcloud/wasm-rquickjs/issues/62 to mark `awaitPromise` as async

vigoo:
To do this it has to be implemented with wasi pollables on the host

mschuwalow:
Can you explain the reason for this change a bit more?

Currently, our await implementation will suspend the agent immediately if the promise is not ready https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/durable_host/golem/v1x.rs#L132-L150, so I don't think a returned pollable would ever complete.

vigoo:
Ah that's a good point. The motivation is that I tried explaining how our `fork` API can be used together with golem promises to later join the forked agents back, and if you would be able to async await these golem promises, then you can implement more interesting patterns with these such as forking multiple agents and awaiting them all or the first one that succeeds etc. 

That said, the current behavior of suspending and resuming on completion is good for the simple case of for example implementing human-in-the-loop and making sure the agent that works for the slow human is not consuming resources unnecessarily. 

I think the best would be to make this API always a promise in JS, and detect when all the program is blocked on awaiting golem promises (only) and _then_ automatically suspend. I don't think we should try to do this for 1.3, but as a starting step, I would like to try to expose an alternative await pollable API that returns a `pollable` - the only thing that can be async awaited from the guest. I'd do this in a timeboxed way, if it turns out to be too hard, then leave it for a future release. Also we should keep the current await and maybe rename it to something like `awaitSleeping()`.

PR #2159, opened 2025-09-30
Turn promises into pollables & only suspend if all pollables are promises

mschuwalow:
part of #2147


golem-worker-executor/src/durable_host/io/poll.rs
```diff
@@ -56,6 +57,29 @@ impl<Ctx: WorkerCtx> HostPollable for DurableWorkerCtx<Ctx> {
 
 impl<Ctx: WorkerCtx> Host for DurableWorkerCtx<Ctx> {
     async fn poll(&mut self, in_: Vec<Resource<Pollable>>) -> anyhow::Result<Vec<u32>> {
+        // check if all pollables are promise backed. In this case we can suspend immediately
```
vigoo, line 60: Interesting. I expected this will have to be somehow supported in the underlying wasmtime-wasi poll in our fork. Maybe the "suspend for sleep" thing can also be implemented in a way that we can revert the fork to the its original state - but let's not deal with that now

golem-worker-executor/src/durable_host/golem/v1x.rs
```diff
@@ -859,6 +818,50 @@ impl<Ctx: WorkerCtx> HostGetOplog for DurableWorkerCtx<Ctx> {
...
+        resource: Resource<GetPromiseResultEntry>,
+    ) -> anyhow::Result<Resource<Pollable>> {
+        self.observe_function_call("golem::api::promise-result", "subscribe");
+        let handle = self.table().get(&resource).unwrap().handle.clone();
+        let dyn_pollable = subscribe(self.table(), resource, None)?;
+        self.state
+            .promise_backed_pollables
+            .write()
+            .await
+            .insert(dyn_pollable.rep(), handle);
```
vigoo, line 832: What is going to remove these?

vigoo (changes requested):
This looks very nice. I have one concern about the pollable handle table not being cleaned up (but maybe I just missed something)

And it would be nice to somehow explicitly test both cases:
- awaiting a pollable AND something else (can be an explicit subscribe_duration with wasi:clocks, a wasi-http (fetch in ts) or an rpc call  
- await a single promise and verify in the test that the worker is suspended (maybe this already exists)
- await multiple promises (and only promises) and verify the worker is suspended and can still wake up when all are completed (this was not possible before)

vigoo (approved):

PR #2207, opened 2025-10-14
Launch demo fixes

### LINKED ISSUE #2205: Public oplog mapping issue
Debugging #2204 with `golem agent oplog` fails with

```
error: Worker Service - Error: 500 Internal Server Error, Unknown error: Oplog entry #63 - golem_agent::get_agent_type input: Failed to deserialize value: UnexpectedVariant { type_name: "AnalysedType", allowed:
     Range { min: 0, max: 21 }, found: 111 }
```

Reproducer: same as #2204

mschuwalow:
resolves #2205 

Also allows invoking (but not invoking) the same worker in rpc.


vigoo:
I don't think this PR is necessary - the root cause was fixed in the TS SDK. I'll have a better (covering one more missing case) fix for the oplog error in another one

PR #2215, opened 2025-10-16
Allow white spaces in agent ids

### LINKED ISSUE #2211: Allow spaces in agent IDs
For example the following should be valid:

```
my-agent("hello world")
```

To be released as quickfix for 1.3.0

afsalthaj:
Fix #2211 

reworked (@noise64 here) after discussions:
- no validation on GRPC API
- validate and normalize on HTTP for agent workers
- keep only the slash and length restriction for legacy test workers
- ~increased max length to 2048~ remains 512, adding followup card for it

### DIFF
--- a/golem-worker-executor/src/model/mod.rs
+++ b/golem-worker-executor/src/model/mod.rs
@@ -71,19 +71,33 @@ pub struct WorkerConfig {
 
 impl WorkerConfig {
     pub fn new(
-        worker_id: WorkerId,
-        agent_id: &Option<AgentId>,
-        target_component_version: u64,
         worker_args: Vec<String>,
-        mut worker_env: Vec<(String, String)>,
+        worker_env: Vec<(String, String)>,
         deleted_regions: DeletedRegions,
         total_linear_memory_size: u64,
         component_version_for_replay: u64,
         created_by: AccountId,
         initial_wasi_config_vars: BTreeMap<String, String>,
     ) -> WorkerConfig {
+        WorkerConfig {
+            args: worker_args,
+            env: worker_env,
+            deleted_regions,
+            total_linear_memory_size,
+            component_version_for_replay,
+            created_by,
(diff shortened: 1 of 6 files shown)


noise64:
I think we have to add "normalization" on the server side too, to make this realiable, because otherwise now we allow WAVE parts that describe the same data to be used as differently formatted strings resulting in different agents that should have the same identity.

The CLI does this normalization, but we also have other sources now, especially console and desktop. So in general all APIs:
- either should only accepts normal forms
- or normalize on the fly

noise64:
So this is not just about allowing spaces, but we have to carefully touch the identity / format part, otherwise we just shift the bugs to other places, and it would worth to also test with multimodal too (including newlines and so on)

afsalthaj:
> The CLI does this normalization, but we also have other sources now, 

Looks like REPL too didn't go through the normalization.  

Let's wait for @vigoo to analyze this state. Normalization at cli side is fine, but having it at console / desktop and did we forget REPL?  - looks a bit questionable altogether.

Out of the two options from @noise64  , `either should only accepts normal forms` looks reliable.

vigoo:
The normalization should be in the REPL too: https://github.com/golemcloud/golem/blob/main/cli/golem-cli/src/command_handler/rib_repl.rs#L326-L332

The other frontends are agreed to be problematic in this regard. However I don't want to introduce one more "unnecessary step" in the hot path (I rather want to eliminate many of the existing ones :) - there is no need to do it for example if the invocation is happening through RPC etc.

So that said, maybe the logic should be:
- do the normilization like the one linked above in all the worker-service HTTP endpoints
- never do them on the gRPC endpoints (assuming that internally we call them correctly)

what do you think?

noise64:
@vigoo I'm happy with that solution, i will do it in this PR soon, as we agreed with @afsalthaj

vigoo (approved):

Issue #1210, opened 2024-12-24
First-Class Windows Development Environment


jdegoes:
In the Golem project currently, Windows is not a first-class development experience. The problems are many and range from:

1. How Golem is built and tested
2. Which Rust libraries are used, and to what degree they support Windows
3. Random scripts and tools that may not support Windows fully or at all

This ticket is to enhance Golem so that developers who are using Windows may have a first-class development experience, without having to install Windows Subsystem for Linux (WSL). 

When completed, a clean checkout of the Golem code base will:

1. Fully build
2. Reliably pass all tests
3. Require no more setup than development on Linux or Mac

Moreover, Github-based CI will automatically run tests on Windows to ensure no future regressions.

zelosleone:
Truth to be told the major problem is wasmtime package versioning making it hard to build-test, linking version is extremely hard to pass unless default features of wasmtime is completely disabled, which could make it unable to connect on HTTP via python. But there are also at least 4 years old crates like poule that is not updated recently, I am pretty sure some golem repo code could be changed to fit the newer/more maintained crates.

zelosleone:
> @zelosleone The issue of conflicting wasm appearing in both Windows and Linux after the merge of #1200. Before it worked fine without any conflicts of wasm versions. The problem stems from wasm-wave likely pulling older version of wasm as its dependency



Nope, its the wib-parser that pulls the old versioning, which breaks the wib module of the golem if upgraded. 0.6.0 is p old for wib-parser crate since it went more than 5 updates from that time, so golem code needs to change and be tested that "Yeah, this proves upgrading it didnt break anything after we updated the code and it works like old"

zelosleone:
> @zelosleone The wit parser conflict, I think not under consideration of this issue. It happening with other os linux too not sure what about Mac though.
> 
> I have clearly discussed with jdegoes and vigoo for the compatibility of windows and the alternatives and following them to solve keeping in mind of future work of Golem

Yeah I was working for 10 days on windows with golem so i had a headstart, thats fine, I almost finished the PR.

Aditya-Choudhry:
Hi, I would like to work on this issue. Could you assign it to me?

parvit:
/attempt #1210 

@jdegoes Do you require for the second point of eliminating docker/WSL requirement to provide also scripts for setup of services (eg. postgres) or should this change assume the services already installed on the test machine ? thank you.

laz-001:
[...]

PR #2222, opened 2025-10-16
Windows support dev

parvit:
Working PR for #1210, most of the required changes are done but some integration tests were not fixed ready.

Requires https://github.com/golemcloud/moonbit-component-generator/pull/7 to work.


vigoo:
I'm not sure what you did in this PR but it basically changes the whole repository:
image

So I'm closing it.

If you'd like to get in Windows related fixes feel free to open a new one *with only the relevant changes*, but note that the linked bounty is closed (it was never only about making Golem compile and tests run on Windows, but much more - and it was opened a long time ago)

parvit:
I'm unsure why it signals so much difference, but it is still a lot of changes nonetheless.
It was basically only to signal that i was nearly done to keep the ticket going and not intended as the final proposal.

I will prepare it again as all these months of work may not be lost.

PR #2243, opened 2025-10-23
Start moving rust sdk from prototype

### LINKED ISSUE #2182: Initial minimal code-first SDK into golem-rust
We want to develop the Rust SDK as part of `golem-rust` (living in the golem repo) from the beginning to make it easier to test and integrate.

The goal of this ticket is to extend the existing `golem-rust` and `golem-rust-macro` crates with a bare minimum to start defining and testing code-first agents. 

Part of this ticket should be to add a new rust agent template to `golem-templates`.

afsalthaj:
Initial setup for Rust SDK

Fixes #2182 

I have tried to avoid too many details except for agent implementation that will internally use registry..which is all TODOs..
and so are the macros. The `AgentArg` is not brought in either as well..just to reduce the discussions..

Draft PR until Vigoo reviews

### DIFF
--- a/sdks/rust/golem-rust/src/agentic/agent_registry.rs
+++ b/sdks/rust/golem-rust/src/agentic/agent_registry.rs
@@ -1,5 +1,3 @@
-use std::sync::Arc;
-
 use crate::{
     agentic::{agent_initiator::AgentInitiator, agent_type_name::AgentTypeName, ResolvedAgent},
     golem_agentic::exports::golem::agent::guest::AgentType,
@@ -13,9 +11,7 @@ pub fn get_all_agent_definitions() -> Vec<AgentType> {
 
 // The registry should hold the initiator instances for each agent type
 // TODO; Implement registration of initiators and retrieval
-pub fn get_agent_initiator(
-    agent_type_name: &AgentTypeName,
-) -> Option<Arc<dyn AgentInitiator + Send + Sync>> {
+pub fn get_agent_initiator(agent_type_name: &AgentTypeName) -> Option<Box<dyn AgentInitiator>> {
     todo!(
         "Unimplemented function to get agent initiator of type {}",
         agent_type_name.0
--- a/sdks/rust/golem-rust/src/agentic/agent_initiator.rs
+++ b/sdks/rust/golem-rust/src/agentic/agent_initiator.rs
@@ -4,6 +4,6 @@ use crate::{agentic::ResolvedAgent, golem_agentic::exports::golem::agent::guest:
  * An initiator for an agent implementation.
  * This is used to create an instance of the agent with the given parameters.
  */
-pub trait AgentInitiator: Send + Sync {
+pub trait AgentInitiator {
     fn initiate(&self, params: DataValue) -> ResolvedAgent;
 }


sdks/rust/golem-rust/src/agentic/agent.rs
```diff
@@ -0,0 +1,43 @@
...
+// ```
+//
+//  ```
+//  struct MyWeatherAgent;
+//
+//  #[agent_implementation]
+//  impl WeatherAgent for MyWeatherAgent {fn get_weather(&self, location: String) -> String } }
+//  ```
+// There is no need to implement `Agent` anywhere, as it is automatically implemented by the `[agent_implementation]` attribute.
+pub trait Agent: Send + Sync {
```
vigoo: There should not be a `Send` constraint as we are single threaded. (And only add `Sync` if it turns out to be really needed)

sdks/rust/golem-rust/src/agentic/agent_registry.rs
```diff
@@ -0,0 +1,25 @@
...
+// The registry should hold the initiator instances for each agent type
+// TODO; Implement registration of initiators and retrieval
+pub fn get_agent_initiator(
+    _agent_type_name: &AgentTypeName,
+) -> Option<Arc<dyn AgentInitiator + Send + Sync>> {
+    todo!("Unimplemented function to get agent initiator")
+}
+
+// At any point, there should be only one active agent instance
+pub fn get_agent_instance() -> Option<Arc<crate::agentic::ResolvedAgent>> {
```
vigoo: I don't think this needs any reference counting, it can be stored in a static `RefCell<Option<ResolvedAgent>>` and then this can return `&Option<ResolvedAgent>`

sdks/rust/golem-rust/src/agentic/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+
+mod agent;
+mod agent_export;
+mod agent_initiator;
+mod agent_registry;
+mod agent_type_name;
+
+#[derive(Clone)]
+pub struct ResolvedAgent {
+    pub agent: ::std::sync::Arc<dyn agent::Agent + Send + Sync>,
```
vigoo: - If we need ref counting, use `Rc` not `Arc`, as we don't have threads. 
- The `Agent` already has the `Send + Sync` constraints (although I commented there that it should not) so don't repeat it here
- I don't think we need to wrap in `Rc` at all, we just store a single instance in a global and use that

sdks/rust/golem-rust/src/agentic/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+
+mod agent;
+mod agent_export;
+mod agent_initiator;
+mod agent_registry;
+mod agent_type_name;
+
+#[derive(Clone)]
+pub struct ResolvedAgent {
```
vigoo, line 23: What is a `ResolvedAgent` going to add on top of an `Agent`? storing it's id?

sdks/rust/golem-rust/src/lib.rs
```diff
@@ -125,6 +125,24 @@ pub mod save_snapshot {
     pub use __export_golem_rust_save_snapshot_impl as export_save_snapshot;
 }
 
+#[cfg(feature = "export_golem_agentic")]
```
vigoo, line 128: :+1: let's put it behind a feature flag so `golem-rust` can be still used in libraries (like `golem-ai`) without problems

sdks/rust/golem-rust/src/agentic/agent_impl.rs
```diff
@@ -0,0 +1,90 @@
...
+pub struct Component;
+
+impl Guest for Component {
+    fn initialize(agent_type: String, input: DataValue) -> Result<(), AgentError> {
+        let agent_types = agent_registry::get_all_agent_definitions();
+
+        let agent_type = agent_types
+            .iter()
+            .find(|x| x.type_name == agent_type)
+            .expect(
```
vigoo, line 33: This should be an `AgentError` not panic

sdks/rust/golem-rust/src/agentic/agent_impl.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        let agent_type_name = AgentTypeName(agent_type.type_name.clone());
+
+        let agent_initiator = agent_registry::get_agent_initiator(&agent_type_name);
+
+        if let Some(agent) = agent_initiator {
+            agent.initiate(input);
+            Ok(())
+        } else {
+            panic!(
```
vigoo: This also can be an `AgentError`.

sdks/rust/golem-rust/Cargo.toml
```diff
@@ -1,31 +1,41 @@
...
-serde = { version = "1", optional = true }
-serde_json = { version = "1", optional = true }
-uuid = { version = "1", features = ["v4"] }
-golem-wasm-rpc = { version = "1.3.0-rc9pre", default-features = false, features = [
-    "stub",
-] }
-wit-bindgen = { version = "=0.40.0" }
+golem-rust-macro = {path = "../golem-rust-macro", version = "0.0.0", optional = true}
+
+# TODO; This needs to be updated to golem-wasm
```
vigoo, line 17: Have to start publishing new `dev` releases but I don't know what version to use :)

sdks/rust/golem-rust/src/agentic/mod.rs
afsalthaj: It will have a few more details like AgentId. Yes, in prototype it is "string" representation (which is wrong), so I didn't move it, but pretty sure we want it.

sdks/rust/golem-rust/src/agentic/agent_impl.rs
afsalthaj: Oh yes.

sdks/rust/golem-rust/src/agentic/agent_impl.rs
afsalthaj: Yes.

PR #2254, opened 2025-10-27
Implement agent definition and agent implementation

afsalthaj:
Fix #2183  and #2184 

It took some time for this [test](https://github.com/golemcloud/golem/blob/rust_agents_with_tests/cli/golem-cli/tests/app/agents.rs#L11C1-L41C2) that proves agent_impl works all the way from CLI.

And ensuring compilation [here](https://github.com/golemcloud/golem/blob/rust_agents_with_tests/sdks/rust/golem-rust/tests/agent.rs#L21) at sdk level

[Integration test ](https://github.com/golemcloud/golem/pull/2259)is on its way adding 1 more test to cli side.  

In a way this PR has also got some starting points to solving #1819 

If you just need to see macros, [here](https://github.com/golemcloud/golem/tree/rust_agents_with_tests/sdks/rust/golem-rust-macro/src/agentic)

The template is ready.

 `generic` template never used to really work or can run tests using it. David fixed it within this PR, along with configurable SDK versions for rust (path or version).

### Agent Template example:

https://github.com/user-attachments/assets/e684f1e1-eed7-4dfb-9f03-a7bc0de0d271

## TODO
- [x] Use of `AgentError`. No panics in macros
- [x] cli tests to work, since default profile is now agent. 
- [x] Resolve some comments I put regarding the `cargo component build` as part of creating a new app. (discussed with David)
- [x] Cleanups
- [x] Schema derivation needs to ensure using `golem_rust's` type classes and nothing else. But currently it uses `golem_wasm::IntoValue`. Please allow me to do this as part of type support where this derivation expands.

PR #2257, opened 2025-10-28
Make SDKs in templates configurable

noise64:
- make Rust and Ts SDK version or path configurable from CLI env vars (it reuses the current ones for rust, and added similars for TS)
- also foward override env vars in CLI integ tests, so it should be easy to switch locally when testing
- removed exclude / include template metadata properties, and made the logic fully driven by template kinds and filenames:
  - these propertied were already not respected properly, as there were exceptions based on the template type
  - and with the current templates they were not really needed (compared to previous iterations)
- added no-capture to CLI tests
- CLI integ tests now recongise the "QUIET" evn var, which silences the CLI output, and starts the server in silent mode

### DIFF
--- a/cli/golem-templates/templates/rust/rust-app-component-default/components-rust/component-name/src/lib.rs
+++ b/cli/golem-templates/templates/rust/rust-app-component-default/components-rust/component-name/src/lib.rs
@@ -1,29 +1,28 @@
 use golem_rust::{Schema, agent_definition, agent_implementation};
 
 #[agent_definition]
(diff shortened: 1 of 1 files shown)


afsalthaj:
I tried locally with the new rust counter [test](https://github.com/golemcloud/golem/pull/2254/commits/f7a005a70930aa3ca441b70c6eafa4258d3bc78d) and looks good to me. Just merging this PR to mine now.

image

Issue #1502, opened 2025-04-08
Benchmark result investigation: investigate the regression of large input benchmarks


vigoo:
Possibility: max oplog payload size and fallback to blob storage?
But maybe not because it affects remote RPC more than local.

Issue #1501, opened 2025-04-08
Benchmark result investigation: explain why local RPC invocations became significantly slower


vigoo:

vigoo:
This is not true anymore - however, there was a bug in the measurement itself (not sure if only in the new benchmark implementation or before too) which wrongly reported remote cases as local (not waiting for sharding manager to see all worker executors)

Issue #1499, opened 2025-04-08
Benchmark result investigation: confirm that reduced max size is due to the memory scheduling


vigoo:

vigoo:
This seems to be no longer the case - probably by previous fixes to the memory-based scheduling in the executor.
However, running the `sleep` benchmark with `--size=10000` does not seem to work and in one occasion I've seen a panic in the executor trying to read a non-existing oplog entry. I could not reproduce this and want to revisit the sleep 10000 case once other optimizations (including the atomic deployment) are done.

Issue #1433, opened 2025-03-19
Investigate performance issues in worker-executor / rpc


mschuwalow:
Running these rpc calls in a for loop is surprisingly slow:
https://github.com/mschuwalow/golem-hackathon-17.03.25/blob/main/backend/components-rust/chatroom-room/src/lib.rs#L50-L61

Running this on top of the single executable downloaded from a github release prints the following:
```
> golem worker stream chatroom:room/global
Fuzzy matched pattern chatroom:room as chatroom:room
Connecting to worker chatroom:room/global
[2025-03-19T15:40:41.525Z] [STDOUT  ] Fowarding message to user_1742398810853
[2025-03-19T15:40:41.926Z] [STDOUT  ] Fowarding message to user_1742398806608
```

Notice the timestamps between the log messages, this is about 400 ms for one invocation in the for loop.

---

The goal of this ticket is to investigate the cause of the performance issue and fix it if possible

vigoo:
To be revisited after the wRPC migration and the other benchmark investigations

vigoo:
This is no longer the case, see:

```
[2025-10-31T10:24:56.340Z] [INVOKE  ] STARTED  chatroom:room-exports/chatroom-room-api.{join} (7b26aee2-0f90-4aea-aa5f-99a1b8aef145)
[2025-10-31T10:24:56.340Z] [INVOKE  ] FINISHED chatroom:room-exports/chatroom-room-api.{join} (7b26aee2-0f90-4aea-aa5f-99a1b8aef145)
[2025-10-31T10:25:01.749Z] [INVOKE  ] STARTED  chatroom:room-exports/chatroom-room-api.{send} (c4228a9d-9363-4626-821a-43cb1c0e8127)
[2025-10-31T10:25:01.756Z] [STDOUT  ] Fowarding message to user_1761906296329
[2025-10-31T10:25:01.764Z] [INVOKE  ] FINISHED chatroom:room-exports/chatroom-room-api.{send} (c4228a9d-9363-4626-821a-43cb1c0e8127)
[2025-10-31T10:25:34.705Z] [INVOKE  ] STARTED  chatroom:room-exports/chatroom-room-api.{join} (1cd8cf8d-83c1-4adc-a05e-946d1aaeff87)
[2025-10-31T10:25:34.705Z] [INVOKE  ] FINISHED chatroom:room-exports/chatroom-room-api.{join} (1cd8cf8d-83c1-4adc-a05e-946d1aaeff87)
[...]

PR #2266, opened 2025-10-31
Benchmark framework and performance improvements

vigoo:
Fixes and improvements by analyzing the benchmark framework's results and past performance issues

In details:

### Postgres stats
Enabled Postgres statistics, and printing the 10 most frequent SQL statements at the end of the benchmarks. This is already useful and can be further improved to quickly detect things to be indexed and/or cached.

### Waiting for shard manager registration
The benchmarks using multiple executors were only waiting for each executor to start up, but not for the shard manager to know about them. This caused false reports in the RPC benchmarks where the distinction between "local RPC" and "remote RPC" is made by the routing table from the shard manager, which in these cases did not reflect the truth. Now the benchmarks (and integration tests, if using spawned processes) are waiting for the routing table to match the desired state before proceeding.

### Tracing improvements
The previous benchmark PR introduced support for exporting traces to Jaeger from benchmarks, but they were not really usable. Here we update the tracing stack, configure it properly so traces are propagated between the services, and also have multiple improvements in how spans are created and relate to each other. The result is that when benchmarks are ran with the `--oltp` are enabled, the results in Jaeger are actually usable to find performance issues.

### Killing benchmarks in case a process panics
There is a bug in the worker executor that sometimes happen during benchmark runs, causing a panic in the executor process. Previously the benchmarks did not detect this and took a lot of time to eventually time out. Now if the executor process dies, it kills the benchmark run as well. (This should be extended to all spawned processes but for now it's only implemented for worker executor processes).

### Worker service optimization
One interesting observation from the benchmark results was that invocation through Rib was much faster than using the invocation API directly. After investigating this, it turned out that the difference was caused by the _agent id normalization_ that currently requires agent metadata to parse and reprint the id. With rib, we know the component version the rib was compiled with, so we can get a specific component metadata that can be cached. In the direct invocation case it was always fetching the latest component metadata, which cannot be cached. This is the primary reason for the difference in performance. I've made some optimizations so the normalization tries to use the latest cached item first, and only reaches out to component service if that fails. 

### Durability overhead benchmark changes
I have changed the durability overhead benchmark a bit:
- added a new measured case, where an explicit oplog commit is called after each iteration of the inner loop. This simulates the case when a worker is doing "remote write style" oplog-heavy work where things need to be committed as soon as possible
- changed that it calls the wasi random host function and not the time one, because the time one was behaving specially in ephemeral cases

### Different commit strategy in persist-nothing regions
Applied a change so the oplog implementation knows when we are in a persist-nothing region, and can do optimizations based on this. The actual change I did is that in persist-nothing regions it commits less frequently (buffers more).  Although it did not cause much change in performance results, I'm leaving it may help help with further improvements later.

### Redis pipelined XADD
The redis indexed store uses XADD to add oplog entries to redis streams one by one. (There is no batch add operation in redis). However there is something called pipelineing, which I am now using to bulk-add multiple oplog entries in a single redis network message. This significantly improved the durability-overhead benchmark results.

### Updated fred
As part of this I also updated to the latest `fred` (and some minor updates to other libraries).

### Less oplog commits
Changed the following three oplog entries to not cause a commit (and by that, status recalculation):
- grow memory
- start span
- finish span

### Other tickets covered:
- https://github.com/golemcloud/golem/issues/1502 I did not find anything suspicious about large inputs
- https://github.com/golemcloud/golem/issues/1500 The primary reason why turning off persistence is not leading to faster execution now compared to the first results in 2024 is that back then it really turned off writing to the oplog - while now we still write everything, just don't use it during replays.
- https://github.com/golemcloud/golem/issues/1501 I did not find anything wrong or suspicious with local RPC now
- https://github.com/golemcloud/golem/issues/1433 tested this and it's not slow anymore
- https://github.com/golemcloud/golem/issues/1499 this does not seem to be the case anymore either, but while testing it I've found the earlier mentioned executor panic that is going to be investigated separately

### DIFF
--- a/golem-worker-executor/src/durable_host/mod.rs
+++ b/golem-worker-executor/src/durable_host/mod.rs
@@ -402,7 +402,7 @@ impl<Ctx: WorkerCtx> DurableWorkerCtx<Ctx> {
             // let current_size = self.update_worker_status();
             self.public_state
                 .worker()
-                .add_and_commit_oplog(OplogEntry::grow_memory(delta))
+                .add_to_oplog(OplogEntry::grow_memory(delta))
                 .await;
 
             self.public_state.worker().increase_memory(delta).await?;
@@ -1684,7 +1684,7 @@ impl<Ctx: WorkerCtx> InvocationContextManagement for DurableWorkerCtx<Ctx> {
         if is_live {
             self.public_state
                 .worker()
-                .add_and_commit_oplog(OplogEntry::start_span(
+                .add_to_oplog(OplogEntry::start_span(
                     span.start().unwrap_or(Timestamp::now_utc()),
                     span.span_id().clone(),
                     Some(parent.clone()),
@@ -1720,7 +1720,7 @@ impl<Ctx: WorkerCtx> InvocationContextManagement for DurableWorkerCtx<Ctx> {
         if self.is_live() {
             self.public_state
                 .worker()
-                .add_and_commit_oplog(OplogEntry::finish_span(span_id.clone()))
+                .add_to_oplog(OplogEntry::finish_span(span_id.clone()))
                 .await;
         } else {
             crate::get_oplog_entry!(self.state.replay_state, OplogEntry::FinishSpan)?;
--- a/golem-worker-executor/src/services/component.rs
+++ b/golem-worker-executor/src/services/component.rs
@@ -459,10 +459,6 @@ mod filesystem {
                 .metadata
                 .values()
                 .map(|local_metadata| {
-                    warn!(
-                        "all_cached_metadata returning: {} ({})",
-                        local_metadata.component_id, local_metadata.component_name
-                    );
                     golem_service_base::model::Component::from(local_metadata.clone())
                 })
                 .collect()


.github/workflows/benchmark.yaml
```diff
@@ -41,7 +41,7 @@ jobs:
           path: benchmark-results
 
       - name: Run benchmark suite
-        run: cargo run --profile benchmarks -p integration-tests --bin benchmarks '--' --primary-only suite --add-to-json benchmark-results/results/results.json integration-tests/benchmark_suites/ci.yaml spawned --mute-child
+        run: cargo run --profile benchmarks -p integration-tests --bin benchmarks '--' --primary-only suite --add-to-json benchmark-results/results/results.json integration-tests/benchmark_suites/ci.yaml spawned
```
noise64, line 44: is mute child intentionally removed (as in do we want to see it on main too?)

golem-worker-executor/src/services/oplog/mod.rs
```diff
@@ -168,7 +168,8 @@ pub trait Oplog: Any + Debug + Send + Sync {
     /// Adds a single entry to the oplog (possibly buffered), and returns its index
     async fn add(&self, entry: OplogEntry) -> OplogIndex;
 
-    async fn add_safe(&self, entry: OplogEntry) -> Result<(), String> {
+    /// A variant of add that can inject failures in tests. TO BE REMOVED
```
noise64, line 171: i'm not sure i understand how this can inject failures (and when it should be removed?)

mschuwalow (approved):

.github/workflows/benchmark.yaml
vigoo: yes it is not adding that much extra output, but with it even the executor panic was hidden. (this --quiet / --verbose / --mute-child combination of flags is not the best but I did not want to reinvent it now)

golem-worker-executor/src/services/oplog/mod.rs
vigoo: This was added by Peter so he can inject failures where `add` would be called by overriding the oplog implementation in his rdbms tests. I just renamed it now because it was the opposite of safe :) (And I don't think that's the best way to do what those tests are trying to do, but that's a different topic)

PR #2268, opened 2025-11-02
Feat : creation of MCP server  [WIP]

### LINKED ISSUE #1926: Incorporate MCP Server into Golem CLI
With a new command, Golem CLI enters serve mode:

```bash
%> golem-cli --serve --serve-port 1232
%> golem-cli running MCP Server at port 1232
```

In this mode, Golem CLI creates an MCP Server that exposes different commands as different tools, and exposes relevant resources (the manifest file in the current, ancestor, and children directories) as resources.

When this ticket is completed, it should be possible to use an agent such as Claude Code to perform anything that Golem CLI can do. Moreover, all individual tools and resources must be end-to-end tested with an MCP Client that interacts with the MCP Server.

[This library](https://github.com/rust-mcp-stack/rust-mcp-sdk) looks like the one to use for Rust, but perhaps there are better options available.

Abiji-2020:
Currently created a  single tool to build the app and tested E2E and it is working as intended, and with the similar way I will be extending functionalities for all the commands into tools.

closes #1926 
/claim #1926

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3039,9 +3039,9 @@ dependencies = [
 
 [[package]]
 name = "desert_core"
-version = "0.1.3"
+version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e5260e1dc286b2bb7e4de5f6475ac25261b90b670aad4d82eb2d22c5c7372a5e"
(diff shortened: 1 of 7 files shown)


vigoo:
This bounty is cancelled

PR #2282, opened 2025-11-05
Migration to desert_rust and better oplog entry types

### LINKED ISSUE #1616: Introduce a new oplog format (v3) using desert-rs
Once it is done, verify the performance impact by running some of the benchmarks.

### LINKED ISSUE #2191: Oplog specs improvements
Oplog is a core entity and has multiple data types and trait implementations attached to it. It is currently quite verbose and redundant. 

This ticket is about cleaning up this a bit, possibly by using some helper macros, to make sure there is a single (or as few as possible) source of truth about what an oplog entry is and what information it stores.

Part of this is the untyped payloads. We want to keep the untyped payloads for known entries for performance reasons, but the currenty string-based switch is something continuously goes out of sync. We should find a better way to do that too.

vigoo:
Resolves #1616 
Resolves #2191 

**THIS IS BREAKING THE OPLOG FORMAT (and other binary payloads)**
**ALSO BREAKING THE WIT INTERFACE**

Switches from `bincode` to `desert-rust` for all binary serialization, primarily because it supports schema evolution, and changes how oplog entries and their payloads are defined to make it more type safe. To reduce the boilerplate it comes with, some declarative macros have been introduced.

In details:

### Desert
desert-rust is the port of my Scala library desert, and it provides a derivable binary codec that is comparable in size and performance to bincode (although a little bit slower at the moment). The big advantage is that if we decide to keep the format stable, we can still modify our data types quite freely using desert's schema evolution capabilities. Until we decide so we should not use these features though, as using them has a performance (and size) penalty. 

For most of the cases the difference is just a different typeclass in derive. Newtypes are supported with `#[desert(transparent)]`. For types with hand-written custom codec, the difference is bigger but nothing surprising.

### Definition of oplog entries
Oplog entries (the type `OplogEntry`) are the direct representation of an oplog entry that gets stored in the indexed storage. OplogEntries should be serialized as concise as possible and not having anything unnecessary.
Another type, `PublicOplogEntry` defines the same logical cases but with the intention of used in public APIs, such as REST APIs. These contain more verbose information that can be derived (with additional context) from the stored oplog entries.
For some oplog entry cases there are also _payloads_ in them that depend on the entry and in some cases (host function call or custom durability) they can be even user defined. The change to payloads is discussed in a separate section.

Previously `OplogEntry` and `PublicOplogEntry` were defined by hand and kept in sync by hand. In addition to that, a lot of boilerplate had to be written when adding a new entry (or changing an existing one):
- constructor functions 
- "rounding" functions for tests
- "is_hint" and timestamp extraction
- for `PublicOplogEntry` cases, separate inner types had to be defined to make it OpenAPI compatible
- etc.

This PR introduces a helper macro called `oplog_entry!` with the intention to be the single source of truth for what an oplog entry is.

The syntax is the following:

```rust
oplog_entry! {
    /// The worker has completed an invocation
    ExportedFunctionCompleted {
        hint: false
        raw {
            response: OplogPayload<Option<ValueAndType>>,
            consumed_fuel: i64,
        }
        public {
            response: Option<golem_wasm::ValueAndType>,
            consumed_fuel: i64,
        }
    },
    /// Worker suspended
    Suspend {
        hint: true
        raw {}
        public {}
    },
   ...
```
`hint` decides if the entry is "hint" or not (this is not the best name but this PR is not changing it). Basically "hint" entries are the ones that can get into the oplog in a nondeterministic way and non-hint ones are the ones the replay goes through.
`raw` contains the fields for `OplogEntry` and `public` the fields for `PublicOplogEntry`. The macro generates the required inner types for the public entry, and all the constructor and other methods for oplog entry. The derived codecs for both types are:
- `OplogEntry`: `Clone, Debug, PartialEq, desert_rust::BinaryCodec`
- `PublicOplogEntry`: `Clone, Debug, PartialEq, serde::Serialize, serde::Deserialize, golem_wasm::derive::IntoValue, poem_openapi::Union`

### Oplog payload
A couple of oplog entries have payloads. These were arbitrary byte arrays before, but their contents was always serialized/deserialized with the "default" serializer (which was bincode), except in one single case where it was protobuf. Another property of these payloads are that their size has a limit, and if they are above the limit, they are not directly stored in the oplog entry, but uploaded to the blob storage.

When we look at an oplog entry through the `PublicOplogEntry` "view", we want to see structured information about these payloads. However, in the "real" `OplogEntry`, the payload should be as concise as possible, for performance reasons.

This was handled in a very arbitrary way before; every host function implementation just saved whatever binary-serializable payload it decided to save, together with a fully-qualified function name string. Then when converting oplog entries to `PublicOplogEntry` values, a huge, string-based switch tried to reverse engineer these binary serialized blobs and convert them to structured data (`ValueAndType`). And finally, in the debugging service there was a need to somehow convert back these `ValueAndType` values into the original oplog entry format, which may worked once when it was written but since then it has been out of sync completely, as there were no types helping these conversions in any way.

This PR introduces a type-safe way to define all the above. The only disadvantage is that many types that are really "worker executor implementation details" had to be moved to `golem-common`, where the oplog entry definitions live. I think this is an acceptable trade-off.

#### OplogPayload is typed
The first related change is that `OplogPayload` is now typed with the type it holds. This also makes it possible to have unserialized payloads in memory.
For example:
```rust
    ExportedFunctionCompleted {
        hint: false
        raw {
            response: OplogPayload<Option<ValueAndType>>,
``` 

defines now that `ExportedFunctionCompleted` entry's response payload MUST contain an `Option<ValueAndType>`.

#### HostRequest and HostResponse
The more interesting oplog entry is `ImportedFunctionInvoked`, which is the one emitted for host function calls (or custom durability calls), containing two payloads:

```rust
            request: OplogPayload<HostRequest>,
            response: OplogPayload<HostResponse>,
```

These are the ones that were previously ad-hoc in each host function implementation. Now two types define all the possibilities - both are enums that contain a `Custom` case that is reserved for **custom durability**:

```rust
#[derive(Clone, Debug, PartialEq, BinaryCodec)]
pub enum HostRequest {
    Custom(golem_wasm::ValueAndType),
    // ...
}
```

The other-than-custom cases of these two types are also generated by a helper macro, because they require some boilerplate for conversion to/from various representations.
It looks like this:

```rust
oplog_payload! {
    HostRequest => {
        NoInput {},
        BlobStoreContainer {
            container: String
        },
        BlobStoreContainerAndObject {
            container: String,
            object: String
        },
        BlobStoreContainerAndObjects {
            container: String,
            objects: Vec<String>
        },
        // ...
```

So what does this macro generate?

- the `HostRequest` (or `HostResponse`) enum itself
- a separate struct for each case, for example `HostRequestBlobStoreContainer`
- each such struct derives `Clone, Debug, PartialEq, BinaryCodec, IntoValue, FromValue`
- generates a top-level pair of `IntoValueAndType` and `FromValue` implementations for the enum, which delegates the call to the inside so these "payload types" are not visible for the user in the public oplog entries (for example for an oplog entry operating on a blob store container, the request field is going to be a struct with a `container` field, and not a huge `ValueAndType` describing the whole enum and a value as one of its cases. This is very important for keeping the public oplog entries reasonably sized)

With these types, host function implementations can persist data in a typed way:

```rust
        let result = if durability.is_live() {
            let value = Host::get_random_u64(&mut self.as_wasi_view()).await?; // this supposed to never fail
            durability
                .persist(self, HostRequestNoInput {}, HostResponseRandomU64 { value })
                .await
        } else {
            durability.replay(self).await
        }?;

        Ok(result.value)
```

#### Host payload pairs
The above types still don't specify what kind of "pairs" of request-response host payloads are possible, and how they relate to the function names that were previously identifying these oplog entries.

This brings us to the third (and last) helper macro called `host_payload_pairs!`:

```rust
pub mod host_functions {
    use crate::host_payload_pairs;

    host_payload_pairs! {
        (RdbmsMysqlDbConnectionExecute => "rdbms::mysql::db-connection", "execute", GolemRdbmsRequest, GolemRdbmsRowCount),
        (RdbmsMysqlDbConnectionQuery => "rdbms::mysql::db-connection", "query", GolemRdbmsRequest, GolemRdbmsResult),
        // ...
```

This "table" contains an identifier (such as `RdbmsMysqlDbConnectionQuery`) and four parameters:
- interface name
- function name
- host request type
- host response type

For each row, it defines a type with the identifier that implements the following trait:

```rust
pub trait HostPayloadPair {
    type Req: Into<HostRequest>;
    type Resp: Into<HostResponse> + TryFrom<HostResponse, Error = String>;

    const INTERFACE: &'static str;
    const FUNCTION: &'static str;
    const FQFN: &'static str;

    const HOST_FUNCTION_NAME: host_functions::HostFunctionName;
}
```

Requiring `HostPayloadPair` in `Durability` makes it sure that the function names and request/response pairs are now strictly connected by the type system. 

`HostFunctionName` is a simple enum containing all the possible fully-qualified function names (plus a custom(String) case) which makes it possible to store a single byte in the oplog entries for built-in cases and only store actual strings when the custom durability API is used.

### Lot of optimizations and cleanups
While reworking the payload handling, I've eliminated a lot of unnecessary conversions and clones from the hot path. 

### WIT changes
Two changes in the WIT interfaces:
- removed the ability to store arbitrary byte arrays in custom durability - it always has to be structured data (the raw one was there only for "ourselves" when we thought we'll do durability-by-composition)
- the WIT version of `PublicOplogEntry` now also contains `ValueAndType` fields not just `WitValue`, which makes it not loosing any data when converting to/from WIT vs native

I've changed the WIT package versions as well, as we did not do that for a while, but it does not matter from a compatibility point of view (not changing the version but changing the interface is just as breaking as the other)

This requires rebuilding test components and possibly some rounds of publishing/testing where not everyting passes, unfortunately.

### DIFF
--- a/golem-worker-executor/tests/indexed_storage.rs
+++ b/golem-worker-executor/tests/indexed_storage.rs
@@ -847,14 +847,7 @@ async fn drop_prefix_no_match(
         .await
         .unwrap();
 
-    check!(
-        result
-            == vec![
-                (10, value1),
-                (11, value2),
-                (12, value3)
-            ]
-    );
+    check!(result == vec![(10, value1), (11, value2), (12, value3)]);
 }
 
 #[test]


vigoo:
Merging with failing CLI tests, and will create a follow-up PR once the external dependencies are published

PR #2284, opened 2025-11-06
Implement RPC in code first agent

### LINKED ISSUE #2186: Code-first Rust RPC
Make sure it includes the "trigger" and "schedule" variants from the start, as adding them as an afterthought may be difficult.

afsalthaj:
Fixes #2186 

The existing integration test fully worked with Rust code first RPC

- [x] Async support with wstd: macro expansion down below
- [x] generate trigger and schedule variants of every method
- [x] reuse wasm-rpc client (PR review comment)
- [x] use wstd

### DIFF
--- a/sdks/rust/golem-rust-macro/src/agentic/client_generation.rs
+++ b/sdks/rust/golem-rust-macro/src/agentic/client_generation.rs
@@ -111,6 +111,27 @@ fn get_remote_method_impls(tr: &ItemTrait, agent_type_name: String) -> proc_macr
                 syn::ReturnType::Default => quote! { () },
             };
 
+            let process_invoke_result = match &method.sig.output {
+                syn::ReturnType::Type(_, ty) => {
+                    let is_unit = matches!(**ty, syn::Type::Tuple(ref t) if t.elems.is_empty());
+
+                    if !is_unit {
+                        quote! {
+                            let element_value = golem_rust::golem_agentic::golem::agent::common::ElementValue::ComponentModel(wit_value);
+                            let element_schema = <#ty as golem_rust::agentic::Schema>::get_type();
+                            <#ty as golem_rust::agentic::Schema>::from_element_value(element_value, element_schema).expect("Failed to deserialize rpc result to return type")
+                        }
+                    } else {
+                        quote! {
+                            ()
+                        }
+                    }
+                },
+                syn::ReturnType::Default => quote! {
+                    ()
+                },
+            };
+
             // Depending on the input parameter type and output parameter type generate different implementations
             let input_parameter_type = get_input_param_type(&method.sig);
 
@@ -134,12 +155,7 @@ fn get_remote_method_impls(tr: &ItemTrait, agent_type_name: String) -> proc_macr
 
                           let wit_value = golem_rust::agentic::unwrap_wit_tuple(rpc_result_ok);
 
-                          let element_value = golem_rust::golem_agentic::golem::agent::common::ElementValue::ComponentModel(wit_value);
-
-                          let element_schema = <#return_type as golem_rust::agentic::Schema>::get_type();
-
-                          <#return_type as golem_rust::agentic::Schema>::from_element_value(element_value, element_schema).expect("Failed to deserialize rpc result to return type")
-
+                          #process_invoke_result
                         }
 
                         pub fn #trigger_method_name(#(#inputs),*) {
--- a/sdks/rust/golem-rust/src/agentic/schema.rs
+++ b/sdks/rust/golem-rust/src/agentic/schema.rs
@@ -78,28 +78,3 @@ impl<T: IntoValue + FromValueAndType> Schema for T {
         }
     }
 }
-
-impl Schema for () {
-    fn get_type() -> ElementSchema {
-        let analysed_type = AnalysedType::Result(TypeResult {
-            name: None,
-            owner: None,
-            ok: None,
-            err: None,
-        });
-        let wit_type = WitType::from(analysed_type);
-        ElementSchema::ComponentModel(wit_type)
-    }
-
-    fn to_element_value(self) -> Result<ElementValue, String> {
-        let value = Value::Result(Ok(None));
-        Ok(ElementValue::ComponentModel(WitValue::from(value)))
-    }
-
-    fn from_element_value(_value: ElementValue, _schema: ElementSchema) -> Result<Self, String>
-    where
-        Self: Sized,
-    {
-        Ok(())
-    }
-}
--- a/sdks/rust/golem-rust-macro/src/agentic/agent_definition_impl.rs
+++ b/sdks/rust/golem-rust-macro/src/agentic/agent_definition_impl.rs
@@ -155,9 +155,13 @@ fn get_agent_type_with_remote_client(
                     match &trait_fn.sig.output {
                         syn::ReturnType::Default => (),
                         syn::ReturnType::Type(_, ty) => {
-                            output_parameters.push(quote! {
-                                ("return-value".to_string(), <#ty as golem_rust::agentic::Schema>::get_type())
-                            });
+                            let is_unit = matches!(**ty, syn::Type::Tuple(ref t) if t.elems.is_empty());
+
+                            if !is_unit {
+                                output_parameters.push(quote! {
+                                    ("return-value".to_string(), <#ty as golem_rust::agentic::Schema>::get_type())
+                                });
+                            }
                         }
                     };
                 },
@@ -177,9 +181,10 @@ fn get_agent_type_with_remote_client(
             };
 
             let output_schema = match output_param_type.param_type {
-                ParamType::Tuple => quote! {
-                    golem_rust::golem_agentic::golem::agent::common::DataSchema::Tuple(vec![#(#output_parameters),*])
-                },
+                ParamType::Tuple =>
+                    quote! {
+                        golem_rust::golem_agentic::golem::agent::common::DataSchema::Tuple(vec![#(#output_parameters),*])
+                    },
                 ParamType::Multimodal => quote! {
                     golem_rust::golem_agentic::golem::agent::common::DataSchema::Multimodal(vec![#(#output_parameters),*])
                 },
--- a/sdks/rust/golem-rust-macro/src/agentic/agent_implementation_impl.rs
+++ b/sdks/rust/golem-rust-macro/src/agentic/agent_implementation_impl.rs
@@ -177,7 +177,7 @@ fn build_match_arms(
 
             let post_method_param_extraction_logic = match output_param_type.param_type {
                 ParamType::Tuple => match output_param_type.function_kind {
-                    FunctionKind::Async => Some(quote! {
+                    FunctionKind::Async if !output_param_type.is_unit => Some(quote! {
                         let result = self.#ident(#(#param_idents),*).await;
                         <_ as golem_rust::agentic::Schema>::to_element_value(result).map_err(|e| {
                             golem_rust::agentic::custom_error(format!(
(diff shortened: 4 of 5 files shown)


afsalthaj:
Progress with async here... just need to fix a few minor bugs.. https://github.com/golemcloud/golem/compare/add_rpc...async_support?expand=1

cli/golem-cli/tests/app/agents.rs
```diff
@@ -336,7 +336,7 @@ async fn test_rust_code_first_with_rpc_and_all_types() {
 
     run_and_assert(
         &ctx,
-        "rust:agent/bar-agent.{fun-option-complex}",
+        "rust:agent/foo-agent.{fun-option-complex}",
```
afsalthaj, line 355: now we have two agents In the code. `foo` agents always delegates to bar agent's methods through rpc.

cli/golem-templates/templates/rust/rust-app-common/.vscode/settings.json
```diff
@@ -1,3 +1,3 @@
 {
-    "rust-analyzer.server.extraEnv": { "CARGO": "cargo-component" }
+    "rust-analyzer.procMacro.enable": true
```
afsalthaj, line 2: This setting is rarely needed but important.
server.extraEnv with cargo component was slowing down VS Code builds by 3–4 minutes, and sometimes never finished. It took me some time to identify the culprit, so I changed it. This likely started after we switched to code-first.

I also re-installed VS Code to ensure macro expansions weren’t causing false compiler errors. Sometimes these had to be manually set, so including this setting in the template is safer

Rustrover note: I tested with Rustrover.It has good defaults and macro expansions on by default, so false compiler errors shouldn't happen hopefully.

sdks/rust/golem-rust/Cargo.toml
```diff
@@ -13,11 +13,14 @@ path = "src/lib.rs"
...
 golem-rust-macro = {path = "../golem-rust-macro", version = "0.0.0", optional = true}
-
 golem-wasm = {version = "1.3.2-dev.2", default-features = false, features = ["stub"]}
+
+async-trait = {version = "0.1.89"}
 serde = {version = "1", optional = true}
 serde_json = {version = "1", optional = true}
 uuid = {version = "1", features = ["v4"]}
+wasi = {version = "=0.14.7+wasi-0.2.4"}
+wasi-async-runtime = { git = "https://github.com/golemcloud/wasm-http-tools.git" }
```
afsalthaj: The published wasi-async-runtime depends on old wasi, and it had conflicts with the one through golem-wasm. 

May be there is a way to still depend on published wasi-async-runtime which is 0.1.2, and yet makes sure to use recent wasi versions. But I felt more comfortable bumping up the version in wasi-async-runtime, and kept it as our fork. We can raise PR to upstream if needed.

Note that, the dependency combinations used in places like wasm-rquickjs won't work here, because we have dependency to golem-wasm.

afsalthaj:
### Macro expansion for the following:

Pasting only the client generation (as rest for already done before)

Note that the expansion, if its multimodal is different. It was better of branching off to generate completely separate code for multimodal (rather than embedding everything in one code)

```rust

use golem_rust::agentic::Agent;

#[agent_definition]
trait AssistantAgent {
    async fn new(init: MyId) -> Self; // doesn't need to be async
    async fn ask(&self, question: String) -> String;
}

struct AssistantAgentImpl {
    id: MyId,
}

#[agent_implementation]
impl AssistantAgent for AssistantAgentImpl {
    async fn new(id: MyId) -> Self {
        AssistantAgentImpl { id }
    }
    async fn ask(&self, question: String) -> String {
        let client = WeatherAgentClient::get("input".to_string());
        let result = client.get_weather(question).await;
        format!("Assistant Agent (ID: {}) reporting weather: {}", self.get_agent_id(), result)
    }
}

#[derive(Schema)]
struct MyId {
    id: String,
}

#[agent_definition]
trait WeatherAgent {
    async fn new(input: String) -> Self;
    fn get_weather(&self, location: String) -> String;
}

struct WeatherAgentImpl;

#[agent_implementation]
impl WeatherAgent for WeatherAgentImpl {
    async fn new(_input: String) -> Self {
        WeatherAgentImpl
    }

    fn get_weather(&self, location: String) -> String {
        format!("The weather in {} is Sunny. Reported by {}", location, self.get_agent_id())
    }
}
```

### Client generation 

```rust

pub struct WeatherAgentClient {
    agent_id: golem_rust::wasm_rpc::AgentId,
    wasm_rpc: golem_rust::wasm_rpc::WasmRpc,
}
impl WeatherAgentClient {
    pub fn get(input: String) -> WeatherAgentClient {
        let agent_type = golem_rust::golem_agentic::golem::agent::host::get_agent_type(
                "WeatherAgent",
            )
            .expect("Internal Error: Agent type not registered");
        let element_values = <[_]>::into_vec(
            ::alloc::boxed::box_new([
                golem_rust::agentic::Schema::to_element_value(input)
                    .expect("Failed to convert constructor parameter to ElementValue"),
            ]),
        );
        let data_value = golem_rust::golem_agentic::golem::agent::common::DataValue::Tuple(
            element_values,
        );
        let agent_id_string = golem_rust::golem_agentic::golem::agent::host::make_agent_id(
                "WeatherAgent",
                &data_value,
            )
            .expect("Internal Error: Failed to make agent id");
        let agent_id = golem_rust::wasm_rpc::AgentId {
            agent_id: agent_id_string,
            component_id: agent_type.implemented_by.clone(),
        };
        let wasm_rpc = golem_rust::wasm_rpc::WasmRpc::new(&agent_id);
        WeatherAgentClient {
            agent_id: agent_id,
            wasm_rpc: wasm_rpc,
        }
    }
    pub fn get_agent_id(&self) -> String {
        self.agent_id.agent_id.clone()
    }
    pub async fn get_weather(&self, location: String) -> String {
        let wit_values: Vec<golem_rust::wasm_rpc::WitValue> = <[_]>::into_vec(
            ::alloc::boxed::box_new([
                golem_rust::agentic::Schema::to_wit_value(location).expect("Failed"),
            ]),
        );
        let rpc_result_future = self
            .wasm_rpc
            .async_invoke_and_await("weather-agent.{get-weather}", &wit_values);
        let rpc_result: Result<
            golem_rust::wasm_rpc::WitValue,
            golem_rust::wasm_rpc::RpcError,
        > = golem_rust::agentic::await_invoke_result(rpc_result_future).await;
        let rpc_result_ok = rpc_result
            .expect(
                ::alloc::__export::must_use({
                        ::alloc::fmt::format(
                            format_args!(
                                "rpc call to {0} failed",
                                "weather-agent.{get-weather}",
                            ),
                        )
                    })
                    .as_str(),
            );
        let wit_value = golem_rust::agentic::unwrap_wit_tuple(rpc_result_ok);
        let element_value = golem_rust::golem_agentic::golem::agent::common::ElementValue::ComponentModel(
            wit_value,
        );
        let element_schema = <String as golem_rust::agentic::Schema>::get_type();
        <String as golem_rust::agentic::Schema>::from_element_value(
                element_value,
                element_schema,
            )
            .expect("Failed to deserialize rpc result to return type")
    }
    pub fn trigger_get_weather(&self, location: String) {
        let wit_values: Vec<golem_rust::wasm_rpc::WitValue> = <[_]>::into_vec(
            ::alloc::boxed::box_new([
                golem_rust::agentic::Schema::to_wit_value(location).expect("Failed"),
            ]),
        );
        let rpc_result: Result<(), golem_rust::wasm_rpc::RpcError> = self
            .wasm_rpc
            .invoke("weather-agent.{get-weather}", &wit_values);
        rpc_result
            .expect(
                ::alloc::__export::must_use({
                        ::alloc::fmt::format(
                            format_args!(
                                "rpc call to trigger {0} failed",
                                "weather-agent.{get-weather}",
                            ),
                        )
                    })
                    .as_str(),
            );
    }
    pub fn schedule_get_weather(
        &self,
        location: String,
        scheduled_time: golem_rust::wasm_rpc::golem_rpc_0_2_x::types::Datetime,
    ) {
        let wit_values: Vec<golem_rust::wasm_rpc::WitValue> = <[_]>::into_vec(
            ::alloc::boxed::box_new([
                golem_rust::agentic::Schema::to_wit_value(location).expect("Failed"),
            ]),
        );
        self.wasm_rpc
            .schedule_invocation(
                scheduled_time,
                "weather-agent.{get-weather}",
                &wit_values,
            );
    }
}
```

sdks/rust/golem-rust/tests/agent.rs
```diff
@@ -68,7 +67,63 @@ mod tests {
         }
     }
 
-    #[derive(Schema)]
+    #[agent_definition]
```
afsalthaj, line 71: This is just extending "does it compile" test in golem-rust

sdks/rust/golem-rust/Cargo.toml
vigoo: We cannot publish the golem-rust crate this way so we cannot really merge the PR with it.
Upstream PR is also make no sense, as this crate is "dead" and it's successor is `wstd`.

sdks/rust/golem-rust-macro/src/agentic/client_generation.rs
```diff
@@ -0,0 +1,190 @@
...
+            let output_parameter_type = get_output_param_type(&method.sig);
+
+            match (input_parameter_type.param_type, output_parameter_type.param_type) {
+                (ParamType::Tuple, ParamType::Tuple) =>
+                    Some(quote!{
+                        pub async fn #method_name(#(#inputs),*) -> #return_type {
+                          let wit_values: Vec<golem_rust::wasm_rpc::WitValue> =
+                            vec![#(golem_rust::agentic::Schema::to_wit_value(#input_idents).expect("Failed")),*];
+
+                          let wasm_rpc = golem_rust::wasm_rpc::WasmRpc::new(&self.agent_id);
```
vigoo: It would be more efficient to create the `WasmRpc` resource once when the client is constructed, and store it as a field, instead of creating a new one for each invocation.

sdks/rust/golem-rust/Cargo.toml
vigoo: I wanted to do the wstd migration separately (very soon) but if there is no way to use the published `wasi-async-runtime` for now, then it has to be done for this crate now. 
If that's the case make sure to use the `=0.5.4` because from 0.5.5 it introduces a 'static constraint that is problematic. The WASI crate it surely works with is ` "=0.14.1+wasi-0.2.3"`. 
You can check the `internal.rs` diff in https://github.com/golemcloud/wasm-rquickjs/pull/66 for an example of migrating to wstd.

sdks/rust/golem-rust-macro/src/agentic/client_generation.rs
afsalthaj: Makes sense

vigoo:
In addition to the above, missing (as mentioned in private too):
- trigger rpc call
- schedule rpc call

sdks/rust/golem-rust-macro/src/agentic/client_generation.rs
afsalthaj: Resolved. See updated expansion (only client generated) here: https://github.com/golemcloud/golem/pull/2284#issuecomment-3509483725

afsalthaj:
https://github.com/golemcloud/golem/pull/2284#issuecomment-3512660393

Done, and updated the macro expansion: https://github.com/golemcloud/golem/pull/2284#issuecomment-3509483725

And extended tests to use schedule and trigger variants of a function. Manually tested too

sdks/rust/golem-rust/Cargo.toml
afsalthaj: Resolved: https://github.com/golemcloud/golem/pull/2284/commits/519d585341b67bf90618416abea1ffe8349ee8c2

Tests working too

sdks/rust/golem-rust/src/agentic/schema.rs
```diff
@@ -77,3 +78,28 @@ impl<T: IntoValue + FromValueAndType> Schema for T {
         }
     }
 }
+
+impl Schema for () {
+    fn get_type() -> ElementSchema {
+        let analysed_type = AnalysedType::Result(TypeResult {
```
vigoo: This type represents `Result<(), ()>` and not `()`. The schema for `()` should be the empty tuple data schema, same as we did in the TS SDK. There is no way to express `()` as a component model type

afsalthaj: Yes. I will fix it very soon

afsalthaj: Done.

PR #2295, opened 2025-11-13
Support unstructured text in rust code first agent

afsalthaj:
## Implemented idea

```rust
#[agent_defintiion]
trait Foo {
    fn new(b: String) -> Self;
     // without language code
    async fn process(&self, input: UnstructuredText);
    // with language code
    async fn process_lc(&self, input: UnstructuredText<MyLanguage>);
}

#[derive(AllowedLanguages)]
enum MyLanguage {
    En,
   #[code("de")]
    German
}

```

## Alternative idea (not implemented in this PR, but happy to discuss)

There is an alternative idea too, where SDK exposes only unconstrained UnstructuredText, and for constrained one, users can do. 

```rust

#[agent_defintiion]
trait Foo {
    fn new(b: String) -> Self;
    // without language code
    async fn process(&self, input: UnstructuredText);
    // with  language code
    async fn process_lc(&self, input: MyTextType);
}

#[derive(UnstructuredText)]
enum MyTextType {
   En(UnstrucutedText), 
    #[code("de")]
   German(UnstrcutedText),
}   

```

### Comparison

It's comparable, but I kind of liked the one in this PR for some reason, after thinking about both.

#### Implemented version
* Looks more compiler driven. Need a custom language ? - pass a language code (LC) too when calling `fromInline` and that will make it `UnstructuredText<LC>` , and will be asked to provide `AllowedLanguages` trait - `#[derive(AllowedLanguages)]`

* More correct data representation. In principle, the implemented version allows users to use more correct data structure:  factorising `(A * X + B * X)` (alternative)  to `(A + B) * X`. Language code is `A + B` and `X` is text, no duplication. Former is alternative version.  

* Feels more orthogonal natuarlly. i.e, `UnstructuredText` is unconstrained, `UnstructuredText<MyText>` is constrained. 

#### alternative version

* languages are not separated. If language restriction is needed, its a custom type of enum
* every variant in enum should hold exactly `unstructured-text`  in it.  Meaning, users need to define it in a "specific" way.  
* if they just need unstructured-text out of their custom type, they are forced to pattern match. They will end up writing helpers may be. more things to do.

### DIFF
--- a/cli/golem-cli/tests/app/agents.rs
+++ b/cli/golem-cli/tests/app/agents.rs
@@ -378,28 +378,28 @@ async fn test_rust_code_first_with_rpc_and_all_types() {
         "rust:agent/foo-agent.{fun-unstructured-text}",
         &[r#"url("foo")"#],
     )
-        .await;
+    .await;
 
     run_and_assert(
         &ctx,
         "rust:agent/foo-agent.{fun-unstructured-text}",
         &[r#"inline({data: "foo", text-type: none})"#],
     )
-        .await;
+    .await;
 
     run_and_assert(
         &ctx,
         "rust:agent/foo-agent.{fun-unstructured-text-lc}",
         &[r#"url("foo")"#],
     )
-        .await;
+    .await;
 
     run_and_assert(
         &ctx,
         "rust:agent/foo-agent.{fun-unstructured-text-lc}",
         &[r#"inline({data: "foo", text-type: some({language-code: "en"})})"#],
     )
-        .await;
+    .await;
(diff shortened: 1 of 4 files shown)


test-components/benchmarks/wit/deps/golem-agent/host.wit
```diff
@@ -0,0 +1,24 @@
+package golem:agent;
+
+interface host {
```
afsalthaj: I don't know why it came even after git ignore. Will delete this

PR #2296, opened 2025-11-13
Agent mode

### LINKED ISSUE #2285: Add durability mode to agent type and remove it from component metadata
We move the current component type (durable vs ephemeral) to the agent metadata and call it durability mode. 
This is not a very big change but has some important consequences:
- Legacy workers (non-agents) can no longer be ephemeral. This will break some tests, and those tests need to be migrated to be agents.
- We can no longer decide durability mode just based on the component metadata, but we also need the worker (agent) id too - as that's what selects the agent type. I don't expect this to be a problem

Parts of this task:
- Move the actual metadata
- Add a way to the TS and Rust SDKs to configure the durability mode
- Change the agent and host WITs
- Change the test components where needed
- Remove component type from CLI / app manifest

vigoo:
Resolves #2285 

Removes the `component-type` component property and defines an "agent mode" agent property.

The user-facing way to specify it:

Rust:
```rust
#[agent_definition(ephemeral)]
trait Xyz { 
  // ...
}

// or

#[agent_definition(mode = ephemeral)]
trait Xyz { 
  // ...
}
```

TypeScript:
```typescript
@agent({ mode = "ephemeral" })
class Xyz extends BaseAgent {
  // ...
}
```

Corresponding TS SDK change: https://github.com/golemcloud/golem-agentic/pull/200/

### DIFF
--- a/cli/golem-templates/src/lib.rs
+++ b/cli/golem-templates/src/lib.rs
@@ -45,7 +45,7 @@ static APP_MANIFEST_HEADER: &str = indoc! {"
 "};
 
 static GOLEM_RUST_VERSION: &str = "1.9.0";
-static GOLEM_TS_VERSION: &str = "0.0.59";
+static GOLEM_TS_VERSION: &str = "0.0.60";
 
 fn all_templates(dev_mode: bool) -> Vec<Template> {
     let mut result: Vec<Template> = vec![];


jdegoes:
> @agent({ durabilityMode = "ephemeral" })

Maybe we can make this just `mode` to match with Rust?

noise64 (approved):

PR #2326, opened 2025-11-19
Multimoda base

### LINKED ISSUE #2297: Add a simple multimodal type to both Rust and TS SDKs
This type should be simply unconstraint unstructured text | unconstrained unstructured binary.
Maybe another variant that adds a single custom type T.

These should be defined on top of the existing multimodal type support as a "basic mode", not replacing those.

afsalthaj:
Fixes #2297 

https://github.com/golemcloud/golem/blob/multimoda_base/sdks/rust/golem-rust/src/agentic/multimodal.rs#L208

### Multimodal

```rust
let multimodal_data: Multimodal = Multimodal::new([
    BasicModality::text("foo".to_string()),
    BasicModality::binary(vec![1, 2, 3], "image/png"))
]);

```

### MultimodalCustom

```rust
let multimodal_data: MultimodalCustom<MyCustomType> = MultimodalCustom::new([
    CustomModality::text("foo".to_string()),
    CustomModality::binary(vec![1, 2, 3], "image/png"),
    CustomModality::Custom(MyCustomType { x: "bar".to_string(), y: 42 }),
]);

#[derive(Schema)]
struct MyCustomType {
    x: String,
    y: i32,
}

```

### MultimodalAdvanced

```rust

let multimodal_data: MultimodalAdvanced<Input> = MultimodalAdvanced::new([
    Input::Text("foo".to_string()),
    Input::Image(vec![1, 2, 3])
]);

#[derive(MultimodalSchema)]
enum Input {
    Text(String),
    Image(Vec<u8>),
}

```

### Simplifying macros, bug fixes

**User outcome of this change:** They can do their own type aliases for built-in multimodal types we expose, and so for unstructured-text and binary

Details:

Also, removed the odd logic of "is_unstructured_text" or "is_unstructured_binary" in macros, making it more reliable.  Why? what if its a type alias?

With multimodal_base, what really happened was introduction of type alias. This questioned "is_multimodal_type" function which checks if a type is really Multimodal. And it compile time failed because MultimodalBasic is not Multimodal. It was easy to add 1 more or-logic. Thus, I got the PR green but ended up confirming its terribly wrong. 

So I took the pain in this PR, to get rid of `is_multimodal_type`. So in macros, we don't inspect the type by its name at all!

For this what I thought the most cleanest solution is to bring all the pain into Schema type class, and multimodal will have a schema. i.e, schema type class will know the semantics of multimodal. It's hard to "explain" further, but I am confident of this change - except for some namings. 

With these changes macro code is reduced by ending up [deleting](https://github.com/golemcloud/golem/pull/2326/commits/21e81a7f7f115007d6824a66f47fe607cc9c9234) a lot of branching logic between default and multimodal and all the integration tests work too

In short, the addition of new code in this PR, is outside macros, and deleting code from macros

### DIFF
--- a/sdks/rust/golem-rust/src/agentic/unstructured_binary.rs
+++ b/sdks/rust/golem-rust/src/agentic/unstructured_binary.rs
@@ -147,7 +147,7 @@ pub trait AllowedMimeTypes {
 
 impl AllowedMimeTypes for String {
     fn all() -> &'static [&'static str] {
-        &["*"] // in the schema, it would imply you can pass any mime type
+        &[]
     }
 
     fn from_string(mime_type: &str) -> Option<Self> {


vigoo (commented):
The changes look good.
- I am not sure I like the name (`Multimodal<>` for the fully custom, `MultimodalBasic` for the simple one). Maybe the `MultimodalBasic` should be called `Multimodal` and the other one something else? 
- Should we have a third one, that is like `MultimodalBasic` but also parametrized with exactly one arbitrary user type? So it's `text | binary | t`? 

@jdegoes what do you think?

afsalthaj:
@vigoo  I like the idea of Multimodal for the basic but then I am not sure what should be the name of Multimodal<T>..Let's wait for @jdegoes

sdks/rust/golem-rust/src/agentic/multimodal.rs
```diff
@@ -167,9 +204,451 @@ impl<T: MultimodalSchema> Multimodal<T> {
...
+
+    fn to_wit_value(self) -> Result<WitValue, String>
+    where
+        Self: Sized,
+    {
+        match self {
+            CustomModality::Basic(basic) => basic.to_wit_value(),
+            CustomModality::Custom(custom) => custom.to_wit_value(),
+        }
+    }
```
afsalthaj, line 653: The boiler plate of having to define schema can actually be avoided if use type aliases for new variants of advanced multimodal. However, we lose clean API boundaries and user will end up having to see the advanced variants especially when relying on ide to auto complete type. I don't think they ever need to get exposed to advanced variants when they are using any of these simple variants.

sdks/rust/golem-rust/src/agentic/unstructured_binary.rs
```diff
@@ -138,15 +138,29 @@ impl<T: AllowedMimeTypes> UnstructuredBinary<T> {
...
     where
         Self: Sized;
 
-    fn to_string(&self) -> &'static str;
+    fn to_string(&self) -> String;
+}
+
+impl AllowedMimeTypes for String {
+    fn all() -> &'static [&'static str] {
+        &["*"] // in the schema, it would imply you can pass any mime type
```
vigoo: Where is this handled? (And represented as an empty restriction in the schema, instead of this string?)

afsalthaj: No this should be empty. Will fix it.

afsalthaj: Fixed: it was a carry over of an initial code before raising this PR.

afsalthaj: @vigoo  https://github.com/golemcloud/golem/blob/main/sdks/rust/golem-rust/src/agentic/unstructured_binary.rs#L150

PR #2328, opened 2025-11-19
Load and save snapshot

### LINKED ISSUE #2305: Save/load snapshot support in code-first Rust
Similar to the TS SDK, we should have a way to implement save/load snapshot on the agent level, as methods on the agent implementation struct. 
The exact details of what's the best way to do this in the Rust SDK is part of the ticket. 

Defining these should be optional for the user (but the SDK itself should always export the save/load functions, like the TS SDK)

afsalthaj:
Fixes #2305 

Resolves #2305 

image

(during implementation of agent, those 2 functions available to optionally implement)

### DIFF
--- a/.github/workflows/windows-daily-build.yaml
+++ b/.github/workflows/windows-daily-build.yaml
@@ -35,4 +35,4 @@ jobs:
         env:
           QUIET: true
         run: cargo make --profile ci build-all-bins-including-tests
-        timeout-minutes: 60
+        timeout-minutes: 120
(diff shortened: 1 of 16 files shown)


sdks/rust/golem-rust/src/agentic/agent_impl.rs
```diff
@@ -83,7 +83,31 @@ impl LoadSnapshotGuest for Component {
 
 impl SaveSnapshotGuest for Component {
     fn save() -> Vec<u8> {
-        vec![]
+        with_agent_instance_async(|resolved_agent| async move {
+            // TODO; just referring to TS, verify
+            let agent_id_bytes = resolved_agent.agent_id.agent_id.as_bytes();
```
vigoo: No need to save the agent-id in the snapshot, I removed it from the TS one too. It was written when we did not have a agent-id=worker-name correspondence.

PR #2350, opened 2025-11-24
Add AI in golem rust sdk

### LINKED ISSUE #2304: Add the Golem AI libraries to the Rust SDK
Similar to the TS SDK, we need to add all the Golem AI libraries to the WIT world implemented by the Rust SDK, and reexport the generated bindings as submodules. This way the users can use these libraries without any custom WIT machinery.

afsalthaj:
Fixes #2304 

The following works now:

```rust
use golem_rust::golem_ai::golem::llm::llm::Config;

#[agent_definition]
trait FooAgent {
    fn new(llm_config: Config) -> Self;
    fn echo_llm(&mut self, llm_config: Config) -> Config;
    async fn remote_agent(&self) -> Config;
}

struct FooAgentImpl {
    llm_config: Config,
}

#[agent_implementation]
impl FooAgent for FooAgentImpl {
    fn new(llm_config: Config) -> Self {
        FooAgentImpl { llm_config }
    }

    fn echo_llm(&mut self, llm_config: Config) -> Config {
        llm_config
    }

    async fn remote_agent(&self) -> Config {
        let client = RemoteClient::get(self.llm_config.clone());
        client.get_config().await
    }
}


#[agent_definition]
trait Remote {
    fn new(llm_config: Config) -> Self;
    fn get_config(&self) -> Config;
}
```

### Schema derivation for golem AI types

* We need to derive `Schema` for wit generated types mainly for users to AI types. This is how the above constructor or method parameters that use llm.Config worked. 

* But for this to work,  `Config` needs schema instance, and this can be configured as one of the options of wit-bindgen generate function. https://docs.rs/wit-bindgen/latest/wit_bindgen/macro.generate.html#options-to-generate. Hence we listed  `[Schema]`  in `additonal_derives`. But this alone didn't make it work. 

* Note macro generations refers to `golem_rust` types is as if users always add `golem_rust` as a crate dependency. But in the above case, the `generate` is expanded at the golem_rust itself and this will fail with a compile time error

* One of the tricks could  be define `mod golem_rust` re-exporting its own types.This works, but breaks when the implementation are expanded at modules at deep nested levels.

* Second solution is to allow us to configure derivation to either put `crate` or `golem_rust` depending on the context. But having to change the derivation schema (user facing API) is bad. And that's when learned about [proc-macros-crate](https://crates.io/crates/proc-macro-crate) which helps exactly the problem. 

* It allows us to configure the identifier of the golem_rust crate, depending on how it is used - such as if the derive happens within the golem_crate itself, or user has golem_crate has an external crate, or cargo rename or in workspace etc. 

### Smaller things that I need some inputs for this PR

* Implementation of IntoValue and FromValueAndType for `WitValue`, `WitType` etc. I don't know if these are needed (but compiler forces us to give an implementation).  I panicked for things like get the wit-type of wit-type etc.

### DIFF
--- a/sdks/rust/golem-rust/src/agentic/multimodal.rs
+++ b/sdks/rust/golem-rust/src/agentic/multimodal.rs
@@ -465,7 +465,7 @@ impl<T: Schema> MultimodalCustom<T> {
     /// Create a Multimodal input data for agent functions with basic types: Text and Binary.
     ///
     /// # Example
-    /// ```
+    /// ```ignore
     /// use golem_rust::agentic::*;
     /// use golem_rust::MultimodalSchema;
     /// use golem_rust::Schema;


sdks/rust/golem-rust/src/lib.rs
```diff
@@ -148,6 +148,51 @@ pub mod golem_agentic {
...
+
+#[cfg(feature = "export_golem_ai")]
+pub mod golem_ai {
+    use wit_bindgen::generate;
+    generate!({
+        path: "wit",
+        world: "golem-ai",
+        generate_all,
+        generate_unused_types: true,
+        additional_derives: [crate::FromValueAndType, crate::IntoValue],
```
afsalthaj: Doesn't really work, because the tokens of this refers to `golem_rust::agent::...` and these instances reside deep inside modules `golem::agent::common::...`  and Rust always resolves it to an external crate, even if at the root we report the current crate as "golem_rust" somehow. This is what is pending...

sdks/rust/golem-rust/src/value_and_type/mod.rs
```diff
@@ -556,3 +556,94 @@ impl<K: FromValueAndType + Eq + Hash, V: FromValueAndType> FromValueAndType for
         Ok(HashMap::from_iter(items))
     }
 }
+
+// golem_wasm instances
```
afsalthaj, line 560: Implementation of IntoValue and FromValueAndType for WitValue, WitType etc. I don't know if these are really needed  for users (but compiler forces us to give an implementation). I panicked for things like get the wit-type of wit-type etc. Obviously if users directly use these types as method or constructor parameters, it will fail. Happy to implement, but felt confusing.

PR #2352, opened 2025-11-24
Add support for golem AI in rust SDK

### LINKED ISSUE #2304: Add the Golem AI libraries to the Rust SDK

afsalthaj:
Fixes #2304 

The following works now:

```rust
use golem_rust::golem_ai::golem::llm::llm::Config;

#[agent_definition]
trait FooAgent {
    fn new(llm_config: Config) -> Self;
    fn echo_llm(&mut self, llm_config: Config) -> Config;
    async fn remote_agent(&self) -> Config;
}

struct FooAgentImpl {
    llm_config: Config,
}

#[agent_implementation]
impl FooAgent for FooAgentImpl {
    fn new(llm_config: Config) -> Self {
        FooAgentImpl { llm_config }
    }

    fn echo_llm(&mut self, llm_config: Config) -> Config {
        llm_config
    }

    async fn remote_agent(&self) -> Config {
        let client = RemoteClient::get(self.llm_config.clone());
        client.get_config().await
    }
}


#[agent_definition]
trait Remote {
    fn new(llm_config: Config) -> Self;
    fn get_config(&self) -> Config;
}
```

### Schema derivation for golem AI types

* We need to derive `Schema` for wit generated types mainly for users to AI types. This is how the above constructor or method parameters that use llm.Config worked. 

* But for this to work,  `Config` needs schema instance, and this can be configured as one of the options of wit-bindgen generate function. https://docs.rs/wit-bindgen/latest/wit_bindgen/macro.generate.html#options-to-generate. Hence we listed  `[Schema]`  in `additonal_derives`. But this alone didn't make it work. 

* Note macro generations refers to `golem_rust` types is as if users always add `golem_rust` as a crate dependency. But in the above case, the `generate` is expanded at the golem_rust itself and this will fail with a compile time error

* One of the tricks could  be define `mod golem_rust` re-exporting its own types.This works, but breaks when the implementation are expanded at modules at deep nested levels.

* Second solution is to allow us to configure derivation to either put `crate` or `golem_rust` depending on the context. But having to change the derivation schema (user facing API) is bad. And that's when learned about [proc-macros-crate](https://crates.io/crates/proc-macro-crate) which helps exactly the problem. 

* It allows us to configure the identifier of the golem_rust crate, depending on how it is used - such as if the derive happens within the golem_crate itself, or user has golem_crate has an external crate, or cargo rename or in workspace etc. 

### Smaller things that I need some inputs for this PR

* Implementation of IntoValue and FromValueAndType for `WitValue`, `WitType` etc. I don't know if these are needed (but compiler forces us to give an implementation).  I panicked for things like get the wit-type of wit-type etc.

### DIFF
--- a/.github/workflows/windows-daily-build.yaml
+++ b/.github/workflows/windows-daily-build.yaml
@@ -10,9 +10,9 @@ permissions:
 
 jobs:
   build-windows:
-    runs-on: windows-latest
+    runs-on: blacksmith-16vcpu-windows-2025
     env:
-      CARGO_BUILD_JOBS: 5
+      CARGO_BUILD_JOBS: 10
     steps:
       - name: Checkout
         uses: actions/checkout@v5
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12832,9 +12832,9 @@ dependencies = [
 
 [[package]]
 name = "wasm-rquickjs"
-version = "0.0.20"
+version = "0.0.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "562779248cf2a46fca8382069ba25ad2ba961b8e4b76c9f8e462429563be607c"
+checksum = "41878e9c938687b9a6a3bc9b784a52f9fcfeb39ad6646039df112c6a8ddc0c26"
 dependencies = [
  "anyhow",
  "camino",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -322,7 +322,7 @@ walkdir = "2.5.0"
 warp = "0.3.7"
 wasm-encoder = "0.235.0"
 wasm-metadata = { version = "0.235.0", features = ["oci"] }
-wasm-rquickjs = "0.0.20"
+wasm-rquickjs = "0.0.21"
 wasm-wave = "0.235"
 wasmparser = "0.229.0"
 wasmprinter = "0.229.0"
--- a/cli/golem-cli/test-data/ts-code-first-snippets/main.ts
+++ b/cli/golem-cli/test-data/ts-code-first-snippets/main.ts
@@ -1,4 +1,4 @@
-import {BaseAgent, Result, agent, UnstructuredText, UnstructuredBinary, WithRemoteMethods, Multimodal} from '@golemcloud/golem-ts-sdk';
+import {BaseAgent, Result, agent, UnstructuredText, UnstructuredBinary, WithRemoteMethods, MultimodalAdvanced} from '@golemcloud/golem-ts-sdk';
 
 import * as Types from './model';
 import {
@@ -27,8 +27,8 @@ import {
 } from './model';
 
 
-export type InputText = { val: string };
-export type InputImage = Uint8Array;
+export type InputText = { val: string, tag: "text" };
+export type InputImage = { val: Uint8Array; tag: "image" };
 
 @agent()
 class FooAgent extends BaseAgent {
@@ -244,7 +244,7 @@ class FooAgent extends BaseAgent {
         return await this.barAgent.funUnstructuredBinary(unstructuredText);
     }
 
-    async funMultimodal(multimodal: Multimodal<InputText | InputImage>): Promise<string> {
+    async funMultimodal(multimodal: MultimodalAdvanced<InputText | InputImage>): Promise<string> {
         return await this.barAgent.funMultimodal(multimodal);
     }
 
@@ -415,7 +415,7 @@ class BarAgent extends BaseAgent {
         return "foo"
     }
 
-    async funMultimodal(multimodal: Multimodal<InputText | InputImage>): Promise<string> {
+    async funMultimodal(multimodal: MultimodalAdvanced<InputText | InputImage>): Promise<string> {
         return "foo"
     }
 
--- a/cli/golem-cli/tests/app/agents.rs
+++ b/cli/golem-cli/tests/app/agents.rs
@@ -620,7 +620,7 @@ async fn test_ts_code_first_with_rpc_and_all_types() {
     run_and_assert(&ctx, "fun-unstructured-binary", &["url(\"foo\")"]).await;
 
     // Multimodal
-    run_and_assert(&ctx, "fun-multimodal", &["[input-text({val: \"foo\"})]"]).await;
+    run_and_assert(&ctx, "fun-multimodal", &["[text(\"foo\")]"]).await;
 
     // Union that has only literals
     run_and_assert(&ctx, "fun-union-with-only-literals", &["bar"]).await;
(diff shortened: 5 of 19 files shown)


sdks/rust/golem-rust/src/value_and_type/mod.rs
```diff
@@ -556,3 +556,94 @@ impl<K: FromValueAndType + Eq + Hash, V: FromValueAndType> FromValueAndType for
...
+            .string()
+            .ok_or_else(|| "Expected ComponentId as string".to_string())
+            .and_then(|s| {
+                s.parse::<golem_wasm::ComponentId>()
+                    .map_err(|e| format!("Failed to parse ComponentId: {}", e))
+            })
+    }
+}
+
+impl IntoValue for ValueAndType {
```
vigoo, line 623: These are wrong (as you mention in the PR description). WitValue and WitType and `ValueAndType` are valid WIT types, defined in the golem:rpc/types WIT package. These instances have to encode them like that.

You just have to describe the (quite complex) wit types with a `TypeNodeBuilder`. Here is how it's defined on the host side:

https://github.com/golemcloud/golem/blob/main/golem-wasm/src/value_and_type/into.rs#L567-L572
https://github.com/golemcloud/golem/blob/main/golem-wasm/src/value_and_type/into.rs#L584-L590
https://github.com/golemcloud/golem/blob/main/golem-wasm/src/value_and_type/into.rs#L687-L728
https://github.com/golemcloud/golem/blob/main/golem-wasm/src/value_and_type/into.rs#L567-L572
https://github.com/golemcloud/golem/blob/main/golem-wasm/src/value_and_type/into.rs#L687-L727

Maybe we can reuse these definitions, if we somehow change `TypeNodeBuilder` to support directly adding a type (not with the builder dsl). I'm not 100% sure.

For the value, of course `WitValue` is already a `WitValue` - but similarly, we need to somehow be able to inject it into the parent builder.
And for `WitType`, we need to convert it to a `WitValue` using the builder, just like in the host side (https://github.com/golemcloud/golem/blob/main/golem-wasm/src/value_and_type/into.rs#L594-L685 etc)

sdks/rust/golem-rust/src/value_and_type/mod.rs
```diff
@@ -556,3 +556,94 @@ impl<K: FromValueAndType + Eq + Hash, V: FromValueAndType> FromValueAndType for
         Ok(HashMap::from_iter(items))
     }
 }
+
+// golem_wasm instances
```
afsalthaj: This is where we need some care before merging the PR

sdks/rust/golem-rust/src/value_and_type/mod.rs
```diff
@@ -556,3 +556,94 @@ impl<K: FromValueAndType + Eq + Hash, V: FromValueAndType> FromValueAndType for
...
+
+impl FromValueAndType for WitValue {
+    fn from_value_and_type(value_and_type: ValueAndType) -> Result<Self, String> {
+        Ok(value_and_type.value)
+    }
+
+    fn from_extractor<'a, 'b>(
+        _extractor: &'a impl WitValueExtractor<'a, 'b>,
+    ) -> Result<Self, String> {
+        Err("extractor is not needed to implement FromValueAndType for WitValue".to_string())
```
afsalthaj: Not sure if anyone practically passes `WitValue` or `WitType`  as a agent constructor or method parameter. If they do so, they get runtime error. So.. that's the implication, just to be clear.

sdks/rust/golem-rust/src/value_and_type/mod.rs
afsalthaj: Thanks Vigoo

afsalthaj: Makes sense

vigoo:
One important note about the WIT deps. Should not add them to the root `wit/deps`. That's our root WIT packages, this way you create a circular dependency because `golem-ai` fetches the wit deps from here. 
These `golem-ai-*` wit packages have to be only added to the Rust SDK (similary how they are added to the TS one in the golem-agentic repo).

afsalthaj:
I am dealing with errors related to generic templates. The test `adding_and_changing_rpc_deps_retriggers_build` fails. 

The last one is , and I have no clue why generic apps have to fail. `golem_rust` has macros enabled by default. golem_rust depends on 


```
 use golem_rust_macro::{FromValueAndType, IntoValue};
> golem-cli - stderr:    |     ^^^^^^^^^^^^^^^^ use of unresolved module or unlinked crate `golem_rust_macro`
> golem-cli - stderr:    |
> golem-cli - stderr:    = help: if you wanted to use a crate named `golem_rust_macro`, use `cargo add golem_rust_macro` to add it to your `Cargo.toml`
> golem-cli - stderr:
> golem-cli - stderr: For more information about this error, try `rustc --explain E0432`.
> golem-cli - stderr: error: could not compile `golem-rust` (lib) due to 1 previous error
```

vigoo:
> I am dealing with errors related to generic templates. The test `adding_and_changing_rpc_deps_retriggers_build` fails.
> 
> The last one is , and I have no clue why generic apps have to fail. `golem_rust` has macros enabled by default. golem_rust depends on
> 
> ```
>  use golem_rust_macro::{FromValueAndType, IntoValue};
> > golem-cli - stderr:    |     ^^^^^^^^^^^^^^^^ use of unresolved module or unlinked crate `golem_rust_macro`
> > golem-cli - stderr:    |
> > golem-cli - stderr:    = help: if you wanted to use a crate named `golem_rust_macro`, use `cargo add golem_rust_macro` to add it to your `Cargo.toml`
> > golem-cli - stderr:
> > golem-cli - stderr: For more information about this error, try `rustc --explain E0432`.
> > golem-cli - stderr: error: could not compile `golem-rust` (lib) due to 1 previous error
> ```

This means that you are incorrectly implementing the `macro` feature. Because it is enabled by default, it compiles with default feature flags, but does not compile when the macro feature is off, like in the generic template:
```
golem-rust = { GOLEM_RUST_VERSION_OR_PATH, features = ["export_load_snapshot", "export_save_snapshot", "export_oplog_processor"] }
```

Make sure all feature combinations (including empty feature set) compiles.

afsalthaj:
https://github.com/golemcloud/golem/pull/2352#issuecomment-3588370404


No that was unused bad import in sdk which was not picked by any clippy.. just removed it and it worked. All good

PR #2355, opened 2025-11-25
fixes in rust code first based on testing and doc updates

afsalthaj:
I was testing everything just to be sure when [documenting](https://github.com/golemcloud/docs/pull/184) Rust, and found some residual issues

### Fixes

All fixes are user facing. 

1. `agent definitions` and `agent implementations` can be in separate modules and it always worked,  **_but if they are in separate crates (obviously within the same workspace) it didn't work_**.  The exports never showed agent methods but simply dynamic methods, indicating there is an agent-type registration issue. This is fixed properly, and ensured users don't need to do anything to get this going.

2. Making sure `prompt` and `description` works. `prompt` was always empty, and `description` was empty because of bug.** . Vscode don't complain about unknown attributes too, but Rustrover does. This is fixed too.

3. Type mismatch when dealing with host types (expected golem::api::PromiseId, got something else). This was simple - just making it consistent with other modules with wit-bindgen generate in golem_rust where types are properly mapped.

4. Avoid any compiler warnings for users when the agent trait is public and has async functions. 

5. The rust code in the template was not consistent with typescript. Fixed so that the documentation site  has same explanation for the code in Rust and Typescript in quick-start

### DIFF
--- a/cli/golem-cli/tests/app/plugins.rs
+++ b/cli/golem-cli/tests/app/plugins.rs
@@ -426,7 +426,7 @@ async fn plugin_installation_test2(_tracing: &Tracing) {
             flag::YES,
             cmd::AGENT,
             cmd::INVOKE,
-            &format!("test:rust1/counter({{ id: \"{uuid}\" }})"),
+            &format!("test:rust1/counter-agent(\"{uuid}\")"),
             "increment",
         ])
         .await;


afsalthaj:
image

sdks/rust/golem-rust-macro/src/agentic/agent_definition_impl.rs
```diff
@@ -95,29 +95,26 @@ pub fn agent_definition_impl(attrs: TokenStream, item: TokenStream) -> TokenStre
...
             } = agent_type_with_remote_client;
 
-            let register_fn_name = get_register_function_ident(&item_trait);
-
-            // ctor_parse! instead of #[ctor] to avoid dependency on ctor crate at user side
-            // This is one level of indirection to ensure the usage of ctor that is re-exported by golem_rust
-            let register_fn = quote! {
-                ::golem_rust::ctor::__support::ctor_parse!(#[ctor]fn #register_fn_name() {
+            let my_registration_function: syn::TraitItem = syn::parse_quote! {
+                fn __register_agent_type() {
```
afsalthaj, line 99: fixing registration issues

PR #2363, opened 2025-11-27
Interrupt improvements

### LINKED ISSUE #2172: Agents deadlocked on "self" calls are not deletable
If an agent calls itself, that results in deadlock, which is the specified and expected behaviour for now, but users cannot delete these workers to undo the mistake.

vigoo:
Resolves #2172 

- Make sure to interrupt running workers on delete-worker call
- Interrupt now broadcasts a signal that can be used in potentially long-running host function implementations to race with
- Using that in `poll` and `invoke-and-await` for now
- Worker executor test doing a long RPC call and deleting the worker while it's running

Extra:
- Updated to latest wasm-rquickjs and TS SDK
- Using bigger executors for the daily windows CI job (as it always fails with oom)
- A fix in how stderr/warn/log entries are collected for agent errors

### DIFF
--- a/golem-worker-executor/src/durable_host/mod.rs
+++ b/golem-worker-executor/src/durable_host/mod.rs
@@ -1279,7 +1279,7 @@ impl<Ctx: WorkerCtx> StatusManagement for DurableWorkerCtx<Ctx> {
     fn check_interrupt(&self) -> Option<InterruptKind> {
         let execution_status = self.execution_status.read().unwrap();
         match &*execution_status {
-            ExecutionStatus::Interrupting { interrupt_kind, .. } => Some(interrupt_kind.clone()),
+            ExecutionStatus::Interrupting { interrupt_kind, .. } => Some(*interrupt_kind),
             _ => None,
         }
     }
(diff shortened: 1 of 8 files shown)


golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -1321,16 +1321,22 @@ impl<Ctx: WorkerCtx> StatusManagement for DurableWorkerCtx<Ctx> {
         match current_execution_status {
             ExecutionStatus::Running { .. } => {}
             ExecutionStatus::Suspended { agent_mode, .. } => {
+                let (tx, _) = tokio::sync::broadcast::channel(128);
```
noise64, line 1324: can this queue size be bigger then 1 in practice? if it can, what happens when it is full, will interrtuption requests block?

vigoo: you can flood the server with interrupt requests and then I think there is a chance to have more than 1. But as soon as the execution status switches to Interrupting (which is basically the next step) it is no longer possible so I would not worry about overflowing this.

Issue #2353, opened 2025-11-25
Docker Compose Failed to Start


khajavi:
```
milad@nixos~/t/g/docker-compose> docker compose -f compose.yaml up
[+] Running 8/8
 ✔ golem-cloud-service Pulled                                                                                                                3.3s
 ✔ golem-shard-manager Pulled                                                                                                                3.4s
 ✔ golem-component-service Pulled                                                                                                            3.3s
 ✔ golem-worker-executor Pulled                                                                                                              3.1s
 ✔ golem-component-compilation-service Pulled                                                                                                3.4s
 ✔ golem-debugging-service Pulled                                                                                                            3.4s
 ✔ golem-worker-service Pulled                                                                                                               3.4s
 ✔ router Pulled                                                                                                                             3.2s
[+] Building 0.0s (0/0)                                                                                                            docker:default
[+] Running 3/0
 ✔ Container docker-compose-postgres-1             Created                                                                                   0.0s
 ✔ Container docker-compose-redis-1                Running                                                                                   0.0s
 ✔ Container docker-compose-golem-shard-manager-1  Running                                                                                   0.0s
Attaching to golem-cloud-service-1, golem-component-compilation-service-1, golem-component-service-1, golem-debugging-service-1, golem-shard-manager-1, golem-worker-executor-1, golem-worker-service-1, postgres-1, redis-1, router-1
postgres-1                             | Error: in 18+, these Docker images are configured to store database data in a
postgres-1                             |        format which is compatible with "pg_ctlcluster" (specifically, using
postgres-1                             |        major-version-specific directory names).  This better reflects how
postgres-1                             |        PostgreSQL itself works, and how upgrades are to be performed.
postgres-1                             |
postgres-1                             |        See also https://github.com/docker-library/postgres/pull/1259
postgres-1                             |
postgres-1                             |        Counter to that, there appears to be PostgreSQL data in:
postgres-1                             |          /var/lib/postgresql/data (unused mount/volume)
postgres-1                             |
postgres-1                             |        This is usually the result of upgrading the Docker image without
postgres-1                             |        upgrading the underlying database using "pg_upgrade" (which requires both
postgres-1                             |        versions).
postgres-1                             |
postgres-1                             |        The suggested container configuration for 18+ is to place a single mount
postgres-1                             |        at /var/lib/postgresql which will then place PostgreSQL data in a
postgres-1                             |        subdirectory, allowing usage of "pg_upgrade --link" without mount point
[...]

PR #2370, opened 2025-11-29
docker example - variables for postgres, redis, nginx image versions

justcoon:
fixes: #2353

added env variables for docker image versions for: 
- redis
- postgres 
- nginx

to avoid situation like reported in #2353, where new major version of postgres was released (18), and related setup is not compatible with previous major one

NOTE: GOLEM_IMAGES_VERSION has currently value: latest, this may have similar problem, ideally there should be specific version, which should be updated with each release there


vigoo (approved):
Nice, thanks!

Issue #2371, opened 2025-12-01
Code first Rust SDK: Schema derivation panicked for recursive structures


afsalthaj:

afsalthaj:
One solution done [here](https://github.com/golemcloud/golem/pull/2405) here was to define a generic type which is a culmination of wit-type and wit-value into 1 structure and use it as a generic representation (from which encoding and decoding was possible) for all recursive types.   But it repeats "wit-value" like things and was pointed out in PR review comments. Invoking from outside was also dynamic in this solution - which was my concern. 

Another solution - was  done [here](https://github.com/golemcloud/golem/compare/main...recursion) which looks to be more simple compared to first solution.  This involves transforming recursive type to its non recursive version and use schema of the non recursive version. Invoking from outside (say Rib, which strongly holds to wit semantic and hence not possible to write recursive values) was better compared to first solution, because except for the recursive holes, the type looked exactly like the user-written type. May be type safety from outside invocation is a less important thing for now, yet good to point out that this solution was more typesafe than first one. In TS, this can be difficult to achieve though. And it has the same problem of two step encoding and decoding like the first one.

Third solution tried [here]( https://github.com/golemcloud/golem/commit/1ca6bac6061566784a271517658af4d67ae1da7c
) and probably ideal is to re-use as much as current set up of builders and existing wit-value- i.e,  identifying recursive spots within current macros of  `IntoValue` and `FromValueAndType` derivation such that it doesn't produce "recursive limits" at compile time and works correctly.  Finding recursive spots for type generation was kind of ok and done (but not at all clean given the current builder patterns doesn't natively has the concept of recursion stack) and eliminating this error for value building was even tougher, because the builder here is mainly index-first manner.  Mainly in the current abstraction, "For every `T: NodeBuilder`, I can consume self and return exactly `T::Result`, it implies it's a single pass. One is a type level state machine per se, and the other recursion which is mostly at runtime, and I logically felt, this is probably incompatible. May be there are work arounds. But this point, I decided to get Vigoo's help because it touch many parts which he owns and maintain, whereas  vision was to just detect recursion and then piggyback on these instances as before and then ensure to represent these recursive types as wit-type itself across the board.


And finally some less concerning, yet important enough notes:

[...]

PR #2405, opened 2025-12-08
Handle recursive types in Rust agents

afsalthaj:
Handle recursive types in agents, mainly intended to get this going  for golem-timeline, where every internal data is recursive. Typescript to follow later.

* User experience: No new annotations in user side. `[derive(Schema)]` as before.
* A generic type that represents all recursive types which already will have a schema instance.
* Spot recursive types in schema derivation macros and serialise and deserialise to graph and use graphs schema instance. Otherwise, do things as before.
* No new changes into existing functionalities in the last minute before 1.4

Continuing to write some reasonable number of tests


vigoo:
We already have a representation for this, `WitValue`. It was defined exactly to support recursive data types. We also have a relatively normal JSON representation for it, rib support, etc. 
Defining a new one on top of it does not feel right. I suggest to modify this to remove your custom graph type and use WitValue which is the same thing (of having nodes, and referring to them by indexes within the nodes).

afsalthaj:
Closing this due to the following reason

https://github.com/golemcloud/golem/issues/2371#issuecomment-3659878823

(Other related works involved: https://github.com/golemcloud/golem/issues/2371#issuecomment-3652709069

PR #2407, opened 2025-12-08
Atomic CLI cleanups

### LINKED ISSUE #1370: Reserve or escape slash characters for entity names
Names that it affects: account, project, component, worker

Either we disallow them in cloud, or we do escaping in host functions and CLI.

noise64:
Resolves #1370


mschuwalow:
@noise64 Makes sense, already deleted my comment by the time I saw your reply :D

PR #2484, opened 2025-12-10
Separated sqlite dbs for KVS and IS

### LINKED ISSUE #2475: Separate sqlite DB per agent for KVStore+IXStore in local server

vigoo:
Resolves #2475

### DIFF
--- a/cli/golem/src/launch.rs
+++ b/cli/golem/src/launch.rs
@@ -281,11 +281,13 @@ fn worker_executor_config(
         compiled_component_service: CompiledComponentServiceConfig::Enabled(
             CompiledComponentServiceEnabledConfig {},
         ),
-        shard_manager_service: ShardManagerServiceConfig::Grpc(Box::new(ShardManagerServiceGrpcConfig {
-            host: args.router_addr.clone(),
-            port: shard_manager_run_details.grpc_port,
-            ..ShardManagerServiceGrpcConfig::default()
-        })),
+        shard_manager_service: ShardManagerServiceConfig::Grpc(Box::new(
+            ShardManagerServiceGrpcConfig {
+                host: args.router_addr.clone(),
+                port: shard_manager_run_details.grpc_port,
+                ..ShardManagerServiceGrpcConfig::default()
+            },
+        )),
         registry_service: golem_service_base::clients::registry::GrpcRegistryServiceConfig {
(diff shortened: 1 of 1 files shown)


golem-worker-executor/src/services/golem_config.rs
```diff
@@ -500,10 +524,18 @@ impl SafeDisplay for IndexedStorageConfig {
                 let _ = writeln!(&mut result, "sqlite kv-store:");
                 let _ = writeln!(&mut result, "{}", inner.to_safe_string_indented());
             }
+            IndexedStorageConfig::KVStoreMultiSqlite(inner) => {
+                let _ = writeln!(&mut result, "multi-sqlite kv-store:");
+                let _ = writeln!(&mut result, "{}", inner.to_safe_string_indented());
+            }
             IndexedStorageConfig::Sqlite(inner) => {
```
mschuwalow, line 604: Do we still have a usecase for the old single-sqlite mode?

golem-worker-executor/src/storage/keyvalue/multi_sqlite.rs
```diff
@@ -0,0 +1,377 @@
...
+        self.cache
+            .get_or_insert_simple(&db, async move || {
+                Self::init_storage(max_connections, foreign_keys, db_path).await
+            })
+            .await
+    }
+
+    async fn namespace_to_db(&self, namespace: &KeyValueStorageNamespace) -> String {
+        match namespace {
+            KeyValueStorageNamespace::RunningWorkers => "kv-running_workers".to_string(),
```
mschuwalow, line 104: No need to do in this pr, but if we could consider namespacing the running workers by environment. iirc we never need to /should enumerate more than one environment at a time.

golem-worker-executor/src/services/golem_config.rs
vigoo: I don't think so, but I kept the config for it for now to be able to quickly switch back if needed. I guess we can clean it up soon.

golem-worker-executor/src/storage/keyvalue/multi_sqlite.rs
vigoo: Sounds good, but probably not a bottleneck for local server so let's do it separately if needed

PR #2529, opened 2025-12-19
Fix agent-id with complex parameter types

vigoo:
The string representation of the agent-id uses WAVE, so we always need to use the WIT encoding for things like record field names and variant constructor names. The agent metadata, however, for user defined types contains the names as they appear in the original source language (for example camelCase field names coming from TypeScript). This is intentional but the conversion was missing in the AgentId parsing/printing code.


noise64 (approved):

golem-common/src/model/agent/mod.rs
```diff
@@ -983,17 +984,21 @@ pub struct AgentId {
...
     pub phantom_id: Option<Uuid>,
     wrapper_agent_type: String,
+    as_string: String,
 }
 
 impl AgentId {
     pub fn new(agent_type: String, parameters: DataValue, phantom_id: Option<Uuid>) -> Self {
         let wrapper_agent_type = agent_type.to_wit_naming();
-        Self {
+        let mut result = Self {
```
noise64, line 993: could have been a constructor

PR #2572, opened 2026-01-06
Stub out old http-api apis

### LINKED ISSUE #2571: Comment out existing routes tests and redirect cli to a stub api
To simplify development of the new features, it will be simpler to disable the old code for now so we don't need to maintain two parallel implementations.

mschuwalow:
resolves #2571 

requires #2573

This is temporary until 1.5, where these apis/features/tests will instead be implemented using the new code-first routes


cli/golem-cli/src/command_handler/api/definition.rs
```diff
@@ -110,7 +110,7 @@ impl ApiDefinitionCommandHandler {
                         .golem_clients()
                         .await?
                         .environment
-                        .list_deployment_http_api_definitions(
+                        .list_deployment_http_api_definitions_old(
```
afsalthaj: A minor comment:

Can we choose some other name instead of this `old` suffix. Just feeling a bit awkward even if its temporary in our code?

how about `..._legacy`?

vigoo (approved):

Issue #2577, opened 2026-01-09
Improve the PoC agent invocation method


vigoo:
As part of the bridge epic we've created a new agent invocation endpoint, that relies on a new gRPC endpoint in registry service to find the current deployed agent metadata of a given app/env by names. Review the implementation of these new PoC endpoints and make them production ready (introduce caches / new queries if needed)

mschuwalow:
The registry service side is also missing some permission checks [here](https://github.com/golemcloud/golem/pull/2578/files#diff-19ca20f4932f4ef31b42acb725a1ec3b625db7913b8616766b5773ef2a516b3fR379)

The actual users token needs to be sent to the registry service. While it can use AuthCtx::System internally to simplify the resolution logic, it needs to check that the final resolved agent type (via environment ownership + shares) is accessible to the requesting user.

noise64:
while this is done it should be tested:
- with trigger (currently results in cancellation)
- with schedule (currently results in "not implementd")

noise64:
After doing a review check, these are solved now. Closing it

Issue #2459, opened 2025-12-09
Agentic invocation APIs


vigoo:
In this step we change all invocation APIs:

- Worker executor’s gRPC API for invocation
- Worker service’s APIs for invocation
- the golem:rpc host interface

The changes:

- taking agent function names as function names, not the kebab-cased fully qualified wit export names
- taking `DataValue` as input
- returning `DataValue` as output
- Only taking the `agent-id`, without the `component-id`.

<aside>
💡

To make the steps smaller, in this step we will not change yet the fact that invoking agents require a component-id to be specified.

</aside>

Adjusting the SDKs (code-first RPC, generated bridges) and the CLI is part of this  step.

In the worker executor this requires changing the invocation loop to support invoking only a few special built-in WIT exports (`initialize`, `invoke`, `save-snapshot`, `load-snapshot`).

PR #2578, opened 2026-01-09
JS/TS bridge package generator

### LINKED ISSUE #2413: Create the JS bridge generator
The JS bridge generator takes an AgentType as input, and generates a self-contained, valid source code of an NPM package exporting an API to get/create and invoke agents through the public Golem API.

The package should only depend on `fetch` and potentially some shared helper package to avoid redundancy in generated bridge packages for different agents.

There must be a way to configure what environment to target in the generated clients, and this has to be made in a way that when using the bridge for the REPL, we can somehow enforce a specific one "from the outside".

Integrating the bridge generator into build flows / CLI commands is not part of this ticket.

vigoo:
Resolves #2413 

- Implements a NPM package generator for creating an external client for an agent
- Depends on a new PoC agent-invoke API. I've created an initial ticket for improving the implementation of it (#2577) and also we are going to review and change it during the #2459 epic 
- Started using the `AgentTypeName` wrapper type in more places (it already existed but in many places we were just passing raw Strings)

### DIFF
--- a/cli/golem-cli/src/model/agent/test.rs
+++ b/cli/golem-cli/src/model/agent/test.rs
@@ -18,8 +18,8 @@ use golem_common::model::agent::{
     NamedElementSchemas, TextDescriptor,
 };
 use golem_wasm::analysis::analysed_type::{
-    bool, case, chr, field, list, option, r#enum, record, result, s32, str, tuple, u8, unit_case,
-    unit_result, variant,
+    bool, case, chr, field, list, option, r#enum, record, result, s32, str, tuple, u32, u8,
+    unit_case, unit_result, variant,
 };
 
 pub fn single_agent_wrapper_types() -> Vec<AgentType> {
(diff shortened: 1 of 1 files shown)


golem-registry-service/src/services/application.rs
```diff
@@ -237,8 +237,14 @@ impl ApplicationService {
         name: &ApplicationName,
         auth: &AuthCtx,
     ) -> Result<Application, ApplicationError> {
+        tracing::warn!("get_in_account {name} enter, account-id = {}", account_id.0);
```
noise64: i think these were added as "debug", and not really warns (and i remember you mentioned, that we do not properly log auth / api errors, so i guess it was added while debugging that)

vigoo: yes that's a temporary debug line (warn so it is easy to notice in the log ;)

PR #2582, opened 2026-01-12
Use principal instead of auth-context in wit

### LINKED ISSUE #2581: Update auth-context data type in WIT to embed the concept of `Principal`
The current `auth-ctx` structure is very specific, and can be easily perceived as a leak with very less semantic clarity, even though it is the least jump required to solve the problem of including authentication of code first routes. Please find the details of the proposed changes as given below. These changes are already agreed within the team internally.

* Users should be exposed with a structure which makes sense "all the time" and not something which becomes "valid" only when they annotate with endpoint details
* This structure is basically `principal` (auth context is more of ceremony which may or may not contribute to the principal, who is basically "who gave me the job". The structure of principal can be a variant (more developer friendly), or a flat record simply depicting what a principal is (more friendly to configuring things in CLI easily). I personally don't have much opinion about this on how it should look in WIT. 
However, as John recommended, anything we configure in CLI in future related to `Principal` should be user friendly.  

```rust

allowed-issuers:
  - https://accounts.google.com/
  - https://mycompany.okta.com/
  - golem:agent
  - golem:user
  # omit golem:system to deny

```

and not

```rust
allowed-identities:
  oidc:
    - google
    - okta
  agents: true
  golem-users: true
  system: false
```

* Proposed structure in WIT (@mschuwalow or @afsalthaj   can choose whatever we are comfortable with as far as we stick on to how it should look like in CLI). Please make sure to read through the implications of choosing either of it instead of quickly loving Variant :) 

     *  There are 2 options, one is a variant, and the other is a unified structure. The unified structure reflects the  idea that these variants may get converged in near future. Example: `agent-identity` may have email in future. Or agent-to-agent permission model can reuse claims and so on.
     * At this point, we are repeating most of the fields in all variants, and usual resolution of such a data representation in any language leads to unified structure. 
     * The SDK will always be exposing a variant (in Rust and TS) and the only reason is it allows pattern matching on the issuer (the unified structure has issuer as a string, and the variant structure has variant terms representing the issuer). 
     * Given WIT is not directly exposed to users, these can be changed. Just that we should just choose a structure that doesn't end up requiring changing a lot of internals in future. Also, regardless of what we chooser in WIT, we  should ensure what we expose in CLI-manifest/SDK doesn't go through too many changes and sticks on to the above examples. See above example of how CLI manifest should look like regardless of the structure we choose internally.
     * Not a matter of this task, but SDK should expose a helper to its `Principal` called `issuer`, which produce the corresponding `issuer `string`  explained below

```rust

  variant principal {
      oidc(oidc-identity), // issuer string is provider URL
      agent(agent-id), // issuer string is golem:agent:agent-id, or tenant specific like golem:agent:jdgoes:agent-id
      golem-user(golem-user-identity), // golem:user:account-id
      system // the issuer is golem:system
  }

  record oidc-identity { 
    sub:      string,
    provider: string, // provider url
    email:    string,
    name:     string,

    email-verified: option<bool>,
    given-name:     option<string>,
    family-name:    option<string>,
    picture:        option<string>,
    username:       option<string>,

    claims: string,
  }

  record agent-identity { 
    agent-id: agent-id, // note that the issuer string is `golem:agent:agent-id` or `golem:agent:jdgoes:agent-id`
  }

  record golem-user-identity { 
    account-id: account-id // note that the issuer string is `golem:user:account-id`
  }


```

Unified Representation (see the bullet points above). This is not the structure exposed in SDK, but could be the one in WIT, given the fields in each type of identity may converge. We should quickly choose either one of it.

```rust

record caller-identity {
  sub: string,
  issuer: string,
  email: option<string>,
  name: option<string>,
  email-verified: option<bool>,
  given-name: option<string>,
  family-name: option<string>,
  picture: option<string>,
  username: option<string>,
  claims: string,
}
```

afsalthaj:
resolves #2581


afsalthaj:
I think we will change this to variant structure @mschuwalow  as we discussed

cli/golem-cli/agent_wrapper/guest.mbt
```diff
@@ -4,18 +4,18 @@
 pub fn initialize(
   agent_type : String,
   input : @common.DataValue,
-  auth : Option[@common.AuthContext],
+  principal : @common.Principal,
```
afsalthaj, line 7: 👍

cli/golem-cli/agent_wrapper/guest.mbt
```diff
@@ -4,18 +4,18 @@
...
-  @guest.initialize(agent_type, input, auth)
+  @guest.initialize(agent_type, input, principal)
 }
 
 ///|
 pub fn invoke(
   method_name : String,
   input : @common.DataValue,
-  auth : Option[@common.AuthContext],
+  principal : @common.Principal,
```
afsalthaj, line 16: 👍

cli/golem-cli/wit/deps/golem-agent/common.wit
```diff
@@ -100,21 +100,36 @@ interface common {
     required: bool,
   }
 
-  record auth-context {
+  variant principal {
+    oidc(oidc-principal),
+    agent(agent-principal),
+    golem-user(golem-user-principal),
+    anonymous
+  }
```
afsalthaj, line 108: 👍

cli/golem-cli/wit/deps/golem-rpc/wasm-rpc.wit
```diff
@@ -17,6 +17,11 @@ interface types {
     uuid: uuid,
   }
 
+  /// Represents a Golem account
+  record account-id {
+      uuid: uuid
+  }
```
afsalthaj, line 23: wasn't it already existing? Why is this coming up as a new thing?

golem-api-grpc/proto/golem/component/agent.proto
```diff
@@ -221,15 +223,40 @@ message AgentHttpAuthDetails {
...
   string sub = 1;
-  string provider = 2;
-  string email = 3;
-  string name = 4;
+  string issuer = 2;
+
+  optional string email = 3;
+  optional string name = 4;
   optional bool email_verified = 5;
+
```
afsalthaj, line 242: 👍

sdks/ts/packages/golem-ts-sdk/types/golem_agent_common.d.ts
```diff
@@ -73,18 +75,40 @@ declare module 'golem:agent/common' {
...
+    tag: 'agent'
+    val: AgentPrincipal
+  } |
+  {
+    tag: 'golem-user'
+    val: GolemUserPrincipal
+  } |
+  {
+    tag: 'anonymous'
+  };
```
afsalthaj, line 111: 👍

afsalthaj:
Thanks @mschuwalow  for the updates

Issue #2415, opened 2025-12-08
Automatic bridge generation based on manifest


vigoo:
By a simple switch (like `enable-bridge-generation: true`) we should be able to automatically run the bridge generator for all agent types within the unit that has it enabled for (whole app, per component, per agent-type etc). When enabled, _building_ the bridges should be part of `golem build` with proper up-to-date checks etc.

Issue #2414, opened 2025-12-08
Create a CLI command to run a bridge generator


vigoo:
An explicit command to generate bridge for a specific agent type in a specific language into a specific target directory:

```
golem-cli bridge-gen MyAgent  —language typescript ./my-agent-bridge-ts
```

The command must compile all components and extract all agent types to find the provided agent types by name.

Issue #2201, opened 2025-10-14
Cache extracted agent types in CLI


vigoo:
The "extracting agent types" step is slow and reexecuted on every build/deploy etc. operation. We should cache the results based on the component hash and reuse it.

noise64:
note: we _should_ cache, but with release build it seems reasonably fast

PR #2583, opened 2026-01-12
Bridge Generator CLI integration

noise64:
Resolves https://github.com/golemcloud/golem/issues/2414
Resolves https://github.com/golemcloud/golem/issues/2415
Resolves https://github.com/golemcloud/golem/issues/2201

#### Adds `golem generate-bridge` command to the CLI, which can be used for "one-time" SDK generation.

```
$ golem generate-bridge --help
Generate bridge SDK(s) for the selected agent(s)

Usage: golem generate-bridge [OPTIONS]

Options:
      --language <LANGUAGE>
          Selects the target language for the generated bridge SDK, defaults to the agent's language
      --component-name <COMPONENT_NAME>
          Optional filter for component names; can be defined multiple times
      --agent-type-name <AGENT_TYPE_NAME>
          Optional filter for agent type names; can be defined multiple times
      --output-dir <OUTPUT_DIR>
          Optional output directory for the generated SDK, when not specified, will use separate temporary directories in the application's directory
```

Without arguments the command will generate SDKs for all agent using their lanugage into golem-temp.

#### Adds bridge section to the manifest, which controls SDK generation as part of the build command:

```yaml
# Generating all ts SDKs 
bridge:
  ts: "*"

# Generating some rust SDKs based on agent name
bridge:
  rust: CounterAgent, DispatcherAgent

# Generating ts SDKs based on component name
bridge:
  ts: app:main

# Customizing output directory:
bridge:
  ts: "*"
  outputDir: bridge-sdk/ts
  
# The above component and agent type names matchers can be mixed, and one can also use YAML lists
bridge:
  ts:
  - CounterAgent
  - app:other-component
```

#### Schema changes: 

https://github.com/golemcloud/golem/pull/2583/commits/3a97252cd9f73be15dc8ef3cbfd310a7cd4536f7

The schema is deployed to S3.

#### Smaller changes and cleanups
- agent type extraction is cached now
- agent types are normalized (order by agent and method name, etc) after extraction:
  - this solves the "phantom agent type" diffs happening on some code changes
  - and also makes agent type related task operations deterministic (e.g. bridge sdk generations)
- some command doc related cleanups
- agent types are now stored behind an RWLock in app context, which enabled using "new style" up-to-date checks everywhere, also some modul cleanups were done aroudn these
- fixed some confliting command flag issues, so planning can be used now we rollbacks too

#### Follow ups
Currently the `rust` generator is in progress, and the `ts` one has some bugs. Once those are finalized and fixed, we can add the bridge sdk generation step to the "build all templates" CLI integ test.

We might also want to add some hints for the bridge generation in the manifest.

And lastly, we could provide customization of the generated package versions, but i think it would be nice not to do it in a static way, rather connect it with the not-yet-done user version support of the application.

### DIFF
--- a/cli/golem-cli/src/app/build/gen_bridge.rs
+++ b/cli/golem-cli/src/app/build/gen_bridge.rs
@@ -15,8 +15,6 @@ use golem_common::model::agent::wit_naming::ToWitNaming;
 use golem_templates::model::GuestLanguage;
 use heck::ToKebabCase;
 use itertools::Itertools;
-use std::collections::BTreeSet;
-use strum::IntoEnumIterator;
 
 pub async fn gen_bridge(ctx: &mut ApplicationContext) -> anyhow::Result<()> {
     let mut targets = vec![];
@@ -84,18 +82,10 @@ pub async fn gen_bridge(ctx: &mut ApplicationContext) -> anyhow::Result<()> {
             }
         }
         None => {
-            for target_language in GuestLanguage::iter() {
-                let Some(lang_sdks) = ctx.application.bridge_sdks().for_language(target_language)
-                else {
-                    continue;
-                };
-
-                let mut matchers = lang_sdks
-                    .agents
-                    .clone()
-                    .into_vec()
-                    .into_iter()
-                    .collect::<BTreeSet<_>>();
+            for (target_language, sdks_targets) in
+                ctx.application.bridge_sdks().for_all_used_languages()
+            {
+                let mut matchers = sdks_targets.agents.clone().into_set();
 
                 if matchers.is_empty() {
                     continue;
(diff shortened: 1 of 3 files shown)


vigoo:
When customizing the output dir:
```
bridge:
  ts: "*"
  outputDir: bridge-sdk/ts
```
will it create a subdirectory per agent-type-name within `bridge-sdk/ts`? how exactly the final name is generated?

noise64:
@vigoo 
after considering multiple options / smartness depending on the number of targets, i decided to make it simple, and always include the agent name in kebab-case as a directory created during the generation.

so when somebody uses the generator command without params, it will result in this:

```
golem-temp
  bridge-sdk
    ts
      counter-agent
      other-agent
    rust
      my-rust-agent
```

In the manifest basically you can customize the `golem-temp/bridge-sdk/ts` and `golem-temp/bridge-sdk/rust` paths.
I think this should be enough for most, and if somebody need more custom setup, they can use the standalone command multiple times.

As I wrote, i tried other options, but always ended up with weird case without always adding the agent-name to the path.

vigoo (approved):

PR #2585, opened 2026-01-13
Allow crud on code-first routes http api deployments

### LINKED ISSUE #2562: Allow using code-first routes in deployments
Create either a new endpoint or adjust the existing apis such that api deployments can use code first routes (by listing agent-types) instead of rib-based api-definitions.

No need for webhooks or advanced features at this point, just enough for the routes to be referenced in a deployment.

mschuwalow:
resolves #2562

### DIFF
--- a/golem-registry-service/src/services/deployment/rib.rs
+++ b/golem-registry-service/src/services/deployment/rib.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Fixme: code-first routes
+
 // use golem_service_base::custom_api::HttpCors;
 // use rib::{Expr, GetLiteralValue, RibCompiler, RibInput};
 
--- a/golem-registry-service/src/services/deployment/write.rs
+++ b/golem-registry-service/src/services/deployment/write.rs
@@ -46,6 +46,7 @@ use rib::RibCompilationError;
 use std::collections::{BTreeMap, HashSet};
 use std::sync::Arc;
 
+// Fixme: code-first routes
 // macro_rules! ok_or_continue {
 //     ($expr:expr, $errors:ident) => {{
 //         match ($expr) {
@@ -524,6 +525,7 @@ impl DeploymentContext {
         Ok(agent_types)
     }
 
+    // Fixme: code-first routes
     // fn validate_http_api_deployments(&self) -> Result<(), DeploymentWriteError> {
     //     let mut errors = Vec::new();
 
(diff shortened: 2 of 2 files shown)


golem-registry-service/src/api/http_api_deployments.rs
```diff
@@ -0,0 +1,445 @@
...
+        //         domain.0,
+        //         auth,
+        //     )
+        //     .instrument(record.span.clone())
+        //     .await;
+
+        // record.result(response)
+    }
+
+    // async fn get_http_api_deployment_in_deployment_internal(
```
noise64, line 342: could you please add a comment todo (eg. `// TODO: code-first-routes`) for cases where there are no unimplemented or todo macros? just to make it a bit easier to find leftovers

noise64: oh, i see you started using `Fixme: code-first routes` in other places, let's just add it consistently

golem-client/build.rs
```diff
@@ -322,6 +322,20 @@ fn generate(yaml_path: PathBuf, out_dir: OsString) {
...
+                "LegacyHttpApiDeployment",
+                "golem_common::model::http_api_deployment_legacy::LegacyHttpApiDeployment",
+            ),
+            (
+                "LegacyHttpApiDeploymentCreation",
+                "golem_common::model::http_api_deployment_legacy::LegacyHttpApiDeploymentCreation",
+            ),
+            (
+                "LegacyHttpApiDeploymentUpdate",
+                "golem_common::model::http_api_deployment_legacy::LegacyHttpApiDeploymentUpdate",
```
afsalthaj, line 337: Just out of curiosity. This is all for legacy, so why this mapping is required now?

golem-registry-service/db/migration/postgres/002_agent_type_http_api_deployment.sql
```diff
@@ -0,0 +1,5 @@
+DELETE FROM deployment_http_api_deployment_revisions;
+DELETE FROM http_api_deployment_revisions;
+DELETE FROM http_api_deployments;
+
+ALTER TABLE http_api_deployment_revisions RENAME COLUMN http_api_definitions TO agent_types;
```
afsalthaj, line 5: So this will be list of of agent_type correct? and it can still be a TEXT as before yes?

mschuwalow: Yes. Sqlite does not support arrays.
As we don't need to join / etc. on this column using a json array is the most convenient option.

PR #2586, opened 2026-01-14
Add mount and http details in TS

afsalthaj:
Fixes #2549 

Merged #2594  here, which fixes #2550

PR #2594, opened 2026-01-15
Http Endpoint details in TS

### LINKED ISSUE #2550: Code First Routes: Create endpoint decorator in TS
- `endpoint` decorator is at agent method level, and it should instantly fail if there is no corresponding `http-mount-details`
- Similar to `AgentDecoratorOptions` , add `HttpEndpointOptions` that replicates the `http-endpoint-details` in WIT.
- It should mainly capture the `http-method` and `path-suffix`
- `path-segment`, `header-variable` etc are already defined as part of mount details, and it should just reuse those types
- There is `agentMethodRegistry.ts` which should now hold the `http-endpoint-details`.
- Add tests in `http.decorator.tests.ts` for `endpoint` decorator.
- Manual test with REPL or invoke showcasing the agent-type methods to hold endpoint details too.

afsalthaj:
Fixes #2550

### DIFF
--- a/sdks/ts/packages/golem-ts-sdk/src/index.ts
+++ b/sdks/ts/packages/golem-ts-sdk/src/index.ts
@@ -26,7 +26,7 @@ export { AgentId } from './agentId';
 export { description } from './decorators/description';
 export { agent, AgentDecoratorOptions } from './decorators/agent';
 export { prompt } from './decorators/prompt';
-export { endpoint } from './decorators/httpEndpoint';
+export { endpoint, EndpointDecoratorOptions } from './decorators/httpEndpoint';
 
 export * from './newTypes/either';
 export * from './newTypes/agentClassName';
@@ -37,8 +37,6 @@ export * from './newTypes/multimodalAdvanced';
 export { Client } from './baseAgent';
 export { AgentClassName } from './newTypes/agentClassName';
 export { TypescriptTypeRegistry } from './typescriptTypeRegistry';
-
-export * from './options';
 
 export * from './host/hostapi';
 export * from './host/guard';
--- a/sdks/ts/packages/golem-ts-sdk/src/options.ts
+++ b/sdks/ts/packages/golem-ts-sdk/src/options.ts
@@ -1,1 +1,0 @@
-import { AgentMode } from 'golem:agent/common';


sdks/ts/packages/golem-ts-sdk/src/decorators/agent.ts
```diff
@@ -82,10 +90,34 @@ import { validateHttpMountWithConstructor } from './http/validation';
...
  * - `mode`: Agent durability mode, either "durable" or "ephemeral" (default: "durable")
  *
- * Example:
+ * ### HTTP Mount Options
+ * Agents can optionally expose an HTTP API using a base mount path. The following options are available:
+ * - `mount`: The base HTTP path to expose agent methods (e.g., `'/api/weather'`). A path can have path variables (example: `'/api/{city}/weather'`),
+ *    or system variables (e.g., `'/api/{agent-type}/status'`) or both.
+ * - `headers`: Default HTTP headers mapped to constructor parameters (e.g., `{ 'X-Api-Key': 'apiKey' }`).
+ *    Note that the value of the header is the name of the constructor parameter to which it maps. In this case `apiKey` is one of the constructor parameters.
+ * -  Note that all the parameters in the constructor must be provided either via header variables or path variables.
```
mschuwalow, line 99: why the exclusion of query parameters here?
The wit also implies that agent-wide query parameter configuration is possible https://github.com/golemcloud/golem/blob/main/wit/deps/golem-agent/common.wit#L42

sdks/ts/packages/golem-ts-sdk/src/decorators/httpEndpoint.ts
```diff
@@ -0,0 +1,205 @@
...
+import { AgentMethodRegistry } from '../internal/registry/agentMethodRegistry';
+import { parsePath } from '../http/path';
+import { parseQuery } from '../http/query';
+
+export type EndpointDecoratorOptions = {
+  get?: string;
+  post?: string;
+  put?: string;
+  delete?: string;
+  custom?: string;
```
mschuwalow, line 33: how does custom work here as we will need both the method name and the path as the value?

Note that we are also using a few http method cases like patch here. Check https://github.com/golemcloud/golem/blob/code-first-route-deployment-compilation/wit/deps/golem-agent/common.wit for the updated list we should support.

sdks/ts/packages/golem-ts-sdk/src/decorators/httpEndpoint.ts
```diff
@@ -0,0 +1,205 @@
...
+ * ```
+ *
+ * ### HTTP Methods
+ * - Specify **one** of `get`, `post`, `put`, `delete`, or `custom`.
+ * - The value of the option is the endpoint path.
+ * - Examples:
+ *   - Simple GET: `{ get: '/status' }`
+ *   - GET with path variables: `{ get: '/rooms/{roomId}/messages/{messageId}' }`
+ * - Path variables must exactly match method parameters.
+ * - No "foreign" path variables are allowed.
```
mschuwalow, line 73: What does "foreign path variables" mean?

sdks/ts/packages/golem-ts-sdk/src/decorators/httpEndpoint.ts
```diff
@@ -0,0 +1,205 @@
...
+ * ```ts
+ * @endpoint({ get: '/secure-data', auth: true })
+ * getSecureData(userId: string) { ... }
+ * ```
+ *
+ * ### Errors / Validation
+ * This decorator will throw an error if:
+ * - No HTTP method is specified.
+ * - Path does not start with `/`.
+ * - Path contains query parameters (`?`) when using the path property.
```
mschuwalow, line 120: what does "Path contains query parameters (`?`) when using the path property." mean?

afsalthaj: I removed error/validation comment.

sdks/ts/packages/golem-ts-sdk/src/decorators/httpEndpoint.ts
afsalthaj: path variables that cannot be `bind` to a parameter in method.
Note that every variable in the path is corresponding to something in method parameter. An error there is better than user giving a typo and then a runtime error to not able to bind to a parameter.

So in the backend. Every variable that you see will be corresponding to a method parameter. We reduce the surface area of the semantics. That's all.

It's the same behaviour with mount (the other PR) - that whatever you see in the mount will be corresponding to the constructor

sdks/ts/packages/golem-ts-sdk/src/decorators/agent.ts
afsalthaj: Note that this is related to base path (and not final endpoint)
I am not sure it's a good idea to mount it. I wanted to ask this to John as well.

Example: base path is `/foo?bar={baz}` and then endpoint (method levvel) configured as `get = /qux`. What does it really mean? `/foo?bar={baz}/baz`.  Not sure if that makes sense.

We can actually allow. This leads to semantics such as if the base path consist of query parameters, then users cannot use any more suffix in the endpoint. 

I wanted to discuss this.  Let's pair up for 2 mins, and take it to John as well, before merging the final PR which is https://github.com/golemcloud/golem/pull/2586.

sdks/ts/packages/golem-ts-sdk/src/decorators/httpEndpoint.ts
afsalthaj: That's true. I didn't worry about the custom now. I will fix that once merged to the original PR.

PR #2595, opened 2026-01-15
Code first route deployment compilation & wit updates

### LINKED ISSUE #2561: Extract and store code-first routes during deployment
The goal of this ticket is to extract code-first routes similarly as rib-based routes during deployment.
The information about code-first routes will be stored as part of the [agent-type](https://github.com/golemcloud/golem/blob/main/golem-common/src/model/agent/mod.rs#L239) once the wit changes are done.

During deployment, all necessary validations are done on all routes in all agent types that are referenced by api deployments.
The resulting routes are flattened into a (domain, sercurity_scheme_id, ?,  route) representation and stored in the db for efficient lookup at routing time.

### LINKED ISSUE #2563: Create a grpc api to lookup deployed code-first routes for a given domain
Similarly to the existing rib-based route support, we need a new (or adjusted) grpc api that allows the worker service to lookup all active routes for a given domain. Make sure this grpc method is reasonably performant as it will be in the hot path of worker service request handling logic.

mschuwalow:
resolves #2561
resolves #2563

Also removes http-api-definition related code from the backend

### DIFF
--- a/sdks/ts/packages/golem-ts-sdk/src/decorators/agent.ts
+++ b/sdks/ts/packages/golem-ts-sdk/src/decorators/agent.ts
@@ -53,7 +53,6 @@ export type AgentDecoratorOptions = {
   mount?: string;
   cors?: string[];
   auth?: boolean;
-  headers?: Record<string, string>;
   webhookSuffix?: string;
 };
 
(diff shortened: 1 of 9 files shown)


cli/golem-cli/wit/deps/golem-agent/common.wit
```diff
@@ -61,9 +61,14 @@ interface common {
...
-    put,
+    head,
     post,
+    put,
     delete,
+    connect,
+    options,
+    trace,
+    patch,
     custom(string),
```
afsalthaj, line 70: I can see you changed wit here.

mschuwalow: Yes. The other ones are debatable, but we at least need to support patch.

afsalthaj (approved):

PR #2608, opened 2026-01-19
Clean up schema discovery and add Principal in TS SDK

### LINKED ISSUE #2551: Code First Routes: Support of AuthContext  in TS SDK and CLI
This ensures SDK really supports authentication context along with CLI manifest support configuring security provider. Whether it is 1 security provider per component, or per agent can be decided. Regardless of whether we support fine grained configuration, a user should be able to set a global security provider during the deployment of  the manifest. 

- From spec: 
”The absence of auth details on a constructor implies access to the agent through HTTP is unsecured” . This is doubtful, as we can have unsecured initialization of an agent but accessing a method may require authenticated.

```jsx
@agent({
    mount: "/chats/{agentId}",
    cors: ["https://app.acme.com", "https://staging.acme.com"],
    auth: true,
    headers: { "X-Request-Id": "requestId" }
})

class AssistantAgent extends BaseAgent {
    // the constructor  don't take any auth-context 
    // Implies initialize is unsecured
    constructor(agentId: string, requestId: string, ctx: AuthContext) {}

    // the method is authenticated since it takes a authentication context
    @endpoint({ get: "/ask?q={input}" })
    async ask(input: string, ctx: AuthContext): Promise<string> {
        // ...
    }
}
```

- If the auth is set to false at either mount level or at method level, and then if constructors or methods use `AuthContext` as one of the parameters, then it should fail.
- If the auth context is set to true, then the manifest should define the security provider name. It could be security provider per agent or for the whole deployment. This can be discussed during implementation
- The registry service should persist this information for the backend to finally support. It is important to see whether we can reuse the existing services and implementation in this space

### LINKED ISSUE #2613: TS SDK Bug: agent can be successfully initialized even if argument size don't match
This is more of JS being ok with accepting less number of arguments than invoked when dynamically invoked.
The bug was captured as part of code first route implementation.

afsalthaj:
Fixes #2551 
Fixes #2613 

While adding support of principal, i made a lot of cleanups. While certain parts was just code moving aaround, certain other parts mainy `schema` module was heavily refactored. And then it was much easier to add support for Principal.

Also note that, I couldn't do all cleaning (so certain code that was moved around might still look to be something that can be improved). Allow me to do follow ups for that, because it took a lot of time for me to figure out how i broke integration tests: https://github.com/golemcloud/golem/actions?query=is%3Afailure+branch%3Afeature_2551 , and finally it is green.

Apart from the whole schema module being refactored, the following specific clean ups was done.

* There was TypeDetails, TypeInfoInternal which are internal representation of a type (a method parameter type, constructor parameter type or the return type). Removing TypeDetails just led to a few changes. We need only 1 internal type.
* TypeInfoInternal now holds WitType as well, which makes it easy to find the ElementSchema, or DataSchema for any multimodal type or generally any return type - makes things easy to comprehend.
* Finally, start a `schema` package and separate out functions that's dedicated to constructor, method input and method output.
* Move `http` module inside `internal`.
* This leads to simply adding 1 more variant to just TypeInfoInternal called Principal and work backwards

### DIFF
--- a/sdks/ts/packages/golem-ts-sdk/src/internal/typeInfoInternal.ts
+++ b/sdks/ts/packages/golem-ts-sdk/src/internal/typeInfoInternal.ts
@@ -162,7 +162,7 @@ export function getReturnTypeDataSchemaFromTypeInternal(
       if (analysed.kind === 'tuple' && analysed.emptyType) {
         return Either.right({
           tag: 'tuple',
-          val: []
+          val: [],
         });
       }
 


afsalthaj:
There are other bugs which is already fixed. The main one was #2613 . the last two things spotted is:

1. For multimodal, more than 1 parameter can exist as far as the rest is `Principal`.
2. Handle constructor parameters or method parameters with `?` during auto injection

sdks/ts/packages/golem-ts-sdk/src/internal/http/validation.ts
```diff
@@ -0,0 +1,200 @@
...
+function collectHttpMountVariables(agentMount: HttpMountDetails): Set<string> {
+  const vars = new Set<string>();
+
+  for (const segment of agentMount.pathPrefix) {
+    if (segment.tag === 'path-variable') {
+      vars.add(segment.val.variableName);
+    }
+  }
+
+  return vars;
```
afsalthaj, line 199: Not sure why this is showing, all files in `http` should have been a file renaming.

sdks/ts/packages/golem-ts-sdk/src/internal/schema.ts
afsalthaj, line 1: It was getting overly complicated, hence refactored the code along with re-organisation

afsalthaj:
1 more integration test failure to fix which is 

```ts
    async funOptionalQMark(param1: string, param2?: number, param3?: string) {
        return self.barAgent(param1, param2, param3)
    }
```

After auto injection, the RPC is failing for these question mark parameters.

sdks/ts/packages/golem-ts-sdk/src/internal/mapping/values/dataValue.ts
```diff
@@ -52,47 +55,85 @@ export type ParameterDetail = {
...
+      const elementsCount = elements.length;
+
+      let dataValueElementIdx = 0;
 
       return Either.all(
-        elements.map((elem, idx) => {
-          switch (elem.tag) {
-            case 'unstructured-text':
-              const parameterDetail = paramTypes[idx];
+        paramTypes.map((parameterDetail) => {
```
afsalthaj, line 77: we mapped data value elements before. This time we are mapping parameter types and then lookup data value elements. This is an important change to get errors

afsalthaj:
While adding support of principal, i made a lot of cleanups. While certain parts was just code moving aaround, certain other parts mainy schema module was heavily refactored. And then it was much easier to add support for Principal.

Also note that, I couldn't do all cleaning (certain bits within schema module too, and some code that you see was just code moved around and might still look to be something that can be improved). Allow me to do follow ups for that, because it took a lot of time for me to figure out how i broke integration tests - to finally - make it green. https://github.com/golemcloud/golem/actions?query=is%3Afailure+branch%3Afeature_2551 , and finally it is green.

sdks/ts/packages/golem-ts-sdk/src/internal/clientGeneration.ts
```diff
@@ -591,12 +605,18 @@ function deserializeRpcResult(
...
+          dataValueText,
+          [
+            {
+              name: 'return-value',
+              type: typeInfoInternal,
+            },
+            // Deserializing rpc result doesn't require principal context,
+            // In this case typeInfoInternal is 'unstructured-text', so Principal type cannot appear here
+          ],
+          { tag: 'anonymous' },
```
afsalthaj, line 618: We can do something about it - to be more clearer. but I have already done a lot of cleanups elsewhere to avoid these kind of situations to add Principal feature, and can clean ups in follow up PR.

Issue #2598, opened 2026-01-16
Ensure unique naming of conflicting type names in bridge SDK generators


noise64:
Related issue: https://github.com/golemcloud/golem/issues/2597

Issue #2597, opened 2026-01-16
Use type owner for type name generation in bridge SDKs


noise64:
Related issue: https://github.com/golemcloud/golem/issues/2598

PR #2609, opened 2026-01-19
Bridge tests and fixes

noise64:
The PR was originally only about enabling existing CLI integ tests for bridges, but turned out to be a bigger set of fixes and changes around Bridges.

### Reworking tests relying on "code first snippets" agent types
During 1.4 development multiple tests started using the "code_first_snippets" test method, which was a half manually done extraction of agent types from the CLI "code_first_snippets" test. Now we started using those test types in the bridge generators too, but even previously there were instances of that type not being "up-to-date" with the current extraction, and as turned out, they even contained encodings of old buggy behaviours.

To help with this situation but still trying to keep the unit tests fast, this PR adds golden file creation for the "code snippets" CLI integ tests' agent types, and uses those types - loaded from the golden files - in the unit tests. This is easy now, given we do save and cache agent types to the `golem-temp` folder during the normal `golem build` flow.

NOTE: To help with updating those files (eg. for the upcoming http binding changes, or the planned name and owner extraction improvements), a cargo make command was added:
```shell
cargo make cli-integration-tests-update-golden-files
```

Updating the used agent types resulted in highlighting many small bugs and even _nonsense tests_, like testing result types with a string literal.

### Type naming improvements
Once the cross languge bridge tests were enabled it created more instances of type name collisions. This is something we wanted to do soon, so i included in this PR.

Before this PR the TS generator simply used inline types in case of missing type names, while the Rust one generated numebered type names like "AnonType12".

To improve and unify the situation, a new `TypeNaming` type and module was introduced, which can be customized based on the language, and can be configured based on the type, whether it must have a type name, even if the extraction does not gave it a name. The name generation now uses _type location paths_: these are segments constructed starting from the agent constructors and methods, and then containing any possible named info we have, e.g.: parameter, field, case, inner types names. During name generation we back track these segments until we find a new unqiue name. This heuristic is not perfect - and can be improved later - but for most cases it should provide better experience then the "anon names". Simple examples which are generated for `y(x: <inline-complex-type>): <inline-complex-type>`

```ts
readonly y: base.RemoteMethod<[YInputX], YOutputReturnValue>
```

```rust
pub async fn y(&self,x: YInputX) -> Result<YOutputReturnValue, golem_client::bridge::ClientError>
```

The type naming also handles:
- owners: once we start extracting them, they will appear
- custom remapping: there are plans e.g. in rust to re-use custom / builtin types, the type naming can also define such remaps

Resolves: #2597 
Resolves: #2598

### Bridge related fixes and changes
- more combinations of source and target languages were added to the bridge gen unit tests
- many smaller code cleanups around bridge unit tests
- enabled testing rust bridge gen in CLI integ tests (where TS was already tested)
- use kebab-case for the generated Ts package name and Rust crate name
- the rust bridge generator now recognizes the "GOLEM_RUST_SDK_PATH"  environment variable and uses overrides for the generated bridge based on that, so one can avoid publishing anywhere while e.g. testing golem-common change
- rust bridge gen:
  - fixed missing semicolon in result type alias
  - fixed multimodel language enum conversion methods
  - fixed missing tuple and result instances for FromValue and IntoValue in golem-common
- ts bridge gen / tests:
  - added tests for void like (null, undefined, void) result types, and adapted generated test functions, so they can be tested
  - changed non-compiling result type, and related generated parts (returning undefined, instead of null), the new representation is ``
  - fixed multimodal return values
  - reverted the recent TS SDK output name change (from return-type to return-value), as it is now can take part in name generation, and it is a parameter name, not a type one
  - updated tests with many cases of:
    - wrong variant names, now TS will use type names in them, when available
    - number is F64, not S64, NOTE: the  encoding / decoding tests will always include fractional parts on one side (even for .0), and omit them, if it's 0, so to make these tests pass easily all the literals are using non integers.
    - wrong or nonsense result encodings

Resolves: #2610

### Other changes
- fixed a missing check in the "new style" task result marker up-to-date check
- added makefile task for cleaning the ts sdk and change the build one to also trigger on missing dist dir
- added common makefile task for setting up common SDK overrides: the intention is that we should be easily switch on and off these both on main or locally for all tasks
- failed golem-cli command outputs are now dumped in CLI integ tests, even with QUITE mode it true, so we can avoid retriggering with other flags, or running the test locally for seeing the problems

### DIFF
--- a/sdks/ts/packages/golem-ts-sdk/src/internal/mapping/values/dataValue.ts
+++ b/sdks/ts/packages/golem-ts-sdk/src/internal/mapping/values/dataValue.ts
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import {
+  isEmptyType,
   isOptionalWithQuestionMark,
   TypeInfoInternal,
 } from '../../typeInfoInternal';
@@ -79,6 +80,10 @@ export function deserializeDataValue(
 
           if (dataValueElementIdx >= elementsCount) {
             if (isOptionalWithQuestionMark(parameterType)) {
+              return Either.right(undefined);
+            }
+
+            if (isEmptyType(parameterType)) {
               return Either.right(undefined);
             }
 
@@ -293,6 +298,13 @@ export function serializeToDataValue(
 ): Either.Either<DataValue, string> {
   switch (typeInfoInternal.tag) {
     case 'analysed':
+      if (isEmptyType(typeInfoInternal)) {
+        return Either.right({
+          tag: 'tuple',
+          val: [],
+        });
+      }
+
       return Either.map(
         WitValue.fromTsValueDefault(tsValue, typeInfoInternal.val),
         (witValue) => {
--- a/sdks/ts/packages/golem-ts-sdk/src/internal/typeInfoInternal.ts
+++ b/sdks/ts/packages/golem-ts-sdk/src/internal/typeInfoInternal.ts
@@ -53,6 +53,17 @@ export function isOptionalWithQuestionMark(
   return false;
 }
 
+export function isEmptyType(typeInfoInternal: TypeInfoInternal): boolean {
+  if (typeInfoInternal.tag === 'analysed') {
+    const analysed = typeInfoInternal.val;
+    if (analysed.kind === 'tuple' && analysed.emptyType) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
 // Except for multimodal, all types can be converted to ElementSchema
 export function convertTypeInfoToElementSchema(
   typeInfoInternal: TypeInfoInternal,
--- a/sdks/ts/packages/golem-ts-sdk/tests/agentsInit.ts
+++ b/sdks/ts/packages/golem-ts-sdk/tests/agentsInit.ts
@@ -12,6 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// This file represents the bootstrap code that actually runs
+// in real agents. Any `test` that requires an agent to be registered (the outcome of @agent decorator)
+// in global registry must make use of this test setup and avoid manual imports within tests.
+// The way to make use of this setup is adding new agents in the `ts` file which is imported here
+// or by creating new files with agents and importing them here.
+
 const SampleAgentModuleName = './sampleAgents';
 
 import { TypeMetadata } from '@golemcloud/golem-ts-types-core';
--- a/sdks/ts/packages/golem-ts-sdk/tests/decorator.test.ts
+++ b/sdks/ts/packages/golem-ts-sdk/tests/decorator.test.ts
@@ -765,7 +765,7 @@ describe('Agent decorator should register the agent class and its methods into A
     expect(barAgent.methods.length).toEqual(25);
     expect(barAgent.constructor.inputSchema.val.length).toEqual(6);
     expect(barAgent.typeName).toEqual('my-complex-agent');
-    expect(simpleAgent.methods.length).toEqual(44);
+    expect(simpleAgent.methods.length).toEqual(45);
     expect(simpleAgent.constructor.inputSchema.val.length).toEqual(1);
   });
 
--- a/sdks/ts/packages/golem-ts-sdk/tests/invoke.test.ts
+++ b/sdks/ts/packages/golem-ts-sdk/tests/invoke.test.ts
@@ -502,6 +502,21 @@ test('Invoke function that takes and returns inbuilt result type', () => {
   );
 });
 
+test('Invoke function that returns unit type', () => {
+  overrideSelfAgentId(new AgentId('foo-agent()'));
+  const classMetadata = TypeMetadata.get(FooAgentClassName.value);
+  if (!classMetadata) {
+    throw new Error('FooAgent type metadata not found');
+  }
+
+  const resolvedAgent = initiateFooAgent('foo', classMetadata);
+
+  testInvoke('fun45', [['param', 'foo']], resolvedAgent, undefined, false, {
+    tag: 'tuple',
+    val: [],
+  });
+});
+
 test('Invoke function that takes and returns custom result type with void', () => {
   overrideSelfAgentId(new AgentId('foo-agent()'));
   const classMetadata = TypeMetadata.get(FooAgentClassName.value);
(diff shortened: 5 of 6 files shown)


cli/golem-cli/src/model/agent/test.rs
```diff
@@ -606,2248 +609,28 @@ pub fn unit_result_type() -> Vec<AgentType> {
...
-    ])
-    .named("UnionWithLiterals");
-
-    let union_with_only_literals = r#enum(&["foo", "bar", "baz"]).named("UnionWithOnlyLiterals");
-
-    let anonymous_union_with_only_literals = r#enum(&["foo", "bar", "baz"]);
-
-    let anonymous_union_with_only_literals2 = r#enum(&["foo", "bar", "baz", "baz2"]);
+// Use `cargo make cli-integration-tests-update-golden-files` to update the reference extracted types
+pub fn code_first_snippets_agent_types(language: GuestLanguage) -> Vec<AgentType> {
```
afsalthaj, line 613: Nice - removal of this big chunk of agent types and relying on golden files. Thanks for that!

cli/golem-cli/src/model/agent/test.rs
```diff
@@ -606,2248 +609,28 @@ pub fn unit_result_type() -> Vec<AgentType> {
...
-                input_schema: DataSchema::Tuple(NamedElementSchemas { elements: vec![] }),
-                output_schema: DataSchema::Tuple(NamedElementSchemas { elements: vec![] }),
-                http_endpoint: vec![],
-            }],
-            dependencies: vec![],
-            mode: AgentMode::Durable,
-            http_mount: None,
-        },
-    ]
+pub fn code_first_snippets_agent_type(language: GuestLanguage, agent_name: &str) -> AgentType {
```
afsalthaj, line 626: 👍

cli/golem-cli/tests/app/agents.rs
```diff
@@ -549,10 +560,12 @@ async fn test_ts_code_first_with_rpc_and_all_types() {
     .unwrap();
 
     let outputs = ctx.cli([cmd::BUILD]).await;
-    assert!(outputs.success());
+    assert!(outputs.success_or_dump());
+
+    check_agent_types_golden_file(ctx.cwd_path(), GuestLanguage::TypeScript).unwrap();
```
afsalthaj, line 565: 👍

afsalthaj: So generation of golden file is about running this test. We don't have a situation where we need to generate and not test? (not a big deal though)

cli/golem-cli/tests/app/agents.rs
```diff
@@ -108,10 +115,12 @@ async fn test_rust_code_first_with_rpc_and_all_types() {
     .unwrap();
 
     let outputs = ctx.cli([cmd::BUILD]).await;
-    assert!(outputs.success());
+    assert!(outputs.success_or_dump());
+
+    check_agent_types_golden_file(ctx.cwd_path(), GuestLanguage::Rust).unwrap();
```
afsalthaj, line 120: 👍

golem-wasm/src/value_and_type/from.rs
```diff
@@ -249,30 +262,38 @@ impl<T: FromValue> FromValue for Vec<T> {
...
+impl_from_value_for_tuples!(A, B, C);
+impl_from_value_for_tuples!(A, B, C, D);
+impl_from_value_for_tuples!(A, B, C, D, E);
+impl_from_value_for_tuples!(A, B, C, D, E, F);
+impl_from_value_for_tuples!(A, B, C, D, E, F, G);
+impl_from_value_for_tuples!(A, B, C, D, E, F, G, H);
+impl_from_value_for_tuples!(A, B, C, D, E, F, G, H, I);
+impl_from_value_for_tuples!(A, B, C, D, E, F, G, H, I, J);
+impl_from_value_for_tuples!(A, B, C, D, E, F, G, H, I, J, K);
+impl_from_value_for_tuples!(A, B, C, D, E, F, G, H, I, J, K, L);
```
afsalthaj, line 296: 👍

golem-wasm/src/value_and_type/into.rs
```diff
@@ -230,6 +231,19 @@ impl<S: IntoValue> IntoValue for Result<S, ()> {
...
+        match self {
+            Ok(_) => Value::Result(Ok(None)),
+            Err(_) => Value::Result(Err(None)),
+        }
+    }
+
+    fn get_type() -> AnalysedType {
+        unit_result()
+    }
+}
```
afsalthaj, line 245: Ok, I understand the need of being very specific with unit here.
Don't we need `Result<S, ()>` and `Result<(), S>` too? I think so

cli/golem-cli/src/bridge_gen/typescript/tests.rs
```diff
@@ -524,70 +517,69 @@ fn bridge_tests_optionalqmark(#[tagged_as("code_first_snippets")] pkg: &Generate
...
+            "b": 1.1,
             "c": true,
             "d": {
                 "a": "nested",
-                "b": 2,
+                "b": 2.2,
                 "c": false
             },
-            "e": { "tag": "case1", "val": 5 },
+            "e": { "tag": "union-type2", "val": 5.5 },
```
afsalthaj, line 535: 👍

afsalthaj:
TypeNaming is a huge improvement, along with golden file approach for testing. LGTM.

cli/golem-cli/tests/app/agents.rs
noise64: I would say, it is even the other way around: if we detect that golden files shifted, then you want to update them, see that RPC and all other things all still working _and_ then rerun all the unit tests (and everything) too. We could have a more optimized "update task", but if you already generated and built the snippest application, i think it is not a waste to actually run the rest of the test, and on CI i would not separate them, as CLI tests are relatively slow.

golem-wasm/src/value_and_type/into.rs
noise64: they are already there just two lines above! :D
image

cli/golem-cli/tests/app/agents.rs
afsalthaj: ok 👍

golem-wasm/src/value_and_type/into.rs
afsalthaj: Ah! :) my bad!

PR #2618, opened 2026-01-24
Minor clean up within in method schema derivation (as follow up to #2608)

afsalthaj:
As mentioned, one of th follow ups to https://github.com/golemcloud/golem/pull/2608. Also, part of #2623 
This is a very minor hygiene work in method package before adding the following features into code first routes
* handling catch-all params
* removing validations of the absence left over parameters as it can be part of request body


afsalthaj:
Taking the liberty to merge this PR since it is minor and doesn't change any behaviour

PR #2619, opened 2026-01-24
Avoid boolean based logic in constructor schema

### LINKED ISSUE #2623: Clean up constructor schema derivation in schema package in TS SDK
This is not cleaning up actual mapping logic (because that's a different beast on its own), but there is a whole lot that's happening even before hitting `mapping` package, which is under schema.

The flow that exist now is `decorator -> schema -> mapping`. `schema` package decides on a lot of things such as "if the input is multimodal", "eliminating principal, yet storing it in an internally" , identifying "unstructured-text" vs "unstructured-binary". This is not great as of now, with lot of conditional logic, which is very hard to comprehend. This became evident as we added more things related to `Principal`.

To overcome this, I am going to do the following - which is very internal to `schema/constructor.ts`, which is `handler` interface. A handler interface is simply having two functions `canHandle` returning a `boolean` and then the real `handle` which gets called if `canHandle` is true. 

Much easier to understand, and easy to add new handlers. In this way, we choose the first handler that returns true and then simply call handle. I think, this will be atleast first level of necessary clean up

afsalthaj:
Fixes #2623 

A follow up to the clean up happening in schema derivation part

### DIFF
--- a/sdks/ts/packages/golem-ts-sdk/src/internal/schema/constructor.ts
+++ b/sdks/ts/packages/golem-ts-sdk/src/internal/schema/constructor.ts
@@ -59,15 +59,16 @@ export function getAgentConstructorSchema(
   );
 }
 
-type ConstructorParamHandler = {
+interface ConstructorParamHandler {
   canHandle(param: ConstructorArg): boolean;
+
   handle(
     agentClassName: string,
     param: ConstructorArg,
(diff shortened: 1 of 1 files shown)


afsalthaj:
Looks like I am back with these merge commit magic.. will fix in next PRs

PR #2632, opened 2026-01-27
Clean up Type mapping in TS SDK

### LINKED ISSUE #2629: Clean up type mapping (similar to schema module) in TS SDK

afsalthaj:
Fixes #2629 

- [x]  Better management of each type mapping. Many times its pattern matching on `kind` of type and then depending on it, access the field parameters of `Type`, but this became so ugly especially when we needed to split functions with passing around these random parameters so much. The PR finds a way out a way to separate out them as different packages. Example: `union.ts` to understand how schema is derived for a union type, while avoiding any cyclic dependency in modules in TS, as we are handing recursion. 

- [x] Once done, clean up some very complicated handlers such as union

- [x] verify all mappings, every logic  and try and simplify. Example, see this simplification:  [removing wrapping in option multiple times](https://github.com/golemcloud/golem/pull/2632/changes/a2282f5928c7bc2148d7073160178e5cae8d7d2a), [avoid scattered caching of anonymous union types](https://github.com/golemcloud/golem/pull/2632/commits/55915d864528d0533a114c980640c0b3cc3ec2f8) etc

- [x]  Remove option type. This is because there is significant inconsistency in its use with many contributing to SDK: Between, `Option<A>` versus `A | undefined`, and there are no strong reasons anymore with `Option` 

This is early draft PR

During clean ups, there are possibilites of bug fixes that happened which I will document

Notes:
* I am not refactoring for value mapping refactorin in this  PR. type mapping is the part where bridge code may have dependency on, and therefore I cleaned up only type mapping only for now, and it really took some time for myself to comprehend. 
* I have put detaile comments especially `union` handling, because many things are really very subtle. Also documentation of variant name generations. 
* I know `Either` usage is verbose, but it really is giving us more control and avoid propagating context top down for error messages.  I removed `Option` usage though
* Also note that, some parts are kept as it is. I know they could also be improved but it isnt hurting much as logic that exist for complex handlers. 

Every refactoring done is very manual, with re-verification of how mapping is done, and it took time!

### DIFF
--- a/sdks/ts/packages/golem-ts-sdk/src/internal/mapping/types/union.ts
+++ b/sdks/ts/packages/golem-ts-sdk/src/internal/mapping/types/union.ts
@@ -34,31 +34,44 @@ type UnionCtx = Ctx & { type: Extract<TsType, { kind: "union" }> };
 
 export function handleUnion(
   ctx: UnionCtx,
-  mapper: TypeMapper
+  mapper: TypeMapper,
+): Either.Either<AnalysedType, string> {
+
+  const unionType = ctx.type;
+
+  const isAnonymous = !unionType.name;
+
+  const cacheKey = isAnonymous ? hashUnion(unionType) : undefined;
+
+  const analysedType = handleUnionInternal(ctx, mapper, cacheKey);
+
+  cacheAnonymousUnionType(cacheKey, analysedType);
+
+  return analysedType
+}
+
+export function handleUnionInternal(
+  ctx: UnionCtx,
+  mapper: TypeMapper,
+  cacheKey: string | undefined
 ): Either.Either<AnalysedType, string> {
   const { type } = ctx;
 
-  const cacheKey = hashUnion(type);
-  const isAnonymous = !type.name;
-
   // Reuse cached anonymous unions
-  const cached = reuseAnonymousUnionCache(ctx, cacheKey);
+  const cached = reuseAnonymousUnionCache(cacheKey);
   if (cached) return cached;
 
   // Try if it's inbuilt Result<T, E>
   const inbuiltResult = tryInbuiltResultType(ctx, mapper);
-  if (inbuiltResult) {
-    cacheIfAnonymous(cacheKey, isAnonymous, inbuiltResult);
-    return inbuiltResult;
-  }
+  if (inbuiltResult) return inbuiltResult;
 
   // Try if its union of only literals, and if so convert to enum
-  const enumResult = tryEnumUnion(ctx, cacheKey);
+  const enumResult = tryEnumUnion(ctx);
   if (enumResult) return enumResult;
 
   // Try if all variants in union are tagged. Example: `{tag : 'a", val: 1} | {tag: 'b', val: 2}`.
   // If they are tagged, it further `process` it such as verifying if it corresponds to `result` type
-  const taggedResult = tryTaggedUnionAndProcess(ctx, mapper, cacheKey);
+  const taggedResult = tryTaggedUnionAndProcess(ctx, mapper);
   if (taggedResult) return taggedResult;
 
   // Try Optional union (null | undefined | void) and convert to `option` WIT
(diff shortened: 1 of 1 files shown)


sdks/ts/packages/golem-ts-sdk/src/internal/mapping/types/union.ts
```diff
@@ -0,0 +1,397 @@
...
+import { fromTsTypeInternal} from './typeMapping';
+import { Ctx } from './ctx';
+
+type TsType = CoreType.Type;
+
+const AnonymousUnionTypeRegistry = new Map<string, AnalysedType>();
+
+type UnionCtx = Ctx & { type: Extract<TsType, { kind: "union" }> };
+
+// TODO; Refactor more here, (added only comments for now to avoid losing changes)
```
afsalthaj: For now, I separated each handler to be a separte module. 
So to handle `union` for type discovery, we just need to care this `union.ts` under `internal/type` package.
For now, I put some comments on what's going on. Helps with improving this further with better hygiene before merging the PR.

I tried to simplify the logic but they all are valid corner cases, backed by tests which is good.

sdks/ts/packages/golem-ts-sdk/src/internal/mapping/types/alias.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { buildJSONFromType, Node, Type as CoreType } from '@golemcloud/golem-ts-types-core';
+import * as Either from "../../../newTypes/either";
+import { Ctx } from './ctx';
+import { AnalysedType } from './analysedType';
+import { TypeMapper } from './typeMapper';
+
+type TsType = CoreType.Type;
+
+type AliasCtx = Ctx & { type: Extract<TsType, { kind: "alias" }> };
```
afsalthaj, line 23: We have multiple handlers. `alias`, `tuple`, `union` etc,  where a handler (on a high level) is a function from `context` to `Either<AnalysedType, string>`

Now `context` should be enriched to avoid having to check if a `type.kind` is `union` or `tuple` etc. Instead they can defined 

```ts

type AliasCtx = Ctx & { type: Extract<TsType, { kind: "alias" }> };

type UnionCtx = Ctx & { type: Extract<TsType, { kind: "union" }> };

// etc
```
And they can simply refer to the fields of that type directly. Previously we check the kind and then individually pass around each field values (since a standalone function missed this context)

sdks/ts/packages/golem-ts-sdk/src/internal/mapping/types/handlers.ts
```diff
@@ -0,0 +1,87 @@
...
+  "class": unsupportedWithHint("class", "Use object instead."),
+  "promise": handlePromise,
+  "map": handleMap,
+  "literal": handleLiteral,
+  "alias": handleAlias,
+  "others": handleOthers,
+  "unresolved-type": handleUnresolved,
+  "array": handleArray,
+}
+
```
afsalthaj, line 69: Same as `switch(type.kind)` for the most part - just more concise with no returns or break or default requirements and more readable that it matches on each kind and return the corresponding handler, which can be called separately.

sdks/ts/packages/golem-ts-sdk/src/internal/mapping/types/analysedType.ts
```diff
@@ -0,0 +1,405 @@
+// Copyright 2024-2025 Golem Cloud
```
afsalthaj, line 1: With new `mapper` analysedType package consist of just `AnalysedType`.

sdks/ts/packages/golem-ts-sdk/src/internal/mapping/types/analysedType.ts
```diff
@@ -0,0 +1,405 @@
...
+): AnalysedType {
+  return {
+    kind: 'handle',
+    value: {
+      name: convertOptionalTypeNameToKebab(name),
+      owner: undefined,
+      resourceId,
+      mode,
+    },
+  };
```
afsalthaj, line 404: Removed lambda definitions in favor of simple functions

sdks/ts/packages/golem-ts-sdk/src/internal/mapping/types/interface.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  const interfaceFields = interfaceResult.val;
+
+  if (interfaceFields.length === 0) {
+    return Either.left(`Type ${type.name} is an object but has no properties. Object types must define at least one property.`);
+
+  }
+
+  return Either.right(record(type.name, interfaceFields))
```
afsalthaj, line 69: I know `Either` usage is verbose, but it really is giving us more control and avoid propagating context top down for error messages. 

I removed `Option` usage though

sdks/ts/packages/golem-ts-sdk/src/baseAgent.ts
```diff
@@ -76,17 +75,20 @@ export class BaseAgent {
    */
   getAgentType(): AgentType {
     if (!this.cachedAgentType) {
-      const agentType = AgentTypeRegistry.get(this.agentClassName);
+      const agentType: AgentType | undefined = AgentTypeRegistry.get(
```
noise64, line 78: is there a special reason for explicit types here? i do not saw them in other places

sdks/ts/packages/golem-ts-sdk/src/internal/try.ts
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as Either from '../newTypes/either';
+
+export type Try<T> = Either.Either<T | undefined, string>;
```
noise64, line 17: this name seems to be a bit confusing, given this has two error channels

afsalthaj: I will need to think about a better name there. It's about trying to convert it to a type where it identifies its of the type but ther are reasons to fail. Or it is not of that that type at all. Please give me some time to think through about a better naming in follow up PRs

PR #2634, opened 2026-01-29
Follow up 2629: A few more mechanical clean ups

afsalthaj:
Small improvements:

* Avoid random spaces in error messages. It gives a less enterprise grade user experience
* Making sure to propagate the type scope when it comes to Promise, although we reject the use of `Promise` anywhere else other than return type of methods

Other than these two "changes", everything else is a code level mechanical cleanups with zero changes in logic.


afsalthaj:
Taking the liberty to merge this.

PR #2698, opened 2026-02-05
Add create-webhook host function

mschuwalow:


mschuwalow:
closing in favor of #2699 2699

PR #2699, opened 2026-02-05
Add create-webhook host function

mschuwalow:
The reason for not making it a struct with a promise-id in it is that it forces move of promise-id from golem-host to golem-rpc (in order to avoid golem-agent depending on all of host). This in turn breaks composition of the benchmarks_agent_ts_large test component with the golem-ai library -> need to update golem-ai.

---
The reason for moving the registered-agent-type to common is that it's impossible to specify just specific resources in with when using wit-bindgen.

With the old structure, we would need to generate some of the bindings for golem:agent/host (for registered-agent-type) in golem-common and then reuse them + a new binding for golem:agent/host/agent-webhook in the worker-executor. This does not seem to be supported.

### DIFF
--- a/sdks/rust/golem-rust/src/lib.rs
+++ b/sdks/rust/golem-rust/src/lib.rs
@@ -12,10 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::fmt;
-use std::fmt::{Display, Formatter};
-use std::str::FromStr;
-
 #[cfg(test)]
 test_r::enable!();
 


afsalthaj:
Let me play around this change with my change first, before merging @mschuwalow. Let's have another round of discussion today.


Just to be safe, I am moving back to draft, to avoid other merging it.

cli/golem-cli/wit/deps/golem-agent/host.wit
```diff
@@ -23,4 +25,11 @@ interface host {
     /// Parses an agent-id (created by `make-agent-id`) into an agent type name and its constructor parameters
     /// and an optional phantom ID
     parse-agent-id: func(agent-id: string) -> result<tuple<string, data-value, option<uuid>>, agent-error>;
+
+    /// Creates a webhook structure that can be used to call webhook driven apis.
+    /// Blocking on the returned pollable will block until the returned url is called with a get request.
+    /// Note the following behaviours:
+    /// * Only agents whoose agent types are _currently_ deployed via an http api are allowed to create a webhook. Calling this function while the agent
+    ///    is not deployed via an http api will trap.
+    create-webhook: func() -> agent-webhook;
```
afsalthaj: I never thought this is a host function. Not sayiny its incorrect.

This is what I thought. I was expecting only `get_webhook_url` than going into any other details at the host level.

This implies, `create_webhook` in SDK is simply about calling `get_webhook_url` and store the `PromiseId` from it. I felt that it's the least invasive change to do.

afsalthaj:
With the PR change, I can get PromiseId from the URL (although a small bit of ceremony of summoning PromiseId which I can await on using await_promise/await_promise_json, or I also have this subscribe to get Pollable which I can await on,, but still I need to rely on the original PromiseId to access the final value. Implies, I may not need this subscribe as such? slightly confusing semantics at the client side. 

Also a few thoughts:

Obviously we both decided to have a struct that has just a URL and/or promise-id. `and/or` because URL already should have the promise-id in it.

So the change I was expecting was just one host function which is as follows:

```rust
 get_webhook_url() 
```
The implementation of this will be simple appending of domain (an agent-type can only be in 1 domain at a time, as we discussed) and the webhook suffix (which is available in the agent type information), the webhook prefix (deployment time data), and finally the output of `create_promise` which is promise_id.(for context for others reading it)


Also this: https://github.com/golemcloud/golem/pull/2699/changes#r2771768638

Quoting PR description
>The reason for not making it a struct with a promise-id in it is that it forces move of promise-id from golem-host to golem-rpc (in order to avoid golem-agent depending on all of host). This in turn breaks composition of the benchmarks_agent_ts_large test component with the golem-ai library -> need to update golem-ai.

 `move of promise-id from golem-host to golem-rpc`. I thought it is a very less invasive thing to do as well. Anyway, I have to understand how this is going to be complicated (or not) back in SDK implementation.

If not already, let's get an opinion from @vigoo

cli/golem-cli/wit/deps/golem-agent/host.wit
```diff
@@ -2,12 +2,14 @@ package golem:agent;
...
 
-    /// Associates an agent type with a component that implements it
-    record registered-agent-type {
-      agent-type: agent-type,
-      implemented-by: component-id
+    /// A completable webhook owned by the current agent and an associated url that can be called
+    /// via a get request to complete the webhook.
+    resource agent-webhook {
+      get-callback-url: func() -> string;
+      subscribe: func() -> pollable;
```
afsalthaj: With the PR change, I can get PromiseId from the URL (although a small bit of ceremony of summoning `PromiseId`)  which I can await on using `await_promise` or I also have this subscribe to get `Pollable` which I can await on,, but still I need to rely on the original PromiseId to access the final value. 

Implies, I may not need this subscribe as such? slightly confusing semantics at the client side

cli/golem-cli/wit/deps/golem-agent/host.wit
```diff
@@ -2,12 +2,14 @@ package golem:agent;
...
     }
```
afsalthaj: Also note that, in TS, there is no easier way to go from promise-id string to promise-id unlike in Rust. So host may need to expose this parsing..

cli/golem-cli/wit/deps/golem-agent/host.wit
afsalthaj: See TODO here: https://github.com/golemcloud/golem/pull/2702

cli/golem-cli/wit/deps/golem-agent/host.wit
vigoo: I don't really like this approach because it "duplicates" (a part of) the promise API and the SDKs are already building on the existing promise API providing all kind of stuff like async awaiting them, completing them, getting out the result etc. 

With this the use of Golem Promises for webhooks becomes a hidden implementation detail which can be argued for being a feature and not a problem, but I thought we wanted to do the opposite - and just build a webhook feature on top of Golem Promises.

I think there are some advantages to do it in the way it was originally planned:
- The promises (exposed as webhooks) can be completed through the complete-promise SDK calls from code, not just from outside
- no confusion between what is a promise and what is a webhook and how they are related (thinking of documenting patterns of coordinating agents with golem promises and the outside triggers, for example)
- golem promises are also designed to be completed from the outside (as well) through our promise API; having a completely different guest-side API for working with both seems unnecessarily complicated and confusing

afsalthaj: I agree with @vigoo too. @mschuwalow you can see the semantic and implementation level duplication in the other 2 PRs of mine which is in progress. Just 1 function to get the final URL will be enough and infact it should have PromiseId along with it in struct when I was trying to work this In TS SDK. in both SDKs we should avoid any parsing into promise-Id. It's very reliable to have PromiseId and a separate URL as we discussed.

cli/golem-cli/wit/deps/golem-agent/host.wit
mschuwalow: Whether you need the actual promise-id or not depends on whether we want to expose the webhook payload or not. Definitely we don't want to do any parsing of the url in the sdk side!

afsalthaj: We will need to expose the actual payload.

cli/golem-cli/wit/deps/golem-agent/host.wit
```diff
@@ -26,10 +18,10 @@ interface host {
...
 
-    /// Creates a webhook structure that can be used to call webhook driven apis.
-    /// Blocking on the returned pollable will block until the returned url is called with a get request.
+    /// Creates a webhook that can be used to integrate with webhook driven apis.
+    /// When the created url is called with a post request, the provided promise-id is completed with the body of the post request.
     /// Note the following behaviours:
     /// * Only agents whoose agent types are _currently_ deployed via an http api are allowed to create a webhook. Calling this function while the agent
     ///    is not deployed via an http api will trap.
-    create-webhook: func() -> agent-webhook;
+    create-webhook: func(promise-id: promise-id) -> string;
```
afsalthaj, line 26: May be `create-webhook-url` instead of `create-webhook`?

mschuwalow:
While technically you can call the webhook without ever creating the url using the host function to complete the promise, I want to treat that as an edge case of our implementation.

Potentially we can adjust this in the future such that only promises for which create-webhook has been called can be completed by calling the webhook url. My reasoning is that being able to complete arbitrary promises is a bit of a security smell I would like to be rid of. In that case the create-webhook-url naming would no longer fit the behaviour.

Not a strong reason not to change it, just explaining my reasoning. Can adjust it if you prefer the other name.

PR #2701, opened 2026-02-06
Rust and TS SDK changes for webhook

### LINKED ISSUE #2569: Code First Routes: Add web-hook support for code-first routes
Similarly to cors/auth, a new synthetic route needs to be added during deployment or routing time.
The web-hooks should be handled directly in the worker-service without creating an intermediary agent.

It might be necessary to create a new host api in order to back the createWebhook function used by agents.
The current webhook prefix / deployment domain needs to come from the _current deployment_ not the agent component version.

afsalthaj:
Fixes #2569


sdks/rust/golem-rust/src/agentic/webhook.rs
```diff
@@ -0,0 +1,90 @@
...
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output>>>;
+
+    fn into_future(self) -> Self::IntoFuture {
+        Box::pin(async move { self.wait().await })
+    }
+}
+
+pub struct WebhookRequestPayload {
+    payload: Vec<u8>,
+}
```
afsalthaj, line 60: SDK assume worker service to complete with the request body. But if it is assuming a specific structure there, then that structure needs to be reused somehow between server and SDK

afsalthaj: This will be done differently soon

PR #2722, opened 2026-02-08
code-first-routes: Webhooks backend

### LINKED ISSUE #2569: Code First Routes: Add web-hook support for code-first routes

mschuwalow:
resolves #2569


cli/golem-cli/wit/deps/golem-agent/host.wit
```diff
@@ -23,5 +23,7 @@ interface host {
     /// Note the following behaviours:
     /// * Only agents whoose agent types are _currently_ deployed via an http api are allowed to create a webhook. Calling this function while the agent
     ///    is not deployed via an http api will trap.
+    /// * Only the agent type that created the promise is allowed to create a webhook for it. Using this host function
+    ///   from a different agent type will trap.
```
afsalthaj, line 27: 👍

golem-common/src/base_model/http_api_deployment.rs
```diff
@@ -53,6 +55,13 @@ declare_structs! {
...
         pub hash: diff::Hash,
         pub agents: BTreeMap<AgentTypeName, HttpApiDeploymentAgentOptions>,
+        pub webhooks_url: String,
         pub created_at: DateTime<chrono::Utc>,
     }
 }
+
+impl HttpApiDeploymentCreation {
+    pub fn default_webhooks_url() -> String {
+        "/webhooks/".to_string()
```
afsalthaj, line 65: Good to be more specific ? `default_webhook_url_prefix` ? Can be confusing with prefix, suffix and the fully qualified.

Issue #2725, opened 2026-02-09
Having same function name and type name result in wrapper generation errors


afsalthaj:
```
    fn json_response(&self) -> JsonResponse {
        JsonResponse {
            value: "ok".to_string(),
        }
    }


```

This is an example of the situation. And it fails with

```
world agent-wrapper {
  import golem:agent/guest;
  import golem:api/save-snapshot@1.3.0;
  import golem:api/load-snapshot@1.3.0;
  import wasi:logging/logging;
  export golem:agent/guest;
  export golem:api/save-snapshot@1.3.0;
  export golem:api/load-snapshot@1.3.0;
  export cors-agent;
  export http-agent;
  export types;
}
: name `json-response` is defined more than once
     --> wrapper.wit:40:3
      |
   40 |   json-response: func()
```

vigoo:
We are no longer generating wrappers

Issue #2724, opened 2026-02-09
SDK DX: Forgetting agent_implementation annotation should somehow warn users


afsalthaj:
Currently in Rust SDK, if we forget to write `#[agent_implementation]`  things would still compile and build, which I think is a bit of a trouble. It can result in users not seeing the agent at all, integration tests failing with "agent type not found" etc. I think this is not a great DX (both for users and for us).

First question is, do we care? The answer is very subjective. I bumped into this couple of times already.
If we do care, then the solution to this problem is not that easy and requires some wordy explanation. Minimum, it serves as an explanation for why we didn't do it if someone bumps into it and has questions in future.  

Breaking it down starting with a bit of context to reason about a possible solution, and it's feasibility (almost infeasible, but kicks off thoughts)

* An agent is an agent only if it's implementation exist. i.e, the actual registration function is generated as part of `agent_implementation`. Why? Let's say, the registration function is generated as part of agent definition, and iwe keep agent definition in a common crate.  Then, any generated registration functions as part of `definition` will not run. Also even if we made this work somehow (which is impossible actually), then agent type would exist for clients to discover, but when initiating it, it will result in "no initiator for this agent type" runtime error.

* Therefore `agent_implementation` will generate entry level functions that registers the agent type into agent-type registry. If this implementation doesn't exist, then there is no agent , and the `get_definitions` called from anywhere will not see agents, and that's good because otherwise they are able to initialise an agent resulting in runtime error as mentioned before.  Note that as part of `agent_implementation` we register the actual agent-type into agent-type registry as well agent-initiator into the initiator registry too. Agent inititor registry is a registry that has every agent-type and its corresponding `AgentInitiator` instance.  Once the client successfully initiate, then it gets to the real `Agent` (and SDK set it globally) , so that the subsequent `invoke` methods make use of it.

* Now these entry point functions of `registering an agent type, and then registering how to initiate it` is generated per macro annotation of `agent_implementation`, and at load time, these entry-points run, and the order doesn't matter, and there is no guarantee in the order too. This is to say that, we shouldn't rely on any entry point generation for any sort of global validation. Rust SDK is carefully devoid of any load-time-global-validation anywhere.

* To repeat the problem, if developer forgets to do `#[agent_implementation]` but still have an implementation or if they completely forget to write the implementation (which can be rare, yet) - the  result is their agents are not registered or seen anywhere, but still builds successfully. CLI cannot judge at any point whether every agent definition has a corresponding implementation.

[...]

PR #2726, opened 2026-02-09
Add test for rust code first routes and fix bugs related to unstructured binary

### LINKED ISSUE #2558: Code First Routes: Implement integration test for agents with code first route
We need to decide on how to implement the integration tests for agents with code first routes

Given typescript SDK and backend is implemented by this time, we should aim to have atleast a full end to end working test with TS.

Later once, Rust is implemented, we should update these tests with Rust agents.

afsalthaj:
Fixes #2558 

A few observations:

https://github.com/golemcloud/golem/issues/2725
https://github.com/golemcloud/golem/issues/2724

And 2 more small issues which I will fix in a PR soon
* endpoint parameter syntax errors should result in compilation errors. I think its not working for headers
* disallow adding endpoints to static functions

### DIFF
--- a/test-components/agent-http-routes-rust/Cargo.toml
+++ b/test-components/agent-http-routes-rust/Cargo.toml
@@ -8,7 +8,7 @@ lto = true
 
 [workspace.dependencies]
 
-golem-rust = { path = "/Users/afsalthaj/projects/ribbb/golem/sdks/rust/golem-rust", features = [
+golem-rust = { path = "../../sdks/rust/golem-rust", features = [
     "export_golem_agentic",
 #    "golem_ai" # Uncomment to use Golem AI libraries
 ] }


integration-tests/tests/custom_api/http_test_context.rs
```diff
@@ -0,0 +1,191 @@
...
+        client,
+        base_url,
+        user,
+        env_id: env.id,
+        deployment_revision: deployment.revision,
+        host_header,
+    })
+}
+
+pub async fn run_webhook_callback_test(
```
mschuwalow: Personally, I would not have extracted this one.

afsalthaj: I had some doubts on it... may be I will revert it back

PR #2745, opened 2026-02-11
Open API export

### LINKED ISSUE #2567: Add openapi export for code-first routes
Implement openapi export for code-first routes. At this point this should just be basic support, it's fine to leave more complex features as followup (especially if they need sdk support)

afsalthaj:
Fixes #2567

### DIFF
--- a/golem-worker-service/src/custom_api/openapi/http_openapi_spec.rs
+++ b/golem-worker-service/src/custom_api/openapi/http_openapi_spec.rs
@@ -140,7 +140,9 @@ fn add_route_to_paths(
             &inner.constructor_parameters,
             &inner.method_parameters,
         ),
-        RichRouteBehaviour::WebhookCallback(_) => vec![("promise-id", false, &PathSegmentType::Str)],
+        RichRouteBehaviour::WebhookCallback(_) => {
+            vec![("promise-id", false, &PathSegmentType::Str)]
+        }
         _ => Vec::new(),
     };
     let query_params_raw = match &route.behavior {
@@ -206,7 +208,8 @@ fn create_schema_from_path_segment_type(
     let analysed_type = AnalysedType::from(path_segment_type);
     let mut schema = create_schema_from_analysed_type(&analysed_type);
     if is_catchall_segment {
-        schema.schema_data.description = Some("Parameter represents the remaining path, including slashes.".to_string())
+        schema.schema_data.description =
+            Some("Parameter represents the remaining path, including slashes.".to_string())
     }
     schema
 }


golem-worker-service/src/custom_api/request_handler.rs
```diff
@@ -123,6 +128,47 @@ impl RequestHandler {
                     .await
             }
 
+            RichRouteBehaviour::OpenApiSpec(OpenApiSpecBehaviour { open_api_spec }) => {
+                match self
+                    .route_resolver
+                    .get_enriched_routes_for_domain(&resolved_route.domain)
```
mschuwalow: There should not be any http requests during route evaluation.

Instead you could keep the data you need for the openapi generation around when going from CompiledRoute -> RichCompiledRoute

golem-worker-service/src/custom_api/request_handler.rs
```diff
@@ -157,5 +203,14 @@ fn route_execution_result_to_response(
         ResponseBody::UnstructuredBinaryBody { body } => Ok(response_builder
             .body(body.data)
             .set_content_type(body.binary_type.mime_type)),
+
+        ResponseBody::OpenApiSchema { body } => {
```
mschuwalow: We want to render a swagger ui here if I'm not mistaken, not just the raw api document

golem-worker-service/src/custom_api/route_resolver.rs
```diff
@@ -81,6 +82,20 @@ impl RouteResolver {
         }
     }
 
+    pub async fn get_routes_for_domain(
```
mschuwalow: where is this used?

golem-registry-service/src/services/deployment/route_compilation.rs
```diff
@@ -256,6 +257,87 @@ pub fn add_webhook_callback_routes(
...
+            build_openapi_spec_for_component(&component_id, &agents, compiled_routes);
+
+        if open_api_spec.is_empty() {
+            continue;
+        }
+
+        // Note: This is currently a fixed path,
+        // but it can be part of the http api deployment configuration
+        let path = vec![PathSegment::Literal {
+            value: "openapi.json".to_string(),
```
mschuwalow, line 276: We need to have one api spec per deployment, not per component.

golem-registry-service/src/services/deployment/route_compilation.rs
```diff
@@ -256,6 +257,87 @@ pub fn add_webhook_callback_routes(
...
+
+        component_agents
+            .entry(agent.implemented_by.component_id)
+            .or_default()
+            .push(agent);
+    }
+
+    for (component_id, agents) in component_agents {
+        let open_api_spec =
+            build_openapi_spec_for_component(&component_id, &agents, compiled_routes);
```
mschuwalow: We cannot select the routes that are part of the openapi spec here. They might get filtered out later because of missing security schemes, etc.
Instead we should base the openapi scheme on the final set sent to the worker-service

mschuwalow (changes requested):

PR #2754, opened 2026-02-12
Initial implementation of recovering from snapshots

### LINKED ISSUE #2423: Implement recovery from saved snapshots
Recovery will work very similarly to how manual update works. We locate the latest Snapshot oplog entry and read the payload. Then call the load-snapshot function, and replay the remaining of the oplog.

If `load-snapshot` fails, we try recovering from the beginning with no snapshots. If that fails too, the agent goes into permanent failed state.

Part of this is adding some new worker executor tests verifying the snapshotting behavior.

vigoo:
Follow-up step for https://github.com/golemcloud/golem/pull/2752

Do not merge yet

Resolves #2423

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -221,7 +221,7 @@ jobs:
 
       - name: Worker Executor integration tests
         run: cargo make --profile ci worker-executor-tests-${{ matrix.group.name }}
-        timeout-minutes: 10
+        timeout-minutes: 15
       - name: Publish Test Report
         uses: ctrf-io/github-test-reporter@v1
         if: always()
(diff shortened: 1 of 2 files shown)


golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2749,7 +2951,21 @@ impl PrivateDurableWorkerState {
...
+            let snapshot_skip =
+                DeletedRegionsBuilder::from_regions(vec![OplogRegion::from_index_range(
+                    OplogIndex::INITIAL.next()..=snapshot_idx,
+                )])
+                .build();
+            regions.set_override(snapshot_skip);
+            regions
+        } else {
+            deleted_regions
+        };
```
noise64, line 2967: just to check if i understand correctly: after a snapshot, everything before it accounted as deleted?

vigoo: This is the same mechanism that is used for snapshot-based updates to skip the oplog's beginning, so recovery continues from the snapshotting point and ignores what happened before. It's just an override so it is not persisted to the oplog as an actual deleted region.

PR #2760, opened 2026-02-12
fix: desktop app CLI arg fixes and integration tests

Myestery:
## Summary
- **fix: remove invalid positional args from deploy and ephemeral invoke CLI calls**
- **fix: use 'new' as root-level command for app creation** (was incorrectly using "root")
- **fix: normalize case when matching ephemeral agent constructor names** — CLI returns PascalCase (`HumanAgent`) but function names use kebab-case (`human-agent`), causing ephemeral invocations to omit constructor params
- **test: add integration tests for all golem-cli service invocations** — 61 new tests across 8 service test files, plus a unit test project in vitest config


vigoo:
We are not doing an 1.4.3 release, so closing this for now

PR #2764, opened 2026-02-13
Default snapshotting implementation in the TS SDK

### LINKED ISSUE #2427: Default snapshotting implementation in the TS SDK
The default implementation for save/load should save the agent class as JSON.

vigoo:
Resolves #2427 

Do not merge yet

Built on top of #2754

### DIFF
--- a/sdks/ts/packages/golem-ts-sdk/src/index.ts
+++ b/sdks/ts/packages/golem-ts-sdk/src/index.ts
@@ -123,19 +123,31 @@ async function save(): Promise<{ data: Uint8Array; mimeType: string }> {
   }
 
   const { data: agentSnapshot, isCustom } = await resolvedAgent.saveSnapshot();
+  const principal = initializationPrincipal ?? { tag: 'anonymous' };
 
-  const principalJson = JSON.stringify(initializationPrincipal ?? { tag: 'anonymous' });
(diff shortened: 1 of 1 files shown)


sdks/ts/packages/golem-ts-sdk/src/baseAgent.ts
```diff
@@ -91,19 +91,44 @@ export class BaseAgent {
...
+    const state: Record<string, unknown> = {};
+
+    for (const [k, v] of Object.entries(this)) {
+      if (k === 'cachedAgentType' || k === 'agentClassName') continue;
+      if (typeof v === 'function') continue;
+      state[k] = v;
+    }
+
+    const payload = { version: 1, state };
+    return new TextEncoder().encode(JSON.stringify(payload));
```
noise64, line 131: i don't think we have to handle all the possible problems with JSON.stringify, but maybe Date would be nice to handle

noise64: as in:
- function is already handled
- bigint / recursive should fail, so users will know that they have to customize
- undefined should be fine
- symbol / non-enumerable are not that common, and probably should not be part of serailization any way

noise64 (approved):

PR #2762, opened 2026-02-13
fix bug related to Principal being the only parameter

afsalthaj:
Fixes #2753 

While the following already works:

```ts

function foo(a: string, b: Principal, c: number)

// first param (string) → fetch index 0, increment count
// second param (Principal) → return available Principal, no count change
// third param (number) → fetch index 1, all good
```

A specific but basic scenario was not handled correctly:

```ts
function foo(a: Principal)

```

This was because, before processing for every parameter type, it has to check if the counter is greater than or equal to the total length of the data-value, to handle special cases such as empty type. But then forgot to add Principal here. 

Fixed now

image

Added a separate integration test for basic principal injection without any auth. Let's keep it separate here and not conflate with other custom http api integration tests.

PR #2770, opened 2026-02-13
fix bug related to Principal being the only parameter (2)

mschuwalow:
Continuation of #2762, as I merged that one before it was ready.

### DIFF
--- a/integration-tests/tests/api/deployment.rs
+++ b/integration-tests/tests/api/deployment.rs
@@ -234,17 +234,19 @@ async fn full_deployment(deps: &EnvBasedTestDependencies) -> anyhow::Result<()> 
         .await?;
 
     let expected_hash =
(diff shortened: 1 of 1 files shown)


test-components/agent-principal-ts/common-ts/src/lib.ts
```diff
@@ -0,0 +1,3 @@
+export function exampleCommonFunction() {
```
afsalthaj: @mschuwalow  I did this in the other PR, as a separate component (which you merged here), but that needs to be changed now and the way to do is to update an existing component with this new agent that deals with `Principal`. I had a discussion around this with @vigoo

afsalthaj: I am doing it: https://github.com/golemcloud/golem/pull/2775

PR #2775, opened 2026-02-16
Remove principal agent

afsalthaj:


afsalthaj:
Merging this to original PR since it passed in my local

image

PR #2796, opened 2026-02-18
Component level wasi:config settings

### LINKED ISSUE #2789: Wasi-config support on the level of components

mschuwalow:
resolves #2789


vigoo:
One initial comment (could not properly review yet): 
let's not call this things `wasi_config_vars` etc - I think I've already renamed in some places by dropping `wasi_`. Our users should not see `wasi` anywhere and I think it's better to be consistent in our internal namings too.

PR #2819, opened 2026-02-20
Agent invocation

### LINKED ISSUE #1093: Raw / enriched oplog entry types on the host interface
We should only send the raw oplog entries to oplog processor plugins by default, and provide host functions to enrich them to the current "public oplog entry" type when necessary.

### LINKED ISSUE #2459: Agentic invocation APIs

### LINKED ISSUE #2461: Make the MoonBit static wrapper generator an optional feature
We no longer need to generate static WIT wrappers for agents, so we remove it from the build flows, but can keep it as an interoperability feature to package agents for other WASM environments.

vigoo:
Resolves #2459
Resolves #2461 (by completely removing it instead of making it optional)
Resolves #1093

Replaces the old invocation model and APIs with a new, agent oriented one on all levels.

### HTTP APIs
Worker-service no longer has the old `invoke-and-await` and `invoke` endpoints. The only way to invoke is the new `invoke-agent` endpoint (already created as part of the bridge SDK epic, now extended/fully implemented)

The `launch-new-worker` endpoint (creates an agent without invoking anything on it) now accepts invocation context and principal. 

Note: further API changes to be expected in a next step where we try to remove the need to explicitly use component-ids when it can be inferred from agent type, but that's intentionally NOT part of this PR.

### gRPC APIs
The old invocation gRPC endpoints are removed from both the worker-exeuctor and the worker-service. A single new endpoint is created that matches the `invoke-agent` http endpoint, directly using agentic types (agent-type, data-value, etc) and supporting trigger, schedule and await together.

All the endpoints that potentially create an agent instance have been extended with an optional principal parameter.

### WIT changes
Restructured our WIT packages significantly:
- common types moved to `golem:core`
- `golem:rpc` does not exist anymore, became part of `golem:agent/host`
- All golem packages bumped to be `@1.5.0`

The old `wasm-rpc` resource is now agentic - its constructor gets agent-type and data-value, and its methods are invoking agent methods using the agent data model too.

Significant changes to the oplog (described below) are applied in the WIT version of the oplog too, and now we also have two types of oplog entries on WIT level just like in code: `oplog-entry` and `public-oplog-entry`. This was an old ticket (#1093) that reduces the data travelling in oplog processor calls. An `enrich-oplog-entries` host function has been added to convert between them if needed.

### Oplog changes
The most important oplog changes:

- `ImportedFunctionInvoked` has been renamed to `HostCall`
- `ExportedFunctionInvoked` and `ExportedFunctionCompleted` is now `AgentInvocationStarted` and `AgentInvocationFinished`
- More importantly all invocation-related oplog entry (the above ones and `PendingAgentInvocation`) are now using the `AgentInvocation` and `AgentInvocationPayload` types to talk about invocations.

`AgentInvocation` is a sum type describing all the possible invocations that we support now:
- `AgentInitialization` means the agent's `initialize` is being called
- `AgentMethod` calls an arbitrary agent method through the agent `invoke` interface
- `SaveSnapshot` and `LoadSnapshot` are special, non-agentic calls for snapshotting
- `ProcessOplogEntries` is a special non-agentic call for oplog processor plugins
- `ManualUpdate` is a placeholder in the invocation queue to execute a snapshot-based update

### Executor internals

- The worker executor's internals are now everywhere passing around the above mentioned `AgentInvocation` and its related types. This becomes _lowered_ to raw wasmtime invocation just at the end of the invocation chain. 
- The special wasi-http invocation type has been removed. 
- Principal is propagated everywhere it's needed
- RPC, worker-proxy and scheduler all using the new invocation APIs

### Others

- Significant improvements to the `IntoValue`/`FromValue` derivation macros, and explicit unit test verifying that they produce the same type that's in WIT (for `oplog-entry`)
- The static moonbit wrapper generator has been completely removed from the CLI (and all templates)
- As the CLI's invoke command no longer works through the static WIT wrapper, we lost the ability to call endpoints with non-component-model parameters (unstructured text/binary and multimodal) as they cannot be encoded using the invoke command's WAVE syntax. Note that we can still invoke these things through the new REPLs (and repl-based "eval" like commands)

### Out of scope
The agent-id format does not change by this PR - agent-type name is still kebab-cased and the arguments are using the WAVE syntax extended with custom syntax for unstructured/multimodal. The invocation API itself is also expecting agent type name to be the kebab-cased variant. 
The next task in the epic will consider changing these (#2835)

### DIFF
--- a/golem-registry-service/src/services/domain_registration/aws_config.rs
+++ b/golem-registry-service/src/services/domain_registration/aws_config.rs
@@ -25,7 +25,7 @@ impl AwsConfig {
         Self { sdk_config }
     }
 
-    pub async fn new(_token: &str, _role_arn: &str) -> Self {
+    pub async fn new() -> Self {
         // In tests this is called with dummy values; the default credential chain
         // will resolve credentials from the environment (including web identity token).
         let sdk_config = aws_config::load_defaults(BehaviorVersion::latest()).await;
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -332,7 +332,7 @@ jobs:
         env:
           QUIET: true
         run: cargo make --profile ci ${{ matrix.group.name }}
-        timeout-minutes: 15
+        timeout-minutes: 20
       - name: Publish Test Report
         uses: ctrf-io/github-test-reporter@v1
         if: always()
--- a/cli/golem-cli/src/model/invoke_result_view.rs
+++ b/cli/golem-cli/src/model/invoke_result_view.rs
@@ -15,6 +15,7 @@
 use crate::log::log_error;
 use anyhow::{anyhow, bail};
 use golem_client::model::AgentInvocationResult;
+use golem_common::model::agent::wit_naming::ToWitNaming;
 use golem_common::model::agent::{AgentType, DataSchema, DataValue, ElementSchema};
 use golem_common::model::IdempotencyKey;
 use golem_wasm::{print_value_and_type, ValueAndType};
@@ -76,7 +77,7 @@ impl InvokeResultView {
                 }
             };
             let analysed_type = match &first_schema.schema {
-                ElementSchema::ComponentModel(cm) => cm.element_type.clone(),
+                ElementSchema::ComponentModel(cm) => cm.element_type.to_wit_naming(),
                 _ => {
                     log_error("Non-ComponentModel output schema not supported for result display");
                     return None;
@@ -133,7 +134,7 @@ impl InvokeResultView {
             .ok_or_else(|| anyhow!("Empty output schema"))?;
 
         let analysed_type = match &first_schema.schema {
-            ElementSchema::ComponentModel(cm) => cm.element_type.clone(),
+            ElementSchema::ComponentModel(cm) => cm.element_type.to_wit_naming(),
             _ => bail!("Non-ComponentModel output schema not supported for WAVE formatting"),
         };
 
--- a/cli/golem-cli/tests/app/agents.rs
+++ b/cli/golem-cli/tests/app/agents.rs
@@ -411,54 +411,55 @@ async fn test_rust_code_first_with_rpc_and_all_types() {
 
     run_and_assert(&ctx, "fun-enum-with-only-literals", &["a"]).await;
 
-    run_and_assert(
-        &ctx,
-        "rust:agent/foo-agent.{fun-multi-modal}",
-        &[r#"[text("foo"), text("foo"), data({id: 1, name: "foo"})]"#],
-    )
-    .await;
-
-    run_and_assert(
-        &ctx,
-        "rust:agent/foo-agent.{fun-multi-modal-basic}",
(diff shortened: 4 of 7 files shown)


golem-registry-service/src/services/domain_registration/aws_config.rs
```diff
@@ -12,81 +12,35 @@
...
-            duration_seconds: None,
-            policy: None,
-            policy_arns: None,
-        };
-
-        AwsConfig {
-            credentials_provider,
-            region: Region::default(),
-        }
+    pub async fn new(_token: &str, _role_arn: &str) -> Self {
```
mschuwalow:
```suggestion
    pub async fn new() -> Self {
```

golem-service-base/tests/blob_storage.rs
```diff
@@ -112,12 +113,19 @@ impl Debug for S3Test {
...
 impl GetBlobStorage for S3Test {
     async fn get_blob_storage(&self) -> Arc<dyn BlobStorage + Send + Sync> {
-        let container = tryhard::retry_fn(|| MinIO::default().start())
-            .retries(5)
-            .exponential_backoff(Duration::from_millis(10))
-            .max_delay(Duration::from_secs(10))
-            .await
-            .expect("Failed to start MinIO");
+        let container = tryhard::retry_fn(|| {
+            GenericImage::new("minio/minio", "RELEASE.2025-01-20T14-49-07Z")
```
mschuwalow, line 117: why switch to a specific image? This is another things that will need to be updated

Makefile.toml
```diff
@@ -133,22 +134,7 @@ private = true
 script_runner = "@duckscript"
 script = """
 components = array
-array_push ${components} agent-counters/.wit
```
mschuwalow, line 136: Looking at the manifests of the test-components these are still needed, no?

vigoo: WIT dirs are not needed anymore. That should be updated in the manifests

golem-service-base/tests/blob_storage.rs
vigoo: The old one (which was `Minio::default()`) was failing with the updated aws crates (which I updated to reduce some redundant transitive dependencies to try to make the build faster).
Maybe it could have been updated by updating `testcontainers_modules` instead but that affects more things

PR #2833, opened 2026-02-24
Mcp integration

### LINKED ISSUE #2792: Implement Server layer with Handshake protocol
This will cover actual server exposing, JSON RPC message handling `initialize`, `notifications/initialized` , and `tool/list` methods . See the prototype and how it is done using rmcp as well as manual handshake implementation  (both works)

The actual protocol talks about streamable http, SSE (with lots of features) and single http response too. My experimentation shows to ensure single http response works too and also shows streamable http works out of the box with Rust SDK. 

Anyway, here is the bare minimum protocol that has to be going regardless of whether we use a library or do it by ourselves. 

https://github.com/afsalthaj/simple-mcp-server/blob/master/README.md#manul-server-handshake-details-the-bare-minimum-that-has-to-be-there-in-golem-regardless-of-using-sdk-or-not

```rust
cloudflared tunnel --url http://localhost:8080

# returning https://telecom-exam-chances-developer.trycloudflare.com
```

### Initialize  Request from Open AI client

```rust
curl -i -X POST https://telecom-exam-chances-developer.trycloudflare.com/my-agent/mcp \
                            -H "Content-Type: application/json" \
                            -H "Accept: application/json" \
                            -d '{
                          "jsonrpc": "2.0",
                          "id": 1,
                          "method": "initialize",
                          "params": {
                            "protocolVersion": "2025-06-18",
                            "capabilities": {},
                            "clientInfo": {
                              "name": "curl-client",
                              "version": "1.0.0"
                            }
                          }
                        }'
HTTP/1.1 200 OK
content-type: application/json
mcp-session-id: f05ad43b-7641-4467-8202-b3e4ed4e9e1d
content-length: 169
date: Wed, 18 Feb 2026 05:32:17 GMT

{"id":1,"jsonrpc":"2.0","result":{"capabilities":{"tools":{"listChanged":false}},"protocolVersion":"2025-06-18","serverInfo":{"name":"golem-mcp-dev","version":"0.1.0"}}}⏎
```

### Handling Notification/Initialized Handshake from Client

```rust
curl -i -X POST https://telecom-exam-chances-developer.trycloudflare.com/my-agent/mcp \
  -H "Content-Type: application/json" \
  -H "Accept: application/json" \
  -H "Mcp-Session-Id: 59fcd8cf-0399-481a-97ed-24284327b880" \
  -d '{
    "jsonrpc": "2.0",
    "method": "notifications/initialized",
    "params": {}
  }'

```

### Capability Listing

```rust
curl -X POST https://telecom-exam-chances-developer.trycloudflare.com/my-agent/mcp \
                            -H "Content-Type: application/json" \
                            -H "Accept: application/json" \
                            -H "Mcp-Session-Id: f05ad43b-7641-4467-8202-b3e4ed4e9e1d" \
                            -d '{
                          "jsonrpc": "2.0",
                          "id": 2,
                          "method": "tools/list",
                          "params": {}
                        }'

{"id":2,"jsonrpc":"2.0","result":{"tools":[{"description":"Returns a greeting","inputSchema":{"properties":{"name":{"type":"string"}},"required":["name"],"type":"object"},"name":"hello_agent","title":"Hello Agent"}]}}⏎

```

Since we are exposing handlers and the actual server, this task should handle how to encode the constructor parameters which directly relates to **Statefulness.** 

More about Statefulness: In all the above example the MCP server URL is holding only the agent name. However we need to see how to encode the constructor parameters that dictates the statefulness of the MCP server. The actual tool call is when the api-gateway/worker-service actually interact with an agent. 

### Actual Tool Call (not in this task/PR, because this requires more work)

See task 4 for more details, not a concern of this PR

```rust
curl -X POST https://telecom-exam-chances-developer.trycloudflare.com/my-agent/mcp \
  -H "Content-Type: application/json" \
  -H "Accept: application/json" \
  -H "Mcp-Session-Id: f05ad43b-7641-4467-8202-b3e4ed4e9e1d" \
  -d '{
    "jsonrpc": "2.0",
    "id": 3,
    "method": "tools/call",
    "params": {
      "name": "hello_agent",
      "arguments": {
        "name": "Alice"
      }
    }
  }'

{"id":3,"jsonrpc":"2.0","result":{"content":[{"text":"Hello, Alice!","type":"text"}],"isError":false}}⏎
```

In summary, the outcome should be a bare minimum handshake protocol so that golem MCP servers can be connected with standard clients such as https://platform.openai.com/chat/edit?models=gpt-4.1

### LINKED ISSUE #2793: Implement tool/call handler in MCP
Once the handshake is done, we are already able to connect to MCP server and list capabilities. Now we need to implement `tool/calls` handler. Note

```rust
curl -X POST http://localhost:8080/**my-agent**/mcp \
  -H "Content-Type: application/json" \
  -H "Accept: application/json" \
  -H "Mcp-Session-Id: f05ad43b-7641-4467-8202-b3e4ed4e9e1d" \
  -d '{
    "jsonrpc": "2.0",
    "id": 3,
    "method": "tools/call",
    "params": {
      "name": "counter",
      "arguments": {
        "number": 1
      }
    }
  }'

{"id":3,"jsonrpc":"2.0","result":{"content":[{"text":"Hello, Alice!","type":"text"}],"isError":false}}⏎
```

The handler implementation is actually invoke the agent method and return a response.

Also try in an actual MCP client (don’t create any manual one), and golem tool invocation should work (something like below in playgrounds)

For example: This should be the output in openai playground once tool support is implemented

Image

### LINKED ISSUE #2837: CLI integration for MCP

afsalthaj:
Fixes #2792 
Fixes #2793
Fixes #2837  (I had to do this straight away, to ensure whatever I did in the backend to make mcp part of the plan etc works). 

Also merged tool schema mapping into this: https://github.com/golemcloud/golem/pull/2847. This is simple tool schema mapping and value mapping (value mapping is almost noop - we just use existing value_and_type.to_json_value) under the hood, to avoid any duplicate/vunmaintainable logic. JSON schema mapping will adhere to whatever golem_wasm code does in terms of converting a json value to/from ValueAndType.  Also note that text/binary/multimodal will be done once I am done with resource ticket

Bootstrapping MCP server, the flattened version as documented.

And the whole thing works with prominent AIs such as : https://platform.openai.com/chat/edit?models=gpt-4.1-mini

# Testing with Open AI

A new golem rust app with the following in golem.yaml

```yaml
mcp:
  deployments:
    local:
       # public simply forwarding to local mcp server 
      - domain: shoot-vincent-attached-med.trycloudflare.com
        agents:
          counter-agent: {}
```
* golem deploy

image

The name of the tool is a bit odd, bu I will be fixing it as part of cleaning things up. The above screenshot successfully handshaked with Golem and listed tools

# User experience using MCP Server 

The actual usage looks very reasonable. The fact that constructor parameters being part of the tool itself seems to be very reasonable as well as flexible. 

For example see this specific situation. Here name is a constructor argument 

image

Here I just prompted to increment by 10, but forgot to mention which counter (the value of the constructor param) and client tell us the user to give more details.

## Statefulness 

The statefulness is about being able to hit the same identity of the agent, that's it.

image

The agent that was used:

```rust
use golem_rust::{agent_definition, agent_implementation, endpoint};

#[agent_definition]
pub trait CounterAgent {
    fn new(name: String) -> Self;

    fn increment_by(&mut self, num: u32) -> u32;
}

struct CounterImpl {
    _name: String,
    count: u32,
}

#[agent_implementation]
impl CounterAgent for CounterImpl {
    fn increment_by(&mut self, num: u32) -> u32 {
        self.count += 1;
        self.count
    }
}


```

### Input Schema and Output Schema and different AI behaviours.

Note that in OpenAI, I couldn't see it really considering output-schema. It did care about input-schema, and I could actually see the final response (json value). This questions whether our output schema works. This may or may not be our problem, and the only way to test this manually doing a handshaking following MCP protocol and see if the response schema is detailed. And for us it works.

 See this https://github.com/golemcloud/golem/pull/2833#issuecomment-3956579180

### DIFF
--- a/golem-registry-service/tests/repo/common.rs
+++ b/golem-registry-service/tests/repo/common.rs
@@ -1646,17 +1646,15 @@ pub async fn test_mcp_deployment_list_and_delete(deps: &Deps) {
 
     assert!(deployments.len() == 2);
 
-    // Try to delete with old revision (should fail with ConcurrentModification)
     let delete_with_old_revision = deps
         .mcp_deployment_repo
-        .delete(user.revision.account_id, deployment_id, 0)
+        .delete(user.revision.account_id, deployment_id, 1)
         .await;
 
     let_assert!(Err(McpDeploymentRepoError::ConcurrentModification) = delete_with_old_revision);
 
-    // Delete with correct revision
     deps.mcp_deployment_repo
-        .delete(user.revision.account_id, deployment_id, 1)
+        .delete(user.revision.account_id, deployment_id, 2)
         .await
         .unwrap();
 


golem-registry-service/db/migration/postgres/004_mcp_deployment.sql
```diff
@@ -0,0 +1,58 @@
...
+CREATE INDEX mcp_deployment_revisions_latest_revision_by_id_idx
+    ON mcp_deployment_revisions (mcp_deployment_id, revision_id DESC);
+
+CREATE TABLE deployment_compiled_mcp
+(
+    account_id                UUID   NOT NULL,
+    environment_id            UUID   NOT NULL,
+    deployment_revision_id    BIGINT NOT NULL,
+    domain                    TEXT   NOT NULL,
+    agent_type_implementers   TEXT   NOT NULL,
```
afsalthaj: Will change this.

afsalthaj:
The way we are going to write integration test is as follows:
We acting as a client and we following the MCP protocol

### Initialization

```sh
curl -i -X POST https://restaurant-hamburg-sims-los.trycloudflare.com/mcp \
           -H "Content-Type: application/json, text/event-stream" \
           -H "Accept: application/json, text/event-stream" \
           -d '{
         "jsonrpc": "2.0",
         "id": 1,
         "method": "initialize",
         "params": {
           "protocolVersion": "2024-11-05",
           "capabilities": {},
           "clientInfo": {
             "name": "curl-client",
             "version": "1.0.0"
           }
         }
       }'
```


```
HTTP/2 200
date: Wed, 25 Feb 2026 03:51:51 GMT
content-type: text/event-stream
server: cloudflare
cf-ray: 9d343f1f3d78d87d-MEL
cf-cache-status: DYNAMIC
cache-control: no-cache
mcp-session-id: 20718598-6ac9-4503-a683-80e999db5044

data:
id: 0
retry: 3000

data: {"jsonrpc":"2.0","id":1,"result":{"protocolVersion":"2024-11-05","capabilities":{"prompts":{},"resources":{},"tools":{}},"serverInfo":{"name":"rmcp","version":"0.16.0"},"instructions":"This server provides  tools related to agent in golem and prompts. Tools: increment, decrement, get_value, say_hello, echo, sum. Prompts: example_prompt (takes a message), counter_analysis (analyzes counter state with a goal)."}}
```

### Inititalized Notification request
```sh
curl -i -X POST https://restaurant-hamburg-sims-los.trycloudflare.com/mcp \
           -H "Content-Type: application/json" \
           -H "Accept: application/json, text/event-stream" \
           -H "Mcp-Session-Id: 20718598-6ac9-4503-a683-80e999db5044" \
           -d '{
         "jsonrpc": "2.0",
         "method": "notifications/initialized",
         "params": {}
       }'
```


```
HTTP/2 202
date: Wed, 25 Feb 2026 03:52:59 GMT
content-length: 0
server: cloudflare
cf-ray: 9d3440c56bbef0ca-MEL
cf-cache-status: DYNAMIC

```

### And finally, a simple tool/list call

```sh
curl -X POST http://restaurant-hamburg-sims-los.trycloudflare.com/mcp \
           -H "Content-Type: application/json" \
           -H "Accept: application/json, text/event-stream" \
           -H "Mcp-Session-Id: 20718598-6ac9-4503-a683-80e999db5044" \
           -d '{
         "jsonrpc": "2.0",
         "id": 2,
         "method": "tools/list",
         "params": {}
       }'
```


```

data: {"jsonrpc":"2.0","id":2,"result":{"_meta":{"tool_meta_key":"tool_meta_value"},"tools":[{"name":"CounterAgent-increment_by","description":"","inputSchema":{"properties":{"num":{"type":"integer"}},"type":"object"},"outputSchema":{"properties":{"return-value":{"type":"integer"}},"type":"object"}}]}}
id: 0/0
```

golem-common/src/model/diff/mcp_deployment.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+impl Diffable for McpDeploymentAgentOptions {
+    type DiffResult = McpDeploymentAgentOptionsDiff;
+
+    fn diff(_new: &Self, _current: &Self) -> Option<Self::DiffResult> {
+        // TODO: Implement diff when configuration options are added
+        None
+    }
+}
```
afsalthaj, line 38: These are kept for consistency, and also we will have such configs coming soon in MCP similar to http

afsalthaj:

afsalthaj:
Note that further improvements are on its way as part of resolving next set of tickets: https://github.com/golemcloud/golem/pull/2847

I will update this branch with main to change the invoke etc soon. (Trying to fix other bugs before that)

cli/golem/src/command_handler.rs
```diff
@@ -52,6 +52,7 @@ impl CommandHandlerHooks for ServerCommandHandler {
                     router_addr: args.router_addr().to_string(),
                     router_port: args.router_port(),
                     custom_request_port: args.custom_request_port(),
+                    mcp_port: args.mcp_port(),
```
vigoo, line 55: Does it have to be a separate port? I imagined it's a special part of the deployed APIs (going on `custom_request_port`)

golem-api-grpc/proto/golem/mcp/core.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package golem.mcp;
+
+import "golem/common/environment.proto";
+import "golem/common/account_id.proto";
+
+message AgentTypeImplementer {
```
vigoo: This protobuf type already exists (`RegisteredAgentTypeImplementer`)

golem-worker-service/src/mcp/invoke.rs
```diff
@@ -0,0 +1,284 @@
...
+
+    let worker_id = WorkerId {
+        component_id: mcp_tool.component_id,
+        worker_name: agent_id.to_string(),
+    };
+
+    let function_name = "golem:agent/guest.{invoke}".to_string();
+
+    let result = worker_service
+        .invoke_and_await_owned_agent(
```
vigoo: This method does not exist anymore :) please merge to the new agent invocation

cli/golem/src/command_handler.rs
afsalthaj: It has to be separate port based. Custom request port is dedicated to http apis

golem-worker-service/src/mcp/invoke.rs
afsalthaj: Done and it worked

golem-worker-service/src/mcp/invoke.rs
```diff
@@ -34,7 +32,9 @@ pub async fn agent_invoke(
     let constructor_params = extract_parameters_by_schema(
         &args_map,
         &mcp_tool.constructor.input_schema,
-        |value_and_type| value_and_type,
+        |value_and_type| ComponentModelElementValue {
```
vigoo, line 35: Are you going to add in a next step support for multimodal and unstructured text / binary? My understanding was that we added those primarily for this mcp server

vigoo:
I would like to ask @mschuwalow or @noise64 to review the registry service parts and @noise64 for the cli integration

golem-registry-service/src/api/mod.rs
```diff
@@ -58,7 +60,7 @@ pub type Apis = (
         EnvironmentsApi,
         EnvironmentSharesApi,
     ),
-    HttpApiDeploymentsApi,
+    (HttpApiDeploymentsApi, McpDeploymentsApi),
```
mschuwalow: I would not group these unless you are hitting the limit of number of api tuple elements

golem-registry-service/src/repo/model/deployment.rs
```diff
@@ -223,6 +258,7 @@ impl DeploymentIdentity {
                     )
                 })
                 .collect(),
+            mcp_deployments: Default::default(),
```
mschuwalow: This is wrong. You need to construct the diffing representation here.

golem-registry-service/src/repo/model/deployment.rs
```diff
@@ -407,6 +468,50 @@ impl DeploymentRevisionCreationRecord {
     }
 }
 
+#[derive(Debug, Clone, BinaryCodec)]
+pub struct CompiledMcpData {
+    pub implementers: golem_service_base::mcp::AgentTypeImplementers,
+}
+
+#[derive(FromRow)]
+pub struct DeploymentMcpCapabilityRecord {
```
mschuwalow: Let's match the table names where possible, so DeploymentCompiledMcpRecord

golem-registry-service/src/repo/mcp_deployment.rs
```diff
@@ -0,0 +1,468 @@
...
+        self.db_pool
+            .with_tx_err(METRICS_SVC_NAME, api_name, f)
+            .await
+    }
+}
+
+#[trait_gen(PostgresPool -> PostgresPool, SqlitePool)]
+#[async_trait]
+impl McpDeploymentRepo for DbMcpDeploymentRepo<PostgresPool> {
+    async fn create(
```
mschuwalow, line 269: For manifest managed entities we tend to undelete an existing entity if it has matching environment-level unique key (domain in your case).
Check components or http-api-deployments for examples.

golem-registry-service/src/repo/mcp_deployment.rs
```diff
@@ -0,0 +1,468 @@
...
+
+    async fn delete(
+        &self,
+        user_account_id: Uuid,
+        mcp_deployment_id: Uuid,
+        revision_id: i64,
+    ) -> Result<(), McpDeploymentRepoError> {
+        self.with_tx_err("delete", |tx| {
+            async move {
+                // Check that the current revision matches the provided revision
```
mschuwalow: Don't do that. The service layer is responsible for checking this when computing the update.

The repo should only check for concurrent updates using optimistic concurrency control (i.e. unique index violations because of duplicate id + revision)

golem-registry-service/db/migration/sqlite/004_mcp_deployment.sql
```diff
@@ -0,0 +1,74 @@
...
+    mcp_deployment_id TEXT      NOT NULL,
+    revision_id       INTEGER   NOT NULL,
+
+    hash              BLOB      NOT NULL,
+
+    created_at        TIMESTAMP NOT NULL,
+    created_by        TEXT      NOT NULL,
+    deleted           BOOLEAN   NOT NULL,
+
+    domain            TEXT      NOT NULL,
```
mschuwalow: domain is immutable and should not be part of the revision table

golem-registry-service/src/repo/mcp_deployment.rs
```diff
@@ -0,0 +1,468 @@
...
+                        .bind(&domain)
+                        .bind(&revision.audit.created_at)
+                        .bind(revision.audit.created_by),
+                    )
+                    .await
+                    .to_error_on_unique_violation(McpDeploymentRepoError::McpDeploymentViolatesUniqueness)?;
+
+                let revision = revision.with_updated_hash();
+
+                let revision: McpDeploymentRevisionRecord = tx
```
mschuwalow: Add an insert_revision helper like the other repos, please

golem-registry-service/src/repo/mcp_deployment.rs
```diff
@@ -0,0 +1,468 @@
...
+                        "# })
+                        .bind(revision.revision_id)
+                        .bind(revision.mcp_deployment_id),
+                    )
+                    .await?;
+
+                Ok(McpDeploymentExtRevisionRecord {
+                    environment_id: mcp_deployment.0,
+                    domain: revision.domain.clone(),
+                    entity_created_at: revision.audit.created_at.clone(),
```
mschuwalow: this is wrong. This needs to be the created at time of the parent entity (i.e. creation time of the first revision)

golem-registry-service/src/repo/mcp_deployment.rs
```diff
@@ -0,0 +1,468 @@
...
+                        .bind(revision.revision_id)
+                        .bind(revision.hash)
+                        .bind(&revision.domain)
+                        .bind(&revision.data)
+                        .bind(&revision.audit.created_at)
+                        .bind(revision.audit.created_by),
+                    )
+                    .await?;
+
+                // Fetch environment_id from mcp_deployments
```
mschuwalow: this comment is very misleading imo

golem-registry-service/src/repo/mcp_deployment.rs
```diff
@@ -0,0 +1,468 @@
...
+                let mcp_deployment: (Uuid,) = tx
```
mschuwalow: Just fetch the whole McpDeploymentRecord here. The db is touching that row anyway.

golem-registry-service/src/repo/mcp_deployment.rs
```diff
@@ -0,0 +1,468 @@
...
+                        .bind(mcp_deployment_id),
+                    )
+                    .await?;
+
+                if current_revision.0 != revision_id {
+                    return Err(McpDeploymentRepoError::ConcurrentModification);
+                }
+
+                // Get the current domain from the current revision
+                let current_domain: (String,) = tx
```
mschuwalow: The domain should not be part of the revision record and we should not be fetching it here.

golem-registry-service/src/repo/mcp_deployment.rs
```diff
@@ -0,0 +1,468 @@
...
+                    .bind(&deletion_data)
+                    .bind(user_account_id),
+                )
+                .await?;
+
+                // Update the main table to point to the deletion revision
+                tx.execute(
+                    sqlx::query(indoc! { r#"
+                        UPDATE mcp_deployments
+                        SET deleted_at = CURRENT_TIMESTAMP, modified_by = $1, current_revision_id = $2
```
mschuwalow: use the timestamp from the audit.

golem-registry-service/src/services/mcp_deployment.rs
```diff
@@ -0,0 +1,445 @@
...
+                EnvironmentError::EnvironmentNotFound(_) => {
+                    McpDeploymentError::McpDeploymentNotFound(mcp_deployment_id)
+                }
+                other => other.into(),
+            })?;
+
+        auth.authorize_environment_action(
+            environment.owner_account_id,
+            &environment.roles_from_active_shares,
+            EnvironmentAction::ViewHttpApiDeployment,
```
mschuwalow: Please add dedicated environment actions for McpDeployments.

golem-registry-service/src/services/mcp_deployment.rs
```diff
@@ -0,0 +1,445 @@
...
+            &environment.roles_from_active_shares,
+            EnvironmentAction::UpdateHttpApiDeployment,
+        )?;
+
+        if update.current_revision != mcp_deployment.revision {
+            Err(McpDeploymentError::ConcurrentUpdate)?
+        };
+
+        mcp_deployment.revision = mcp_deployment.revision.next()?;
+        if let Some(domain) = update.domain {
```
mschuwalow: Domain should be immutable.

golem-service-base/src/clients/registry.rs
```diff
@@ -173,6 +174,11 @@ pub trait RegistryService: Send + Sync {
         domain: &Domain,
     ) -> Result<CompiledRoutes, RegistryServiceError>;
 
+    async fn get_active_mcp_capabilities_for_domain(
```
mschuwalow: The mcp capability name only shows up for the grpc apis. Let's just call this `get_active_compiled_mcps_for_domain`

golem-common/src/base_model/mcp_deployment.rs
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    pub struct McpDeploymentCreation {
+        pub domain: Domain,
+        pub agents: BTreeMap<AgentTypeName, McpDeploymentAgentOptions>,
+    }
+
+    pub struct McpDeploymentUpdate {
+        pub current_revision: McpDeploymentRevision,
+        pub domain: Option<Domain>,
```
mschuwalow: Domain should not be updatable as it's part of the logical primary key (environment_id + domain) of the entity.

golem-registry-service/src/services/deployment/write.rs
```diff
@@ -249,17 +261,32 @@ impl DeploymentWriteService {
...
+            http_api_deployments.len(),
+            mcp_deployments.len()
+        );
+
+        let account_id = environment.owner_account_id;
+        let deployment_context = DeploymentContext::new(
+            environment,
+            components,
+            http_api_deployments,
+            mcp_deployments.clone(),
```
mschuwalow: I would suggest not cloning here but rather just taking the mcp-deployments back from the context once you are done with it, similar to the other entities.

golem-worker-service/src/mcp/schema/mcp_schema_mapping.rs
```diff
@@ -0,0 +1,31 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::mcp::schema::mcp_schema::McpSchema;
+use golem_common::base_model::agent::DataSchema;
+
+pub trait GetMcpSchema {
```
mschuwalow: Let's not have traits that we are only implementing for a single type. Let's make this a plain function

golem-worker-service/src/mcp/schema/mcp_tool_schema.rs
```diff
@@ -0,0 +1,38 @@
...
+use crate::mcp::schema::mcp_schema::McpSchema;
+use crate::mcp::schema::mcp_schema_mapping::GetMcpSchema;
+use golem_common::base_model::agent::AgentMethod;
+
+pub struct McpToolSchema {
+    pub input_schema: McpSchema,
+    pub output_schema: Option<McpSchema>,
+}
+
+pub trait GetMcpToolSchema {
```
mschuwalow: same here

golem-worker-service/src/mcp/mcp_capabilities_lookup.rs
```diff
@@ -0,0 +1,106 @@
...
+        Self {
+            registry_service_client,
+        }
+    }
+}
+
+#[async_trait]
+impl McpCapabilityLookup for RegistryServiceMcpCapabilityLookup {
+    async fn get(&self, domain: &Domain) -> Result<CompiledMcp, McpCapabilitiesLookupError> {
+        self.registry_service_client
```
mschuwalow, line 71: This needs to be cached. We don't want a request to the registry service on every new mcp session.
Disable the cache in the single-executable to have better dx there.

golem-worker-service/src/mcp/mcp_capabilities_lookup.rs
```diff
@@ -0,0 +1,106 @@
...
+            .ok_or_else(|| {
+                McpCapabilitiesLookupError::InternalError(anyhow::anyhow!(
+                    "Agent type {} not found in MCP for domain {}",
+                    agent_type_name.0,
+                    domain.0
+                ))
+            })?;
+
+        self.registry_service_client
+            .get_agent_type(
```
mschuwalow, line 97: This needs to be part of the compiled mcp structure. There should only be one request per uncached domain to the registry service to fetch all data needed to serve the mcp request.

golem-worker-service/src/mcp/mcp_capabilities_lookup.rs
```diff
@@ -0,0 +1,106 @@
...
+impl SafeDisplay for McpCapabilitiesLookupError {
+    fn to_safe_string(&self) -> String {
+        match self {
+            McpCapabilitiesLookupError::InternalError(_) => "Internal error".to_string(),
+            McpCapabilitiesLookupError::UnknownSite(_) => "Unknown authority".to_string(),
+        }
+    }
+}
+
+// Note: No caching here, the caching is part of MCP session
```
mschuwalow, line 55: mcp session is only one user interaction, no? So this would still result in a request for each new user session.

golem-worker-service/src/mcp/agent_mcp_server.rs
```diff
@@ -0,0 +1,292 @@
...
+                tracing::info!("No session ID found in headers");
+            }
+
+            if let Some(host) = parts.headers.get("host") {
+                let domain = Domain(host.to_str().unwrap().to_string());
+                let tool_router = self.tool_router(&domain).await;
+                for tool in tool_router.list_all() {
+                    self.tools.insert(tool.name.to_string(), tool);
+                }
+                *self.domain.write().await = Some(domain);
```
mschuwalow, line 285: How does this work? If each client gets a clone of the GolemAgentMcpServer you are going to clobber the data in the Arc RwLock here

mschuwalow (changes requested):

golem-common/src/model/diff/http_api_deployment.rs
```diff
@@ -49,6 +49,8 @@ impl Diffable for HttpApiDeploymentAgentOptions {
     }
 }
 
+// TODO; peal off non-user things from mcp-deployment
+//
```
mschuwalow: ### 
```suggestion
```

golem-registry-service/src/grpc/api_impl.rs
```diff
@@ -392,15 +398,35 @@ impl RegistryServiceGrpcApi {
         &self,
         request: GetActiveRoutesForDomainRequest,
     ) -> Result<GetActiveRoutesForDomainSuccessResponse, GrpcApiError> {
+        use golem_api_grpc::proto::golem::customapi::CompiledRoutes;
+
         let domain: Domain = Domain(request.domain);
 
-        let compiled_routes = self
+        let compiled_routes_internal = self
```
mschuwalow: why this naming change?

golem-worker-service/src/mcp/invoke.rs
afsalthaj: Yes

golem-registry-service/src/grpc/api_impl.rs
afsalthaj: No... that was just accidental. Will revert

golem-worker-service/src/mcp/agent_mcp_server.rs
afsalthaj: It's not a clone. StreamableHttpService::new takes a factory closure` || move { Ok(GolemAgentMcpServer::new(..))} }. Each new mcp client session calls this closure and gets a brand new server with its own independent Arc<RwLock... etc. So clients don't share it.

golem-worker-service/src/mcp/mcp_capabilities_lookup.rs
afsalthaj: Will cache it as we discussed. That's a comment I forgot to remove

golem-worker-service/src/mcp/agent_mcp_server.rs
afsalthaj: The derive Clone exist to satisfy `rmcp` and not for any of our internal logic

golem-worker-service/src/mcp/mcp_capabilities_lookup.rs
afsalthaj: Agreed. Could we do that as part of next PR along with resource support ? There is a lot more to come. 

This ticket was more of just bootstrapping things and it's already blown up

golem-worker-service/src/mcp/mcp_capabilities_lookup.rs
afsalthaj: Yes. We can do that. We should do that. I have added caching and extending compiled_mcp in upcoming tickets.

The PR is already sort of ending up to be "implement the whole MCP" which is not the case.

golem-registry-service/src/repo/model/deployment.rs
afsalthaj: Fixed

golem-registry-service/src/api/mod.rs
afsalthaj: Fixed

golem-registry-service/src/repo/mcp_deployment.rs
afsalthaj: Resolved

golem-worker-service/src/mcp/schema/mcp_tool_schema.rs
afsalthaj: It's an opinion not a mandate. Several places, I can find people doing it for discovering capabilities over a type instead of randomly searching the functions available in big code base. But I don't mind :) So fixed

golem-common/src/base_model/mcp_deployment.rs
afsalthaj: Makes sense. Fixed

golem-worker-service/src/mcp/mcp_capabilities_lookup.rs
afsalthaj: I will be caching it in upcoming tickets. Note that, this is not computed per session. It's when you add the tool into the MCP clients for the most AIs. But yes, we should cache it, as we cannot guarantee it (it depends between claude openai etc). But this is tracked as part of details in upcoming tickets.

image

afsalthaj:
@mschuwalow  https://github.com/golemcloud/golem/issues/2864

golem-worker-service/src/bootstrap.rs
```diff
@@ -43,6 +44,7 @@ pub struct Services {
     pub component_service: Arc<dyn ComponentService>,
     pub worker_service: Arc<WorkerService>,
     pub request_handler: Arc<RequestHandler>,
+    pub mcp_capability_lookup: Arc<dyn McpCapabilityLookup + Sync + Send + 'static>,
```
mschuwalow, line 47:
```suggestion
    pub mcp_capability_lookup: Arc<dyn McpCapabilityLookup>,
```

McpCapabilityLookup already requires Send + Sync and trait objects are 'static [by default](https://doc.rust-lang.org/stable/reference/lifetime-elision.html#default-trait-object-lifetimes)

PR #2834, opened 2026-02-24
CLI, first class lang support: on demand common

### LINKED ISSUE #2729: Per-language embedded build files automatically extracted/updated into the app's temp directory
Especially for TS, we have a set of files such as `rollup.config`, `tsconfig.json` etc. These could also be removed from the template and be part of the “embedded language template” in the CLI, automatically extracted during `golem build` to some temp directory (`golem-temp`).

Note that this is a more generic solution than the *“1. Named build templates embedded in CLI”* idea, as that can be viewed as support for referencing an embedded app manifest snippet. So if we do this, we don’t need to do that (at least first).

**Advantages:**

- These files are also hidden from the user, having a clean “only source file(s)” experience
- They can still be observed (in the temp directory) for gaining understanding / debugging purposes, if needed
- Updating the CLI (running build with a new version of the CLI) can update these build instructions automatically

**Disadvantages:**

- Users can no longer customize these (mostly language-specific and not golem-specific)
- Some files *must* be in specific places (such as the project’s root dir, or `project` in Scala, etc)

At this point we eliminated the whole `common-ts/common-rust`  directories from the golem apps.

noise64:
- resolves #2729
- cleans up the app_template module

### DIFF
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -42,10 +42,12 @@ uuid = { workspace = true, features = ["v4"] }
 serde_yaml = { workspace = true }
 
 [dev-dependencies]
-backtrace-on-stack-overflow = { workspace = true }
 goldenfile = { workspace = true }
 test-r = { workspace = true }
 wac-graph = { workspace = true }
+
+[target.'cfg(unix)'.dev-dependencies]
+backtrace-on-stack-overflow = { workspace = true }
 
 [lib]
 harness = false
--- a/integration-tests/benchmark_suites/ci.yaml
+++ b/integration-tests/benchmark_suites/ci.yaml
@@ -16,13 +16,6 @@ benchmarks:
     length: [ 5 ]
     disableCompilationCache: true
 
-  - name: cold-start-unknown-large
-    iterations: 1
-    clusterSize: [ 1 ]
-    size: [ 1, 5, 10 ]
-    length: [ 10 ]
-    disableCompilationCache: true
-
   - name: cold-start-unknown-small
     iterations: 1
     clusterSize: [ 1 ]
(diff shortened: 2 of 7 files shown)


noise64:

cli/golem-cli/src/lib.rs
```diff
@@ -47,6 +48,83 @@ test_r::enable!();
...
+#[macro_export]
+macro_rules! app_manifest_version {
+    () => {
+        "1.5.0-dev.1"
+    };
+}
+static GOLEM_RUST_VERSION: &str = "2.0.0-dev.2";
+static GOLEM_TS_VERSION: &str = "0.1.0-dev.1";
+static GOLEM_AI_VERSION: &str = "v0.5.0-dev.1";
+static GOLEM_AI_SUFFIX: &str = "-dev.wasm";
```
vigoo, line 60: We no longer use the golem-ai wasms (just crate dependencies with versions)

cli/golem-cli/src/context.rs
```diff
@@ -110,6 +105,16 @@ impl Context {
             bail!(ContextInitHintError::CannotUseShortEnvRefWithLocalOrCloudFlags);
         }
 
+        let sdk_overrides = SdkOverrides {
```
vigoo, line 108: Is anything changes regarding the overrides (anywhere) or this is just moving things around?

cli/golem-cli/src/lib.rs
noise64: i will clean them up, just wanted to check with you before: i guess we still should keep the env var hints for rust?

cli/golem-cli/src/context.rs
noise64: just moving for now

cli/golem-cli/src/lib.rs
vigoo: yes those are still used in rust

noise64: i'm cleaning them up in the next first-class-lang PR (and a few more common leftovers, which i found since then)

PR #2840, opened 2026-02-26
agent-first-config: local agent config during component creation

### LINKED ISSUE #2800: Allow storing (partial) local agent configuration as part of a component revision
Local configuration can be provided in 2 places.

- As part of the component metadata
- During agent creation

The goal of this task is to allow specifying a subset of the local configuration when creating a new component revision. All provided values should fit the type of the declared config key/value in agent types and specifying an undeclared configuration key is an error.

mschuwalow:
resolves #2800

### DIFF
--- a/cli/golem-cli/src/command_handler/app/deploy_diff.rs
+++ b/cli/golem-cli/src/command_handler/app/deploy_diff.rs
@@ -742,6 +742,9 @@ fn normalized_diff_deployment(
                             wasm_hash: component.wasm_hash,
                             files_by_path: component.files_by_path.clone(),
                             plugins_by_grant_id: component.plugins_by_grant_id.clone(),
+                            local_agent_config_ordered_by_agent_and_key: component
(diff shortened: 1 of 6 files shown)


golem-common/src/base_model/agent.rs
```diff
@@ -1112,6 +1115,58 @@ pub struct SnapshottingEveryNInvocation {
     pub count: u16,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
vigoo, line 1118: Not very important but my this hierarchy makes my head spin :)

ConfigKeyValueType -> value -> ConfigValueType -> ConfigValueTypeLocal -> value -> AnalysedType 

valuetype-value-valuetype-valuetype-value-type :D

mschuwalow: Yeah, I'm not a huge fan either. Will try to fix the next time I'm touching the wit.

Issue #2836, opened 2026-02-25
Fix the new agent invocation HTTP endpoint's ambiguity


vigoo:
The caller's account-id + app-name + environment name does not work if the environment is not owned by the account just shared with it.
One possibility is to pass the environment's owner account email as well in the request.

To be investigated and fixed.

PR #2842, opened 2026-02-26
Agent invocation fixes

vigoo:
Attemps to fix all the issues that came by merging https://github.com/golemcloud/golem/pull/2819 with broken checks (that were reported as green).

In details:

### Agent Invocation API Fixes

- Fixes idempotency key handling on the new agent invocation API
- Deployment revision support added to the invocation API, this supports invoking older agents (with different agent types)
- Agent metadata casing inconsistency fixes (kebab vs non-kebab in various places)

### Registry / Deployment

- Required some changes to support the version of the invocation API when it has a specific deployment revision.

### Test Infrastructure Improvements

- The integration test framework now captures and re-emits the child process logs properly with all information (log level, nested spans, attributes) so I can finally read the integration test output! 
- Fail on missing test reports in CI (this would have prevented the merge yesterday)
- Fix in cargo-test-r to fail if the child cargo-test was aborted

### Other
- Some changes to AGENTS.md / extracting things to SKILLs
- Deleted unused component-transformer-example1

### DIFF
--- a/golem-worker-executor/tests/revert.rs
+++ b/golem-worker-executor/tests/revert.rs
@@ -287,7 +287,11 @@ async fn revert_auto_update(
 
     // Wait for the worker to finish initialization before reading the oplog
     executor
-        .wait_for_status(&worker_id, WorkerStatus::Idle, std::time::Duration::from_secs(10))
+        .wait_for_status(
+            &worker_id,
+            WorkerStatus::Idle,
+            std::time::Duration::from_secs(10),
+        )
         .await?;
 
     let updated_component = executor


golem-worker-service/src/service/worker/service.rs
```diff
@@ -750,20 +751,43 @@ impl WorkerService {
...
+                    WorkerServiceError::Internal(format!(
+                        "Invalid deployment revision (must be non-negative): {rev}"
+                    ))
+                })?;
+                let deployment_revision = DeploymentRevision::new(rev_u64).map_err(|e| {
+                    WorkerServiceError::Internal(format!("Invalid deployment revision: {e}"))
+                })?;
+                self.registry_service
+                    .resolve_agent_type_at_deployment(
+                        &auth.account_id(),
```
mschuwalow, line 766: This suffers from the same problem of only behaving correctly on the user's own account

golem-api-grpc/proto/golem/registry/v1/registry_service.proto
```diff
@@ -265,6 +266,25 @@ message ResolveLatestAgentTypeByNamesSuccessResponse {
   RegisteredAgentType agent_type = 1;
 }
 
+message ResolveAgentTypeAtDeploymentRequest {
```
mschuwalow, line 269: If the plan is to allow cross-account resolution of accounts / this is ever called with a user provided account id, we need to pass the auth context here. Otherwise, the currently used system auth context is going to bypass permission checks and potentially allow users to access data they should not be able to access.

golem-test-framework/src/config/dsl_impl.rs
```diff
@@ -245,8 +249,11 @@ impl<Deps: TestDependencies> TestDsl for TestUserContext<Deps> {
             .await?;
 
         if self.auto_deploy_enabled {
-            // deploy environment to make component visible
-            self.deploy_environment(&component.environment_id).await?;
+            let deployment = self.deploy_environment(&component.environment_id).await?;
+            self.last_deployments
```
mschuwalow, line 253: The auto_deployment stuff is only to allow the worker-executor style tests (which assume a component is ready as soon as its created) to work with the test dsl.

For integration tests if think an explicit `user.deploy_environment(&environmentId)` is clearer and gives you the deployment revision that you need.

vigoo:

golem-test-framework/src/config/dsl_impl.rs
vigoo: I like being able to copy a worker executor test to an integration test so I don't think this feature hurts

golem-api-grpc/proto/golem/registry/v1/registry_service.proto
vigoo: This mirrors the existing `ResolveLatestAgentTypeByNamesRequest` just has a `deployment_revision` field

golem-worker-service/src/service/worker/service.rs
vigoo: Yes, that's a separate ticket, not trying to solve in this one (https://github.com/golemcloud/golem/issues/2836)

PR #2844, opened 2026-02-26
Ability to invoke other account's environments if they are shared

### LINKED ISSUE #2836: Fix the new agent invocation HTTP endpoint's ambiguity

vigoo:
- Adds an optional owner-account-email filed to the invocation API
- Adds a new gRPC endpoint and repo function for registry-service to make the necessary agent type resolution a single-query operation
- Repo level and integration tests

Resolves #2836


golem-registry-service/src/repo/deployment.rs
```diff
@@ -949,6 +986,87 @@ impl DeploymentRepo for DbDeploymentRepo<PostgresPool> {
...
+    ) -> RepoResult<Option<ResolvedAgentTypeRecord>> {
+        self.with_ro("resolve_agent_type_by_names")
+            .fetch_optional_as(
+                sqlx::query_as(indoc! { r#"
+                    WITH owner AS (
+                      SELECT account_id
+                      FROM accounts
+                      WHERE deleted_at IS NULL
+                        AND (
+                          ($5 IS NOT NULL AND LOWER(email) = LOWER($5))
```
mschuwalow: hmm. We currently do not lower emails before storing accounts.

I would suggest not doing that here, as there might be false hits otherwise. Let's create a ticket and lower emails consistently everywhere. (Note that I believe it's not a spec that emails are case-insensitive, only a convention).

golem-registry-service/src/repo/deployment.rs
```diff
@@ -949,6 +986,87 @@ impl DeploymentRepo for DbDeploymentRepo<PostgresPool> {
...
+                          OR
+                          ($5 IS NULL AND account_id = $1)
```
mschuwalow: We prefer not pushing the IS NULL down to the database as the db needs to plan for both null and non-null values (though I know we still have a lot of this pattern flying around).

There is BindingsStack in golem_service_base/repo that can help with building the query on the server.

golem-registry-service/src/repo/deployment.rs
```diff
@@ -949,6 +986,87 @@ impl DeploymentRepo for DbDeploymentRepo<PostgresPool> {
...
+                        env.owner_account_id,
+                        COALESCE(
+                          $6::bigint,
+                          (SELECT cdr.deployment_revision_id
+                           FROM current_deployments cd
+                           JOIN current_deployment_revisions cdr
+                             ON cdr.environment_id = cd.environment_id
+                            AND cdr.revision_id = cd.current_revision_id
+                           WHERE cd.environment_id = env.environment_id)
+                        ) AS deployment_revision_id,
```
mschuwalow: Same here, if possible better to produce the query on the server as it will need to be part of the plan otherwise

integration-tests/tests/api/environment_share.rs
```diff
@@ -137,3 +142,174 @@ async fn update_environment_shares(deps: &EnvBasedTestDependencies) -> anyhow::R
...
+    // Resolve app and env names from owner's perspective
+    let owner_registry_client = deps.registry_service().client(&owner.token).await;
+    let app = owner_registry_client
+        .get_application(&component.application_id.0)
+        .await?;
+    let env = owner_registry_client
+        .get_environment(&component.environment_id.0)
+        .await?;
+
+    // Owner shares the environment with grantee (Admin role — needed for worker creation/invocation)
```
mschuwalow, line 171: Note: Deployer should be enough iirc, but the worker permissions can use a pass.

golem-registry-service/src/repo/deployment.rs
vigoo: ok, that's surprising to me - i would assume emails are case-insensitive (regardless of it is the spec or not); should we just "fix" this in the other places it involves in this pr?

golem-registry-service/src/repo/deployment.rs
noise64: Agree, let's use query building or separate static queries.

golem-registry-service/src/repo/deployment.rs
noise64: if we want to normalize, let's do it on registration, and before reaching the repo layer. And we should avoid using lower(x) on keys in general (unless there is an index for that form, or a suitable db column type). So i would remove the lower usage here, and add a speparate card for normalization (not in the db).

vigoo: https://github.com/golemcloud/golem/issues/2848

Please add any details you want to this ^

PR #2845, opened 2026-02-27
sdks/ts: enforce no-explicit-any and replace any with proper types

Myestery:
## Summary
- Add `@typescript-eslint/no-explicit-any: 'error'` ESLint rule to all 4 TS SDK packages (sdk, repl, typegen, types-core)
- Replace `any` with proper types (`unknown`, generics, specific interfaces) across the entire TS SDK codebase
- Add generic type parameters to the deserialization pipeline (`deserialize<T>`, `toTsValue<T>`, `deserializeDataValue<T>`, `deserializeRpcResult<T>`) so callers can avoid manual casts
- Extract `AgentClass`, `MethodKeys`, `MutableAgentStatics` helper types to eliminate `eslint-disable` comments and `as unknown as` casts
- Fix serializer bugs: swapped `missingObjectKey` args, incorrect `handleObjectMatch` null check, missing null guards in reference serialization
- Add `types/globals.d.ts` for typed `globalThis.currentAgentId` in tests

## Bug fixes
- `missingObjectKey('tag', tsValue)` — arguments were previously swapped, producing wrong error messages
- `handleObjectMatch` — was comparing object to string literal `'interface'` instead of checking for null
- `serializeTsValueToBinaryReference` / `serializeTsValueToTextReference` — added null guards (`typeof null === 'object'` was passing the guard)

### DIFF
--- a/sdks/ts/packages/golem-ts-sdk/src/internal/mapping/values/dataValue.ts
+++ b/sdks/ts/packages/golem-ts-sdk/src/internal/mapping/values/dataValue.ts
@@ -314,7 +314,9 @@ function _deserializeDataValue(
   }
 }
 
-function constructConfigType(typeInfoInternal: TypeInfoInternal & { tag: 'config' }): Config<Record<string, unknown>> {
+function constructConfigType(
+  typeInfoInternal: TypeInfoInternal & { tag: 'config' },
+): Config<Record<string, unknown>> {
   // safe as the parent node is config
   const properties = (typeInfoInternal.tsType as Type.Type & { kind: 'config' }).properties;
 
@@ -491,13 +493,17 @@ function serializeMultimodalToDataValue(
 
         case 'unstructured-binary': {
           const isObjectBinary = typeof elemVal === 'object' && elemVal !== null;
-          isMatch = isObjectBinary && 'tag' in elemVal && (elemVal.tag === 'url' || elemVal.tag === 'inline');
+          isMatch =
+            isObjectBinary &&
+            'tag' in elemVal &&
+            (elemVal.tag === 'url' || elemVal.tag === 'inline');
           break;
         }
 
         case 'unstructured-text': {
           const isObjectText = typeof elemVal === 'object' && elemVal !== null;
-          isMatch = isObjectText && 'tag' in elemVal && (elemVal.tag === 'url' || elemVal.tag === 'inline');
+          isMatch =
+            isObjectText && 'tag' in elemVal && (elemVal.tag === 'url' || elemVal.tag === 'inline');
           break;
         }
 
--- a/sdks/ts/packages/golem-ts-sdk/src/baseAgent.ts
+++ b/sdks/ts/packages/golem-ts-sdk/src/baseAgent.ts
@@ -175,9 +175,7 @@ export class BaseAgent {
     );
   }
 
-  static getPhantom<
-    T extends new (phantomId: Uuid | undefined, ...args: never[]) => BaseAgent,
-  >(
+  static getPhantom<T extends new (phantomId: Uuid | undefined, ...args: never[]) => BaseAgent>(
     this: T,
     ...args: ConstructorParameters<T>
   ): Client<InstanceType<T>> {
@@ -227,9 +225,7 @@ type MethodKeys<T> = {
 }[keyof T];
 
 export type Client<T> = {
-  [K in MethodKeys<T>]: T[K] extends (
-    ...args: infer A
-  ) => infer R
+  [K in MethodKeys<T>]: T[K] extends (...args: infer A) => infer R
     ? RemoteMethod<GetArgs<A>, Awaited<R>>
     : never;
 };
--- a/sdks/ts/packages/golem-ts-sdk/tests/decorator.test.ts
+++ b/sdks/ts/packages/golem-ts-sdk/tests/decorator.test.ts
@@ -804,8 +804,7 @@ describe('Annotated FooAgent class', () => {
       lowBits: BigInt(5678),
     };
 
-    globalThis.currentAgentId =
-      `foo-agent("hello")[${uuid.highBits}-${uuid.lowBits}]`;
+    globalThis.currentAgentId = `foo-agent("hello")[${uuid.highBits}-${uuid.lowBits}]`;
 
     const fooResult = initiator.initiate(
       {
@@ -871,8 +870,7 @@ describe('Annotated SingletonAgent class', () => {
       val: [],
     };
 
-    globalThis.currentAgentId =
-      `singleton-agent(${JSON.stringify(params)})`;
+    globalThis.currentAgentId = `singleton-agent(${JSON.stringify(params)})`;
 
     const singleton = initiator.initiate(params, { tag: 'anonymous' });
     expect(singleton.tag).toEqual('ok');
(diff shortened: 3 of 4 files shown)


sdks/ts/packages/golem-ts-repl/src/base.ts
```diff
@@ -41,6 +41,52 @@ export type EnvironmentName = string;
 export type AgentTypeName = string;
 export type IdempotencyKey = string;
 
+export type Timestamp = string; // ISO 8601
+
+export type WorkerStatus =
```
vigoo, line 46: These are defined in WIT so the binding generator (of wasm-rquickjs) should generate good types for them. If not, can we improve there?

To decide that please explain why you defined new ones

Myestery: I searched for worker definitions but didn't find one.And the old implementation just had worker name

noise64: the original "mapping" is very lightweight intenionally, and just for looking up agent ids for completion from CLI output. we do have plans (and will soon have tickets) around exploring more typed integration with the CLI, so i see no point on having these here for now, especially that we currently keep changing APIs / CLI interfaces as part of the more agent focused APIs

cli/golem-cli/ts_bridge/base.ts
```diff
@@ -7,8 +7,9 @@ export type GolemServer =
 
 export type AroundInvokeHook = {
     beforeInvoke: (request: AgentInvocationRequest) => Promise<void>;
-    afterInvoke: (request: AgentInvocationRequest, result: JsonResult<AgentInvocationResult, any>) => Promise<void>;
-};
+        afterInvoke: (request: AgentInvocationRequest, result: JsonResult<AgentInvocationResult, unknown>) => Promise<void>;
+    }
+    ;
```
noise64, line 12: this format look off (also this code will be soon moved to the sdk)

sdks/ts/packages/golem-ts-repl/src/config.ts
```diff
@@ -33,10 +33,14 @@ export type ReplCliFlags = {
   streamLogs: boolean;
 };
 
-export type AgentConfig = {
+export type AgentModule = Record<string, unknown> & {
```
noise64, line 36: the bridge SDK package is not really a record of string, it is a package, the plumbing here does not really change type safety compared to before.

sdks/ts/packages/golem-ts-repl/src/repl.ts
```diff
@@ -346,13 +337,25 @@ export class Repl {
     setOutput('stdout');
 
     const script = this.replCliFlags.script;
+
+    // We need to create a temporary repl server to get the default eval and completer
```
noise64, line 341: The pactching of the repl was intenionally used the actual instance. (And there is another patching in place before, that is created by tsx, and we follow the way). Not sure it this breaks anything, but there is no point in changing the current solution.

noise64:
General comment: while moving away from `any` in general is a good thing, but there are many moving parts in the sdk currently, and some which will only have proper integ tests soon (e.g. REPL), so doing such bigger changes to the enforced linter rules is something that we should discuss and decide before doing so, especially, if - in effect -  the changes do not increase type safety for the most parts, but can potentially change some semantics.

sdks/ts/packages/golem-ts-sdk/src/internal/mapping/values/dataValue.ts
```diff
@@ -206,7 +214,7 @@ export function deserializeDataValue(
       const multimodalParamTypes = typeInfo.types;
 
       // These are not separate parameters, but a single parameter of multimodal type
-      const multiModalValue: Either.Either<any[], string> = Either.all(
+      const multiModalValue: Either.Either<unknown[], string> = Either.all(
```
afsalthaj, line 217: This is a generic comment.

There are a few subteilities playing against this to introduce this lint. Firstly I didn't introduce it intentionally, when it comes to TS SDK. The reason is the risk of `any` and the use of `any` are both needed to be taken by SDK. We might be able to reduce it but not for all. So while there are places where we can legitimately replace it with `unknown` (because `unknown` can be  definitely a safer option in many cases since it forces narrowing before use) ,  many of these anys are legitimate logically.

afsalthaj: Legitimate because, conceptually I should be able to do anything with this unknown any type in certain places at least. Or in fact,  in many places.. I should be able to invoke a function on any. Should I narrow it? May be. May be not. 

I am yet to review the full PR, but just not convinced with the `any` hate. (even if I like proper types)

afsalthaj: I would like to write more, but then it become very verbose. But if you could trust, the use of `any` in a lot of places in our TS SDK is actually legitimate.

sdks/ts/packages/golem-ts-sdk/src/internal/mapping/values/deserializer.ts
```diff
@@ -23,7 +23,11 @@ import { Result } from '../../../host/result';
  * @param value
  * @param analysedType
  */
-export function deserialize(value: Value, analysedType: AnalysedType): any {
+export function deserialize<T = unknown>(value: Value, analysedType: AnalysedType): T {
+  return _deserialize(value, analysedType) as T;
+}
+
+function _deserialize(value: Value, analysedType: AnalysedType): unknown {
```
afsalthaj, line 30: Why this redirection? Here `T=unknown` and you are returning `unknown` and once deserialization is done, that `T` is not legitimately used? So conceptually its `any`

sdks/ts/packages/golem-ts-sdk/src/internal/mapping/values/Value.ts
```diff
@@ -115,6 +115,7 @@ export function fromWitValue(wit: WitValue): Value {
 }
 
 function buildTree(node: WitNode, nodes: WitNode[]): Value {
+  let tag = node.tag;
```
afsalthaj, line 118: Why this?

sdks/ts/packages/golem-ts-sdk/src/internal/mapping/values/Value.ts
```diff
@@ -283,9 +284,9 @@ export function toWitValue(value: Value): WitValue {
 function buildNodes(value: Value, nodes: WitNode[]): number {
   const idx = nodes.length;
   nodes.push({
-    tag: 'placeholder',
-    val: undefined,
-  } as any);
+    tag: 'prim-bool',
```
afsalthaj, line 287: I think this is honestly too many changes, that questions what's the final outcome for golem? Is it fixing any bug? or cleaning up?

If there are bug fixes in this PR, can we raise the bug fix ticket separately and fix only that?

Because, apart from that, I don't think this PR brings in better design of the code (which is in our backlog for value mapping) or readability.

So kindly raise any bugs you find, and fix only them as separate PR. If the intend is better design and readability , let's raise a ticket separately for that, and describe the design changes in PR description. Like this one I did: https://github.com/golemcloud/golem/pull/2608 

that PR improved a lot about schema discovery within SDK.

Myestery:
Closing this in favor of three focused PRs based on review feedback:

1. **Bug fixes**: #2855 — serializer argument swap, BigInt conversion, null guards, `isAgentError` type safety
2. **Metadata types**: #2856 — typed metadata JSON interfaces for `loadFromJson`, typegen, and type registry
3. **SDK type improvements**: #2857 — replace `any` with `unknown`/generics across ~15 SDK files

Dropped from scope (per reviewer discussion): REPL types, ESLint config, lock file, `globals.d.ts`, `type-lite.ts`, and `ts_bridge` changes.

PR #2851, opened 2026-02-27
Wasmtime v42.0.1

### LINKED ISSUE #2340: Upgrade to wasmtime 39+

vigoo:
Resolves #2340

### DIFF
--- a/.github/workflows/windows-daily-build.yaml
+++ b/.github/workflows/windows-daily-build.yaml
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   build-windows:
-    runs-on: blacksmith-32vcpu-windows-2025
+    runs-on: windows-latest-large
     env:
       CARGO_BUILD_JOBS: 10
     steps:
--- a/integration-tests/benchmark_suites/ci.yaml
+++ b/integration-tests/benchmark_suites/ci.yaml
@@ -4,68 +4,68 @@ name: CI
 benchmarks:
   - name: cold-start-unknown-small
     iterations: 1
-    clusterSize: [ 1 ]
-    size: [ 1, 5, 10 ]
-    length: [ 2 ]
+    clusterSize: [1]
+    size: [1, 5, 10]
+    length: [2]
     disableCompilationCache: true
 
   - name: cold-start-unknown-medium
     iterations: 1
-    clusterSize: [ 1 ]
-    size: [ 1, 5, 10 ]
-    length: [ 5 ]
+    clusterSize: [1]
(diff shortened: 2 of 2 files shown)


vigoo:
In this PR random worker-executor tests were seemingly timing out more frequently than before. I've analysed the logs of these cases, fixed optional tracing support, analysed whole test suite runs with otlp traces, etc and the only conclusion I could get is that it depends on when the test-components are first get compiled - as the compilation itself uses all cores and can be relatively long in debug builds for large components like the TS ones. 

I've measured the wasmtime v33 vs v42 compilation performance and checked if they changed anything in their cache implementation but there should not be any regression so my conclusion is that the "more flaky" behavior is just a random coincidence of things running slightly differently with the new engine.

So I've made two changes _for worker executor tests_ (integration tests remain as they are - they test the real behavior where compilation happens when and how it would in production):

- first, fixed the agent extraction wasmtime config so it uses _the same flags_ as the executor. this way the wasmtime filesystem cache works as intended (it uses these flags in its cache key)
- latest test-r versions are no longer initializing all test dependencies in a suite, just the ones needed. this made it possible to define the test-components as test dependencies injected to the tests via constructor parameters. The test DSL was changed to accept these `PrecompiledComponent` objects optionally. When used, the test dependency itself does a "warmup" agent type extraction that fills up both the agent-type cache and wasmtime's own filesystem-based compilation cache

This makes the worker-executor tests much more deterministic because they no longer depend on injecting minutes-long compilations in their execution based on the execution order and parallelism. Everything is pre-initialized for them - while test-r's lazy test-dep evaluation guarantees that we can still run just a subset of the tests without penalties.

golem-test-framework/src/dsl/mod.rs
```diff
@@ -91,6 +112,17 @@ pub trait TestDsl {
         StoreComponentBuilder::new(self, *environment_id, name.to_string())
     }
 
+    /// Creates a `StoreComponentBuilder` from a `PrecompiledComponent`, automatically
+    /// setting both the WASM file name and the package name.
+    fn component_dep(
```
mschuwalow, line 117: Can we make this specific to the worker-executor test dsl? This one doesn't make sense for the integration tests from what I can tell.

vigoo: I was thinking maybe it makes sense for them too (but did not do any such change) to have a central definition of the test components and refer to them with test_dep instead of hardcoding their names everywhere. (The precompilation itself should not happen there, of course)

PR #2855, opened 2026-03-02
sdks/ts: fix serializer bugs and agentError null crash

Myestery:
## Summary
- Fix `missingObjectKey` argument swap: arguments were reversed `(tsValue, 'tag')` → `('tag', tsValue)`, producing wrong error messages
- Fix `handleObjectMatch` null check: was comparing to string literal `'interface'` instead of checking for `null`
- Fix u64 BigInt conversion: properly convert number to BigInt instead of casting as `any`
- Add null guards in `serializeTsValueToBinaryReference` and `serializeTsValueToTextReference` (`typeof null === 'object'` was passing the guard)
- Fix `isAgentError` crash on `null` input: add null/non-object guard before accessing properties

## Context
Split from #2845 — this PR contains only the bug fixes, no lint rule changes or type refactoring.

## Test plan
- [ ] Existing SDK tests pass
- [ ] Manual verification that `isAgentError(null)` returns `false` instead of crashing
- [ ] Serializer correctly handles `missingObjectKey` error messages


vigoo (approved):

Myestery:

PR #2856, opened 2026-03-02
sdks/ts: add MetadataJSON types to metadata pipeline

Myestery:
## Summary
- Add `MethodMetadataJSON`, `ClassMetadataJSON`, `MetadataJSON` type definitions to `golem-ts-types-core`
- Replace `any` with `MetadataJSON` in `loadFromJson`, removing all `as any` casts
- Use `ClassMetadataJSON` and `MethodMetadataJSON` in typegen `saveAndClearInMemoryMetadata`
- Use `MetadataJSON` in `TypescriptTypeRegistry.register`

## Context
Split from #2845 — this PR adds proper types to the metadata JSON pipeline only (~40 lines changed across 3 files).

## Test plan
- [ ] Existing SDK tests pass
- [ ] Typegen metadata generation produces the same output


vigoo:
The failing integration tests show that this change breaks the sdk somehow, need to be fixed before merge

PR #2860, opened 2026-03-02
Stability fixes

vigoo:
Trying to stabilize the integration tests

- adds request timeout configuration support but only enables it on a single edge (registry svc calls from executor)
- fixes a race condition in our `Cache` type and adds lot of tests to it
- removes an async recursion from `HttpWorkerLogEventStream` that I've seen causing stack overflow
- make log streaming in the test framework not panic if the connection is dropped from it
- using a deterministic idempotency key for `initialize` to make it impossible to double-initialize (even if some more race condition is not fixed yet)
- updated test-r so it should now better print failures on CI when using CTRF reporting
- fixes an oplog opening race condition that was probably the cause of many issues in the "executor is overloaded" case (not necessarily during tests on CI)
- investigated a lot about a weird flaky deserialization error. It turned out to be a problem in the **oplog query**, which used a wrong agent-type metadata when enriching some of the oplog entries. Whether it ran into the wrong metadata or not was depending on the runner (hash-based, probably). Fixed now
- Found and fixed an issue in the wasi:http durability wrapper that could lead to "unexpected oplog entry" during recovery - but not the same that we've seen on the effect.ai live demo. (None of our wasi:http libraries did this usage pattern in practice)

### DIFF
--- a/golem-worker-executor/src/services/oplog/tests.rs
+++ b/golem-worker-executor/src/services/oplog/tests.rs
@@ -86,11 +86,10 @@ async fn open_add_and_read_back(_tracing: &Tracing) {
         worker_name: "test".to_string(),
     };
     let owned_worker_id = OwnedWorkerId::new(environment_id, &worker_id);
-    let last_oplog_index = oplog_service.get_last_index(&owned_worker_id).await;
     let oplog = oplog_service
         .open(
             &owned_worker_id,
-            last_oplog_index,
+            None,
             WorkerMetadata::default(worker_id.clone(), account_id, environment_id),
             default_last_known_status(),
             default_execution_status(AgentMode::Durable),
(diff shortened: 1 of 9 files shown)


golem-worker-executor/src/services/oplog/tests.rs
```diff
@@ -1783,3 +1786,132 @@ async fn multilayer_scan_for_component(_tracing: &Tracing) {
...
+
+/// Reproducer for the oplog unique key violation panic during recovery.
+///
+/// The race is in `OpenOplogs::get_or_open`: when two tasks concurrently call
+/// it for the same worker_id, both can observe `entry.initial == true` and both
+/// execute `decrement_strong_count`. This can over-decrement the Arc refcount,
+/// causing premature drop, the Weak becoming un-upgradeable, cache eviction,
+/// and creation of a **second** oplog instance for the same worker. Two instances
+/// means two independent `last_committed_idx` counters, leading to duplicate
+/// INSERT attempts and a unique key violation in SQLite.
```
noise64, line 1782: 👏

Issue #2835, opened 2026-02-25
Design and finalize the way AgentId-s are parsed and rendered for 1.5


vigoo:
AgentId-s need to have a canonical language-agnostic rendering. 
Right now this is:
- kebab-ized agent type name
- special syntax for unstructured text/binary and multimodal
- wave syntax for everything else
- phantom-id optionally in the end

To consider:
- use PascalCased agent-type name
- use a JSON representation for the arguments instead

This requires analysing all the usage scenarios of printing/parsing/normalizing agent-ids after the agentic invocation refactor and choosing a solution that's good enough for release 1.5 and points towards future plans (of dropping wave and potentially wit values)

PR #2869, opened 2026-03-02
Agent types changes step 4

vigoo:
Step 4 of #2835 

Removes "kebabizing" from both SDKs
Introduces some helper macros that can be used in integration tests to convert data values where the type is coming from host-side derived analysed types


vigoo:
Yes these are all finalized in the final step

PR #2870, opened 2026-03-02
agent-config: allow specifying local agent config during agent creation

### LINKED ISSUE #2801: Allow specifying local agent configuration when creating an agent
Similar to environment variables and wasi-config keys it should be possible to provide local configuration when creating a new agent. The provided configuration has to match the declared config type in the agent type.

At this point *all* local configuration needs to have been provided (either through component metadata or agent type specific config). Failure to do so should fail the agent creation with a descriptive error.

Specifying undeclared config keys should also fail with a descriptive error.

### LINKED ISSUE #2802: Handle local agent configuration during updates of the component revision in the worker executor
Similar to environment variables and wasi-config keys the worker-executor needs to handle changes to the local agent configuration during component revision updates. It should compute the new materialized set of config keys by combining the config of the new component revision with the config keys provided during agent creation.
If this final set does not fit the config schema of the new component revision the update should fail.

mschuwalow:
resolves #2801
resolves #2802

changes look much bigger than they are because of wit / test-component changes

### DIFF
--- a/golem-common/src/base_model/worker.rs
+++ b/golem-common/src/base_model/worker.rs
@@ -12,17 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::agent::AgentTypeName;
-use super::component_metadata::ComponentMetadata;
 use crate::base_model::account::AccountId;
 use crate::base_model::component::{ComponentFilePermissions, ComponentRevision, PluginPriority};
 use crate::base_model::environment::EnvironmentId;
 use crate::base_model::oplog::WorkerResourceId;
 use crate::base_model::regions::OplogRegion;
 use crate::base_model::{OplogIndex, Timestamp, WorkerId, WorkerResourceDescription, WorkerStatus};
-use crate::model::agent::{ConfigKeyValueType, ConfigValueType};
 use crate::{declare_enums, declare_structs, declare_unions};
-use golem_wasm::ValueAndType;
 use golem_wasm_derive::{FromValue, IntoValue};
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::fmt::{Display, Formatter};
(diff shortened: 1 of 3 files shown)


golem-common/src/base_model/oplog/mod.rs
```diff
@@ -76,6 +77,7 @@ oplog_entry! {
             initial_total_linear_memory_size: u64,
             initial_active_plugins: HashSet<PluginPriority>,
             config_vars: BTreeMap<String, String>,
+            local_agent_config: Vec<ParsedWorkerCreationLocalAgentConfigEntry>,
```
vigoo: Could this be just the raw data? The idea is that all `raw` oplog entries are as concise as possible (for example invocation parameters also are only `Value` not `ValueAndType` etc.) and the `public` one is the enriched, "render friendly" version. If the enrichment can be done on demand based on metadata and oplog entries, then we can use different types in `raw` vs `public`.

golem-worker-executor/src/worker/mod.rs
```diff
@@ -1601,6 +1616,68 @@ impl<Ctx: WorkerCtx> Worker<Ctx> {
                 };
 
                 let worker_env = merge_worker_env_with_component_env(worker_env, component.env);
+
+                let initial_local_agent_config = if let Some(agent_id) = &agent_id {
+                    let agent_type = component.metadata
+                        .agent_types()
+                        .iter()
+                        .find(|at| at.type_name == agent_id.agent_type)
```
vigoo: There is a `find_agent_type_by_name` or similar helper for this, I think

golem-worker-executor/src/worker/mod.rs
```diff
@@ -1821,6 +1903,60 @@ pub fn merge_worker_env_with_component_env(
     result
 }
 
+pub fn effective_local_agent_config(
```
vigoo: Maybe move these config-related helpers to a submodule? This file is getting big :)

golem-common/src/base_model/oplog/mod.rs
mschuwalow: Done. We do need the types in the worker status, so this necessitates fetching the component metadata when going from initial oplog entry to the initial worker metadata.

golem-worker-executor/src/worker/mod.rs
mschuwalow: done

golem-worker-executor/src/worker/mod.rs
mschuwalow: done

PR #2918, opened 2026-03-03
Support MCP resources, multimodal, unstructured-* with integration tests

### LINKED ISSUE #2794: Create integration test with MCP
By the end of #2793 , we have almost everything to write integration test. Refer to basic handshake and request request in task 1 or here [https://github.com/afsalthaj/simple-mcp-server/tree/master?tab=readme-ov-file#manul-server-handshake-details-the-bare-minimum-that-has-to-be-there-in-golem-regardless-of-using-sdk-or-not](https://github.com/afsalthaj/simple-mcp-server/tree/master?tab=readme-ov-file#manul-server-handshake-details-the-bare-minimum-that-has-to-be-there-in-golem-regardless-of-using-sdk-or-not). If this handshake works, it then works with almost all AIs  such OpenAI or claude desktop  (which I tested).

### LINKED ISSUE #2795: Implement listing resource and corresponding handlers
Firstly, we know the list of agent_methods that are really resources.

In the case of resource call handler, there isn’t heavy mappings required as such needed either. 

And mostly, I guess this is about implementing something like this

```rust
use rmcp::handler::ServerHandler

impl ServerHandler for GolemAgentMcpServer {
    async fn read_resource(
        &self,
        ReadResourceRequestParams { meta: _, uri }: ReadResourceRequestParams,
        _: RequestContext<RoleServer>,
    ) -> Result<ReadResourceResult, McpError> { todo!() }
}
```

### LINKED ISSUE #2822: Implement listing prompt and prompt handlers
Prompts are as important as tools (although for simpler examples like counter, prompts don’t play much role). 

MCP Prompts provide pre-written instructions that show users exactly what to ask for when they want to use specific tools. Here is an excerpt from the spec.

<aside>
💡

***Prompts are designed to be user-controlled, meaning they are exposed from servers to clients with the intention of the user being able to explicitly select them for use.***

</aside>

To get a prompt itself there can be optional`input` argument. 

Example: The prompt depends on a username input. If the username is “afsal”, then prompt should be “generate some Kangaroo jokes” and if the user name is Vigoo, then prompt is “generate some Hungarian jokes”, and the user has full control of seeing this prompt and chooses to send this message. 

MCP Client internally sends this prompt (query) along with the list of tools and the LLM will decide and respond with the actual tool name to call and the arguments. The MCP client then  executes it. This is a typical workflow that involves a prompt.

Mapping this exact workflow in Golem can be a bit vague but can be acheived (explained further below). I think this is because we do not have exact 1-1 mapping between agent SDK and other formal SDKs in MCP world. I think this is why John wrote.

<aside>
💡

Golem does not support such distinguished entities, preferring instead to unify all concepts into the concept of an *agent*.

// entities here mean tools, prompts and resources.

</aside>

Here is another (code level) example of using FastMCP with 2 tools and 1 prompt (for better clarity)

```rust
@server.tool()
async def greet_user_formal_tool(name:str) -> str:
    """
    A tool that returns a greeting message in a very formal tone
    Args:
        name (str): The name of the person to greet.
    Returns:
        str: A formal greeting message for the given name.
    """
    return f"Good day to you, {name}. I trust this message finds you well."

@server.tool()
async def greet_user_street_style_tool(name:str) -> str:
    """
    A tool that returns a greeting message in street style
    Args:
        name (str): The name of the person to greet.
    Returns:
        str: A street style greeting message for the given name.
    """
    return f"Yo {name}! Wassup? You good?"

@server.prompt
def greet_user_prompt(name: str) -> str:
    """Generates a message asking for a greeting""" // this is the prompt descripti
    
    // and the actual prompt when the client calls get_prompt with an input `username`
    return f"""
    Return a greeting message for a user called '{name}'. 
    if the user is called 'Laurent', use a formal style, else use a street style.
    """
```

So a standard workflow that involves prompt is, user chooses from an available prompt and in this case `greet_user_prompt` (possible in Claude Desktop (or even in VSCode with copilot).
Afterwards, user  provide the input `Laurent` as input to `get_user_prompt` to get the actual “prompt”,  which will be 

```rust

Return a greeting message for a user called 'Laurent'
if the user is called 'Laurent', use a formal style, else use a street style
```

**Now how do we bring this back to Golem, given Rust and TS SDK with `prompt` in it ?**

Here is the details of `Prompt` fields in MCP protocol. Bare minimum, we populate `name`, `arguments`

```rust
pub struct Prompt {
    // probably the agent method name itself
    pub name: String,
    pub title: Option<String>,
    /// Optional description of what the prompt does
    pub description: Option<String>,
    /// Optional arguments that can be passed to customize the prompt
    pub arguments: Option<Vec<PromptArgument>>,
    /// Optional list of icons for the prompt
    pub icons: Option<Vec<Icon>>,
    /// Optional additional metadata for this prompt
    #[serde(rename = "_meta", skip_serializing_if = "Option::is_none")]
    pub meta: Option<Meta>,
}
```

We will have `list_prompt` (similar to `list_tools`) implementation in the following way:

- We iterate over each agent method
- The name of the prompt is `get_${method_name}_prompt` and the arguments can be the same argument names of the actual method.  The handler of `get_${method_name}_prompt` is simply returning something like the following:  `call ${method_name} the following input: ${parameter_name1}: ${parameter_value1}, and ${parameter_name2}: ${parameter_value2}` . User given prompt (through our SDK) will be prepended to the above message.
- In a flattened MCP server, a typical `tool` will include the constructor parameters tool in the argument. So the above parameter list should will include the constructor parameters tool.
- Also note that, the tool (during list_tool) documentation should already include the aspect of flattened server as mentioned before.
- In this context, I am not sure if user provided constructor-level prompt.
- Another alternative is to aggregate all the prompts across the agent methods, but I am not sure what’s the standard aggregation rule.

From code perspective, the exact starting point in code is to have `prompt_router` similar to `tool_router` in this server. https://github.com/afsalthaj/simple-mcp-server/blob/master/src/mcp_adaptor/agent_mcp_server.rs#L18

Also this: https://github.com/afsalthaj/simple-mcp-server/blob/master/src/mcp_adaptor/agent_mcp_prompt.rs

### LINKED ISSUE #2863: Handle multimodal, unstructured-text and binary in MCP protocol
Recommending this is to be done once we figure out the implementation for #2795 .

afsalthaj:
Fixes #2795 

Fixes #2863 

Fixes #2794 

Fixes #2822   (See https://github.com/golemcloud/golem/pull/2961)

Also updated with #2980  (this required more testing and that took some time)

Note that, I am not doing any kebab case conversions to make things "look" good. I will do that separately, because it will hurt ongoing changes, and need a separate task for it. Things work without that now.

The whole MCP changes right now covers all features except prompt. It took a lot of re-iterations to make things work with MCP clients. OpenAI playground worked with tools but not with resources (because it didn't support). 

Claude Desktop and MCP inspector are the new one I tested. Now I can see `golem`  within Claude Desktop successfully. Claude Desktop actually helped with some real bugs.

Here is a screenshot of Claude configured with Golem (thought it was impossible until I came to know about https://www.npmjs.com/package/mcp-remote). Pasting the config here (might be helpful for docs)

In `claude_desktop_config.json` and restart desktop:

```json
{
  "mcpServers": {
    "golem": {
      "command" : "npx",
      "args" : [ 
         "mcp-remote",
         "http://localhost:9007/mcp",
         "--header",
         "Host: localhost:9007",
         "--allow-http"
       
      ]
    }
  }
}
```

image

image

These were not running for a major time of this PR in the draft state.

MCP inspector also works similar to Claude Desktop. Open AI doesn't because, they don't show options for resources. 

I also tested resources, and MCP inspector is the best test that I could do, and it understood both resource template and concrete resource as mentioned in the MCP protocol.

### Note
Also note that, everything in Golem doesn't have exact one to one with MCP as mentioned in the spec written by John. We are making the best approximations here. And whatever you see in this PR, has been manually tested with popular clients and making changes will need retesting. 

I would also like to test more but only as part of release tests that will be done later.  And as part of bug fixes. 

## Static Resource

It implies the resource is part of a cluster singleton

image

## Template Resource

It implies resource depends on the identity of your agent

image

image

These are tested with official MCP Inspector (the best client for testing), and invoked these tools. Before this PR, those images never existed

## Multimodal - A weather report + weather image together

image

##  UnstructuredBinary:  A snow fall image, given a city

Now it always render images, instead of at-times base64, or at times MCP client being smart and end up doing erroring out.  

image

## Unstrucured-Text report (there is nothing much here)

image

### DIFF
--- a/golem-worker-service/src/mcp/agent_mcp_prompt.rs
+++ b/golem-worker-service/src/mcp/agent_mcp_prompt.rs
@@ -420,6 +420,7 @@ mod tests {
         );
 
         assert!(prompt.prompt_text.starts_with("Run analysis"));
+
         assert!(
             prompt
                 .prompt_text
@@ -427,7 +428,8 @@ mod tests {
             "got: {}",
             prompt.prompt_text
         );
-        assert!(prompt.prompt_text.contains("Output: result: text"));
+
+        assert!(prompt.prompt_text.contains("text"));
     }
 
     #[test]
--- a/golem-worker-service/src/mcp/invoke/tool.rs
+++ b/golem-worker-service/src/mcp/invoke/tool.rs
@@ -366,9 +366,10 @@ mod tests {
             },
         )]);
         let result = map_agent_response_to_tool_result(response, &schema).unwrap();
-        let structured = result.structured_content.unwrap();
-        assert_eq!(structured["report"]["data"], "weather is sunny");
-        assert_eq!(structured["report"]["languageCode"], "en");
+
+        let raw_content = &result.content[0].raw;
+
+        assert_eq!(raw_content, &RawContent::text("weather is sunny"));
     }
 
     #[test]
@@ -379,6 +380,7 @@ mod tests {
                 schema: ElementSchema::UnstructuredBinary(BinaryDescriptor { restrictions: None }),
             }],
         });
+
         let response = UntypedDataValue::Tuple(vec![UntypedElementValue::UnstructuredBinary(
             BinaryReferenceValue {
                 value: BinaryReference::Inline(BinarySource {
@@ -390,9 +392,13 @@ mod tests {
             },
         )]);
         let result = map_agent_response_to_tool_result(response, &schema).unwrap();
-        let structured = result.structured_content.unwrap();
-        assert_eq!(structured["image"]["data"], "AQID");
-        assert_eq!(structured["image"]["mimeType"], "image/png");
+
+        let raw_content = &result.content[0].raw;
+
+        assert_eq!(
+            raw_content,
+            &RawContent::image("AQID", "image/png".to_string())
+        );
     }
 
     #[test]
@@ -434,13 +440,15 @@ mod tests {
             },
         ]);
         let result = map_agent_response_to_tool_result(response, &schema).unwrap();
-        let structured = result.structured_content.unwrap();
-        let parts = structured["parts"].as_array().unwrap();
-        assert_eq!(parts.len(), 2);
-        assert_eq!(parts[0]["name"], "desc");
-        assert_eq!(parts[0]["value"]["data"], "a photo");
-        assert_eq!(parts[1]["name"], "photo");
-        assert_eq!(parts[1]["value"]["data"], "AQID");
+        let contents = &result.content;
+
+        assert_eq!(contents.len(), 2);
+
+        assert_eq!(&contents[0].raw, &RawContent::text("a photo"));
+        assert_eq!(
+            &contents[1].raw,
+            &RawContent::image("AQID", "image/png".to_string())
+        );
     }
 
     #[test]


golem-worker-service/src/mcp/agent_mcp_resource.rs
```diff
@@ -12,10 +12,93 @@
...
+impl AgentMcpResource {
+    pub fn resource_name(agent_type_name: &AgentTypeName, method: &AgentMethod) -> String {
+        format!("{}-{}", agent_type_name.0, method.name)
+    }
+
+    pub fn static_uri(agent_type_name: &AgentTypeName, method: &AgentMethod) -> String {
+        format!("golem://{}/{}", agent_type_name.0, method.name)
+    }
+
+    pub fn template_uri(
```
afsalthaj, line 207: to be cleaned up, but once I know I can invoke it through some clients than my own.

golem-common/src/base_model/agent.rs
```diff
@@ -547,15 +547,45 @@ pub enum UntypedDataValue {
 }
 
 #[derive(Debug, Clone, PartialEq)]
-#[cfg_attr(
-    feature = "full",
-    derive(IntoValue, FromValue, desert_rust::BinaryCodec)
-)]
+#[cfg_attr(feature = "full", derive(desert_rust::BinaryCodec))]
```
vigoo, line 550: Why is this changed to manual implementation? Revert if possible

golem-worker-service/src/mcp/invoke/constructor_param_extraction.rs
```diff
@@ -0,0 +1,85 @@
...
+
+use golem_common::base_model::agent::{
+    ComponentModelElementSchema, ComponentModelElementValue, DataSchema, ElementSchema,
+    NamedElementSchema,
+};
+use golem_wasm::analysis::AnalysedType;
+use golem_wasm::json::ValueAndTypeJsonExtensions;
+use rmcp::model::JsonObject;
+
+pub fn extract_constructor_input_values(
```
vigoo, line 23: We should have unit tests for this (and similar encoding/decoding functions in the `invoke` submodules)

golem-worker-service/src/mcp/invoke/resource.rs
```diff
@@ -0,0 +1,227 @@
...
+                        format!("Failed to extract constructor parameters from URI: {}", e),
+                        None,
+                    )
+                })?
+        }
+    };
+
+    let parsed_agent_id = ParsedAgentId::new(
+        mcp_resource.agent_type_name.clone(),
+        DataValue::Tuple(ElementValues {
```
vigoo, line 58: Is it not supporting multimodal constructor parameters by design? (I don't think it is an important limitation just asking)

golem-worker-service/src/mcp/invoke/resource.rs
```diff
@@ -0,0 +1,227 @@
...
+        tracing::error!("Failed to parse agent id: {}", e);
+        ErrorData::invalid_params(format!("Failed to parse agent id: {}", e), None)
+    })?;
+
+    let method_params_data_value = UntypedDataValue::Tuple(vec![]);
+
+    let proto_method_parameters: golem_api_grpc::proto::golem::component::UntypedDataValue =
+        method_params_data_value.into();
+
+    let principal = Principal::anonymous();
```
vigoo, line 76: Is it always anonymous now? Is there a follow-up task to support authentication?

golem-common/src/base_model/agent.rs
afsalthaj: Automatic didn't work for functions that returned multimodal if I remember correctly

golem-worker-service/src/mcp/invoke/constructor_param_extraction.rs
afsalthaj: Sure

golem-worker-service/src/mcp/invoke/resource.rs
afsalthaj: Firstly it didn't work with clients. Secondly it's intentional and as you said it's not a huge limitation.

golem-worker-service/src/mcp/invoke/resource.rs
afsalthaj: Yes auth in this sprint in my backlog. Auth and prompt support are the next tasks.

test-components/agent-mcp/golem.yaml
```diff
@@ -0,0 +1,19 @@
+# Schema for IDEA:
+# $schema: https://schema.golem.cloud/app/golem/1.5.0-dev.1/golem.schema.json
+# Schema for vscode-yaml:
+# yaml-language-server: $schema=https://schema.golem.cloud/app/golem/1.5.0-dev.1/golem.schema.json
+
+# Field reference: https://learn.golem.cloud/app-manifest#field-reference
+# Creating HTTP APIs: https://learn.golem.cloud/invoke/making-custom-apis
+
+app: agent-mcp
```
vigoo, line 9: Why do we need a separate test app for this, instead of just adding new agents to one of the existing ones?

And if we do, we need to add it to the `buid-components.sh`

afsalthaj: I understand we should try to reduce test components, but I believe one for MCP is a decent choice, and lot more additions into it may come up soon. I mean, it's a significant new feature like http and deserves an exclusive one in my view, instead of changing an existing agent like http to http-and-mcp etc. 

So, I prefer to keep it separate if you don't mind and I will add it to sh file :)

golem-common/src/base_model/agent.rs
afsalthaj: So just re-tested and yes I need this.

vigoo: Please explain "did not work". We can fix the deriver.

afsalthaj: So, WIT defines multimodal as list<tuple<string, element-value>>, and  the actual runtime value is  Value::Tuple. Not Value::Record.

What's in here is `UntypedNamedElementValue {name, value}` and the derive macro for FromVlaue will produce something that expects `Value::Record`.

So, ended up having `Expected Record value with 2 fields, got Tuple([String("Text"), Variant { ... }])`.


I think, its not a bug in the deriver to fix.

afsalthaj: One thing is untyped-named-element-value is used only in multimodal..No other wit types ends up being untyped-named-element-value.. so may be a rename will solve it, but i don't want to do that.

Also note that this is nothing related to MCP. Invoking a function that returns multimodal didn't work at all

afsalthaj:
Also merged https://github.com/golemcloud/golem/pull/2980 into this.

PR #2922, opened 2026-03-04
fix(bridge-gen): include variant case types in generated get_type()

xsistens:
## Summary

The Rust bridge SDK generator (`cli/golem-cli/src/bridge_gen/rust/mod.rs`) produces an `IntoValue` implementation for WIT variant types. The `into_value()` method correctly serialises payload-carrying cases, but the companion `get_type()` method was hardcoded to emit `typ: None` for **every** variant case — including cases that carry a payload.

This causes all bridge SDK invocations involving variant parameters with payloads to fail at runtime with errors like:

```
Type checker error: Agent method parameters type error:
Failed to parse JSON as ComponentModel value: Input null is not string
```

### What happens

1. `into_value()` correctly produces `Value::Variant { case_idx, case_value: Some(payload) }`
2. `get_type()` returns `NameOptionTypePair { name: "case-name", typ: None }` for **all** cases
3. `to_json_value()` (`golem-wasm/src/json/impl.rs:176`) checks `case.typ` — since it's `None`, it outputs `null` instead of the serialised payload
4. The Golem runtime receives `null` where it expects the actual variant payload and rejects the invocation

### The fix

Collect per-case type tokens during variant code generation, using the existing `analysed_type_as_value()` helper (which already handles this correctly for the response-schema path at line 1140+). Payload cases now emit `typ: Some(analysed_type)` with the full `AnalysedType`, while unit cases continue to emit `typ: None`.

The bug was introduced in #2587 where the `wit_type_to_typedef` path was added alongside `analysed_type_as_value` — the latter correctly distinguishes payload vs unit cases, but the former always hardcoded `typ: None`.

## Test plan

- [ ] Verify `cargo build -p golem-cli` succeeds
- [ ] Generate a bridge SDK for a component with variant types that have payloads
- [ ] Confirm the generated `get_type()` includes `typ: Some(...)` for payload cases and `typ: None` for unit cases
- [ ] Invoke an agent method that takes a variant parameter with a payload through the bridge SDK — should succeed instead of returning "Input null is not string"

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -264,6 +264,10 @@ jobs:
             description: "IT #3"
           - name: integration-tests-group4
             description: "IT #4"
+          - name: integration-tests-group5
+            description: "IT #5"
+          - name: integration-tests-group6
+            description: "IT #6"
           - name: cli-integration-tests-group1
             description: "CLI tests #1"
           - name: cli-integration-tests-group2
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -441,27 +441,45 @@ dependencies = [
     "integration-tests-group2",
     "integration-tests-group3",
     "integration-tests-group4",
+    "integration-tests-group5",
+    "integration-tests-group6",
     "cli-integration-tests",
 ]
 
 [tasks.integration-tests-group1]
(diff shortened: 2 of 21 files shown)


xsistens:

PR #2930, opened 2026-03-04
Fix race condition when deleting an agent

vigoo:
The test for this was flaky. 
The theory is that after confirmation of the interruption (`rx.recv`) but before switching to the rejecting deleting state (`start_deleting`) a new invocation could start, picked from the pending invocation queue but never marked as interrupted/finished.

Calling `start_deleting` immediately closes the agent for further incoming invocation before the previously running one is interrupted. 

Also fixes another flaky test by properly waiting for the auto-update to finish.

### DIFF
--- a/golem-worker-executor/tests/api.rs
+++ b/golem-worker-executor/tests/api.rs
@@ -1099,6 +1099,14 @@ async fn component_env_variables_update(
         .auto_update_worker(&worker_id, updated_component.revision, false)
         .await?;
 
+    executor
+        .wait_for_component_revision(
+            &worker_id,
+            updated_component.revision,
+            Duration::from_secs(30),
+        )
+        .await?;
+
     let env = executor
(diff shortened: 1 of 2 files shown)


golem-worker-executor/tests/api.rs
```diff
@@ -1099,6 +1099,14 @@ async fn component_env_variables_update(
         .auto_update_worker(&worker_id, updated_component.revision, false)
         .await?;
 
+    executor
```
mschuwalow, line 1102: As an alternative to this there is await_ready_to_process_commands in the worker (not exposed via api currently), which will wait until replay / update is finished

vigoo: User-facing APIs like invocation supposed to work during those phases too (recovery, update) so this would only be useful for tests like this; on the other hand this kind of waiting based on worker status is what the CLI is also doing if you want to await an update. So I'd keep it like this

PR #2939, opened 2026-03-05
feat: add TLS support for Postgres

### LINKED ISSUE #2936: Enable TLS support for Postgres - golem internal usage
Currently golem does not support Postgeres instances that do not support plaintext connections. 

This issue does not cover RDBMS Host Interface for WASM Workers (Worker Capability) only covers Golem own usage of Posgres

kmatasfp:
Resolves #2936 

No mTLS support. Bare minimum needed for establishing encrypted connections with AWS RDS instances

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9462,9 +9462,9 @@ checksum = "9774ba4a74de5f7b1c1451ed6cd5285a32eddb5cccb8cc655a4e50009e06477f"
 
 [[package]]
 name = "saa"
-version = "5.3.3"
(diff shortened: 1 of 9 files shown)


kmatasfp:

kmatasfp:
yes just a feature flag, for sqlx by default, the SSL mode is [`Prefer`](PgSslMode::Prefer), and the client will first attempt an SSL connection but fallback to a non-SSL connection on failure. And the new test added only supports connecting over ssl, making sure that ssl connection is established

PR #2945, opened 2026-03-06
Agent types changes step 5 (last)

vigoo:
Step 5 of https://github.com/golemcloud/golem/issues/2835

### 1. Language-Native Agent IDs (no more kebab-case conversion)

- Agent type names now use the source language's native casing (e.g. `CounterAgent(...)` instead of `counter-agent(...)`)
- Added `source_language` field to `AgentType` (WIT, protobuf, Rust SDK, TS SDK)
- Removed `wit_naming` module (`ToWitNaming` trait) and all kebab-case conversion from agent type names, method names, field names, and variant case names
- Removed the `compact_value_formatter` module
- Bridge generators (Rust/TS) now use `same_language` flag to skip casing transformations when source and target language match, preserving original identifiers

### 2. Canonical Structural Format (`golem-common/src/model/agent/structural_format/`)

- New positional, language-independent encoding for agent ID parameter values (no field/variant/case names in the canonical form)
- `format_structural` — serialize DataValue + schema → canonical string
- `parse_structural` — parse canonical string + schema → DataValue
- `normalize_structural` — strip whitespace outside string literals
- roundtrip tests and benchmarks

### 3. Schema Evolution Validation (`golem-common/src/model/agent/schema_evolution.rs`)

- Compares old and new agent type schemas on redeployment
- Produces warnings about positionally incompatible changes that could corrupt existing agent IDs

### 4. CLI Agent ID Display (cli/golem-cli/src/agent_id_display/)

- New module with lexer, parser, and renderer for language-specific agent ID display
- Supports Rust(-like) and TypeScript - schema driven, only supports what `DataSchema` can describe

### 5. Removed Agent Wrapper Type

- Removed some remaining code dealing with wrappers / wit wrapper naming
- DB migration `006_drop_agent_wrapper_type_name` removes the column from registry service

### 6. CLI

- Now prints everything with the language-specific renderers 
- Accepts both language-specific and canonical structured forms 
- WIT type renderer and everything else WIT is gone

### 7. SDKs, tests

- Some fixes to the previous steps (there were still some kebabs)
- Tests adjusted to the new names / formats

### DIFF
--- a/golem-registry-service/src/services/deployment/deployment_context.rs
+++ b/golem-registry-service/src/services/deployment/deployment_context.rs
@@ -146,9 +146,11 @@ impl DeploymentContext {
                         .insert(agent_type_name, registered_agent_type)
                         .is_some()
                     {
-                        Err(DeployValidationError::AmbiguousAgentTypeName(
-                            agent_type.type_name.clone(),
-                        ))
+                        Err(DeployValidationError::ConflictingAgentTypeNames {
+                            name1: agent_type.type_name.clone(),
+                            name2: agent_type.type_name.clone(),
+                            normalized: agent_type.type_name.0.to_kebab_case(),
+                        })
                     } else {
                         Ok(())
                     },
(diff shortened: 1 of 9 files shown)


golem-registry-service/src/services/deployment/deployment_context.rs
```diff
@@ -349,3 +369,38 @@ fn validate_final_router(
...
+            if kebab_map.insert(kebab, name).is_some() {
+                has_collision = true;
+            }
+        }
+        assert!(
+            !has_collision,
+            "Should not have collision for distinct kebab names"
+        );
+    }
+}
```
mschuwalow: These tests don't really do anything

golem-registry-service/src/services/deployment/write.rs
```diff
@@ -167,6 +167,14 @@ pub enum DeployValidationError {
     },
     #[error("Invalid http method: {method:?}")]
     InvalidHttpMethod { method: HttpMethod },
+    #[error(
+        "Agent type names '{name1}' and '{name2}' conflict: both normalize to '{normalized}' in kebab-case"
+    )]
+    ConflictingAgentTypeNames {
```
mschuwalow, line 171: This is the same error that the existing AmbiguousAgentTypeName is for. This one is better, the old one can be removed.

golem-common/src/model/agent/schema_evolution.rs
```diff
@@ -0,0 +1,754 @@
...
+            f,
+            "Agent '{}' at {}: {}",
+            self.agent_type_name, self.path, self.description
+        )
+    }
+}
+
+/// Compare old and new agent type lists and produce warnings about
+/// positionally incompatible schema changes.
+pub fn validate_schema_evolution(
```
mschuwalow, line 47: Looks like this is only used as a warning in the cli for now. If we make this an error this should be done on the server side though.

vigoo: Yes I wanted to just have a warning for now.

PR #2949, opened 2026-03-09
wasi-http and error handling improvements

### LINKED ISSUE #2068: Improve Host HTTP implementation
The current (default) HTTP implementation:
- has TODOs around error handling, and those can result in locks when e.g. open file limits are hit
- does no caching of DNS or TCP connections

These problems were mitigated for `golem server run` with increasing open file limits when the server starts, but we probably should adress at least the error handling, and eventually we should reach performance optimizations, where caching should matter.

vigoo:
wasmtime side: https://github.com/golemcloud/wasmtime/pull/5

Resolves #2068

### DIFF
--- a/golem-worker-executor/src/services/golem_config.rs
+++ b/golem-worker-executor/src/services/golem_config.rs
@@ -16,7 +16,7 @@ use anyhow::Context;
 use figment::providers::{Format, Toml};
 use figment::Figment;
 use golem_common::config::{
-    ConfigExample, ConfigLoader, DbSqliteConfig, HasConfigExamples, RedisConfig,
+    ConfigExample, ConfigLoader, DbPostgresConfig, DbSqliteConfig, HasConfigExamples, RedisConfig,
 };
 use golem_common::model::base64::Base64;
 use golem_common::model::RetryConfig;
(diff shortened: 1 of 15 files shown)


golem-worker-executor/src/services/golem_config.rs
```diff
@@ -1228,6 +1237,74 @@ impl Default for AgentTypesServiceGrpcConfig {
...
 }
 
+/// Configuration for the HTTP connection pool used by outgoing wasi:http requests.
+///
+/// When `pool_enabled` is true, a shared connection pool is created at executor startup
+/// and reused across all workers, reducing TCP+TLS connection overhead.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct HttpClientConfig {
+    /// Whether to use a shared HTTP connection pool for outgoing requests.
+    pub pool_enabled: bool,
```
mschuwalow: Can we make it an enum instead to be consistent with our other configs

vigoo: Yes good point

vigoo:
The divergence issue with wasi-http that we experienced earlier was reproducible with the connection pool changes more easily. The fix is included. The problem was that request-body output stream operations were not persisted, so in case writing to the output failed (for example because a race cancelled the request, or any remote error happens) this error was not persisted and during replay caused a divergence:

image

PR #2954, opened 2026-03-09
IndexStorage implementation for postgres

kmatasfp:
Resolves: #2951 

Postgre IndexStorage implementation for worker executor can be enabled with:
```
GOLEM__INDEXED_STORAGE__TYPE="Postgres"
GOLEM__INDEXED_STORAGE__CONFIG__HOST="your-postgres-endpoint"
GOLEM__INDEXED_STORAGE__CONFIG__PORT=5432
GOLEM__INDEXED_STORAGE__CONFIG__DATABASE="golem_worker_executor"
GOLEM__INDEXED_STORAGE__CONFIG__USERNAME="postgres"
GOLEM__INDEXED_STORAGE__CONFIG__PASSWORD="***"
GOLEM__INDEXED_STORAGE__CONFIG__MAX_CONNECTIONS=10
```


golem-worker-executor/db/migration/postgres/001_init.sql
```diff
@@ -0,0 +1,17 @@
...
+    namespace TEXT NOT NULL,
+    key TEXT NOT NULL,
+    id BIGINT NOT NULL,
+    value BYTEA NOT NULL,
+    PRIMARY KEY (namespace, key, id)
+);
+
+CREATE INDEX IF NOT EXISTS idx_index_storage_ns_key ON index_storage (namespace, key);
+
+ALTER TABLE index_storage SET (
```
kmatasfp, line 9: fyi all the config here is to make `drop_prefix` less painful

kmatasfp: prob still need to tune these based on how it behaves in the wild, but should be good starting point

golem-worker-executor/db/migration/postgres/001_init.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE IF NOT EXISTS index_storage (
```
noise64: When we use the migrator (which does migrations in transactions), we intentionally not use "IF NOT EXISTS" in the migration scripts, as the exists case "should not happen", so we want to see if that goes wrong.

golem-worker-executor/db/migration/postgres/001_init.sql
```diff
@@ -0,0 +1,17 @@
...
+    namespace TEXT NOT NULL,
+    key TEXT NOT NULL,
+    id BIGINT NOT NULL,
+    value BYTEA NOT NULL,
+    PRIMARY KEY (namespace, key, id)
+);
+
+CREATE INDEX IF NOT EXISTS idx_index_storage_ns_key ON index_storage (namespace, key);
```
noise64: This should be already covered by the primary key with the same performance, so i don't think we need this, would only "double" the space and bookkeeping.

golem-worker-executor/db/migration/postgres/001_init.sql
noise64: looks reasonable, but i would like to add some periodic montiroing on the stats, vacumming and CPU (as a separate issue, later), so we can see how these perform (and this is just a generic reminder, we also should monitor our other postgres instances)

golem-worker-executor/src/storage/indexed/postgres.rs
```diff
@@ -0,0 +1,397 @@
...
+    async fn scan(
+        &self,
+        svc_name: &'static str,
+        api_name: &'static str,
+        namespace: IndexedStorageMetaNamespace,
+        pattern: &str,
+        cursor: ScanCursor,
+        count: u64,
+    ) -> Result<(ScanCursor, Vec<String>), String> {
+        let key = pattern.replace('*', "%").replace('?', "_");
```
noise64: even though we hopefully do not use such keys or patterns in general, this should be escaped to handle source '%' and '_' (and not sure if we have to support escaping the other way around `*` and `_`, but hopefull not)

golem-worker-executor/src/storage/indexed/postgres.rs
```diff
@@ -0,0 +1,397 @@
...
+        let query = sqlx::query_as::<_, (String,)>(
+            "SELECT DISTINCT key FROM index_storage WHERE namespace = $1 AND key LIKE $2 ORDER BY key LIMIT $3 OFFSET $4;",
```
noise64: `GROUP BY key` might perform better (or might not matter, depending on the actual plan, and if we consistently have only a limited amount of rows, but i think it safer to use it here)

golem-worker-executor/src/storage/indexed/postgres.rs
```diff
@@ -0,0 +1,397 @@
...
+    }
+
+    async fn delete(
+        &self,
+        svc_name: &'static str,
+        api_name: &'static str,
+        namespace: IndexedStorageNamespace,
+        key: &str,
+    ) -> Result<(), String> {
+        let query = sqlx::query("DELETE FROM index_storage WHERE namespace = $1 AND key = $2;")
```
noise64, line 292: i think either:
- we should care about row count and blocking here too
- or we should not care in the drop prefix neither (if we consistently keep removing old entries, and the usual max number of entries are around a reasonable batch size (not more then a few thoushands, then i think that is safe to assume))

golem-worker-executor/src/storage/indexed/postgres.rs
```diff
@@ -0,0 +1,397 @@
...
+        namespace: IndexedStorageNamespace,
+        key: &str,
+        last_dropped_id: u64,
+    ) -> Result<(), String> {
+        let namespace = Self::namespace(namespace);
+        let mut deleted_rows = self.drop_prefix_delete_batch_size;
+
+        while deleted_rows >= self.drop_prefix_delete_batch_size {
+            let query = sqlx::query(
+                "DELETE FROM index_storage WHERE ctid IN (SELECT ctid FROM index_storage WHERE namespace = $1 AND key = $2 AND id <= $3 LIMIT $4);",
```
noise64: see my comment on delete in general, but if we want this, then `DELETE FROM .. t USING (..) s WHERE t.ctid = s.ctid;` might be better

golem-worker-executor/src/storage/indexed/postgres.rs
```diff
@@ -0,0 +1,397 @@
...
+    async fn last(
+        &self,
+        svc_name: &'static str,
+        api_name: &'static str,
+        _entity_name: &'static str,
+        namespace: IndexedStorageNamespace,
+        key: &str,
+    ) -> Result<Option<(u64, Vec<u8>)>, String> {
+        let query = sqlx::query_as::<_, DBIdValue>(
+            "SELECT id, value FROM index_storage WHERE namespace = $1 AND key = $2 ORDER BY id DESC LIMIT 1;",
```
noise64, line 363: if we want to support this efficiently (and i think we do), then we should have an index for `namespace, key, id DESC)`

golem-worker-executor/tests/indexed_storage.rs
```diff
@@ -173,6 +177,67 @@ async fn multi_sqlite_storage(
...
+                .expect("Invalid postgres connection string")
+                .port()
+                .expect("Postgres connection string missing port"),
+            max_connections: 10,
+            schema: None,
+        };
+
+        let config = IndexedStoragePostgresConfig {
+            postgres,
+            drop_prefix_delete_batch_size: 1000,
```
noise64: i see that this is test, but is this intentinally not matching the value / default (1024) used in other places?

golem-worker-executor/src/storage/indexed/postgres.rs
```diff
@@ -0,0 +1,397 @@
...
+            .bind(key)
+            .bind(id as i64)
+            .bind(value);
+
+            tx.execute(query)
+                .await
+                .map_err(|err| err.to_safe_string())?;
+        }
+
+        tx.commit().await.map_err(|err| err.to_safe_string())
```
noise64: For multi statement transactions please use `self.pool.with_rw` of `self.pool.with_rw_err`. Otherwise the monitoring and logging parts for warns and errors can be either lost or have to be manually re-implemented. (This is something that was only recently introduced in the last release, so we might still have examples not using this pattern, but we should also get rid of those soon.)

golem-worker-executor/db/migration/postgres/001_init.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE IF NOT EXISTS index_storage (
+    namespace TEXT NOT NULL,
+    key TEXT NOT NULL,
+    id BIGINT NOT NULL,
+    value BYTEA NOT NULL,
+    PRIMARY KEY (namespace, key, id)
+);
```
noise64, line 7: One thing we might want to consider is partioning by HASH(namespace, key), which is probably easy to add now, and can help us buying more time until we want to use a more scaleable backend. I will think a bit more about this, and we can discuss / add it as a seperate card, if we wan to, until release.

golem-worker-executor/src/storage/indexed/postgres.rs
```diff
@@ -0,0 +1,397 @@
...
+    }
+
+    async fn scan(
+        &self,
+        svc_name: &'static str,
+        api_name: &'static str,
+        namespace: IndexedStorageMetaNamespace,
+        pattern: &str,
+        cursor: ScanCursor,
+        count: u64,
```
noise64, line 160: This is a generic comment, not for this issue and backend: after discussing with @vigoo a bit, we might want to make the ScanCursor string based (instead of integer), that ideally could allow us making scanning perform better, e.g. with progress, we could skip the offset part, and use the index directly.

this is also something, that i would like to think a bit more, and discuss / solve as a separate issue, if needed.

golem-worker-executor/src/storage/indexed/postgres.rs
vigoo: the key is an agent-id and even with the new canonical representation, that definitely can have these characters in it (think of a string constuctor parameter, for example)

noise64: then we need both layers of escaping, do we have any spec for our scan patterns?

noise64: If we do not yet have that, then we should define it, including escaping, and also collect and define the valid usages (as in, do we really want to support wildcards at any place, or only want to do prefix filtering? this can matter a lot on many backends for perfomance); then review all the backends.

golem-worker-executor/tests/indexed_storage.rs
kmatasfp: as this is a test I passed in just some number here, no deeper meaning, in real code its 1024 by default:
https://github.com/golemcloud/golem/pull/2954/changes#diff-8ea693d288c4e5527913f28dcccdaaa7f23a225b07068a5b4d6df008ffc22e9fR733

golem-worker-executor/src/storage/indexed/postgres.rs
kmatasfp: was not sure either what kind of patterns can be passed in here, so did the same as sqlite impl for now

golem-worker-executor/src/storage/indexed/postgres.rs
kmatasfp: not sure if there will be much difference, i think the plans produced will be about the same. I would consider changing it to GROUP BY if we planning to add aggregates sometime soon. My 2 cents lets keep the DISTINCT, clearer indent imho, for now and if we find it behaves poorly can change to GROUP BY later to see if its faster

golem-worker-executor/db/migration/postgres/001_init.sql
kmatasfp: makes sense, removing it

golem-worker-executor/src/storage/indexed/postgres.rs
kmatasfp: I think here adding this index will not help much, and would make write slower. Same reasoning you gave to remove the index we had in migration file. Primary key index should cover it and index allows to go backwards and forwards afaik.

golem-worker-executor/src/storage/indexed/postgres.rs
kmatasfp: done, changed to use CTE

golem-worker-executor/src/storage/indexed/postgres.rs
noise64: sure, this is something easy to change and measure, even later

golem-worker-executor/tests/indexed_storage.rs
kmatasfp: changed to 1024, not to confused anyone else

golem-worker-executor/src/storage/indexed/postgres.rs
kmatasfp: can you elaborate more what do you mean. I create the tx using `self.pool.with_rw`, see here:
```
let mut tx = self
            .pool
            .with_rw(svc_name, api_name)
            .begin()
            .await
            .map_err(|err| err.to_safe_string())?;
```

Also `with_rw_err`  does not seem to be a thing

golem-worker-executor/src/storage/indexed/postgres.rs
kmatasfp: lets keep this as is for not, only change the `drop_prefix` to use CTE to start with

golem-worker-executor/src/storage/indexed/postgres.rs
noise64: sorry, typo, i meant `pool.with_tx`

noise64: https://github.com/golemcloud/golem/blob/3794771f590e860053965614565e671dc2c05892/golem-service-base/src/db/mod.rs#L99

golem-worker-executor/src/storage/indexed/postgres.rs
noise64: i'm adding a separate issue for this

kmatasfp: anyway added more escaping, its extensible now, so can add more to it later

PR #2955, opened 2026-03-10
Fix missing load snapshot after manual snapshot update

mschuwalow:
This fixes two bugs:

### Duplicate enqueued updates 
When the agent does not have any pending work in the active queue and checks pending_updates, it was always performing a manual update when it had a pending update. That is incorrect in two ways:
*  if the update is an automatic update a manual update was enqueued immediately afterwards
*  if the update was a manual update, it enqueued a duplicate manual update. If the manual update made it the pending_updates queue (instead of the pending invocation queue), it means the save-snapshot was already performed and the agent is ready for restart / load-snapshot

This is a bit difficult to reproduce in our current tests as it only shows up in a particular interleaving due to a race with the interupt / restart after an update is enqueued. The bug can be reproduced consistently if that restart is commented out.

### Manual updates do not call load-snapshot on replay start
Manual updates do not call load-snapshot on replay start, instead just resuming from the update-succeeded oplog entry. This results in agents failing due to them being uninitialized (original initialize is skipped due to the update, but load-snapshot is not called). The fix for this is to do the same as for the automatic snapshots and track the last save-snapshot payload that needs to be loaded.

### DIFF
--- a/golem-common/src/model/mod.rs
+++ b/golem-common/src/model/mod.rs
@@ -699,7 +699,7 @@ pub struct AgentStatusRecord {
     pub current_retry_count: HashMap<OplogIndex, u32>,
     /// Index of the last manual update snapshot index. Agent will call load_snapshot
     /// on this payload before starting replay.
-    pub last_manual_snapshot_index: Option<OplogIndex>,
+    pub last_manual_update_snapshot_index: Option<OplogIndex>,
     /// Index of the last automatic snapshot index. Must be >= last_manual_snapshot_index.
     /// Agent will call load_snapshot on this payload before starting replay. If the load_snapshot
     /// fails this will be ignored and a full replay from last_manual_snapshot_index will performed.
@@ -727,7 +727,7 @@ impl Default for AgentStatusRecord {
             deleted_regions: DeletedRegions::new(),
             component_revision_for_replay: ComponentRevision::INITIAL,
             current_retry_count: HashMap::new(),
-            last_manual_snapshot_index: None,
+            last_manual_update_snapshot_index: None,
             last_automatic_snapshot_index: None,
         }
     }
--- a/golem-worker-executor/src/worker/mod.rs
+++ b/golem-worker-executor/src/worker/mod.rs
@@ -2163,14 +2163,16 @@ impl RunningWorker {
             );
 
         let mut skipped_regions = worker_metadata.last_known_status.skipped_regions;
-        let mut last_snapshot_index = worker_metadata.last_known_status.last_manual_snapshot_index;
+        let mut last_snapshot_index = worker_metadata
+            .last_known_status
+            .last_manual_update_snapshot_index;
 
+        // automatic snapshots are only considered until the first failure.
+        // additionally, if there are updates, the automatic snapshot is temporarily ignored to catch issues earlier
         if let Some(snapshot_idx) = worker_metadata
             .last_known_status
             .last_automatic_snapshot_index
         {
-            // automatic snapshots are only considered until the first failure
-            // if there are updates, ignore the automatic snapshot temporarily to catch issues earlier
             if pending_update.is_none()
                 && !parent.snapshot_recovery_disabled.load(Ordering::Acquire)
             {
--- a/golem-worker-executor/src/worker/status.rs
+++ b/golem-worker-executor/src/worker/status.rs
@@ -153,7 +153,7 @@ pub async fn update_status_with_new_entries<T: HasOplogService + Sync>(
         component_revision,
         component_size,
         component_revision_for_replay,
-        last_manual_snapshot_index,
+        last_manual_update_snapshot_index,
         last_automatic_snapshot_index,
     ) = calculate_update_fields(
         last_known.pending_updates,
@@ -162,7 +162,7 @@ pub async fn update_status_with_new_entries<T: HasOplogService + Sync>(
         last_known.component_revision,
         last_known.component_size,
         last_known.component_revision_for_replay,
-        last_known.last_manual_snapshot_index,
+        last_known.last_manual_update_snapshot_index,
         last_known.last_automatic_snapshot_index,
         &deleted_regions,
         &new_entries,
@@ -209,7 +209,7 @@ pub async fn update_status_with_new_entries<T: HasOplogService + Sync>(
         deleted_regions,
         component_revision_for_replay,
         current_retry_count,
-        last_manual_snapshot_index,
+        last_manual_update_snapshot_index,
         last_automatic_snapshot_index,
     };
 
@@ -594,7 +594,7 @@ fn calculate_update_fields(
     initial_revision: ComponentRevision,
     initial_component_size: u64,
     initial_component_revision_for_replay: ComponentRevision,
-    initial_last_manual_snapshot_index: Option<OplogIndex>,
+    initial_last_manual_update_snapshot_index: Option<OplogIndex>,
     initial_last_automatic_snapshot_index: Option<OplogIndex>,
     deleted_regions: &DeletedRegions,
     entries: &BTreeMap<OplogIndex, OplogEntry>,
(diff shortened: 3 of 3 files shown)


golem-worker-executor/src/worker/mod.rs
```diff
@@ -2162,6 +2162,29 @@ impl RunningWorker {
...
 
+        let mut skipped_regions = worker_metadata.last_known_status.skipped_regions;
+        let mut last_snapshot_index = worker_metadata.last_known_status.last_manual_snapshot_index;
+
+        if let Some(snapshot_idx) = worker_metadata
+            .last_known_status
+            .last_automatic_snapshot_index
+        {
+            // automatic snapshots are only considered until the first failure
+            // if there are updates, ignore the automatic snapshot temporarily to catch issues earlier
```
mschuwalow: I assume that is why the pending_update.is_none() case was there

golem-worker-executor/src/worker/invocation_loop.rs
```diff
@@ -327,22 +327,13 @@ impl<Ctx: WorkerCtx> InnerInvocationLoop<'_, Ctx> {
...
-                    owned_agent_id: self.owned_agent_id.clone(),
-                    parent: self.parent.clone(),
-                    instance: self.instance,
-                    store: store.deref_mut(),
-                };
-                match invocation.manual_update(target_revision).await {
-                    CommandOutcome::Continue => continue,
-                    other => break other,
-                }
+            if status.pending_updates.front().is_some() {
```
vigoo, line 333: Are we sure that there is always an AgentInvocation::ManualUpdate enqueued _and processed_ for saving the snapshot before we reach here? Probably this logic became a bit obscure through all the refactorings.
What if there are multiple manual updates enqueued? When we perform the first, and restart, what enqueues the thing in the command queue?

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -1173,6 +1173,13 @@ impl<Ctx: WorkerCtx + DurableWorkerCtxView<Ctx>> DurableWorkerCtx<Ctx> {
             OplogEntry::Snapshot {
                 data, mime_type, ..
             } => (data, mime_type),
+            OplogEntry::PendingUpdate {
```
vigoo, line 1176: I understand this fixes the main issue, and now that we have the snapshot-based recovery this is probably a nice way to fix it but very confused that there was no machinery for this earlier.

I think it was supposed to be something like this:
- once a manual update succeeds, it adds a skipped region from the beginning to the update point so that oplog part is ignored
- so on next recovery, we reach the `OplogEntry::PendingUpdate { SnapshotBased }` again, and the recovery code calls `load-snapshot` so our state is restored
- the rest of the oplog gets replayed

So I don't fully understand why this did not work, and whether it is a problem or not that now we have two ways to recover from a snapshot-based update. (As I think you did not touch the old one)

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2325,11 +2332,25 @@ impl<Ctx: WorkerCtx + DurableWorkerCtxView<Ctx>> ExternalOperations<Ctx> for Dur
                         UpdateDescription::Automatic {
                             target_revision, ..
                         } => {
-                            // snapshot update will be succeeded as part of the replay.
-                            let result = Self::resume_replay(store, instance, false).await;
-                            record_resume_worker(start.elapsed());
+                            let replay_result = async {
+                                if let SnapshotRecoveryResult::Failed =
+                                    Self::try_load_snapshot(store, instance).await
```
vigoo, line 2337: Maybe I misunderstand this part - but for automatic updates we should never use snapshots, but replay always from the beginning.

golem-common/src/model/mod.rs
```diff
@@ -697,7 +697,13 @@ pub struct AgentStatusRecord {
     /// The number of encountered error entries grouped by their 'retry_from' index, calculated from
     /// the last invocation boundary.
     pub current_retry_count: HashMap<OplogIndex, u32>,
-    pub last_snapshot_index: Option<OplogIndex>,
+    /// Index of the last manual update snapshot index. Agent will call load_snapshot
+    /// on this payload before starting replay.
+    pub last_manual_snapshot_index: Option<OplogIndex>,
```
vigoo: The name is a bit confusing to me without the comments, how about
- `last_manual_update_snapshot_index`
- `last_snapshot_index` (or we can have automatic in it but what I'd like is that one is for updates, and the other has nothing to do with updates - and automatic updates is a thing which has nothing to do with this)

golem-worker-executor/src/durable_host/mod.rs
mschuwalow: This is only using snapshots created by manual updates, not the automatic snapshots (the logic for that lives in the WorkerConfig creation).

I don't think we can replay the parts of the oplog that were skipped as part of a manual oplog with a new component version in general (as manual snapshot updates can freely break oplog backwards compat). So I think only replaying from the last succesful manual snapshot update is correct here.

golem-common/src/model/mod.rs
mschuwalow: let's do last_manual_update_snapshot_index + last_automatic_snapshot_index :+1: 

~Will update~ done

golem-worker-executor/src/durable_host/mod.rs
mschuwalow: I believe this was always broken or at least broken for a very long time.

PendingUpdate is a hint entry and I don't see any logic in replay that would replay load-snapshot. This is surprising to me too, maybe I missed something

golem-worker-executor/src/worker/invocation_loop.rs
mschuwalow: Yes. pending_updates is only based on the pending_update oplog entry, which for manual snapshot updates is only created during the following path:

agent receives update request via grpc -> agent writes a manual update pending_agent_invocation oplog entry -> agent processes the invocation -> agent writes a pending_update oplog entry

For automatic updates the pending_update oplog entry is written immediately

golem-worker-executor/src/durable_host/mod.rs
vigoo: My understanding of how it work(ed) on `main`:

- manual update request arrives, it gets enqueued in the invocation queue (because it needs to save snapshot between invocations) (https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/grpc/mod.rs#L1130)
- the invocation loop picks it up and runs `manual_update` - invokes the save-snapshot function, adds a pending update entry with the actual snapshot and restarts the agent (https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/worker/invocation_loop.rs#L718)
- The pending update applies a temporary "skipped region override" that skips everything after `CREATE` until the pending update entry. It's an override and not just a "set", because in case of a failed update attempt we undo it (https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/worker/status.rs#L405-L415)
- because of the pending update entry, the agent reloads with the new component version (https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/worker/mod.rs#L2102-L2116)
- it gets to `prepare_instance`, that finds that we are executing a pending snapshot-based update, and skips the replay, instead goes to `finalize_pending_snapshot_update` (https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/durable_host/mod.rs#L2320-L2323)
- here we call `load-snapshot` (the agent sdk is supposed to internally intialize the agent as part of this) (https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/durable_host/mod.rs#L1048-L1055)
- if it fails we mark it as a failed update, the pending update is gone from the queue, and we restart so we continue with the previous version (or pick a next pending update)
- if it succeeds, we record a successful update oplog entry (https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/durable_host/mod.rs#L1832-L1846) and the agent is considered ready
- The worker status calculation collapses the skipped region (basically making the override final): https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/worker/status.rs#L417-L422

I think the part that's hard to see/understand is the skipped region logic that now 100% lives in the worker status calculation. Previously it was explicitly set in various points of the above update logic, making it easier to follow (but the current way of calculating everything directly from the oplog is definitely the correct way)

mschuwalow: What you are saying is exactly correct, but it's not doing this part in any following replays as pending_updates will not contain the update anymore (as it now has a successful update entry)

```
here we call load-snapshot (the agent sdk is supposed to internally intialize the agent as part of this) (https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/durable_host/mod.rs#L1048-L1055)
```

PR #2958, opened 2026-03-10
agent-config: add support for agent config to cli/manifest and make naming in the api more consistent

### LINKED ISSUE #2803: Expose agent-type specific local configuration in the golem manifest
They should be part of the component metadata, agent-type-specific and should trigger creation / deployment of a new component revision if changes are detected.

### LINKED ISSUE #2818: Adjust the cli/manifest to allow managing defaults for secrets
e backend can figure out the required secrets during component creation / update and upsert the required keys (failing on type mismatch). One feature that would be useful in the manifest is being able to specify default values for secrets to simplify local demoing / development.
When specified in the manifest, only keys declared by at least one component are allowed and the types have to match.

mschuwalow:
resolves #2803
resolves #2818

Wit naming will be updated as part of next pr as that one will need to touch wit anyway

manifest will look like this:

```
secretDefaults:
   local:
      - path: [ "secret" ]
        value: "a"
      - path: [ "nested", "nestedSecret" ]
        value: 1

components:
  golem-it:agent-sdk-ts:
    templates: ts
    agents:
       ConfigAgent:
          config:
          - path: [ "foo" ]
            value: 1
          - path: [ "bar" ]
            value: "bar"
          - path: [ "nested", "a" ]
            value: true
          - path: [ "nested", "b" ]
            value: [1, 2, 3]
          - path: [ "aliasedNested", "c" ]
            value: 1
```

### DIFF
--- a/cli/golem-cli/src/command.rs
+++ b/cli/golem-cli/src/command.rs
@@ -27,7 +27,6 @@ use crate::command::shared_args::{
 use crate::command::worker::AgentSubcommand;
 use crate::config::ProfileName;
 use crate::error::ShowClapHelpTarget;
-use crate::log::LogColorize;
 use crate::model::app::ComponentPresetName;
 use crate::model::cli_command_metadata::{CliCommandMetadata, CliMetadataFilter};
 use crate::model::environment::EnvironmentReference;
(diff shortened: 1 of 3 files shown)


cli/golem-cli/src/command.rs
```diff
@@ -1095,6 +1098,9 @@ pub mod worker {
             /// wasi:config entries visible for the agent
             #[arg(short, long, value_parser = parse_key_val, value_name = "VAR=VAL")]
             config_vars: Vec<(String, String)>,
+            /// agent config for entries
```
vigoo: This is not a very understandable help string

cli/golem-cli/src/command.rs
```diff
@@ -1654,6 +1660,91 @@ fn parse_key_val(key_and_val: &str) -> anyhow::Result<(String, String)> {
     ))
 }
 
+fn parse_worker_agent_config(s: &str) -> anyhow::Result<WorkerAgentConfigEntry> {
```
vigoo: Maybe these parsing functions with their corresponding tests could live in a small submodule in a different file, this one is getting huge

vigoo (approved):

PR #2961, opened 2026-03-11
Add prompts in MCP

### LINKED ISSUE #2822: Implement listing prompt and prompt handlers

afsalthaj:
Fixes #2822


afsalthaj:
Merging with https://github.com/golemcloud/golem/pull/2918, and reduce the number of ongoing PRs. This PR is useful to see how prompts are handled, and the original PR is already reviewed

PR #2965, opened 2026-03-11
agent-config: normalize agent secret paths & also run tests using rust sdk

### LINKED ISSUE #2953: Ensure secrets are exposed in a user friendly format by the api

mschuwalow:
resolves #2953


golem-registry-service/src/services/agent_secret.rs
vigoo: Should this be `DeleteAgentScecret`?

mschuwalow: Yes, good catch!

PR #2966, opened 2026-03-11
RoutingTablePersistence for Postgres

kmatasfp:
Resolves: #2963 

Can be enabled with:
```
GOLEM__PERSISTENCE__TYPE=Postgres
GOLEM__PERSISTENCE__CONFIG__HOST=some-postgres-endpoint
GOLEM__PERSISTENCE__CONFIG__PORT=5432
GOLEM__PERSISTENCE__CONFIG__DATABASE=some-db
GOLEM__PERSISTENCE__CONFIG__USERNAME=postgres
GOLEM__PERSISTENCE__CONFIG__PASSWORD=***
GOLEM__PERSISTENCE__CONFIG__MAX_CONNECTIONS=10
```

### DIFF
--- a/golem-shard-manager/src/persistence.rs
+++ b/golem-shard-manager/src/persistence.rs
@@ -14,6 +14,7 @@
 
 use crate::error::ShardManagerError;
 use crate::model::{RoutingTable, ShardManagerState};
+use anyhow::anyhow;
 use async_trait::async_trait;
 use bytes::Bytes;
 use golem_common::config::DbPostgresConfig;
@@ -153,7 +154,7 @@ impl RoutingTablePersistence for RoutingTablePostgresPersistence {
         if let Some(row) = row {
             let bytes: Vec<u8> = row
                 .try_get("state")
-                .map_err(|err| RepoError::InternalError(anyhow::Error::msg(err)))?;
+                .map_err(|err| RepoError::InternalError(anyhow!(err)))?;
             let shard_manager_state: ShardManagerState =
                 deserialize(&bytes).map_err(ShardManagerError::SerializationError)?;
             Ok(shard_manager_state.get_routing_table())


golem-shard-manager/src/persistence.rs
```diff
@@ -85,6 +95,73 @@ pub struct RoutingTableFileSystemPersistence {
...
+    async fn write(&self, routing_table: &RoutingTable) -> Result<(), ShardManagerError> {
+        let shard_manager_state = ShardManagerState::new(routing_table);
+        let encoded =
+            serialize(&shard_manager_state).map_err(ShardManagerError::SerializationError)?;
+
+        self.pool
+            .with_rw(PERSISTENCE_SVC, "write")
+            .execute(sqlx::query(
+            "INSERT INTO shard_manager_state (id, state) VALUES (1, $1) ON CONFLICT (id) DO UPDATE SET state = EXCLUDED.state",
+        )
```
noise64: did the formatter glitched on this? :D

golem-shard-manager/src/persistence.rs
```diff
@@ -85,6 +95,73 @@ pub struct RoutingTableFileSystemPersistence {
...
+impl RoutingTablePersistence for RoutingTablePostgresPersistence {
+    async fn write(&self, routing_table: &RoutingTable) -> Result<(), ShardManagerError> {
+        let shard_manager_state = ShardManagerState::new(routing_table);
+        let encoded =
+            serialize(&shard_manager_state).map_err(ShardManagerError::SerializationError)?;
+
+        self.pool
+            .with_rw(PERSISTENCE_SVC, "write")
+            .execute(sqlx::query(
+            "INSERT INTO shard_manager_state (id, state) VALUES (1, $1) ON CONFLICT (id) DO UPDATE SET state = EXCLUDED.state",
```
noise64: maybe it would be a bit nicer for the readers to add constant for the hardcoded id.

golem-shard-manager/src/persistence.rs
```diff
@@ -85,6 +95,73 @@ pub struct RoutingTableFileSystemPersistence {
...
+        )
+            .bind(encoded))
+            .await
+            .map_err(|err| ShardManagerError::PostgresError(sqlx::Error::Protocol(err.to_string())))?;
```
noise64: why do we have to repackage an error that is converted to a string? can't we use the error as it is?

kmatasfp: I did it this way, because in `read` when we do `let bytes: Vec<u8> = row.try_get("state")?` it returns `sqlx::Error`,  I can introduce `anyhow::Error`, but imho that is heavier than what I did

golem-shard-manager/src/persistence.rs
kmatasfp: this is how `cargo fmt` thinks it should be formatted? Changed it manually

golem-shard-manager/src/persistence.rs
kmatasfp: what do you mean? I did not add any constants for queries here either https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/storage/keyvalue/postgres.rs

noise64: But there was no hardcoded `1`, like here

golem-shard-manager/src/persistence.rs
noise64: If it returns sqlx, then why do you convert it a string and convert it back to sqlx::Error::Protocol?

golem-shard-manager/src/persistence.rs
noise64: sometimes it "gives up" on some code

golem-shard-manager/src/persistence.rs
kmatasfp: anyway pushed alternative way, can see if you like it better

kmatasfp: > If it returns sqlx, then why do you convert it a string and convert it back to sqlx::Error::Protocol?

Hmm I am not doing that, `let bytes: Vec<u8> = row.try_get("state")` comes after

kmatasfp: anyway see, I changed it

golem-shard-manager/src/persistence.rs
kmatasfp: oh now I see what you mean, was not clear that you meant :)

kmatasfp: also not sure if following is easier to read:
```
self.pool
            .with_rw(PERSISTENCE_SVC, "write")
            .execute(sqlx::query(
                "INSERT INTO shard_manager_state (id, state) VALUES ($1, $2) ON CONFLICT (id) DO UPDATE SET state = EXCLUDED.state",
            )
            .bind(POSTGRES_SHARD_ROW_ID)
            .bind(encoded))
            .await
            .map_err(ShardManagerError::PostgresError)?;
```

noise64: then let's keep it like this for now, we can change it, if we ever need two rows :D

PR #2967, opened 2026-03-12
throw agent-error instead of custom for http mount validations

### LINKED ISSUE #2935: Make sure constructor schema validation for http mount returns agent-error
Make sure to return agent-error instead of throwing error when the mount schema doesn't match constructor 

Image

afsalthaj:
Fixes #2935 

I am feeling not much "outcome" from this change

```
  Adding metadata to my:ts
Preparing deployment
  Extracting my:ts agent types from /Users/afsalthaj/agent-http/components-ts/my-ts/../../golem-temp/agents/my_ts.wasm


thread '<unnamed>' (1) panicked at src/internal.rs:112:42:
Failed to finish importing user module user:
Error: invalid-input: Agent constructor variable 'name' is not provided by the HTTP mount path.
stack backtrace:

```

Previous error is in issue description

### DIFF
--- a/sdks/ts/packages/golem-ts-sdk/src/internal/http/path.ts
+++ b/sdks/ts/packages/golem-ts-sdk/src/internal/http/path.ts
@@ -14,10 +14,11 @@
 
 import { PathSegment } from 'golem:agent/common@1.5.0';
 import { rejectEmptyString } from './validation';
+import { invalidInput } from '../agentError';
 
 export function parsePath(path: string): PathSegment[] {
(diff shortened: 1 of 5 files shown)


vigoo:
The real fix would be to not throw these errors during loading the module, but defer them and return them from the `initialize` call.

afsalthaj:
I am closing this PR for now. @vigoo

afsalthaj:
https://github.com/golemcloud/golem/pull/3032#issue-4099312208

Issue #2916, opened 2026-03-02
Add retry logic for flaky agent interactions


jdegoes:
Coding agents are non-deterministic. Add configurable retry at the step level: retry: { attempts: 2, delay: 10 }. On step failure, retry from scratch (re-send prompt, re-verify). Record all attempts in the report.

Issue #2915, opened 2026-03-02
Add --resume-from flag (skip passed steps)


jdegoes:
When debugging a failing scenario, allow resuming from a specific step ID. Skips prior steps (assumes workspace state is already correct from a previous interrupted run). Requires: workspace path flag (--workspace).

Issue #2913, opened 2026-03-02
Implement Golem state cleanup between scenarios


jdegoes:
Between scenarios, clean up Golem state to prevent leakage: golem deploy --reset --yes or targeted component deletions, verify clean state before next scenario starts, handle cleanup failures gracefully (log warning, continue).

Issue #2912, opened 2026-03-02
Implement aggregated test run report


jdegoes:
After all scenarios complete, generate a single aggregated report combining all per-scenario JSON reports: total scenarios, passed, failed, skipped, duration breakdown, worst failures summary, matrix heat map (which agent x language x os combos failed).

PR #2979, opened 2026-03-13
Add matrix filtering, retry, resume, cleanup, and HTML reports to harness

### LINKED ISSUE #2911: Implement multi-scenario loading and filtering
Load all *.yaml files from the --scenarios directory. Filter scenarios based on current (agent, language, os) tuple against each scenario's matrix block. Sort by dependency order if scenario-level dependencies exist (stretch -- may not be needed initially).

Myestery:
## Summary
- **#2911** Matrix filtering: scenarios declare supported agents/languages/OS via `matrix` field; harness skips non-matching
- **#2916** Retry logic: steps support `retry: { attempts, delay }` with per-attempt result tracking
- **#2915** `--resume-from <step-id>` and `--workspace <path>` flags to resume execution mid-scenario
- **#2913** Golem state cleanup (`golem deploy --reset --yes`) between scenarios, with `--no-cleanup` flag
- **#2912** Aggregated report with `agent`, `language`, `os`, `timestamp` metadata + `--merge-reports` mode
- **#2903** HTML summary report generation with collapsible scenario details and failure summary

## Test plan
- [ ] `cd tests/harness && npm run build` compiles cleanly
- [ ] `npm test` — all new matrix, retry, and HTML report tests pass
- [ ] `npx tsx src/run.ts --help` shows all new flags
- [ ] `--merge-reports` with sample summary.json files produces merged-summary.json and report.html
- [ ] Scenario YAML with `matrix` and `retry` blocks loads correctly
- [ ] `--resume-from` with invalid step ID throws descriptive error

Closes #2911, #2916, #2915, #2913, #2912, #2903


tests/harness/src/run.ts
```diff
@@ -19,40 +20,161 @@ interface ScenarioReport {
   artifactPaths: string[];
 }
 
+async function cleanupGolemState(cwd: string): Promise<void> {
```
vigoo: This is not really cleaning up everything that is created by a test. I think we should restart the whole server with `--clean`. (And maybe warn the user interactively, if not running on github)

PR #2983, opened 2026-03-13
Oplog processor improvements

### LINKED ISSUE #1107: Exactly once semantics for oplog processor plugins

vigoo:
Resolves #1107 

- Makes it possible to build oplog-processor plugins using `golem build`
- Existing oplog processor test passes
- Implements exactly-once delivery for oplog processors. This has a few core elements:
  - Delivery is always done using determinstic batches, and the idempotency key of the plugin call is derived, not random
  - A checkpoint is recorded in the sender oplog whenever a successful batch was sent out
  - Oplog processor plugins are still optimized to be always local RPC, but during shard reassignments this property can temporarily break. We detect this, verify that everything reached the previous (now remote) oplog processor and switch to a new local one
  - Many new tests cases trying to guarantee these
  -


afsalthaj (approved):
Merge conflicts

Issue #2972, opened 2026-03-12
Verify and fix fuel tracking


vigoo:
Check `AtomicResourceEntry` and `ResourceLimitsGrpc` and make sure everything works correctly (there are some TODO comments, and make sure the calculations are correct (verify if `delta` is positive or negative etc)
Increase test coverage if possible.

PR #2988, opened 2026-03-14
Verify and fix fuel tracking

kmatasfp:
Resolves: #2972 

Fixes following bugs:
- `Context::borrow_fuel`: `min_fuel_level` is never updated after a successful borrow, causing the next borrow to charge the cumulative total consumed rather than the incremental since the last borrow. This leads to quite big over-charging across n epoch ticks.
- `Context::borrow_fuel`: `amount_to_borrow` always at least `self.config.limits.fuel_to_borrow`, which means we always borrow fuel even if we did not burn through the fuel we previously borrowed
- `Context::return_fuel`: because `min_fuel_level` stays at`u64::MAX`, `current_level.saturating_sub(min_fuel_level)` is always 0 and unused fuel is never returned to the pool.

Also not sure if this is the behavior we want:
when `send_batch_updates` fails, the consumed fuel for that interval is lost and the account is not charged. During a sustained outage workers run for free.


vigoo:
> when send_batch_updates fails, the consumed fuel for that interval is lost and the account is not charged. During a sustained outage workers run for free.

I think this is acceptable

golem-worker-executor/src/services/resource_limits.rs
```diff
@@ -257,6 +274,56 @@ impl ResourceLimitsGrpc {
...
+
+    async fn refresh_idle_accounts(&self, account_ids: Vec<AccountId>) {
+        let zero_updates: HashMap<AccountId, i64> = account_ids.iter().map(|id| (*id, 0)).collect();
+
+        tracing::debug!(
+            "Refreshing stale resource limits for {} idle account(s)",
+            account_ids.len()
+        );
+
+        match self.client.batch_update_fuel_usage(zero_updates).await {
```
mschuwalow: Can we merge this and the other batch_update_fuel_usage request? RIght now this is 2 request per executor/interval

golem-worker-executor/src/workerctx/default.rs
```diff
@@ -77,12 +77,80 @@ use wasmtime::{AsContextMut, ResourceLimiterAsync};
...
 
+/// Tracks the wasmtime fuel gauge state for a single worker store.
+///
+/// Wasmtime fuel counts down from `u64::MAX`. Workers borrow a fixed amount of
+/// fuel from the shared [`AtomicResourceEntry`] each epoch tick, and return any
+/// unused portion at invocation end.
+pub(crate) struct FuelTracker {
+    /// The wasmtime gauge reading the last time consumption was measured.
+    /// Starts at `u64::MAX` (the gauge's initial value set at store creation).
+    pub(crate) gauge_at_last_return: u64,
```
mschuwalow: pub(self) should be enough to make it visible for tests

kmatasfp: done

kmatasfp:
@mschuwalow to make this `let sum = fuel as i128 - (delta.max(0) as i128 + in_flight.max(0) as i128);` work I think `delta` should only contain fuel that was actually consumed by wasmtime, otherwise we run out if it very fast, 10_000 units per 10ms epoch at current rate, as we only credit back if invocation returns, and invocations can take multiple epochs. Do implement that is bigger change, as all the logic everywhere is based on the borrow model.

golem-worker-executor/src/services/resource_limits.rs
kmatasfp: done

golem-worker-executor/src/services/resource_limits.rs
```diff
@@ -50,9 +52,16 @@ impl AtomicResourceEntry {
             delta: AtomicI64::new(0),
             in_flight_delta: AtomicI64::new(0),
             max_memory: AtomicUsize::new(max_memory),
+            last_refresh_secs: AtomicI64::new(Utc::now().timestamp()),
         }
     }
 
+    fn secs_since_last_refresh(&self) -> i64 {
+        Utc::now()
```
mschuwalow, line 60: Let's use std::time::Instant, as this one is not monotonic.

golem-worker-executor/src/workerctx/default.rs
```diff
@@ -77,12 +77,106 @@ use wasmtime::{AsContextMut, ResourceLimiterAsync};
...
+            fuel_to_borrow,
+        }
+    }
+
+    /// How much fuel to request from the account pool this epoch tick.
+    ///
+    /// Always at least `fuel_to_borrow`. If the gauge has already dropped below
+    /// `prepaid_gauge_floor` (the worker burned through the whole pre-paid amount
+    /// in one epoch), the full deficit is topped up instead.
+    pub(self) fn determine_amount_to_borrow(&self, current_gauge: u64) -> u64 {
```
mschuwalow, line 127: I think we can be a bit smarter here. Two things we can improve:
* I don't think there is a need to borrow more fuel if we still have enough available
* if we fall under the amount of available fuel we should probably borrow more so that we have a buffer.

I would propose that this should operate with an "unused_fuel_target", which is a delta between the current gauge and the borrowed fuel.
So the formula would be something like:
```
let target_fuel_level = current_gauge.saturating_sub(unused_fuel_target);
self.prepaid_gauge_floor.saturating_sub(target_fuel_level)
```

vigoo:
Let's make sure the latest discussion's findings is incorporated before merge (as the approve was from yesterday)

golem-worker-executor/src/services/resource_limits.rs
kmatasfp: non-monotonicity of `Utc::now()` causes at most a temporary delay or a harmless extra refresh, not big of an issue imho. Reason I used it is that  I can easily do lock free AtomicI64 math, which is not true for `std::time::Instant`, I need to use Mutex or  do something else,  not sure if the extra complexity is worth it in this case, we do not need to be super precise with timing here. Keeping `AtomicResourceEntry` lock free imho adds more value here, we are already slow 🐢 as is, no need to make it more so :)

PR #2991, opened 2026-03-15
Add auth for mcp

### LINKED ISSUE #2823: Auth for MCP Server
This requires another documentation on its own. At least to begin with, the expectation is both auth for MCP as well code first routes should be unified. Please allow me to document this as part of the task itself. 

In general this should be unified with what we already have for code-first routes

afsalthaj:
Fixes #2823 

# Setup
```js
golem api security-scheme create my-google-auth \
                     --provider-type google \
                     --client-id KANGAROOS_CLIENTID  \
                     --client-secret KANGAROO_WILL_NEVER_SHARE_PASSWORD \
                     --scope openid \
                     --scope email \
                     --scope profile \
                     --redirect-url http://localhost:9007/mcp/oauth/callback
```

2. golem deploy your app

```yaml
mcp:
  deployments:
    local:
      - domain: localhost:9007
        agents:
          WeatherAgent:
            securityScheme: my-google-auth

```

Now there are two ways to see `auth` working (default,  and direct auth token). 
Working with both `MCP Inspector` as well as `Claude Desktop configured with mcp-remote`

# Auth with MCP Inspector

This is more involved in terms of logging into provider (similar to http) etc.

MCP inspector was the best client to test if this auth works with MCP server. 

The only way to show this is a video (there is nothing else to do other than playing with MCP client)

https://github.com/user-attachments/assets/db1a2ab8-4fea-4e7b-9b0e-15f9f29b63f5

If no authentication done, then

image

# Auth with Claude Desktop 

Configure the following in `claude_desktop_config.json`. Now when you open Claude Desktop, it will start connecting to this MCP server. 

```json
{
  "mcpServers": {
    "golem": {
      "command": "npx",
      "args": [
        "mcp-remote",
        "http://localhost:9007/mcp",
        "--header",
        "Host: localhost:9007",
        "--allow-http"
      ]
    }
  }
}
```

If you have already been using `mcp-remote`, then make sure to delete cache

```
rm -rf ~/.mcp-auth/mcp-remote-0.1.37/
```

(My browser was already logged into my gmail :)) 

https://github.com/user-attachments/assets/32942c39-a1d0-4a95-84f6-67b0e04affef

# Auth using direct token

In a way this has to obviously work, because in the above workflow, bearer token is send as auth header in every subsequent request, fully session managed by the client. 

Obviously, generate your open-id-token yourself (you are not using Golem to do this) - and this is possible in many ways,and if you have configured the redirect-url in your app with the correct details that can do the whole workflow of OpenID (such as authorize, redirect, exchange auth code for tokens etc) and get an open-id-token.  We will document this too, given different MCP clients behave differently.

1. Connect to MCP Inspector, and configure like this

image

2. Click connect and it will be successful

-----------------------------------

PS: How to generate open-id token (for being able to configure bearer token directly)

Given the app in Google is configured with a redirect URL `https:/./developers.google.com/oauthplayground`

1. Open https://developers.google.com/oauthplayground/
2. In settings, set `Use your own OAuth credentials`, and enter client-id and client-secret
3. Scopes being open-id, email, and profile -> `Authorize APIs`
4. Login to your google
5. Click `Exchange authorisation code for tokens`
6. In the response, copy `id_token` (that long bas64 string)

## Amp's validation

image

(I think, this has been really tricky to get it "working", would like to not make much change in the workflow)

## Reuse with http api's things
Note that, this is not an exact copy of http when it comes to OIDC workflow, but obviously there is a lot in common, so I reused as much as possible. In the case of MCP, the clients take part in most of the things - essentially, as par spec, clients (inspector as an example) drives the flow. What does that mean? It mean unlike http apis (where golem does the job of oauth client with the help of infos of client-id and secret), MCP client (claude or inspector) needs to construct the authorisation URLs themselvevs, call `/token` and manage token by itself. But obviously, these clients don't have the info (client_id) so it won't or can't talk to google/microsoft direclty.  This kind of implies, in the MCP case, golem act like more of a proxy.  Claude or Inspector simply drive the flow against golem mcp auth proxy (and it internally does whetever needed with google).


golem-worker-service/src/mcp/auth.rs
```diff
@@ -0,0 +1,743 @@
...
+/// the same token against the identity provider on every request.
+type TokenHash = u64;
+
+/// When a cached token validation expires and must be re-checked.
+type ValidationExpiry = Instant;
+
+/// Per-instance cache of successfully validated tokens.
+/// Each instance maintains its own cache independently; a cache miss
+/// simply triggers a fresh validation against the identity provider.
+type ValidatedTokenCache = Arc<DashMap<TokenHash, ValidationExpiry>>;
```
vigoo: Let's use our `Cache` type and not `DashMap`. It has built-in support for expiration as well.

golem-worker-service/src/custom_api/oidc/session_store.rs
```diff
@@ -433,12 +617,130 @@ impl SessionStore for SqliteSessionStore {
...
+                    .bind(code)
+                    .bind(serialized)
+                    .bind(expiry),
+            )
+            .await?;
+
+        Ok(())
+    }
+
+    async fn take_mcp_proxy_code(
```
vigoo, line 703: Shouldn't this and the other `take_` impl group the SELECT+DELETE in a transaction?

golem-worker-service/src/mcp/auth.rs
afsalthaj: Okay

Issue #2971, opened 2026-03-12
WASM function table limit


vigoo:
We are protecting against memory allocations above a limit in the `Plan` in our `ResourceLimiterAsync` implementation, but we are not doing anything with `table_growing` (table here is a dynamic function pointer table). We should have a limit on that, just to be safe:

- Add a new limit to `Plan` and propagate it to the worker executor, the same way as the memory limit is propagated
- In the limiter, if we exceed that number, fail the same way as `WorkerExceededMemoryLimit` (but with a new one) - make sure everywhere where `WorkerExceededMemoryLimit` is handled, the new error is also handled accordingly
- Figure out a good default value - log the size in some example components, and choose a significantly bigger number based on that

PR #2999, opened 2026-03-17
Wasm function table limits

kmatasfp:
Resolves: #2971 

Function table limits Implementation based on existing memory limits implementation.

For determining the initially configured limit of 16384 I used `wasm-tool` to analyze test-comonents and Peters app and then 10x the biggest value I saw and chose closest number to it that is power of 2,


golem-worker-executor/tests/lib.rs
```diff
@@ -62,6 +63,7 @@ tag_suite!(scalability, group3);
 tag_suite!(revert, group3);
 
 tag_suite!(rdbms_service, rdbms_service);
+tag_suite!(resource_limits, group1);
```
kmatasfp, line 67: not sure what is the criteria how the tests are grouped, so if this is not the correct group let me know

PR #3000, opened 2026-03-17
Fix parameter extraction in rust sdk

mschuwalow:
Follwoup to #2986 

We cannot guarantee method order is the same in the agent implementation (which is used to generate the match arms in invoke) and the agent definition (which is used to extract the agent types). In case these two don't line up the sdk will fail at runtime with cryptic errors.


vigoo:
We could have been just sort by name compilation or initialization time, to avoid the search on the hot path. But let's merge for now

PR #3001, opened 2026-03-17
Event broadcast based cache invalidation

### LINKED ISSUE #2431: Add event based cache invalidation

vigoo:
Resolves #2431 

This PR introduces a new, extensible way how registry service change events can be propagated to other services for cache invalidation purposes. This directly solves one invocation hot-path performance issue but I applied it in other places as well.

The basic architecture is the following:

- Registry service exports a gRPC stream of events
- Worker service / worker executor nodes connect (and maintain the connection via reconnects etc) a single connection to an arbitrary registry service node to receive such events, and react to them by invalidating some caches
- Because in production deployments there can be arbitrary number of load-balanced registry nodes, worker service nodes and executors, the registry service nodes also need to make sure that each node broadcasts every event not just the ones that are directly happening on a call to that particular node. This is done through a notification interface that, when running a single node (in local golem) is a simple in-process notify, and when running with postgres backend is implemented through the database with `pg_notify/listen` 
- The events are persisted transactionally to an events table, and old events got periodically purged

### DIFF
--- a/.agents/skills/db-migration-scripts/SKILL.md
+++ b/.agents/skills/db-migration-scripts/SKILL.md
@@ -46,6 +46,10 @@ CREATE UNIQUE INDEX plugins_name_version_uk ON plugins (account_id, name, versio
   CONSTRAINT accounts_pk PRIMARY KEY (account_id)
   ```
 
+## Column Types
+
+Use the same column types in PostgreSQL and SQLite whenever possible, to make it easier to write queries that work on both databases. Only use database-specific types (e.g., `BIGSERIAL` vs `INTEGER PRIMARY KEY AUTOINCREMENT`, `UUID` vs `TEXT`, `TIMESTAMPTZ` vs `TEXT`, `TEXT[]` vs `TEXT`) when there is no common alternative.
+
 ## Table Style
 
 - Use uppercase SQL keywords (`CREATE TABLE`, `NOT NULL`, `PRIMARY KEY`)


golem-registry-service/src/repo/deployment.rs
```diff
@@ -770,38 +875,55 @@ impl DeploymentRepo for DbDeploymentRepo<PostgresPool> {
...
+                    )
+                    .await?;
 
-                let ext_revision = CurrentDeploymentExtRevisionRecord {
-                    revision,
-                    deployment_version: deployment_revision.version,
-                    deployment_hash: deployment_revision.hash,
-                };
+                    // Query domains and record registry change event in the same transaction
+                    let domains = Self::query_domains_for_environment(tx, environment_id).await?;
```
mschuwalow: Domain registrations do not change as part of a deployment, but rather in a separate crud api. This should be checked there.

golem-api-grpc/proto/golem/registry/v1/registry_service.proto
```diff
@@ -355,3 +360,46 @@ message GetCurrentEnvironmentStateResponse {
...
+// Sent when the client's cursor is older than the server's retention window.
+// The client must flush all caches and use this event's event_id as its new cursor.
+message CursorExpiredEvent {}
+
+// Sent when the current deployment for an environment changes.
+message DeploymentChangedEvent {
+  golem.common.EnvironmentId environment_id = 1;
+  uint64 deployment_revision = 2;
+  // Union of old and new domains — includes removed domains for cache invalidation.
+  repeated string domains = 3;
```
mschuwalow: This is not part of the deployment / should be separate event

golem-api-grpc/proto/golem/registry/v1/registry_service.proto
```diff
@@ -355,3 +360,46 @@ message GetCurrentEnvironmentStateResponse {
...
+
+message SubscribeRegistryInvalidationsRequest {
+  // The event_id of the last event the client successfully processed.
+  // If omitted, the server starts streaming from live events only (no replay).
+  optional uint64 last_seen_event_id = 1;
+}
+
+// Streamed to worker-service subscribers when any registry state changes.
+// Each event carries exactly one payload variant.
+message RegistryInvalidationEvent {
```
mschuwalow, line 374: We will also need security scheme changes for the worker service / http api

golem-api-grpc/proto/golem/registry/v1/registry_service.proto
vigoo: thanks, I took "atomic deployment" too literally, i guess :)

mschuwalow: Everything that is part of the manifest is atomic. There is just certain admin stuff that is handled separately / has its own lifecycle and behaves in a special way.

Right now those are:
* domains
* security schemes
* secrets

golem-registry-service/db/migration/postgres/008_registry_change_events.sql
```diff
@@ -0,0 +1,12 @@
...
+    event_type  SMALLINT NOT NULL DEFAULT 0,
+    environment_id UUID,
+    deployment_revision_id BIGINT,
+    account_id  UUID,
+    grantee_account_id UUID,
+    domains     TEXT[],
+    changed_at  TIMESTAMPTZ NOT NULL DEFAULT now()
+);
+
+CREATE INDEX idx_registry_change_events_id ON registry_change_events (event_id);
```
noise64: Primary keys need no extra indexes (neither in postgres, nor in sqlite), on the other hand, for the deletion we should have and index on changed_at.

PR #3014, opened 2026-03-18
OpenTelemetry exporter plugin, and built-in plugin provision

### LINKED ISSUE #1613: Create an OpenTelemetry oplog processor plugin for golem
Set up the plugin with full build and CI setup, and have a first implementation that is:
- configurable (otel address etc)
- extracts as much contextual information from the oplog as possible

### LINKED ISSUE #1614: Test the OpenTelemetry plugin
After the initial implementation let's set up a complex test scenario with as much different components as possible:
- http calls
- customized incoming http spans
- custom spans
- rpc calls
- using the golem-llm library

etc.

Verify the trace outcome and make adjustments on all levels until it is good enough.

vigoo:
Resolves #1613
Resolves #1614 

Introduces a way to have built-in oplog processor plugins:
- they are stored in the `plugins` dir of this repo, compiled with `golem-cli` 
- we store the compiled wasm in the repo too, so `cargo install` works for `golem/golem-cli`
- they are embedded in the registry service binary
- the built-in plugins are automatically deployed to the system account 
- every new environment automatically gets grants to use them

Adds the first such built-in plugin, the OpenTelemetry exporter:
- supports traces, logs and metrics
- tested with dockerized otel-collector and jaeger

Follow-up: https://github.com/golemcloud/golem/issues/3015

### DIFF
--- a/golem-worker-executor/src/worker/mod.rs
+++ b/golem-worker-executor/src/worker/mod.rs
@@ -1306,12 +1306,13 @@ impl<Ctx: WorkerCtx> Worker<Ctx> {
                 }
             }
         } else {
-            let is_pending = self
-                .pending_invocations()
-                .await
+            let status = self.last_known_status.read().await;
+            let is_pending = status
+                .pending_invocations
                 .iter()
                 .any(|entry| entry.invocation.has_idempotency_key(key));
-            if is_pending {
+            let is_current = status.current_idempotency_key.as_ref() == Some(key);
+            if is_pending || is_current {
                 LookupResult::Pending
             } else {
                 LookupResult::New
@@ -1436,6 +1437,25 @@ impl<Ctx: WorkerCtx> Worker<Ctx> {
                 run_loop_handle.await.expect("Failed to join run loop");
 
                 let mut instance_guard = self.instance.lock().await;
+                let is_deleting = match &*instance_guard {
+                    WorkerInstance::Stopping(stopping) => {
+                        matches!(stopping.final_state, FinalWorkerState::Deleting)
+                    }
+                    _ => false,
+                };
+
+                // After the invocation loop has fully exited, fail any remaining
+                // unresolved invocations (e.g. the currently running one that was
+                // in progress when deletion was requested).
+                if is_deleting {
+                    drop(instance_guard);
+                    self.fail_pending_invocations(WorkerExecutorError::invalid_request(
+                        "Worker is being deleted",
+                    ))
+                    .await;
+                    instance_guard = self.instance.lock().await;
+                }
+
                 match std::mem::replace(
                     &mut *instance_guard,
                     WorkerInstance::Unloaded {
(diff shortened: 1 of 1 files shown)


vigoo:
One thing I forgot to check / implement: how do we provision changed / updated built-in plugins when deploying to prod (but keeping the live db)

vigoo:
> One thing I forgot to check / implement: how do we provision changed / updated built-in plugins when deploying to prod (but keeping the live db)

The single line change I've pushed that makes plugins versioned by the golem version fixes this, I think

golem-registry-service/src/services/builtin_plugin_provisioner.rs
```diff
@@ -0,0 +1,304 @@
...
+            }
+        }
+        Err(other) => {
+            return Err(anyhow::anyhow!(
+                "Failed to look up system environment: {other}"
+            ));
+        }
+    };
+
+    // 3. Upload "otlp-exporter" component, or update it if it already exists
```
mschuwalow: We should use the component / wasm hash here to prevent updating the component on each startup of each instance

golem-worker-executor/src/services/oplog/plugin.rs
```diff
@@ -1049,6 +1083,7 @@ impl ForwardingOplogState {
             .await
         {
             Ok(()) => {
+                // Enqueue succeeded — immediately confirm
```
mschuwalow, line 1097: Does the receiver side protects against receiving the same batch multiple times?

golem-registry-service/src/bootstrap/mod.rs
```diff
@@ -203,6 +203,13 @@ impl Services {
 
         let auth_service = Arc::new(AuthService::new(repos.account_repo.clone()));
 
+        let root_account_id = config
```
mschuwalow: Let's use a dedicated account for this without special roles. I don't want to leak the admin account id to users

golem-registry-service/src/services/builtin_plugin_provisioner.rs
```diff
@@ -0,0 +1,304 @@
...
+                .await
+            {
+                Ok(_) => {}
+                Err(EnvironmentPluginGrantError::GrantForPluginAlreadyExists) => {}
+                Err(e) => {
+                    tracing::warn!("Failed to grant plugin to env {}: {e}", env_entry.name);
+                }
+            }
+        }
+    }
```
mschuwalow: This block only grants access to environments in the root account's applications, which is not useful.

There should not be any applications or environments in that account (outside the one used for the plugin definition)

golem-registry-service/src/services/environment.rs
```diff
@@ -128,9 +137,25 @@ impl EnvironmentService {
...
 
+        let builtin_plugins = self
+            .plugin_repo
+            .list_by_account(self.root_account_id.0)
+            .await?;
+
+        let plugin_grants: Vec<EnvironmentPluginGrantRecord> = builtin_plugins
+            .into_iter()
+            .map(|p| {
+                EnvironmentPluginGrantRecord::creation(
```
mschuwalow, line 148: EnvironmentPluginGrants can be managed by users and deleted as long as they have access to the environment. I don't think we should allow deletion of builtin plugins from environments.
Additionally, these grants will show up in the api and cli for users.

I think builtin plugins want to have special handling and not reuse exactly the same mechanism as user plugins. A couple of options that I think make sense:
* Do not give builtin plugins an id at all, but rather have an enum for builtin plugins. Whereever we can refer to a plugin we allow doing it either by id or by selecting one of the builtin ones
* Store them in the db with an id like now, but either have a dedicated table for them or extend the current table such that they can live outside of the normal account / application / environment hierarchy. All queries that retrieve plugins can join with the buitin plugins. In that case we do not have an automatically generated account for them at all and can use the system account id wherever we do want to display an owner (clearly marking them as system).

Alternatively, if we do keep the current system managed account + grant structure, we should protect the automatically created grants from deletion and make them more descriptive for users.

golem-registry-service/src/services/builtin_plugin_provisioner.rs
vigoo: Yeah that broke in the last commit, previously it was supposed to be named by the version (but that has other issues), thanks

golem-worker-executor/src/services/oplog/plugin.rs
vigoo: Yes, all the "exactly once semantics" changes were done in a separate PR before this. Here I just added this extra logging of process failures back to the caller agent's event stream. 
(For double send, it's solved by having deterministic oplog processor worker name + determinstic idempotency keys)

golem-registry-service/src/bootstrap/mod.rs
vigoo: By leaking, you mean that they can see the ID because of the plugin grants?

golem-registry-service/src/services/environment.rs
vigoo: I would rather keep it simple - that's why I wanted to build it on top of the existing things. They are not special in any way other than we want them to be available for everyone without explicitly installing them. 

I think if we move them to another root account as you asked, and maybe disallow deleting the grants, we have the desired UX and we have minimal complexity on the implementation side (for me the provision step and auto-granting the plugins feels very nice and scoped)

golem-registry-service/src/services/builtin_plugin_provisioner.rs
vigoo: there supposed to be grants created for the built-in plugins whenever someone creates a new environment. i'll double check and remove this unnecessary part

golem-registry-service/src/bootstrap/mod.rs
mschuwalow: yes, exactly

golem-registry-service/src/services/environment.rs
mschuwalow: That sounds good to me :+1:

golem-registry-service/src/config.rs
```diff
@@ -136,6 +144,17 @@ impl Default for RegistryServiceConfig {
                 plan_id: PlanId(uuid!("157dc684-00eb-496d-941c-da8fd1d15c63")),
             },
         );
+        initial_accounts.insert(
+            "builtin-plugin-owner".to_string(),
+            PrecreatedAccount {
+                id: AccountId(uuid!("adb2694f-cd9f-425d-905d-ca2888c9c5de")),
+                name: "Builtin Plugin Owner".to_string(),
+                email: AccountEmail("builtin-plugin-owner@golem.cloud".to_string()),
+                token: TokenSecret::trusted("32d6072d-64e9-4a4a-b8f9-fadf68bb446b".to_string()),
```
mschuwalow, line 153: I would prefer adjusting this so we don't create a token for this account, as it should never be used interactively. But that can easily be done in a followup.

golem-registry-service/src/config.rs
```diff
@@ -136,6 +144,17 @@ impl Default for RegistryServiceConfig {
...
+                role: AccountRole::BuiltinPluginOwner,
```
mschuwalow, line 154: Not sure about having a dedicated "marker" role. So far the roles were actually used for permission checks, but I'm not strongly against it.

golem-registry-service/src/config.rs
```diff
@@ -373,6 +393,11 @@ impl ComponentCompilationEnabledConfig {
     }
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, Default)]
+pub struct BuiltinPluginsConfig {
+    pub enabled: bool,
```
mschuwalow, line 398: Enum would be more consistent

golem-registry-service/src/config.rs
vigoo: ok

golem-registry-service/src/config.rs
vigoo: will do in a next pr

PR #3016, opened 2026-03-18
Attempt to fix a flaky test

vigoo:


golem-service-base/src/db/postgres.rs
```diff
@@ -425,8 +425,7 @@ pub async fn migrate(
     }
 
     let migrator = sqlx::migrate::Migrator::new(migrations).await?;
-    // See: https://github.com/launchbadge/sqlx/issues/954, Send required for golem-cli
-    futures::executor::block_on(migrator.run(&mut conn))?;
+    migrator.run_direct(&mut conn).await?;
```
noise64, line 428: if the blocking is not required anymore, then let's remove it from the sqlite one too

noise64: Or is it still required, but we only use sqlite currently for CLI?

vigoo: the agent lied to me that it's only there for postgres :)

noise64: .agent/skills/noise64-memory

noise64 (approved):

Issue #2970, opened 2026-03-12
File-system quota


vigoo:
The goal of this ticket is to track and limit file-system usage of agents, similar to how their memory usage is handled.

### Details

The `ActiveWorkers` currently has a `worker_memory` semaphore and a machinery that allows workers to acquire memory, and if they cannot, it tries to move out idle (but loaded) agents out of memory. 

Every agent, when created, gets an associated temporary directory in `DurableWorkerCtx::create`. This is dropped when the durable worker context is dropped (when the agent goes out of memory), and this ensure the directory gets cleaned up.

We want two to get layers of file-system usage protection as part this ticket:

- A maximum file system usage quota. This has to be added to `Plan` (in all layers, DB, registry service, gRPC etc) (note: we already have a `storage_limit` in `Plan`, but that's different, that is about server-side storage of component files etc)
- A semaphor-based mechanism very similar to the memory one mentioned above, to limit the simulateneous per-executor  disk usage of agents

The only way an agent can write to its filesystem is through wasi:filesystem/wasi:io. We need to modify our host functions to track file system usage and acquire tokens - if no tokens, the agent gets suspended. If the per-account (Plan) limit is reached, an error is thrown, similar to `WorkerExceededMemoryLimit`

Sub-steps:

1. Add a semaphore to track available disk space in `ActiveWorkers` similar to the memory one. Have an `acquire`/`try_acquire` for disk space (or merge it in the memory one, if that is more fitting), and also a `try_free_up_space` - just follow the same logic that exists, try to unify things if possible to reduce redundant code
2. `WaitingWorker::new` acquires memory tokens based on the known memory usage of the agent. Here we also need to acquire disk tokens based on the known initial disk space, which comes from the _initial filesystem_ attached to the worker (check `update_filesystem` etc to see where those are coming from)
3. Write some helpers in `PrivateDurableWorkerState` or similar place that can be called whenever more disk space is to be used. Call it from all host functions (filesystem, output stream attached to files, etc) accordingly.
4. In this method, we need to handle limits. Make sure the disk space limit is transferred to the executor from the `Plan`. Follow the pattern how the memory limit is already known. If that limit is reached, throw a `DiskSpaceLimitExceeded` error (follow the pattern of the existing WorkerExceededMemoryLimit)
[...]

PR #3033, opened 2026-03-19
File system quota

kmatasfp:
Resolves: #2970 

This PR implements per-agent and executor-wide filesystem storage quotas for Golem workers.

**Two-layer enforcement:**
1. Per-agent quota (`max_disk_space_per_worker` on the plan) — enforced via `check_filesystem_storage_quota` before every write. Exceeding it returns `AgentExceededFilesystemStorageLimit`, a permanent non-retriable error that marks the worker `Failed`.
2. Executor-wide pool (filesystem_storage config, default 10 GB) — a tokio semaphore where 1 permit = 1 KB of filesystem space. Exceeding it returns `NodeOutOfFilesystemStorage`, a retriable error that triggers `ReacquirePermits`: the worker restarts, evicts the oldest idle workers to free space, and retries the write.

**Quota tracking:**
- Every successful write records a `FilesystemStorageUsageUpdate(+delta)` oplog entry.
- Deletions and truncations record `FilesystemStorageUsageUpdate(-delta)` and immediately release semaphore permits via `add_permits`.
- On worker restart, `current_filesystem_storage_usage` is reconstructed by summing the oplog deltas, and that many bytes are pre-acquired from the executor pool before replay begins.

**Initial component files:**
- Read-only initial files are hardlinked from a shared `FileLoader` cache. The semaphore is charged once (on first download) and released when the last worker holding a token drops — subsequent workers sharing the same content hash via hardlink pay nothing additional.
- Read-write initial files are copied per-worker and charged individually.

**Eviction:**
- When the executor pool is exhausted, `try_free_up_filesystem_storage` evicts the oldest-idle workers, tracking freed bytes to ensure enough space is freed for the pending write plus any pre-acquire needed on restart.
- First-time writes (no prior oplog) that hit `NodeOutOfFilesystemStorage` record desired_extra_filesystem_storage on the Worker, which is included in the restart's `acquire_filesystem_storage` call so eviction frees the right amount

**Known shortcoming: transient over-commitment window**
When a worker is evicted, its storage semaphore permits are released the moment `RunningWorker` drops (inside `stop_internal_locked`). This happens slightly before the worker's temp directory is actually deleted from disk. The temp directory lives inside the wasmtime `Store<DurableWorkerCtx>`, which is dropped when the invocation loop task fully unwinds — a detached tokio task. The gap is the async task teardown time: typically sub-millisecond, but can be tens of milliseconds if the worker had many files (`TempDir::drop` calls `blocking std::fs::remove_dir_all`). During this window the semaphore considers the space free and may grant it to a new worker while the old files are still on disk — a transient over-commitment.

This is identical to the existing race in `try_free_up_memory`: memory permits (`RunningWorker.permit`) are also released when `RunningWorker` drops, before the wasmtime `Store` and its linear memory are freed. This has existed since the beginning and has never caused a production problem.

Why it's fine in practice: The acquire loop retries with a configurable delay (default 500ms). Any new worker granted permits during the gap will only start writing to disk after its own invocation begins — by which time the evicted worker's files are long gone.

Mitigation: Assuming the executor's temp directory has a dedicated pod-local volume, set `total_worker_filesystem_storage_bytes` to ~80–90% of that volume's capacity. The headroom absorbs the transient over-commitment window and filesystem metadata overhead. The `FilesystemStorageConfig` doc comment documents this explicitly.


golem-worker-executor/src/services/active_workers.rs
```diff
@@ -32,16 +32,150 @@ use golem_common::model::worker::WorkerAgentConfigEntry;
 use golem_common::model::{AgentId, OwnedAgentId};
 use golem_service_base::error::worker_executor::WorkerExecutorError;
 
+/// Executor-wide storage semaphore. One permit = `FILESYSTEM_STORAGE_PERMIT_SIZE_KB` KB.
```
vigoo: Minor organization thing: to keep things easier to find/read I'd convert this to three files:
```
active_workers.rs/mod.rs
active_workers/fs_semaphore.rs
active_workers/test.rs
```

golem-worker-executor/tests/storage_quota.rs
```diff
@@ -0,0 +1,1830 @@
...
+inherit_test_dep!(Tracing);
+inherit_test_dep!(
+    #[tagged_as("host_api_tests")]
+    PrecompiledComponent
+);
+
+#[test]
+#[tracing::instrument]
+#[timeout("2m")]
+async fn agent_quota_write_within_limit_succeeds(
```
vigoo, line 36: Some minor test things:

- call `executor.check_oplog_is_queryable(&worker_id).await?;` at the end of each test. this is just to make sure we run oplog query for as many scenarios as possible, it caught many issues before (even if not necessary for the actual test)
- I used to always `drop(executor)` before the assertions, because it made it easier to read the error (as cleanup logs were before the assertion). this may not be actually necessary anymore but maybe it is :)

golem-registry-service/src/repo/model/component.rs
```diff
@@ -435,6 +435,7 @@ pub struct ComponentFileRecord {
     pub audit: RevisionAuditFields,
     pub file_content_hash: SqlBlake3Hash,
     pub file_permissions: SqlComponentFilePermissions,
+    pub file_size: i64,
```
mschuwalow: There is NumericU64 in golem-service-base you can use for this

golem-worker-executor/src/services/active_workers.rs
```diff
@@ -32,16 +32,150 @@ use golem_common::model::worker::WorkerAgentConfigEntry;
...
+    /// Release a number of bytes back to the pool without dropping the entire
+    /// permit. Used when a file is partially freed (e.g. truncation).
+    ///
+    /// Adds `storage_bytes` worth of permits back to the semaphore directly.
+    /// The caller is responsible for ensuring they don't release more than they
+    /// acquired (no underflow protection at this level).
+    pub(crate) fn release(&self, storage_bytes: u64) {
+        let permits = bytes_to_filesystem_storage_permits(storage_bytes);
+        if permits > 0 {
+            self.semaphore.add_permits(permits as usize);
```
mschuwalow: This looks incorrect as I don't see any calls to Semaphore::forget (which would prevent double releases). I suspect we are slowly growing the filesystem limit of the executor

kmatasfp: good point, this is actually old method left behind and forgot to delete it during refactor, its superseded by `release_filesystem_storage_space` that uses ` split + drop`. Deleted now

tmp/post_split_tests.txt
```diff
@@ -0,0 +1,65 @@
+warning: golem-common@0.0.0: git describe result: 1.5.0-dev.1-236-gbd9734475
```
mschuwalow: should be removed

golem-worker-executor/src/services/file_loader.rs
```diff
@@ -64,18 +74,30 @@ impl FileLoader {
             cache: Mutex::new(HashMap::new()),
             cache_dir,
             item_counter: AtomicU64::new(0),
+            filesystem_storage_semaphore: StdMutex::new(None),
         })
     }
 
+    /// Wire up the executor-wide storage semaphore. Must be called after both
+    /// `FileLoader` and `ActiveWorkers` have been constructed.
+    pub fn set_filesystem_storage_semaphore(&self, semaphore: Arc<FilesystemStorageSemaphore>) {
```
mschuwalow: Why can this not be passed via new during construction?

tmp/post_split_tests.txt
kmatasfp: done

golem-worker-executor/src/services/file_loader.rs
kmatasfp: I need ActiveWorkers to be available, and they are created after the FileLoader gets created in golem-worker-executor/src/lib.rs, wanted to keep the changes to minimum. No technical reason, just historical

kmatasfp: actually let me change, can get rid of the outer mutex

PR #3039, opened 2026-03-19
Web socket in Rust

### LINKED ISSUE #3053: Implement Websocket for Rust agents

afsalthaj:
Fixes #3053 

### Rust Agent Websocket

```rust
// In test-components/host-api-tests/components-rust/golem-it-host-api-tests/src/websocket.rs
use golem_rust::{agent_definition, agent_implementation};
use golem_rust::{WebsocketConnection, WebsocketMessage};

#[agent_definition]
pub trait WebsocketTest {
    fn new(name: String) -> Self;
    fn echo(&self, url: String, msg: String) -> String;
}

pub struct WebsocketTestImpl { _name: String }

#[agent_implementation]
impl WebsocketTest for WebsocketTestImpl {
    fn new(name: String) -> Self { Self { _name: name } }

    fn echo(&self, url: String, msg: String) -> String {
        let ws = WebsocketConnection::connect(&url, None).expect("connect");
        ws.send(&WebsocketMessage::Text(msg)).expect("send");
        match ws.receive().expect("receive") {
            Message::Text(t) => t,
            Message::Binary(b) => format!("{} bytes", b.len()),
        }
    }
}
```

image

### DIFF
--- a/golem-worker-executor/tests/websocket.rs
+++ b/golem-worker-executor/tests/websocket.rs
@@ -321,9 +321,12 @@ async fn websocket_async_bidirectional_test(
         .into_return_value()
         .ok_or_else(|| anyhow!("expected return value"))?;
 
-    assert_eq!(result, Value::Result(Ok(Some(Box::new(Value::String(
-        "msg-a|msg-b|msg-c".to_string()
-    ))))));
+    assert_eq!(
+        result,
+        Value::Result(Ok(Some(Box::new(Value::String(
+            "msg-a|msg-b|msg-c".to_string()
+        )))))
+    );
 
     executor.check_oplog_is_queryable(&worker_id).await?;
 


vigoo:
Looks good so far. Let's have a per-executor semaphor to constraint the number of parallel websocket connections (we already have it for regular http connections). This way we protect the executor from overloading with ws sockets. (Let the `connect` rather wait than the whole server overloading or running out of sockets)

golem-worker-executor/src/durable_host/websocket/client.rs
```diff
@@ -0,0 +1,242 @@
...
+            Ok(()) => Ok(Ok(())),
+            Err(e) => Ok(Err(Error::SendFailure(e.to_string()))),
+        }
+    }
+
+    async fn subscribe(
+        &mut self,
+        _self_: Resource<WebSocketConnectionEntry>,
+    ) -> anyhow::Result<Resource<wasmtime_wasi::p2::bindings::io::poll::Pollable>> {
+        anyhow::bail!("golem:websocket/client.websocket-connection.subscribe is not yet supported")
```
vigoo: what about this?

integration-tests/tests/worker.rs
```diff
@@ -1842,6 +1842,142 @@ async fn deployment_invalidates_agent_resolution_cache(
     Ok(())
 }
 
+#[test]
+#[tracing::instrument]
+#[timeout("2m")]
+async fn websocket_echo_rust(
```
vigoo: Why are these not worker-executor tests?

afsalthaj: I think I tried to place these tests in ehre, and was failing with registry service errors.. I can try again

golem-worker-executor/src/durable_host/websocket/client.rs
afsalthaj: Done

integration-tests/tests/worker.rs
vigoo: It would be a better place, as we can much quicker and easier iterate on it and this is an executor feature (not depending on any other service)

mschuwalow:
This has no durability at all, right? Would it make sense to only allow using websocket connections within a persist-nothing region? Otherwise we are going to reexecute (the potentially side-effecting) connections during replay

vigoo:
> This has no durability at all, right? Would it make sense to only allow using websocket connections within a persist-nothing region? Otherwise we are going to reexecute (the potentially side-effecting) connections during replay

It will have durability, I asked for adding it in two new phases on top of this.

golem-worker-executor/src/durable_host/websocket/client.rs
```diff
@@ -0,0 +1,248 @@
...
+    writer: Mutex<SplitSink<WsStream, tungstenite::Message>>,
+    reader: Mutex<SplitStream<WsStream>>,
+    /// Held for the lifetime of the connection to limit concurrent WebSocket
+    /// connections per executor. Released when the connection is dropped.
+    _permit: tokio::sync::OwnedSemaphorePermit,
+}
+
+#[async_trait::async_trait]
+impl wasmtime_wasi::p2::Pollable for WebSocketConnectionEntry {
+    async fn ready(&mut self) {}
```
vigoo: If this is no-op, we cannot to async websocket communication in our agents. 
A minimal version would be something like:
- a separate `blocking_receive` (the current `receive`) and non-blocking `receive` (returns `None` if there is no current message to receive)
- this `ready()` function async-blocks until we know that the next call to `receive` will return `Some`

With that it's possible to mix async ws reads with other operations in the agent. (And send is still always blocking).

Depending on how `send` is implemented it may make sense to make an async send too (if there is backpressure by some limited output queue). In that case we may need two different `subscribe`s.

The most advanced and idiomatic version would probably be if a connected websocket would materialize in a wasi `InputStream` and `OutputStream` (with properly implemented `read()` for each - like wasi-http input/output streams do). I don't remember if this is possible without touching the wasmtime fork though.

golem-worker-executor/src/durable_host/websocket/client.rs
```diff
@@ -0,0 +1,248 @@
...
+        self.observe_function_call("golem:websocket/client", "receive-with-timeout");
+
+        let mut view = self.as_wasi_view();
+        let entry = view.table().get(&self_)?;
+        let mut reader = entry.reader.lock().await;
+
+        let timeout = tokio::time::Duration::from_millis(timeout_ms);
+
+        loop {
+            match tokio::time::timeout(timeout, reader.next()).await {
```
vigoo: The timeout should be on the whole `receive_with_timeout` not on the individual messages, because of the loop. Imagine we continuously get some ping or other system message that we loop on, and never a "real" user message that we return.

golem-worker-executor/src/durable_host/websocket/client.rs
```diff
@@ -0,0 +1,248 @@
...
+        Message::Text(text) => tungstenite::Message::Text(text.into()),
+        Message::Binary(data) => tungstenite::Message::Binary(data.into()),
+    }
+}
+
+fn from_tungstenite_message(msg: tungstenite::Message) -> Option<Message> {
+    match msg {
+        tungstenite::Message::Text(text) => Some(Message::Text(text.as_str().to_owned())),
+        tungstenite::Message::Binary(data) => Some(Message::Binary(data.as_slice().to_vec())),
+        tungstenite::Message::Close(_) => None,
```
vigoo: Shouldn't we forward close messages to the client?

golem-worker-executor/src/durable_host/websocket/client.rs
```diff
@@ -0,0 +1,248 @@
...
+    headers: Option<&[(String, String)]>,
+) -> Result<tungstenite::http::Request<()>, String> {
+    use tungstenite::client::IntoClientRequest;
+
+    let mut request = url.into_client_request().map_err(|e| e.to_string())?;
+
+    if let Some(headers) = headers {
+        let req_headers = request.headers_mut();
+        for (name, value) in headers {
+            if let (Ok(header_name), Ok(header_value)) = (
```
vigoo: We should fail on wrong headers

test-components/host-api-tests/src/websocket.rs
```diff
@@ -0,0 +1,73 @@
...
+        let ws = WebsocketConnection::connect(&url, None).expect("connect failed");
+
+        match ws.receive_with_timeout(timeout_ms).expect("receive failed") {
+            Some(WebSocketMessage::Text(t)) => Some(t),
+            Some(WebSocketMessage::Binary(b)) => Some(format!("{} bytes", b.len())),
+            None => None,
+        }
+    }
+
+    fn poll_for_message(&self, url: String, timeout_ms: u64) -> Result<String, String> {
```
vigoo, line 75: Let's write a "real" async test as well, which is an async agent method and uses wstd to convert the pollable to a future, and does real bidiretional messaging (multiple sends, multiple receives, receive being async through wstd)

afsalthaj: done

golem-worker-executor/tests/websocket.rs
```diff
@@ -254,3 +254,81 @@ async fn websocket_polling_test(
...
+            &component,
+            &agent_id,
+            "async_bidi_test",
+            data_value!(format!("ws://localhost:{ws_port}")),
+        )
+        .await?
+        .into_return_value()
+        .ok_or_else(|| anyhow!("expected return value"))?;
+
+    assert_eq!(result, Value::Result(Ok(Some(Box::new(Value::String(
```
vigoo, line 324: Small note (only fix if you do other changes anyway): the only reason we do explicit `drop` of the executor and stop of the servers in the end of the test is to put the assertions _after_ them. this way logs coming from stopping the executor are always before the actual assertion failure

test-components/host-api-tests/src/websocket.rs
```diff
@@ -42,6 +43,35 @@ impl WebsocketTest for WebsocketTestImpl {
...
+
+        let payloads = ["msg-a", "msg-b", "msg-c"];
+        let mut received = Vec::new();
+
+        for payload in payloads {
+            ws.send(&WebSocketMessage::Text(payload.to_string()))
+                .map_err(|e| format!("Send error: {:?}", e))?;
+
+            // Convert websocket pollable to an async future via wstd.
+            let pollable = ws.subscribe();
```
vigoo, line 58: you just told me a few hours ago to do a single subscribe outside the loop :)

afsalthaj: :D hehe yes. I can move that out

PR #3042, opened 2026-03-20
quotas: create default resources during deployment

### LINKED ISSUE #3003: Registry Service quota deployment
Adjust the deployment process to include a number of quotas that should be created as part of the deployment. Quotas should be unique by environment_id + name.
Any quota updates / creations / deletions done as part of deployments need to be written as part of the db transactions at the end of a deployment. No partial state should ever be created.

Deployments track quotas by id and revision. Quotas are updated as part of rollbacks.

mschuwalow:
resolves #3003

### DIFF
--- a/openapi/golem-registry-service.yaml
+++ b/openapi/golem-registry-service.yaml
@@ -8871,6 +8871,7 @@ components:
       - expectedDeploymentHash
       - version
       - agentSecretDefaults
+      - quotaResourceDefaults
       properties:
         currentRevision:
           type: integer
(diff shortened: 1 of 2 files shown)


golem-registry-service/src/services/deployment/deployment_context.rs
```diff
@@ -438,6 +441,37 @@ impl DeploymentContext {
...
+        let resources_existing_in_env: HashSet<&ResourceName> = resource_definitions_in_environment
+            .iter()
+            .map(|s| &s.name)
+            .collect();
+
+        let mut creations = Vec::new();
+        let mut seen_resources = HashSet::new();
+
+        for resource_default in resource_definition_defaults_in_deployment {
+            if !seen_resources.insert(resource_default.name.clone()) {
```
vigoo, line 460: Do we need some canonicalization for this to avoid mixed cased duplicates?

mschuwalow: I think that should be fine for resource names as they are strings in the sdks as well. Will add if it shows up further down the line

PR #3043, opened 2026-03-20
change-notifier: do not double emit in postgres & do not create deployment

mschuwalow:


mschuwalow:
after discussion with @noise64 we decided to close this one and create dedicated tickets / plan for the cleanups

PR #3047, opened 2026-03-20
Support for parallel CLI integ tests

noise64:
Allows running multiple CLI integtest in parallel by using free ports for `golem server`.

For now the PR does not change the parallelism on CI, as the main goal was enabling testing locally in mutliple workspaces, but later we can also experiment with that.

### DIFF
--- a/cli/golem-cli/src/config.rs
+++ b/cli/golem-cli/src/config.rs
@@ -418,9 +418,9 @@ impl HttpClientConfig {
     pub fn new_for_service_calls(allow_insecure: bool) -> Self {
         Self {
             allow_insecure,
-            timeout: Some(Duration::from_secs(10)),
+            timeout: Some(Duration::from_secs(120)),
             connect_timeout: Some(Duration::from_secs(10)),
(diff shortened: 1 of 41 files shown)


cli/golem-cli/src/config.rs
```diff
@@ -28,15 +28,52 @@ use std::fmt::{Display, Formatter};
...
+struct BuiltinLocalUrlState {
+    url: Url,
+    uses_default: bool,
+}
+
+static BUILTIN_LOCAL_URL_ONCE: OnceLock<BuiltinLocalUrlState> = OnceLock::new();
+
+fn builtin_local_url_state() -> &'static BuiltinLocalUrlState {
+    BUILTIN_LOCAL_URL_ONCE.get_or_init(|| {
+        if cfg!(debug_assertions) {
```
vigoo, line 51: Why is this tied to `debug_assertions`?

noise64: i consider this a "test only" feature, so for now i wanted to only enable it in debug like builds.

For users, overriding the "local"  env should happen via the manifest.

Issue #3004, opened 2026-03-17
Shard Manager quota updates


mschuwalow:
The shard manager should subscribe to the cdc of the registry service on startup and listen for quota changes (via deployments). When a new deployment / quota update is detected, the shard manager needs to reconcile the current quota state with the new declaration. 
The goal of this ticket is just to add the client / mechanism to refresh active resource declarations and wire it up on startup.

PR #3060, opened 2026-03-23
quotas: registry-service grpc api for quotas

mschuwalow:
part of #3004

### DIFF
--- a/golem-common/src/macros.rs
+++ b/golem-common/src/macros.rs
@@ -40,6 +40,7 @@ macro_rules! newtype_uuid {
             }
 
             $(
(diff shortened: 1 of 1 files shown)


golem-common/src/macros.rs
```diff
@@ -38,6 +38,15 @@ macro_rules! newtype_uuid {
             pub fn new() -> $name {
                 Self(uuid::Uuid::now_v7())
             }
+
+            $(
+                #[cfg(feature = "full")]
+                pub fn render_proto(proto: Option<$proto_type>) -> Option<String> {
```
vigoo, line 44: ❤️

Issue #2973, opened 2026-03-12
Per-invocation HTTP and RPC call counter and limit


vigoo:
We can add a counter to `PrivateDurableWorkerState` that counts the number of http invocations and number of RPC invocations within a single invocation of an agent.
Once we have that we can put hard limits on these in `Plan`, propagate it to the worker executor, and fail the agent with limit exceeded when necessary.

In addition to that we can apply the same strategy as we do for fuel tracking:
- we optimistically track the total number of http & rpc calls per account
- periodically send batch updates to the registry service from all executor, and get back an updated remaining amount
- if exceeded, suspend the agents

This can be another (2) new fields in `Plan` and should be resetted monthly.

PR #3068, opened 2026-03-24
2973 http and rpc call limits

kmatasfp:
Resolves: #2973

This PR adds two layers of enforcement for outgoing HTTP and RPC calls made by Golem workers.

### Per-invocation hard limits

Two new `Plan` fields (`per_invocation_http_limit`, `per_invocation_rpc_limit`) set a hard cap on the number of outgoing HTTP and RPC calls a worker can make within a single exported function invocation. When the limit is hit, the worker traps immediately with an unrecoverable error — the same mechanism as exceeding the function table limit. Counters are transient, reset per invocation, and only increment during live execution (not replay).

### Monthly account-level soft limits

Two more `Plan` fields (`monthly_http_limit`, `monthly_rpc_limit`) set a monthly budget for HTTP and RPC calls across all workers for an account. Tracking follows the same pattern as fuel:

- Each executor maintains local `unsynced_http_calls` / `unsynced_rpc_calls` counters incremented at the call site
- A background task batch-syncs deltas to the registry service every minute via two new gRPC endpoints (`BatchUpdateHttpCallCount`, `BatchUpdateRpcCallCount`)
- The registry computes `available = limit - usage` and returns it; the executor caches this as `available_http_calls_from_server`
- `remaining = available_from_server - unsynced - syncing`
- When `remaining == 0` the host function (`handle()` for HTTP, `invoke_and_await()` / `invoke()` / `async_invoke_and_await()` for RPC) raises `WorkerMonthlyHttpCallBudgetExhausted` / `WorkerMonthlyRpcCallBudgetExhausted` — a `GolemSpecificWasmTrap` that maps to `TryStop` → the worker is suspended until the budget is replenished (month rollover or plan upgrade). 
- Monthly usage resets automatically via the existing `YYYY-MM` usage key mechanism


golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -3257,6 +3331,46 @@ impl PrivateDurableWorkerState {
...
+    pub fn check_and_increment_http_call_count(&mut self) -> Result<(), GolemSpecificWasmTrap> {
+        if !self.is_live() {
+            return Ok(());
+        }
+        if self.per_invocation_http_call_limit != u64::MAX
+            && self.http_call_count >= self.per_invocation_http_call_limit
+        {
+            return Err(GolemSpecificWasmTrap::WorkerExceededHttpCallLimit);
+        }
+        self.http_call_count += 1;
```
mschuwalow: let's use saturating add here

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -3257,6 +3331,46 @@ impl PrivateDurableWorkerState {
...
+    pub fn check_and_increment_rpc_call_count(&mut self) -> Result<(), GolemSpecificWasmTrap> {
+        if !self.is_live() {
+            return Ok(());
+        }
+        if self.per_invocation_rpc_call_limit != u64::MAX
+            && self.rpc_call_count >= self.per_invocation_rpc_call_limit
+        {
+            return Err(GolemSpecificWasmTrap::WorkerExceededRpcCallLimit);
+        }
+        self.rpc_call_count += 1;
```
mschuwalow: let's use saturating add here

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -3116,6 +3166,22 @@ struct PrivateDurableWorkerState {
...
+    /// Number of outgoing HTTP calls made in the current invocation (live only, not replayed).
+    /// Reset to 0 at the start of each exported function invocation.
+    http_call_count: u64,
+    /// Per-invocation HTTP call limit from the account's Plan. u64::MAX means unlimited.
+    per_invocation_http_call_limit: u64,
+
+    /// Number of RPC calls made in the current invocation (live only, not replayed).
+    /// Reset to 0 at the start of each exported function invocation.
+    rpc_call_count: u64,
+    /// Per-invocation RPC call limit from the account's Plan. u64::MAX means unlimited.
```
mschuwalow: Not sure we need the special handling for u64::max, but fine with having it

golem-worker-executor/src/services/resource_limits.rs
```diff
@@ -126,12 +218,56 @@ impl AtomicResourceEntry {
...
+
+    /// Records one outgoing HTTP call against the monthly account quota.
+    ///
+    /// Returns `false` when the remaining HTTP call budget is zero,
+    /// signalling that the worker should be suspended at the next opportunity.
+    pub fn record_http_call(&self) -> bool {
+        if self.remaining_http_calls() == 0 {
+            return false;
+        }
+        self.unsynced_http_calls.fetch_add(1, Ordering::AcqRel);
```
mschuwalow: I would prefer to have a saturating add using a cas loop here

golem-worker-executor/src/services/resource_limits.rs
```diff
@@ -563,6 +767,314 @@ mod tests {
         assert_eq!(entry.max_table_elements_limit(), 0);
     }
 
+    // -------------------------------------------------------------------------
+    // AtomicResourceEntry — per-invocation limits
+    // -------------------------------------------------------------------------
+
+    #[test]
+    fn new_with_invocation_limits_stores_http_limit() {
```
mschuwalow, line 797: This is getting quite big. Wdyt about moving the tests to a separate file?

golem-worker-executor/src/model/mod.rs
```diff
@@ -277,6 +277,18 @@ impl TrapType {
                             error: AgentError::ExceededTableLimit,
                             retry_from,
                         },
+                        Some(GolemSpecificWasmTrap::WorkerExceededHttpCallLimit) => {
+                            TrapType::Error {
```
mschuwalow, line 281: I think optimally we should schedule an automatic resumption for when the limit is expected to refresh, but let's think about this some more / do it separately

kmatasfp: These do not refresh, this is hard limit. So basically to unblock this user has to create a new version of an agent that does less http/rpc calls per one invocation. What will refresh are:
 * WorkerMonthlyHttpCallBudgetExhausted,
 * WorkerMonthlyRpcCallBudgetExhausted,

golem-worker-executor/src/services/resource_limits.rs
kmatasfp: will do it in a next pr that addresses that adds more tests here

PR #3069, opened 2026-03-24
quotas: shard manager api to acquire/renew/release leases

### LINKED ISSUE #3005: Shard Manager quota interest registration
Add a grpc api that executors can use to declare interest in a quota. TBD whether a quota is referenced by quota id, quota id + revision, or environment_id + name.
Executors will call this quota whenever an agent first tries to use a quota in live mode.

Once interest in a quota is declared the quota state should be held in memory in the shard manager, as it will be needed for future calculations based on the quota. Any allocation changes need to be persisted in the db before returning.

### LINKED ISSUE #3006: Shard Manager quata allocation / leases
Shard manager allocates quotas to all interested executors whenever they poll. In the initial version that is an equal allocation to all executors that expressed interest in the quota. A quota lease is just (quata, worker_executor_id, allocated_budget, lease_expiry, epoch).
Any quota lease updates need to be atomic and tracked in the db.
An initial lease should probably be granted on the first request from the executor that expresses interest in a quota.

Each lease should contain an epoch so it can be deterministically identified. This is also useful in case when need to add fencing support later.

### LINKED ISSUE #3007: Shard Manager quota refresh and reclamation
All executors need to send a quota update / heart beat before the lease_expiry time. The possible reasons that an executor has for sending a hearbeat are:

- they ran out of quota early and need more allocations
- they still have quota, but their lease was about to expiry
- they lost interest in a quota

In order to support all these usecases, the heartbeat should include unused allocation / demand from the executor side. The response from the shard manager should include a new lease with an allocation. The assigned allocation can be one of:

- Infinity. The quota currently does not exist, so the executor can use as much capacity as they want. They still need to send a heartbeat later to request a new lease in case the quota is created
- A fixed amount. The quota exists and has unused capacity. The executor may use this capacity but needs to send another heartbeat when it runs out
- An expected time when capacity will be available. In case the quota exists and has no remaining capacity, the shard manager should give the executor an estimate for when it can poll for a new allocation. This is used for
    - Limiting the number of requests made to the shard manager
    - Allowing the executor to suspend agents in case the expected wait time is large

In case the executor does not send a heartbeat before the lease expires, the lease is reclaimed and becomes available for allocation to other executors.

Any quota allocation / heartbeat should also include the current enforcement action and other quota metadata in order to prevent divergence between shard manager and executor state.

mschuwalow:
resolves #3005
resolves #3006
resolves #3007

### DIFF
--- a/golem-shard-manager/src/error.rs
+++ b/golem-shard-manager/src/error.rs
@@ -129,7 +129,7 @@ impl From<QuotaError> for golem::shardmanager::v1::QuotaError {
                 },
             QuotaError::StaleEpoch { resource_definition_id, current, provided } =>
                 golem::shardmanager::v1::QuotaError {
-                    error: Some(grpc_quota_error::Error::LeaseNotFound(golem::common::ErrorBody { error: format!("Stale epoch provided for {resource_definition_id} (provided: {provided}, current: {current}) ") })),
+                    error: Some(grpc_quota_error::Error::StaleEpoch(golem::common::ErrorBody { error: format!("Stale epoch provided for {resource_definition_id} (provided: {provided}, current: {current}) ") })),
                 },
             QuotaError::InternalError(_) =>
                 golem::shardmanager::v1::QuotaError {


golem-shard-manager/src/error.rs
```diff
@@ -118,6 +119,26 @@ impl From<ShardManagerError> for golem::shardmanager::v1::ShardManagerError {
...
+    fn from(value: QuotaError) -> golem::shardmanager::v1::QuotaError {
+        use golem::shardmanager::v1::quota_error as grpc_quota_error;
+        match value {
+            QuotaError::LeaseNotFound { resource_definition_id } =>
+                golem::shardmanager::v1::QuotaError {
+                    error: Some(grpc_quota_error::Error::LeaseNotFound(golem::common::ErrorBody { error: format!("Did not find lease for {resource_definition_id}") })),
+                },
+            QuotaError::StaleEpoch { resource_definition_id, current, provided } =>
+                golem::shardmanager::v1::QuotaError {
+                    error: Some(grpc_quota_error::Error::LeaseNotFound(golem::common::ErrorBody { error: format!("Stale epoch provided for {resource_definition_id} (provided: {provided}, current: {current}) ") })),
```
vigoo: I think this should be StaleEpoch

PR #3070, opened 2026-03-25
store security scheme name instead of ids during mcp deployments

afsalthaj:


golem-registry-service/src/services/deployment/mcp.rs
```diff
@@ -56,33 +56,37 @@ impl DeployedMcpService {
...
-                let security_scheme_id = deployment.mcp_data.value().security_scheme_id;
-                let environment_id = deployment.environment_id;
                 let mut compiled_mcp = CompiledMcp::try_from(deployment)?;
 
-                if let Some(scheme_id) = security_scheme_id {
-                    let scheme = self
-                        .security_scheme_service
-                        .get(scheme_id, &AuthCtx::system())
+                // Resolve security scheme by name at runtime
+                if let Some(scheme_name) = &compiled_mcp.security_scheme_name {
```
mschuwalow, line 62: Better to do it as part of the join

PR #3071, opened 2026-03-25
Transparent http retries

### LINKED ISSUE #3020: Better HTTP retry handling
This is a more complex task about improving how HTTP retries are handled at various stages of a streaming request. The result is transparent retries (without restarting the agent) in many cases, automatic attempt to resume downloaded streams with `Range` header etc.

vigoo:
Transparent local HTTP retries

Resolves #3020

Built on top of #3036 

Extends the "in-place" retries to the most important places, HTTP and RPC calls.

### DIFF
--- a/golem-worker-executor/src/durable_host/http/inline_retry.rs
+++ b/golem-worker-executor/src/durable_host/http/inline_retry.rs
@@ -19,12 +19,14 @@
 //! metadata and re-sends it in-place — without trapping, without creating a new WASM
 //! instance, and without replaying the oplog.
 //!
-//! # Zones
+//! # Phases
 //!
-//! - **Zone 1**: Retry at `FutureIncomingResponse::get()` — the response hasn't arrived yet,
-//!   or arrived with an error. The outgoing body is fully finished.
-//! - **Zone 2**: Retry during response body reading — the response was partially consumed.
-//!   Requires re-sending the request and verifying the response prefix matches.
+//! - **Awaiting Response**: Retry at `FutureIncomingResponse::get()` — the response
+//!   hasn't arrived yet, or arrived with an error. The outgoing body is fully
+//!   finished.
+//! - **Resuming Response Body**: Retry during response body reading — the response
+//!   was partially consumed. Requires re-sending the request and verifying the
+//!   response prefix matches.
 
 use crate::durable_host::HttpRequestState;
 use crate::durable_host::durability::{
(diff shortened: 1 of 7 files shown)


golem-worker-executor/src/durable_host/http/inline_retry.rs
```diff
@@ -0,0 +1,1439 @@
...
+    HadBodySkip,
+    /// The output stream had subscribe() called, so pollable may be stale after replacement.
+    OutputStreamSubscribed,
+}
+
+/// Which retry zone is being attempted.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RetryZone {
+    /// Retry at FutureIncomingResponse::get() — response not yet consumed.
+    Zone1,
```
noise64: Could these have better names, hinting better on the stages? like "before response consume" etc?

golem-worker-executor/src/durable_host/http/inline_retry.rs
```diff
@@ -0,0 +1,1439 @@
...
+/// A chunk of outgoing body data reconstructed from the oplog.
+///
+/// This avoids materializing the entire body in memory — callers can
+/// stream chunks into a hyper body pipe incrementally.
+#[derive(Debug, Clone)]
+pub enum BodyChunk {
+    /// Actual data bytes written via `write`.
+    Data(Bytes),
+    /// A run of zeroes written via `write_zeroes`.
+    Zeroes(u64),
```
noise64, line 182: Just interested, how common is this case to have a special (i guess) compression? If we have such a thing, shouldn't it use something like deflate / gzip?

Or is it used (too) for chunks with zero length (Zeroes(0))?

golem-worker-executor/src/durable_host/http/inline_retry.rs
```diff
@@ -0,0 +1,1439 @@
...
+    {
+        Some(resp) => resp,
+        None => return Ok(false),
+    };
+
+    let status = response.resp.status().as_u16();
+    let between_bytes_timeout = response.between_bytes_timeout;
+
+    // 7. Handle response status
+    match status {
```
noise64, line 1064: Do we have to handle other 2xx success codes? like 201 (or in general, services could even use custom ones)?

Or maybe this is specialized to idempotent methods?

golem-worker-executor/src/durable_host/http/inline_retry.rs
vigoo: I think it is never used in practice but it is on the wasi stream interface so we have to do something with it

golem-worker-executor/src/durable_host/http/inline_retry.rs
vigoo: Yes I think it should treat any other than 206 and 416 the same.

golem-worker-executor/src/durable_host/http/inline_retry.rs
vigoo: There are many places where this terminology can be found in the PR (function names, debug logs etc). I first wanted to rename then thought that it's not a bad name - but I'm probably biased because this is the terminology that we used during the whole agentic planning session.

vigoo:
@noise64 fixed both concerns

PR #3073, opened 2026-03-25
Enable (global) configurable snapshotting for oplog processor plugins

### LINKED ISSUE #3015: Snapshotting for plugins
Automatic snapshot based recovery is based on agent metadata configuration and currently not working for oplog processor plugins, however it is a very good use case for them and we should actually have snapshot-based recovery enabled for plugins like the otlp export.

vigoo:
Resolves #3015 

Intentionally kept simple for now (no per-plugin config etc)

### DIFF
--- a/golem-worker-executor/tests/durability.rs
+++ b/golem-worker-executor/tests/durability.rs
@@ -676,8 +676,8 @@ async fn ts_default_json_snapshot_recovery(
                     panic!("Snapshot {i} 'count' is not a number: {:?}", state["count"])
                 });
                 assert!(
-                    (1.0..=5.0).contains(&count),
-                    "Snapshot {i} count should be between 1 and 5, got {count}"
+                    (0.0..=5.0).contains(&count),
+                    "Snapshot {i} count should be between 0 and 5, got {count}"
                 );
             }
             PublicSnapshotData::Raw(raw) => {
@@ -794,8 +794,8 @@ async fn ts_default_json_snapshot_recovery_across_multiple_restarts(
                     panic!("Snapshot {i} 'count' is not a number: {:?}", state["count"])
                 });
                 assert!(
-                    (1.0..=6.0).contains(&count),
-                    "Snapshot {i} count should be between 1 and 6, got {count}"
+                    (0.0..=6.0).contains(&count),
+                    "Snapshot {i} count should be between 0 and 6, got {count}"
                 );
             }
             PublicSnapshotData::Raw(raw) => {


golem-worker-executor/src/worker/mod.rs
```diff
@@ -2767,6 +2778,64 @@ fn resolve_snapshot_policy(
     }
 }
 
+fn normalize_snapshot_policy(policy: &SnapshotPolicy) -> SnapshotPolicy {
```
mschuwalow: Can we do this during config loading instead? Not a fan of validating config inline with business logic

golem-worker-executor/src/worker/mod.rs
```diff
@@ -2767,6 +2778,64 @@ fn resolve_snapshot_policy(
...
+    };
+    let has_load = match metadata.load_snapshot() {
+        Ok(Some(_)) => true,
+        Ok(None) => false,
+        Err(err) => {
+            warn!("Failed to check load-snapshot export: {err}");
+            false
+        }
+    };
+    has_save && has_load
```
mschuwalow: has_save and has_load are checked inconsistently with has_oplog_processor here. (not correctness, just style)

golem-worker-executor/src/worker/mod.rs
```diff
@@ -2794,8 +2863,13 @@ pub enum QueuedWorkerInvocation {
...
         sender: oneshot::Sender<Result<(), WorkerExecutorError>>,
     },
-    // Triggers a periodic snapshot save for the worker
-    SaveSnapshot,
+    // Triggers a periodic snapshot save for the worker.
+    // When enqueued by the periodic snapshot timer, `pending_flag` carries the
+    // `AtomicBool` that must be reset to `false` after the snapshot completes so
+    // the timer can schedule the next one.
+    SaveSnapshot {
+        pending_flag: Option<Arc<AtomicBool>>,
```
mschuwalow: I don't see this flag getting read yet. That will come in a follwoup?

vigoo: it should be passed in the invocation loop. let me recheck :)

vigoo: Actually I think the whole time-based snapshotting mode should be done differently (checking between invocations for time since last snapshot, and injecting if reached). I'll change it in this pr - the current background task based one would only make sense if we could do snapshots in the middle of invocations

Issue #3067, opened 2026-03-24
Registry notification cleanups


noise64:
Some fixes and cleanups, to make the stream more reliable:
- always insert the event records as part of the DB transaction
- always do that as the last operation in the transaction
- in postgres: use an advisory lock just before the event record insert to guarantee event ordering
- rework RegistryChangeNotifier::notify: in general, it should not result in duplicated events
- review "replay" handling

PR #3074, opened 2026-03-25
Registry service notification cleanups

noise64:
- resolves https://github.com/golemcloud/golem/issues/3067
- changes notifications to be strictly monotonic (with possible gaps) _inside_ registry-service, to achive this:
  - message creations are now always part of the related transaction, and always executed at the end of such transactions
  - changes postrges message creation to:
    - create advisory lock for message insertion (which orders all events)
    - insert message
    - notify from the transaction (in theory this _can_ fail, but not for our listen patterns in practice)
  - removes "same node fast-path" notifications, which could resulted many duplication and out-of-order patterns for some performance gains
  - to align with the above, "notify" was changed to "signal event might be emmited", and sqlite now also uses a local worker
  - the above pattern also uses a "must use" pattern, to help with not forgetting calling the signal where it is needed
  - extracted common client subscriber patterns into the registry service client, and added filtering with warn logging for duplicated events, as a safety-net, and to have some proof in production, that now our observed streams are virtually deduplicated
  - related to the above, if a client connects with no last seen event id, we do not replay all the existing events, given the signal is intended for cache invalidation, and a new empty client is already "invalidated"
- on top of the above, and as another layer of safety-net: better guarantess for deployment events: the event now includes the current_deployment_revision too, which is guaranteed to be montonic, unlike the deployment_revision, which can go backwards during rollbacks
- to align with the above, did some cleanup on the "agent type by name" GRPC functions: removed the unused (but covered with the new, unified version) functions, and the remaining one, depending on the filters (latest vs specific), now optionally retruns the current_deployment_version too

### DIFF
--- a/golem-registry-service/tests/repo/common.rs
+++ b/golem-registry-service/tests/repo/common.rs
@@ -60,7 +60,7 @@ use golem_registry_service::repo::registry_change::{
 };
 use golem_registry_service::services::registry_change_notifier::RequiresNotificationSignalExt;
 use golem_registry_service::services::registry_change_notifier::{
-    SqliteRegistryChangeNotifier, RegistryChangeNotifier,
+    RegistryChangeNotifier, SqliteRegistryChangeNotifier,
 };
 use golem_service_base::repo::blob::Blob;
 use heck::ToKebabCase;


golem-api-grpc/proto/golem/registry/v1/registry_service.proto
```diff
@@ -319,6 +282,8 @@ message ResolveAgentTypeByNamesSuccessResponse {
   RegisteredAgentType agent_type = 1;
   golem.common.EnvironmentId environment_id = 2;
   uint64 deployment_revision = 3;
+  // Present only when request.deployment_revision is not set.
+  optional uint64 current_deployment_revision = 4;
```
mschuwalow, line 286: Not a fan of conditional schemas. If we do need this, I would prefer having two versions of this endpoint

golem-api-grpc/proto/golem/registry/v1/registry_service.proto
```diff
@@ -44,8 +44,6 @@ service RegistryService {
   // agent types api
   rpc GetAllAgentTypes (GetAllAgentTypesRequest) returns (GetAllAgentTypesResponse);
   rpc GetAgentType (GetAgentTypeRequest) returns (GetAgentTypeResponse);
-  rpc ResolveLatestAgentTypeByNames (ResolveLatestAgentTypeByNamesRequest) returns (ResolveLatestAgentTypeByNamesResponse);
```
mschuwalow, line 47: 🎉

golem-registry-service/src/repo/registry_change.rs
```diff
@@ -222,12 +256,36 @@ impl TryFrom<RegistryChangeEventRow> for RegistryChangeEvent {
...
+) -> RepoResult<Option<RegistryChangeEvent>> {
+    if matches!(row.event_type, RegistryEventType::DeploymentChanged)
+        && row.current_deployment_revision_id.is_none()
+    {
+        warn!(
+            event_id = row.event_id.0,
+            deployment_revision_id = ?row.deployment_revision_id,
+            "Skipping legacy DeploymentChanged registry event without current_deployment_revision_id"
+        );
+        // TODO: remove this compatibility skip path in some of the next release cycles.
```
mschuwalow, line 275: Could you double-check this? We don't even have an rc yet

golem-registry-service/src/services/registry_change_notifier.rs
```diff
@@ -44,33 +46,118 @@ pub trait RegistryChangeNotifier: Send + Sync {
     }
 }
 
+impl<T> RegistryChangeNotifier for Arc<T>
```
mschuwalow, line 49: This should not be needed. The use sites also still use Arc<dyn RegistryChangeNotifier> from what I can tell

golem-registry-service/src/services/registry_change_notifier.rs
```diff
@@ -44,33 +46,118 @@ pub trait RegistryChangeNotifier: Send + Sync {
...
+        notifier.signal_new_events_available();
+        self.into_inner_after_signal()
+    }
+}
+
 /// Local in-process notifier for single-node deployments (e.g., SQLite).
-/// No background tasks needed — notify() directly publishes to the broadcast.
+/// Signal calls enqueue local processing and a background worker fetches and broadcasts
+/// concrete registry change events.
 pub struct LocalRegistryChangeNotifier {
```
mschuwalow: Maybe call this SqliteRegistryChangeNotifier?

golem-api-grpc/proto/golem/registry/v1/registry_service.proto
noise64: I'm not a fan of neither options :) I tried both, and this was the smaller change for now, but happy to get back to this with another TODO issue we have for the release, including reviewing all the paths and combinations (as in current / vs specific auth or non auth) .

golem-registry-service/src/repo/registry_change.rs
noise64: AFAIU, we are already deployed to dev, and wanted to avoid stuck readers (even if there is a low chance for it), and i think it is fine to keep this here as a pattern for later. (As in, here it is okay to treat backward incompatible changes as skips)

golem-registry-service/src/services/registry_change_notifier.rs
noise64: Otherwise you have to use &** on the callsite, or something similar

golem-registry-service/src/services/registry_change_notifier.rs
noise64: Renaming it

golem-registry-service/src/services/registry_change_notifier.rs
mschuwalow: Ah, I see that it's because of the RequiresNotificationSignalExt

I would prefer just adjusting the trait like this, but also fine keeping as is.
```
pub trait RequiresNotificationSignalExt<T> {
    fn signal_new_events_available(self, notifier: &Arc<dyn RegistryChangeNotifier>) -> T;
}

impl<T> RequiresNotificationSignalExt<T> for RequiresNotificationSignal<T> {
    fn signal_new_events_available(self, notifier: &Arc<dyn RegistryChangeNotifier>) -> T {
        notifier.signal_new_events_available();
        self.into_inner_after_signal()
    }
}
```

noise64: that's what i did not like originally :D but i might change back to this, if i go near it in the follow issue, as i know understand, that that is more idiomatic there

PR #3075, opened 2026-03-25
Build check - checking tools and fixing dependencies

noise64:
- resolves https://github.com/golemcloud/golem/issues/2707
- resolves https://github.com/golemcloud/golem/issues/3041
- introduces `manifestVersion` in application manifests
- better `npm install` detection: now runs if golem or the user changed deps
- some CLI styling tweaks

### Example of fixing `package.json`, `tsconfig.json`, `AGENTS.md`
image

### Nicer external command prefix
image

### Primary and secondary styles for decorated nesting
image

### "Collapsed" diff for AGENTS.md
image

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
 
 env:
-  WASM_RQUICKJS_VERSION: "v0.1.3"
+  WASM_RQUICKJS_VERSION: "v0.1.4"
 
 jobs:
   run-benchmark:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,7 +19,9 @@ permissions:
 
 env:
   BUILD_TARGET: "x86_64-unknown-linux-gnu"
-  WASM_RQUICKJS_VERSION: "v0.1.3"
+  WASM_RQUICKJS_VERSION: "v0.1.4"
+  WASI_SDK_VERSION: "25"
+  WASI_SDK_PATH: /opt/wasi-sdk
 
 jobs:
   build-and-store:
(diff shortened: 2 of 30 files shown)


.github/workflows/benchmark.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: "20.x"
+          node-version: "lts/*"
```
vigoo, line 33: we need 22+ for node:sqlite (no idea what `lts` maps to :) )

cli/golem-cli/src/app/build/check/rust.rs
```diff
@@ -0,0 +1,608 @@
...
+use crate::app::context::BuildContext;
+use crate::app::edit;
+use crate::app::edit::cargo_toml::{DependencySpec, DependencyTable};
+use crate::fs;
+use crate::model::GuestLanguage;
+use crate::sdk_overrides::{RustDependency, SdkOverrides};
+use crate::versions;
+use std::path::{Path, PathBuf};
+
+pub(super) fn plan_rust_cargo_fix_steps(
```
vigoo, line 28: Minor thing you may want to verify: agent thinks if you delete let's say a `golem-rust = { workspace = true }` completely from your component `Cargo.toml` but keep the correct entry in the workspace `Cargo.toml`, then this thing does not restore the workspace dependency, but adds an explicit one instead.

.github/workflows/benchmark.yaml
noise64: 24

cli/golem-cli/src/app/build/check/rust.rs
noise64: I think the agent is right, i will do another round on workspace support. Note: currently when one upgrades to multi-component layouts, we do not automatically convert to a cargo workspace, but now we at least mostly support it, if a user manually does so. I will add a follow-up card to make this handle more cases more nicely, and maybe doing the upgrade to workspaces always, when we upgrade to multi-component mode. (I did not do so initially, to simplify a bit the task, and defer thinking about those edge cases.)

PR #3077, opened 2026-03-26
Checkpoint API for Rust and TS

### LINKED ISSUE #1555: Checkpoint/verify API into the language-specific SDKs
Concept:

```rust
let cp1 = golem::checkpoint("pre-json generation");

...

let json = llm.call(prompt);

// Automatically rolls back to last checkpoint if verification fails.
let result = cp1.verify(Result::from_json(json));
```

This should be implementable using the existing transaction API.

We should have the new API in:
- golem-rust
- golem-go
- golem-ts
- golem-py

vigoo:
Resolves #1555


vigoo:
Yes this is just a nicer interface for the existing "time travel jump" feature. I think our primary reason was to have something like this because it may be familiar to users? cc @jdegoes

PR #3078, opened 2026-03-26
quotas: shard-manager persistence

### LINKED ISSUE #3008: Shard Manager quota handling on restart
The following algorithm should be executed by the shard manager on restart and the executors when they detect shard manager restart respectively.

Shard manager startup:

```
1 load quota allocations from DB
2 delete expired leases
3 recompute remaining quota
4 start serving requests
```

Executors:

```
1 reconnect
2 send heartbeat
3 demand/allocation re-evaluated
```

mschuwalow:
resolves #3008

### DIFF
--- a/golem-shard-manager/src/lib.rs
+++ b/golem-shard-manager/src/lib.rs
@@ -20,7 +20,6 @@ mod registry_event_subscriber;
 pub(crate) mod sharding;
 
 use self::grpc::ShardManagerServiceImpl;
-use self::sharding::worker_executor::WorkerExecutorService;
 use crate::config::{HealthCheckK8sConfig, HealthCheckMode};
 use crate::quota::{DbQuotaRepo, GrpcResourceDefinitionFetcher, QuotaService};
 use crate::registry_event_subscriber::ShardManagerRegistryInvalidationHandler;


golem-shard-manager/src/model.rs
```diff
@@ -42,7 +37,14 @@ pub struct Pod {
...
-    #[cfg(test)]
+    pub fn host(&self) -> &str {
+        &self.host
+    }
+
+    pub fn port(&self) -> u16 {
+        self.port
+    }
+
     pub fn new(host: String, port: u16) -> Self {
```
vigoo: You've removed `#[cfg(test)]` from `Pod::new` here and using it in places (`restore_state` in `QuotaService`) - is this intentional? This new method does not seem right at all and I think that's why it was only exposed for some unit tests - always sets `ip` to localhost ip, and sets the `pod_name` to `None` (and that `pod_name` seems to be manipulated from the outside?) 

If you only need a host/port pair probably you should use a different type. There is one (confusingly also called `Pod` in `golem_common`)

golem-test-framework/src/components/rdb/mod.rs
```diff
@@ -100,7 +100,7 @@ impl DbInfo {
                 (
                     "GOLEM__DB__CONFIG__DATABASE".to_string(),
                     db_path
-                        .join(service_namespace)
+                        .join(format!("{service_namespace}.db"))
```
vigoo, line 104: There is a similar path construction above this which you haven't changed, should we keep them in sync?

golem-registry-service/src/config.rs
```diff
@@ -182,7 +182,10 @@ impl Default for RegistryServiceConfig {
             workspace: "release".to_string(),
             http_port: 8081,
             grpc: GrpcApiConfig::default(),
-            db: DbConfig::default(),
+            db: DbConfig::Sqlite(DbSqliteConfig {
+                database: "golem_registy_service.db".to_string(),
```
vigoo:
```suggestion
                database: "golem_registry_service.db".to_string(),
```

integration-tests/tests/sharding.rs
```diff
@@ -560,6 +563,12 @@ mod tests {
         fn flush_redis_db(&self) {
             self.redis().flush_db(0);
         }
+
+        async fn clean_shard_manager_sqlite(&self) {
+            tokio::fs::remove_file(self.temp_directory().join("sqlite/golem_shard_manager.db"))
```
vigoo: Do we have -wal/-shm files as well for this? If yes we should probably delete those too

golem-shard-manager/src/sharding/error.rs
```diff
@@ -0,0 +1,96 @@
...
+    #[error("gRPC: error status: {0}")]
+    GrpcError(#[from] tonic::Status),
+    #[error("No result")]
+    NoResult,
+    #[error("Worker execution error: {0}")]
+    WorkerExecutionError(WorkerExecutorError),
+    #[error("Persistence serialization error {0}")]
+    SerializationError(String),
+    #[error("Redis error {0}")]
+    RedisError(#[from] golem_common::redis::RedisError),
```
vigoo: I guess this can be removed

golem-shard-manager/src/quota/quota_state.rs
```diff
@@ -0,0 +1,399 @@
...
+pub(super) struct QuotaResourceRevision(u64);
+
+impl QuotaResourceRevision {
+    pub const INITIAL: Self = Self(0);
+
+    pub fn next(self) -> anyhow::Result<Self> {
+        if self.0 == i64::MAX as u64 {
+            Err(anyhow::anyhow!(
+                "Cannot increment {} beyond i64::MAX ({})",
+                stringify!($name),
```
vigoo: no `name` here?

golem-test-framework/src/components/rdb/mod.rs
mschuwalow: Which one? I believe all sqlite paths (including the default configs) should now have the .db extension / consistent naming.

Issue #2977, opened 2026-03-12
Maximum concurrent live agents per account per executor


vigoo:
We want to be able to limit the number of _concurrently running agents_ per account. To keep it simple, we do it per executor. 
For that, we can use the same technique that we already use with memory permits, and that we are going to use for disk space permits (see #2970).

Steps:
- Add a new limit to `Plan` 
- We need a per-account semaphor in `ActiveWorkers` and the workers need to acquire it in their (existing) waiting-for-permits phase.
- We need a a method in `ActiveWorkers` that, in case there are not enough permits, tries to free up one by kicking out an idle worker belonging _to the same account_
- If there is nothing to kick, the agent stays in waiting for permit phase until it gets allowance
- Tests for this

PR #3082, opened 2026-03-27
Maximum concurrent live agents per account per executor

kmatasfp:
Resolves: #2977 

This PR adds a plan-level limit on how many agents from the same account can run concurrently on a single executor. When the limit is reached, the executor first tries to evict the oldest idle agent from the same account to free a slot. If there are no idle agents to evict, the new agent waits in the existing `WaitingForPermit` phase until a running agent finishes and returns its permit.

### How it works

The limit is enforced by a new `ConcurrentAgentsSemaphore` — a standalone, per-account semaphore following the same pattern as the existing memory and filesystem storage semaphores. Each account gets its own tokio `Semaphore` sized to its plan limit. Running agents hold an `OwnedSemaphorePermit` for their lifetime; dropping `RunningWorker` automatically returns the permit via Rust's `Drop`, freeing a slot for the next waiting agent.

The limit value flows through the existing resource limits pipeline: `Plan.max_concurrent_agents_per_executor` → `PlanData` proto → `ResourceLimits` → `AtomicResourceEntry` → `ConcurrentAgentsSemaphore`. The sentinel value `1_000_000_000_000_000_000` (10^18, consistent with `monthly_gas_limit`) means unlimited, which is the default for all existing plans.

### DIFF
--- a/golem-worker-executor/tests/resource_limits.rs
+++ b/golem-worker-executor/tests/resource_limits.rs
@@ -270,7 +270,7 @@ async fn concurrent_agent_limit_waits_for_running_agent_to_finish(
         .await?;
 
     // Start a1 using the HttpClient2 agent. Pass the HTTP server port via env.
-    let a1 = agent_id!("HttpClient2", "concurrent-wait-http-1");
+    let a1 = agent_id!("HttpClient2");
     let mut env = HashMap::new();
     env.insert("PORT".to_string(), port.to_string());
     let a1_id = executor


golem-worker-executor/tests/resource_limits.rs
```diff
@@ -156,3 +158,154 @@ async fn table_exceeding_limit_not_retried(
...
+    let context = TestContext::new(last_unique_id);
+    let executor = start_with_concurrent_agent_limit(deps, &context, 1).await?;
+
+    let component = executor
+        .component_dep(&context.default_environment_id, agent_counters)
+        .store()
+        .await?;
+
+    // Start a1 — takes the only permit. Keep it busy with an invocation so it
+    // is Running (not Idle) and cannot be evicted.
```
vigoo: This test is going to be flaky because it depends on `initialize` running slow enough so it is still running when the second agent is created. You should invoke something that is guaranteed to block until the second agent is created (for example it could call back via http and you can gate the http response)

kmatasfp: should be more stable now

PR #3089, opened 2026-03-29
Fix event broadcast in single-executable golem

vigoo:


cli/golem/src/launch.rs
```diff
@@ -294,7 +294,7 @@ fn shard_manager_config(
             ..Default::default()
         },
         registry_service: golem_service_base::clients::registry::GrpcRegistryServiceConfig {
-            host: "localhost".to_string(),
```
mschuwalow, line 297: This is used everywhere, put probably it actually wants to be localhost. (router_addr is the address the router binds to)

noise64: @mschuwalow  I'm also fixing hosts for windows too, in general, so it will be handled better, when 0.0.0.0 is used

Issue #3009, opened 2026-03-17
Worker Executor quota enforcement


mschuwalow:
Executors should ask for quota allocation the first time an agent expresses interest in a quota in live mode and release it when there are no more interested agents.

Any active quota needs to have the periodic heartbeat and demand sent to the shard-manager.

When the reserve host function is called, the requested amount is debited from the current leases remaining allocation. If there is insufficient allocation available in the current lease the policy enforcement action is fired.

When the commit host function is used, the following rules should apply:

- used > reserved: the extra usage is debited from the current quota allocation. If there is no remaining allocation this is a noop. We might not want to debit this in case the quota allocation epoch is not the same as when reserve was called. Doing it in either case errs on the side of preventing quota overuse though
- used < reserved: The unused quota is returned to the allocation only if the current allocation epoch is the same as when reserve was called. Doing otherwise would allow quota overuse

Note that the enforcement on executor works slightly differently for different resources types, e.g. concurrency limited resources should work via semaphores. No special handling for the different resources types should be needed in the shard-manager.

PR #3097, opened 2026-03-30
quotas: wire up quota client in worker executor and clean up bootstrap

mschuwalow:
part of #3009

* create a shared shard manager client in golem-service-base (same as registry service)
* refactor all shard manager clients to use the new shared clients
* adjust bootstrap in worker executor to reduce duplication / simplify logic
* wire up basic quota service with an unlimited version for testing. This will be extended in the next prs


golem-debugging-service/src/config.rs
```diff
@@ -90,9 +89,7 @@ impl DebugConfig {
             http_client: Default::default(),
             http_address: self.http_address,
             http_port: self.http_port,
-            shard_manager_service: ShardManagerServiceConfig::SingleShard(
-                ShardManagerServiceSingleShardConfig {},
-            ),
+            shard_manager: Default::default(),
```
vigoo, line 92: The debugging service must not register itself to the shard manager. This "single shard" config is by design otherwise real agents would be assigned to the debugger instances because they are built on the worker executor.

(This was definitely the case when we created it, but I think it is still true)

golem-test-framework/src/components/worker_executor/mod.rs
```diff
@@ -95,31 +95,15 @@ async fn env_vars(
...
-        )
-        .with_str(
-            "GOLEM__SHARD_MANAGER_SERVICE__CONFIG__RETRIES__MAX_DELAY",
-            "2s",
-        )
-        .with_str(
-            "GOLEM__SHARD_MANAGER_SERVICE__CONFIG__RETRIES__MULTIPLIER",
-            "2",
-        )
+        .with_str("GOLEM__SHARD_MANAGER__RETRIES__MAX_ATTEMPTS", "5")
```
vigoo, line 103: These should be called `GOLEM__SHARD_MANAGER__RETRIES_ON_UNAVAILABLE__MAX_ATTEMPTS` etc

golem-worker-executor/src/services/shard_manager.rs
```diff
@@ -12,135 +12,54 @@
...
-            |_| true,
-        )
-        .await
+impl ShardManagerService for GrpcShardManagerService {
+    async fn register(
+        &self,
+        port: u16,
+        pod_name: Option<String>,
+    ) -> Result<ShardAssignment, ShardManagerError> {
+        self.client.register(port, pod_name).await
```
vigoo: The old code with `with_retries` may behaved differently than the new regarding retrying errors. (Maybe it was just an accidental "double retry" after we've introduced the retrying-capable common grpc clients, but it worth checking)

golem-debugging-service/src/lib.rs
```diff
@@ -144,167 +164,207 @@ impl Bootstrap<DebugContext> for ServerBootstrap {
...
-            join_set,
         )
-        .await
     }
 }
 
-pub async fn run_debug_worker_executor<T: Bootstrap<DebugContext> + ?Sized>(
-    bootstrap: &T,
+#[allow(clippy::too_many_arguments)]
+pub fn debug_create_services(
```
vigoo: This function name is confusing and I had to dig in to understand what it is. What about `create_debugging_service_services`? That also sounds bad :)

golem-debugging-service/tests/debug_mode/mod.rs
```diff
@@ -91,14 +91,16 @@ async fn run_debug_worker_executor(
...
-    let run_details = TestDebuggingServerBootStrap::new(regular_executor_context)
-        .run(
-            golem_config.into_golem_config(),
-            prometheus_registry,
-            runtime,
-            join_set,
-        )
-        .await?;
+) -> Result<golem_debugging_service::RunDetails, anyhow::Error> {
+    let cors_origin_regix = golem_config.cors_origin_regex.clone();
```
vigoo:
```suggestion
    let cors_origin_regex = golem_config.cors_origin_regex.clone();
```

golem-debugging-service/src/config.rs
mschuwalow: At runtime the debugging service is still using the single shard manager

The way I wanted to do it now is that everything that we are "statically" overriding can be done by the bootstrap trait https://github.com/golemcloud/golem/pull/3097/changes/BASE..ec7e22a1ec852e06df625537d50986c1cb114b1a#diff-d00cad162822fa700d9a52d33ec0bad71936b1ed9bb7f5724e36126137ac699eR108.

(In order to not expose useless options to users)

The shard manager config is now just the config to the shard manager grpc endpoint (which I need for both sharding and quotas). This works nicely for all user facing configs (production is always requiring a real shard manager, debugging service never has any), but there is a slight annoyance here when going from shard manager config -> golem config

golem-worker-executor/src/services/shard_manager.rs
mschuwalow: Yeah, let me restore the double retries in golem service base. Looks like they are useful for errors from the shard manager (not just from the grpc connection)

golem-test-framework/src/components/worker_executor/mod.rs
mschuwalow: Readded the retry config under the retries key, so these should be fine now

golem-worker-executor/src/services/shard_manager.rs
mschuwalow: done

vigoo (approved):

PR #3104, opened 2026-03-31
Semantic retries

### LINKED ISSUE #3018: Retry policy evaluation abstraction
Create a better abstraction for evaluation retry policies as a preparation of replacing the per-agent global with the fine-grained new one. The implementation would still just use the existing simple retry policy, but it's API should already be prepared for the new features.

### LINKED ISSUE #3021: Per-environment policy registry
Introduce the new, fine-grained policy model according to the product spec.
Store it per-environment in a mutable way, similar to how shared secret configs are handled.
Have a public API, and gRPC API + host functions to manage them.
Wire the retry policy evaluator to this new model.

### LINKED ISSUE #3022: Allow defining environment-level retry policies at the manifest level

vigoo:
This is the final PR for the semantic retries epic on top of #3071 

Resolves #3018 
Resolves #3021 
Resolves #3022 

It contains the new retry policy engine and integrates it to our services in a similar way as secrets were done - they can be changed on the fly in the registry service, but there are a set of default retry policies associated with the environment coming from the app manifest.

### DIFF
--- a/integration-tests/tests/plugins.rs
+++ b/integration-tests/tests/plugins.rs
@@ -1031,8 +1031,19 @@ async fn oplog_processor_no_duplicates_after_crash(
     )
     .await?;
 
-    // Wait for flush to complete
-    let _ = wait_for_invocations(&received_batches, 4, Duration::from_secs(60)).await;
+    // Phase 1: wait for the worker oplog to show all 4 completed invocations
+    // (1 agent-initialization + 3 add calls).
+    wait_for_oplog_completions(
+        &user,
+        &worker_id,
+        4,
+        Duration::from_secs(120),
+        &received_batches,
+    )
+    .await;
+
+    // Phase 2: wait for the oplog processor callbacks to arrive before crashing.
+    let _ = wait_for_invocations(&received_batches, 4, Duration::from_secs(120)).await;
     tokio::time::sleep(Duration::from_secs(10)).await;
 
     // Crash both user worker and oplog processor plugin worker via simulated_crash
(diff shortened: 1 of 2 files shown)


cli/schema.golem.cloud/app/golem/1.5.0-dev.1/golem.schema.json
```diff
@@ -203,6 +203,16 @@
         }
       }
     },
+    "retryPolicyDefaults": {
```
noise64: in general, when a schema is added, it is also manually deployed to cloudfront, so do not modify exsiting schemas, as they are already deployed, always create new one, and once the PR is ready to merge, deploy it.

(probably should be a skill, except for the deploy, that we should do manually for now)

golem-common/src/base_model/oplog/public_types.rs
```diff
@@ -228,18 +227,48 @@ impl PublicSpanData {
...
-    pub max_delay: Duration,
-    pub multiplier: f64,
-    pub max_jitter_factor: Option<f64>,
+pub struct PublicNamedRetryPolicy {
+    /// Human-readable identifier for this policy.
+    pub name: String,
+    /// Selection priority — higher values are evaluated first.
+    pub priority: u32,
+    /// The predicate serialized as a JSON string.
+    pub predicate_json: String,
```
noise64: shouldn't we use serde_json::Value for these?

cli/schema.golem.cloud/app/golem/1.5.0-dev.1/golem.schema.json
vigoo: fixed

golem-common/src/base_model/oplog/public_types.rs
vigoo: it's now properly typed

cli/golem-cli/src/app/edit/golem_yaml.rs
```diff
@@ -615,12 +615,27 @@ fn reindent_block(source: &str, start: usize, end: usize, new_indent: usize) ->
 }
 
 fn normalize_block_text(source: &str, start: usize, end: usize) -> String {
-    source[start..end]
-        .trim()
-        .replace(['\r', '\n'], " ")
-        .split_whitespace()
+    let text = &source[start..end];
```
noise64, line 618: could we also add a unit test for this? (fine to do it in a separate PR)

vigoo: Sure, but the existing (prop based) unit tests were finding a bug and this change fixes that.

PR #3111, opened 2026-04-01
quotas: basic quota allocation in worker executor

### LINKED ISSUE #3009: Worker Executor quota enforcement

mschuwalow:
resolves #3009

### DIFF
--- a/golem-worker-executor/src/services/quota.rs
+++ b/golem-worker-executor/src/services/quota.rs
@@ -829,7 +829,7 @@ mod tests {
     use super::*;
     use chrono::Duration as ChronoDuration;
     use golem_common::model::resource_definition::{ResourceLimit, ResourceRateLimit, TimePeriod};
-    use golem_common::model::{Pod, RoutingTable, ShardAssignment};
+    use golem_common::model::{Pod, RoutingTable};
     use golem_service_base::clients::shard_manager::ShardManagerError;
     use pretty_assertions::assert_eq;
     use pretty_assertions::assert_matches;
@@ -991,7 +991,7 @@ mod tests {
             &self,
             _port: u16,
             _pod_name: Option<String>,
-        ) -> Result<ShardAssignment, ShardManagerError> {
+        ) -> Result<u32, ShardManagerError> {
             unimplemented!()
         }
 


golem-worker-executor/src/services/quota.rs
```diff
@@ -12,29 +12,1997 @@
...
+        let mut to_renew: Vec<(ResourceKey, Arc<Mutex<LeaseEntry>>)> = Vec::new();
+        let mut to_release: Vec<BoundedLease> = Vec::new();
+
+        self.state
+            .retain_async(|key, slot_mutex| {
+                if let Ok(slot) = slot_mutex.try_lock() {
+                    if slot.interest.strong_count() == 0
+                        && let TrackedLease::Bounded(b) = &slot.lease
+                    {
+                        to_release.push(b.clone());
```
vigoo, line 462: shouldn't this branch return `false` so `retain_async` removes the item from `self.state`? Currently it never removes anything

golem-worker-executor/src/services/quota.rs
```diff
@@ -12,29 +12,1997 @@
...
+        let mut remaining_waiters: Vec<Waiter> = Vec::new();
+
+        for waiter in waiters {
+            if let Some(reservation) =
+                Self::try_grant(lease, key, waiter.amount, resource_definition_id)
+            {
+                let _ = waiter.tx.send(WaiterOutcome::Granted(reservation));
+            } else {
+                let estimated_wait_time =
+                    Self::estimated_wait(waiter.amount, total_available, &resource_limit);
```
vigoo, line 434: shouldn't the `total_available` decreaded in this loop as we allow higher priority waiters to wait inline?

golem-worker-executor/src/services/quota.rs
mschuwalow: Yes. Already fixed in the next one, just not backported https://github.com/golemcloud/golem/pull/3114/changes#diff-4f888ccc9a1023c196afd6bc5e48562ce41e7c6e710e8c8c3c558d49ea9623edR442

golem-worker-executor/src/services/quota.rs
mschuwalow: The total_available here is just a lower bound for when it makes sense again to check for capacity. I will refactor/improve it later, but for now it assumes no other waiters and no other executors.

Issue #3099, opened 2026-03-30
Make sure resource consumption is consitent in case of cross account invoke


kmatasfp:
When an agent from account A calls another agent belonging to account B via RPC, running in environment shared with A, we want to make sure that resources like fuel etc are consumed correctly. Owner (account B) of the agent being called should be charged.

Issue #2976, opened 2026-03-12
RPC authorization for direct invokes


vigoo:
When an agent calls another agent via RPC, it can take two distinct paths:
- If the target agent is on a different worker executor, the call goes through gRPC via the `worker-service`. Authorization is handled on this path. 
- If the target agent is on the same executor, we bypass the network call and directly invoke the other agent. On this path, we are not checking if the targeted agent has `EnvironmentAction::UpdateWorker` rights for the environment the target agent is belonging to.

This need to be fixed and the two paths must act the same from the caller's point of view when they get rejected.

PR #3112, opened 2026-04-02
Add RPC auth check for local direct worker invocations

kmatasfp:
## Enforce authorization and resource limits for local direct RPC invocations

When a worker invokes another worker via RPC and the target is on the **same executor**, the call previously bypassed all authorization and resource limit checks. The remote path (target on a different executor) correctly enforced both through `worker-service`. This PR closes that gap so that RPC authorization and worker limit accounting are consistent regardless of which executor the target worker lives on.

Additionally, `worker-service` was mapping `AuthError` to `InternalError` rather than `Unauthorized`, causing auth denials to surface to the calling WASM component as `rpc-error.remote-internal-error` instead of `rpc-error.denied`. This is also fixed.

### Changes

**Auth enforcement on the local RPC path**

Added `DirectInvocationAuthService` (`golem-worker-executor/src/services/direct_invocation_auth.rs`) which gates `create_demand`, `invoke_and_await`, and `invoke` before `Worker::get_or_create_running` is called:

- **Fast path**: if `caller_account_id == environment_owner_account_id`, allow immediately with no network call
- **Slow path**: fetch `AuthDetailsForEnvironment` from registry-service, cache by `(EnvironmentId, AccountId)` with a configurable TTL, then run `AuthCtx::authorize_environment_action` — the same rule as the remote path in `worker-service`
- Owner identity is sourced from `account_id_owning_environment` in `AuthDetailsForEnvironment`
- `NoOpDirectInvocationAuthService` provided for test environments without a live registry-service
- `check()` returns `EnvironmentOwnerAccountId`
- New `[direct_invocation_auth_cache]` config block on `GolemConfig`

Fixed `WorkerServiceError::AuthError` → `agent_error::Unauthorized` (previously `InternalError`), completing the propagation chain: `AuthError → Unauthorized → WorkerProxyError::Unauthorized → RpcError::Denied`

**Worker limit accounting on the local RPC path**

`create_demand`'s local branch now calls `update_worker_limit(env_owner, agent_id, +1)` after `Worker::get_or_create_running`, mirroring exactly what `worker-service` does after `LaunchNewWorker` on the remote path.

Introduced `WorkerLimitService`, a narrow single-method trait similar to `LimitService` in golem-worker-service, so `DirectWorkerInvocationRpc` does not depend on the full `RegistryService`

**End-to-end integration tests**

Added `RpcAuthTester` agent to `test-components/agent-rpc/golem-it-agent-rpc-rust`. Its `try_call_counter(name) -> RpcCallOutcome` method captures the `RpcError` instead of panicking, returning a `Schema`-derived enum (`Ok`, `Denied`, `NotFound`, `ProtocolError`, `RemoteInternalError`) that integration tests can assert on via `Value::Variant { case_idx }`

Resolves:
* #2976 
* #3099

### DIFF
--- a/golem-registry-service/src/repo/plan.rs
+++ b/golem-registry-service/src/repo/plan.rs
@@ -121,7 +121,6 @@ impl PlanRepo for DbPlanRepo<PostgresPool> {
                             monthly_http_call_limit, monthly_rpc_call_limit,
                             oplog_writes_per_second
                         )
-<<<<<<< HEAD
                         VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18)
                         ON CONFLICT (plan_id) DO UPDATE SET
                             name = $2,
@@ -132,7 +131,6 @@ impl PlanRepo for DbPlanRepo<PostgresPool> {
                             total_app_count = $7,
                             total_env_count = $8,
                             total_component_count = $9,
-<<<<<<< HEAD
                             total_worker_connection_count = $10,
                             total_component_storage_bytes = $11,
                             monthly_gas_limit = $12,
--- a/golem-worker-executor/src/services/oplog/plugin.rs
+++ b/golem-worker-executor/src/services/oplog/plugin.rs
@@ -115,11 +115,14 @@ pub struct PerExecutorOplogProcessorPlugin<Ctx: WorkerCtx> {
     worker_proxy: Arc<dyn WorkerProxy>,
 }
 
+// TODO: Once cross-environment RPC is enabled via host.wit, the cache key should
+// use the component's canonical environment_id rather than the caller-provided one.
+// Currently all plugin workers live in the same environment as their callers so
+// the inconsistency is harmless — a cache miss just triggers a component lookup.
 type WorkerKey = (EnvironmentId, PluginRegistrationId);
 
 #[derive(Debug, Clone)]
 struct RunningPlugin {
-    pub account_id: AccountId,
     pub owned_agent_id: OwnedAgentId,
     pub configuration: BTreeMap<String, String>,
     pub component_revision: ComponentRevision,
@@ -164,16 +167,16 @@ impl<Ctx: WorkerCtx> PerExecutorOplogProcessorPlugin<Ctx> {
                             ))?;
 
                         let agent_id = self.generate_agent_id_for(plugin_component_id).await?;
+                        // Resolve the component's authoritative environment_id.
                         let plugin_component = self
                             .component_service
                             .get_metadata(plugin_component_id, Some(plugin_component_revision))
                             .await?;
                         let owned_agent_id = OwnedAgentId {
-                            environment_id,
+                            environment_id: plugin_component.environment_id,
                             agent_id: agent_id.clone(),
                         };
                         let running_plugin = RunningPlugin {
-                            account_id: plugin_component.account_id,
                             owned_agent_id: owned_agent_id.clone(),
                             configuration: plugin.parameters.clone(),
                             component_revision: plugin_component_revision,


sdks/scala/wit/deps/golem-websocket/websocket.wit
```diff
@@ -0,0 +1,51 @@
+package golem:websocket@1.5.0;
```
kmatasfp, line 1: was missing, added here by `cargo make wit`

sdks/scala/wit/deps/golem-1.x/golem-oplog.wit
```diff
@@ -367,6 +367,9 @@ interface oplog {
         out-of-memory,
         exceeded-memory-limit,
         internal-error(string),
+        deterministic-trap(string),
```
kmatasfp, line 434: updated by `cargo make wit`

golem-worker-executor/src/services/golem_config.rs
```diff
@@ -1132,6 +1141,38 @@ impl SafeDisplay for EnvironmentStateServiceConfig {
     }
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct DirectInvocationAuthCacheConfig {
+    pub cache_capacity: usize,
+    pub cache_ttl: Duration,
```
vigoo, line 1109: let's use humantime format for this too

golem-worker-executor/src/services/rpc.rs
```diff
@@ -755,6 +775,12 @@ impl<Ctx: WorkerCtx> Rpc for DirectWorkerInvocationRpc<Ctx> {
             )
             .await?;
 
+            // Mirror the remote path: worker-service calls update_worker_limit(+1) after
+            // create regardless of whether the worker was newly created or already existed.
+            self.worker_limit_service
```
vigoo: First, what does it mean "always +1 the worker limit regardless of the worker already existed"? Maybe I misunderstand but if that means that every rpc call to an existing agent counts as "creating a new one" that is a very bad behavior and we must fix it.

A second concern that this somewhat eliminates the perf gain we get from direct invocation by always making a registry service request. We should apply the same technique as for fuel limits, and do batched updates to keep the happy path as fast as possible.

kmatasfp: Maybe I misunderstood this part as well, let me recheck how this works on remote path

mschuwalow: I do not think it's possible to even do rpc to an agent outside the current environment (as the only way to construct a wasm-rpc resource [is via an agent type name](https://github.com/golemcloud/golem/blob/main/wit/deps/golem-agent/host.wit#L49), which is always resolved within the environment).

Auth within the same environment will always allow the rpc, so the check does not add much value currently afaik.

mschuwalow:
Looking at the tests the enforcement here is about who created the agent (i.e. an agent created by the environment owner might behave differently than an agent created by somebody the environment was shared). I do not think that is the right auth model for us and behaves very differently than all other entities we have so far.

The model that we used so far is that permissions / shares are enforced when you are crossing an environment boundary (i.e. you create an agent in an environment that was shared with you). Within an environment all entities behave the same way.

With this change we would start attaching permissions to a particular agent determined by who created the agent. In addition to making the auth model very complex for little benefit, it also breaks our upsert semantics to some degree. I should not care whether an agent exists or not before invoking it -- now I would need to make sure the agent was created by the environment owner, as it otherwise would be lacking permissions.

golem-worker-executor/src/services/rpc.rs
kmatasfp: this code is only called when rpc resource is initally created after`CreateWorker` permission check, so not during actual invoke. Actual check is in direct_invocation_auth.rs @mschuwalow . And yes you are right that currently our users cannot do cross environment calls, `WasmRpc::new(...)` does not allow this, I also pointed this out in the integration tests, but the actual rpc rust code allows cross env calls, so I just mirrored what the remote rpc call was already doing for the direct call. 

@vigoo +1 here means that this counter is always incremented even if agent was already running and we did not need to create it  before we constructed the rpc "resource/connection",  behaves the same way when we do remote rpc. So we probably need to check if agent is actually running before incrementing this

kmatasfp:
@mschuwalow direct path here is mirroring the existing remote auth model, not introducing per-target-agent creator permissions

Both remote and direct rpc call behave the same:
- Authorization is evaluated against the target environment (`environment_id`) and caller account for the relevant action (`CreateWorker` for ensure/start, `UpdateWorker` for invoke).
- It uses environment owner + shares/roles, same as other entities.
- The target agent’s `created_by` is not used in the authorization decision.

Remote rpc:
- `invoke_agent` authorizes `EnvironmentAction::UpdateWorker` on target env in `golem-worker-service/src/service/worker/service.rs:732`.
- `create_with_component` (launch/ensure) authorizes `EnvironmentAction::CreateWorker` in `golem-worker-service/src/service/worker/service.rs:120`.

Direct rpc:
- local `create_demand` checks `CreateWorker` in `golem-worker-executor/src/services/rpc.rs:753`.
- local `invoke_and_await` / `invoke` checks `UpdateWorker` in `golem-worker-executor/src/services/rpc.rs:820` and `golem-worker-executor/src/services/rpc.rs:897`.

Let me see if I can make the tests clearer do avoid any confusion

golem-worker-executor/src/services/rpc.rs
kmatasfp: improved this a bit, now we only increment if worker was actually created, direct path is best effort, still can have a race issue here, making it 100% correct requires a followup ticket

vigoo: I think @kmatasfp is right about the new code mirrors what happens in the remote case in the direct invocation case, so that looks correct to me (even if in practice it is always going to be allowed because of not being able to do cross-environment calls which is a known and accepted limitation).

I'm still not happy about the worker counting and I don't think we should merge an intermediate version of it.

### Tracking issues
- Checking the worker's existence in advance is incorrect (best-effort, as you also mention). Even the newly introduced `created` result flag is implemented the same way, so both the remote and the direct invocation paths are incorrect. 
- Actually the same problem exists for deleting workers too, it is not checking if the worker was actually deleted just decreases the counter always

For these I think the proper solution is that _only_ worker executor tracks these numbers and fully integrated into the agent upsert implementation so it's not a best effort but the truth. 

### Performance
It still (currently, and also if it would be done as what I propose in the previous section) introduces one more worker service call per creation which I believe should be mitigated by a batched update as I recommended above.

### Conceptually
I am not sure now that this limit is conceptually a good thing to have. Especially as we also count ephemeral agents, and we encourage using many small agents for modelling the app domain and ephemeral ones for implementing APIs etc. Having a hard limit (10000 by default now) on the number of agents seems very limiting with not having any automatic way to clean them up.

I think it would be a bit better if we would not count ephemeral agents at all into this counter, but even then, I don't like if the user's app reaches this hard limit it's going to crash (with limit exceeded rpc errors, for example). At that point even if they reach out to us and ask/pay for bigger limit, their app will collapse because of these limit errors, and there is no easy (manual or automatic) way to delete unused old agents or something like that.

### Conclusion
For these reasons I think we should completely remove this particular limit for 1.5 (including the broken counting logic). @jdegoes what do you think?

kmatasfp:
@vigoo Regarding RPC-call optimization, with the current direct-RPC implementation we have:

- Owner fast-path hit, agent already exists -> **0 RPC calls**
- Owner fast-path hit, agent newly created -> **1 RPC call** (`update_worker_limit` only)
- Auth cache hit (non-owner), agent already exists -> **0 RPC calls**
- Auth cache hit (non-owner), agent newly created -> **1 RPC call** (`update_worker_limit` only)
- Auth cache miss, agent already exists -> **1 RPC call** (`get_auth_details_for_environment`)
- Auth cache miss, agent newly created -> **2 RPC calls** (`get_auth_details_for_environment` + `update_worker_limit`)

If we remove limit reporting from this path, all “newly created” rows above drop by 1.

To make sure I understood your suggestion correctly: were you proposing
1) batching/deferring `update_worker_limit` calls (e.g. flush every 60s),
2) proactively refreshing auth cache entries (e.g. every 60s to keep hot keys warm),
3) both?


**Other thing I want to point out here so it is documented**

When account A's worker invokes an agent that belongs to account B for the **first time** (lazy creation via RPC), the newly created worker gets `created_by = A` and `environment_id = A's environment`. 
This means:
- All storage metrics (`golem_storage_bytes_written_total`, etc.) for that worker will be attributed to **account A**, not account B
- Storage quotas and other limits will be counted against account A

This is a pre-existing issue in how `OwnedAgentId` is constructed during cross-environment RPC in `wasm_rpc/mod.rs` — the caller's `environment_id` is inherited by the target worker rather than resolving the target's own environment.

This will be fixed by #3099.

vigoo:
Ok let's remove completely the tracking and limiting of total number of agents, we don't want that. The new limits like maximum parallel agents are much better.

kmatasfp:
Direct-path RPC counts after this cleanup:
- Owner fast-path hit, already exists -> 0
- Owner fast-path hit, newly created -> 0
- Auth cache hit (non-owner), already exists -> 0
- Auth cache hit (non-owner), newly created -> 0
- Auth cold miss, already exists -> 1 (get_auth_details_for_environment)
- Auth cold miss, newly created -> 1 (get_auth_details_for_environment)

golem-worker-service/src/service/worker/connect.rs
```diff
@@ -54,28 +37,3 @@ impl Stream for ConnectWorkerStream {
...
-            account_id = %self.account_id,
-            "Dropping worker {} connections",
-            self.agent_id
-        );
-        let limit_service = self.limit_service.clone();
-        let agent_id = self.agent_id.clone();
-        let account_id = self.account_id;
-        tokio::spawn(async move {
-            let result = limit_service
-                .update_worker_connection_limit(account_id, &agent_id, false)
```
vigoo, line 70: I think we should not have removed this limit (number of parallel webstream connections to agents) just the "total number of agents" one

vigoo:
- The websocket connection limit does not have to be removed
- The total agent count limit should be completely removed (from the registry service layer too)

golem-worker-executor/src/grpc/mod.rs
```diff
@@ -209,6 +209,25 @@ impl<Ctx: WorkerCtx, Svcs: HasAll<Ctx> + UsesAllDeps<Ctx = Ctx> + Send + Sync +
         self.shard_service().check_worker(agent_id.as_ref())
     }
 
+    /// Validate that the request carries a well-formed `auth_ctx` field.
+    ///
+    /// TODO: Now that `account_id` is resolved from the component owner
```
kmatasfp, line 232: do we still need this?

golem-worker-executor/src/services/worker_fork.rs
```diff
@@ -519,6 +518,12 @@ impl<Ctx: WorkerCtx> DefaultWorkerFork<Ctx> {
 
         let initial_source_worker_metadata = source_worker_instance.get_initial_worker_metadata();
 
+        // TODO: Should `created_by` use `component.account_id` (the component
+        // owner) instead of `fork_account_id` (the caller)?  The same issue
```
kmatasfp: this potentially has the same bug as described here: #3099

kmatasfp:
Now also contains fix for #3099

golem-worker-service/src/service/worker/connect.rs
kmatasfp: restored

golem-worker-executor/src/grpc/mod.rs
vigoo: even if we don't use at the moment I think it's important to propagate who the caller is through all service calls, so keep it

golem-worker-executor/src/services/worker_fork.rs
vigoo: and there is a consistency bug already in it - the `AgentMetadata` is set to he `created_by: fork_account_id` but the `created_by` field is not set to that in the `target_initial_oplog_entry`. This inconsistency should be fixed, and I think you are right that the fix can be that the `target_worker_metadata`'s `created_by` is set to the same account-id as the source worker's.

vigoo (approved):
The merging of the semantic retries PR caused some merge conflicts and I'd like that single created-by changed in the fork before this gets merged, otherwise looks good!

PR #3114, opened 2026-04-02
quotas: worker-executor host functions

### LINKED ISSUE #3023: Agent quota intent & allocation
(Note: we might want to split desired_rate into request + limit to make it more flexible)
A quota token declares the intent of an agent to use a certain amount of a resource. 

It is purely a (resource, desired_rate) and acts in the following ways:

- As a hard, **local** rate limit. I.e. if have a desired rate of 10 rps, I cannot use more even if more is available.
- As a hint to the executor how much of the resource the agent is actually going to use.

When an agent creates a token, it should be registered in the executor that owns the token. At this point some host level state is created that tracks the credit that the token has for a given resource. This credit is generally used to decide which agents gets to use the next available resource (highest credit wins and refill rate is controlled by desired_rate).

Whenever an agent tries to use a quota token, that should go through the executor local scheduler. This should be event-based and fine grained. I.e.
`reservation = token.reserve(n)`

Should behave internally as

`loop n times:`

`wait until scheduler grants 1 unit` 

With the scheduler giving the next unit to the token with the highest credit.
Notes:

- Only tokens actively competing for a resource should be considered. This prevents under-utilization
- Credit should have a cap in order to limit the size of bursts that an individual token can produce

mschuwalow:
resolves #3023

### DIFF
--- a/golem-worker-executor/src/durable_host/quota/mod.rs
+++ b/golem-worker-executor/src/durable_host/quota/mod.rs
@@ -16,7 +16,9 @@ pub mod types;
 
 use crate::durable_host::quota::types::{LeaseInterestHandle, QuotaTokenEntry, ReservationEntry};
 use crate::durable_host::{Durability, DurabilityHost, DurableWorkerCtx};
-use crate::preview2::golem::quota::host::{Host, HostQuotaToken, HostReservation, FailedReservation};
+use crate::preview2::golem::quota::host::{
+    FailedReservation, Host, HostQuotaToken, HostReservation,
+};
 use crate::services::quota::LeaseInterest;
 use crate::workerctx::WorkerCtx;
 use chrono::{DateTime, TimeDelta, TimeZone, Utc};
@@ -73,17 +75,16 @@ impl<Ctx: WorkerCtx> HostQuotaToken for DurableWorkerCtx<Ctx> {
         resource_name: String,
         expected_use: u64,
     ) -> anyhow::Result<Resource<QuotaTokenEntry>> {
-        DurabilityHost::observe_function_call(
-            self,
-            "golem::quota::quota-token",
-            "[constructor]quota-token",
-        );
-
         let env_id = self.owned_agent_id.environment_id;
         let rn = ResourceName(resource_name.clone());
-        let is_live = self.state.is_live();
 
-        let token_entry = if is_live {
+        let durability = Durability::<host_functions::GolemQuotaTokenNew>::new(
+            self,
+            DurableFunctionType::WriteLocal,
+        )
+        .await?;
+
+        let token_entry = if durability.is_live() {
             let svc = self.state.quota_service.clone();
             let interest = svc
                 .acquire(env_id, rn, expected_use, 0, None)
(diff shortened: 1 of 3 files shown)


golem-worker-executor/src/durable_host/quota/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+    }
+}
+
+impl<Ctx: WorkerCtx> HostQuotaToken for DurableWorkerCtx<Ctx> {
+    async fn new(
+        &mut self,
+        resource_name: String,
+        expected_use: u64,
+    ) -> anyhow::Result<Resource<QuotaTokenEntry>> {
+        DurabilityHost::observe_function_call(
```
vigoo: This is redundant for durable calls, it's just a small helper for those that are _not_ using `Durability::new`.

golem-worker-executor/src/durable_host/quota/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+        let token_entry = if is_live {
+            let svc = self.state.quota_service.clone();
+            let interest = svc
+                .acquire(env_id, rn, expected_use, 0, None)
+                .await
+                .map_err(|e| anyhow::anyhow!("quota acquire failed: {e}"))?;
+
+            let credit_at_ms = interest.last_credit_value_at.timestamp_millis();
+
+            let durability = Durability::<host_functions::GolemQuotaTokenNew>::new(
```
vigoo: Let's follow the general durable host pattern:

- `Durability::new` first, and only once
- then `durability.is_live()` check
- in one branch `persist`, the other branch `replay`

golem-worker-executor/src/durable_host/quota/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+    async fn drop(&mut self, rep: Resource<QuotaTokenEntry>) -> anyhow::Result<()> {
+        DurabilityHost::observe_function_call(self, "golem::quota::quota-token", "drop");
+        self.table().delete(rep)?;
+        Ok(())
+    }
+}
+
+impl<Ctx: WorkerCtx> HostReservation for DurableWorkerCtx<Ctx> {
+    async fn commit(&mut self, self_: Resource<ReservationEntry>, used: u64) -> anyhow::Result<()> {
+        DurabilityHost::observe_function_call(self, "golem::quota::reservation", "commit");
```
vigoo: Same as above

golem-worker-executor/src/durable_host/quota/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+            ReservationEntry::Live { reservation, token } => {
+                let token_resource: Resource<QuotaTokenEntry> = Resource::new_own(token.rep());
+                let svc = self.state.quota_service.clone();
+
+                let interest = get_live_lease_interest(self, &token_resource).await?;
+                svc.commit(interest, reservation, used).await;
+                let credit_after = interest.last_credit_value;
+                let credit_after_at_ms = interest.last_credit_value_at.timestamp_millis();
+
+                let durability = Durability::<host_functions::GolemQuotaReservationCommit>::new(
```
vigoo: I'd like it better to have the common pattern here as well (it was even a single call with higher order functions originally, it was just too painful to use in rust): 
durability::new first, is_live check, it can panic on illegal state (durability.is_live() + ReservationEntry::Replay etc)

Issue #2975, opened 2026-03-12
Verify and extend storage related metrics


vigoo:
We should have metrics tracking per-account and per-environment storage in oplog, blob-store, component-store, etc.

PR #3117, opened 2026-04-03
Verify and extend storage related metrics

kmatasfp:
Resolves: #2975


golem-worker-executor/src/services/oplog/primary.rs
```diff
@@ -540,6 +556,32 @@ impl PrimaryOplogState {
                     self.key
                 )
             });
+
+        if entry_count > 0 {
+            // Estimate serialized size from pairs_ref
+            let serialized_bytes: u64 = pairs_ref
+                .iter()
+                .map(|(_, entry)| {
+                    golem_common::serialization::serialize(entry)
```
vigoo: We must not double-serialize, this is on the hot path. I think the correct way to do this is to do it in the indexed storage implementation itself (it can be a wrapper layer so it does not have to be implemented for each type). The `account_id` and `environment_id` can be added to `IndexedStorageNamespace::OpLog` so it has all information needed

golem-worker-executor/src/services/file_loader.rs
```diff
@@ -355,18 +363,56 @@ struct InitializedCacheEntry {
...
+            Ok(()) => {
+                // File deleted successfully — disk space is freed, return permits to the pool.
+                // SAFETY: we are in Drop, this value will not be used again.
+                let permit = unsafe { ManuallyDrop::take(&mut self.filesystem_storage_permit) };
+                if permit.is_some() && self.filesystem_storage_permit_bytes > 0 {
+                    record_filesystem_pool_released(self.filesystem_storage_permit_bytes);
+                }
+                drop(permit);
+            }
+            Err(e) => {
```
vigoo: I don't think it's correct to ignore these errors and just never return the permits. If we cannot delete these files then something is broken and we should just crash the server.

golem-worker-executor/src/services/file_loader.rs
```diff
@@ -355,18 +363,56 @@ struct InitializedCacheEntry {
...
 impl Drop for InitializedCacheEntry {
     fn drop(&mut self) {
         debug!("Removing file {}", self.path.display());
-        if let Err(e) = std::fs::remove_file(&self.path) {
-            tracing::error!("Failed to remove file {}: {}", self.path.display(), e);
+        match std::fs::remove_file(&self.path) {
+            Ok(()) => {
+                // File deleted successfully — disk space is freed, return permits to the pool.
+                // SAFETY: we are in Drop, this value will not be used again.
+                let permit = unsafe { ManuallyDrop::take(&mut self.filesystem_storage_permit) };
```
vigoo: Regardless of the comment below, we don't need `unsafe` for this - you can just `take` the permit, and if you don't want it to be dropped, you can call `std::mem::forget()` on it - but as I wrote below, I don't think we should do this

golem-worker-executor/src/services/oplog/blob.rs
```diff
@@ -91,7 +101,7 @@ impl OplogArchiveService for BlobOplogArchiveService {
         idx: OplogIndex,
         n: u64,
     ) -> BTreeMap<OplogIndex, OplogEntry> {
-        let archive = self.open(owned_agent_id).await;
+        let archive = self.open(owned_agent_id, &AccountId::SYSTEM).await;
```
vigoo: This (and the other occurrence) of `AccountId::SYSTEM` should be either replaced with a real one, or explained in detail why is it system (and even then, probably it would make sense to pass it explicitly at the call sites, so if we introduce a new one we can pass a real account if we can)

golem-worker-executor/src/services/active_workers/fs_semaphore.rs
```diff
@@ -80,6 +82,7 @@ impl FilesystemStorageSemaphore {
                         self.semaphore.available_permits(),
                         permit.num_permits()
                     );
+                    record_filesystem_pool_acquired(storage_bytes);
```
vigoo: We use the exact bytes here, but calculate the size to be released from the permits, which if I understand correctly, is rounded.

golem-worker-executor/src/services/oplog/compressed.rs
vigoo, line 344: Should we have a `record_storage_objects_deleted` call here too?

golem-service-base/src/service/compiled_component.rs
```diff
@@ -192,7 +197,13 @@ impl CompiledComponentService for DefaultCompiledComponentService {
                     component_revision,
                     format!("Could not store compiled component: {err}"),
                 )
-            })
+            });
+        if result.is_ok() {
```
vigoo, line 201: Shouldn't use empty string as account-id here

golem-worker-executor/src/services/oplog/compressed.rs
kmatasfp: `InstrumentedOplogArchive` layer handles this now

golem-worker-executor/src/services/oplog/blob.rs
kmatasfp: done

golem-worker-executor/src/services/oplog/primary.rs
kmatasfp: done

golem-worker-executor/src/services/file_loader.rs
kmatasfp: Yeah better to crash here, this means fs is in inconsistent state

golem-worker-executor/src/services/active_workers/fs_semaphore.rs
kmatasfp: done

golem-service-base/src/service/compiled_component.rs
kmatasfp: done

PR #3121, opened 2026-04-03
Fix oidc for safari

### LINKED ISSUE #1395: Make sure OpenID security scheme auth flow works in safari browser

vigoo:
Fixes #1395


golem-worker-service/src/custom_api/oidc/handler.rs
```diff
@@ -120,10 +120,12 @@ impl OidcHandler {
             .store_authenticated_session(&session_id, session)
             .await?;
 
+        let is_https = scheme.redirect_url.starts_with("https://");
```
noise64: i think this should use `scheme.redirect_url.url().scheme() == "https"` to avoid tricky cases altogether

PR #3147, opened 2026-04-10
quotas: rpc

### LINKED ISSUE #3024: Support for splitting and merging tokens
A quota token is just a (resource, desired_rate).

This makes it naturally linear, splitting the token is just dividing the desired_rate between the children and merging is summing them back together. There is no global shard-manager state attached to the tokens that needs to be considered.

On the level of the executor we can decide what to do with the outstanding credit when splitting / merging. One good option is to define max_credit as a fraction of the requested amount of a token and split/merge both outstanding and max credit according to their relative amounts.

Actually sending tokens to a different agent should be done as a new case of ElementSchema/ElementValue on the wit level and with a new built-in type in the sdks.

When an agent sends a token to another agent there are two cases:

- Both agents live on the same executor: No work for the executor. Tokens change ownership, but still refer to the same entity in the local scheduler
- The agent lives on another executor: We need to remove the local entry from the scheduler and reconstruct it on the receiving side. This behaves exactly like creating a new quota-token on the new executor, except that we want to carry over the remaining credit to the next executor.

mschuwalow:
resolves #3024


sdks/scala/wit/dts/golem_quota_1_5_0_types.d.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Host interface for the Golem quota system.
+ * Agents use quota-tokens to declare intent to consume a named resource and to
+ * reserve / commit actual usage.
+ */
+declare module 'golem:quota/types@1.5.0' {
```
vigoo, line 6: Delete the old `_host.d.ts`

sdks/scala/core/js/src/main/scala/golem/host/QuotaApi.scala
```diff
@@ -143,19 +191,60 @@ object QuotaApi {
...
+        override def encode(value: QuotaToken)          = recordSchema.encode(value.toRecord())
+        override def decode(value: golem.data.StructuredValue)   = recordSchema.decode(value).map(fromRecord)
+        override def encodeElement(value: QuotaToken)   = recordSchema.encodeElement(value.toRecord())
+        override def decodeElement(value: golem.data.ElementValue) = recordSchema.decodeElement(value).map(fromRecord)
+      }
+    }
   }
 
   // --- WIT constructor import ---
   // `quota-token` has a WIT constructor, which jco maps to a JS class.
```
vigoo: We are not using `jco`!! This confuses readers (human and agent).  I know it's from a previous PR, but please let's get rid of any mention of jco :)

vigoo (approved):
Asking for two small cleanups otherwise looks good!

PR #3151, opened 2026-04-10
Skill harness fixes and improvements 1

vigoo:
A large batch of skill test harness improvements

- Always use the compiled golem and SDKs from the current repository
- Integrates the skills into the golem templates
- The test harness manages the golem server’s lifecycle
- Isolated workspcae for each test case
- Repo level skill for adding skills to the golem cli
- Tons of logging improvements
- Not failing on additional activated skills if there is no explicit allow list
- AmpCode driver (for local testing, not enabled on CI)
- Integrated to the main CI flow
- Skill catalog in the per-language agents.md files
- Real `golem-create-project`, `golem-add-npm-package`, `golem-add-rust-crate`, `golem-add-scala-dependency` skills with real test scenarios
- Create project scenario step 
- Lint/format
- Idle timeout and retry of flaky tests
- Test OpenCode with multiple models (gemini is often timing out, seems to be a known issue)


Nanashi-lab:
`AGENTS.md` is not a universally autoloaded. 

For example, Claude Code does not autoload `AGENTS.md` and only loads`CLAUDE.md`. Similarly, tooling conventions differ for skills, Claude Code expects skills under `.claude/skills` and also requires `commands` to be added, OpenCode uses `.opencode/skills`, and Codex uses `.codex`. 

There are also inconsistencies in terminology—Claude uses “commands” and “skills,” while others refer to “workflows” or "prompts" and “skills.”

A potential solution is to incorporate this into `golem new` command and allow the user to interactively choose which (multiple choice) coding agent they are using. 

Something like below image. 

Screenshot from 2026-04-11 19-39-46

vigoo:
> `AGENTS.md` is not a universally autoloaded.

We want to avoid having such an interactive choice if possible, that's why this skill harness is verifying with multiple agents and models that the skills are used and the tasks are performed properly (much more scenarios coming). Of course we may run into the issues you mention, but I'd rather have something that "just works" than having to choose - especially because that it is not guaranteed that everyone working on the same golem agent codebase is using the same set of tools.

vigoo:
Merging this now, and continue with more scenarios and skills in a separate PR

PR #3154, opened 2026-04-11
Fix concurrent agent limiting

vigoo:
It was not retrying to evict idle agents so could wait forever to get some permits
Also updated to the fresh published golem-wasm in the rust sdk so now it will be publishable


noise64 (approved):

vigoo:
I'm actually no longer sure this is the right fix :) maybe the idle agents are not giving back their permits. I'll check again tomorrow. *DO NOT MERGE*

PR #3158, opened 2026-04-11
fix: release concurrent agent permit when agent goes idle

kmatasfp:
Idle agents should release `ConcurrentAgentSemaphore` permits


vigoo (approved):

vigoo:
Merged this as a quick fix but a follow-up PR will introduce better scheduling

PR #3164, opened 2026-04-12
Better concurrent agent scheduling

vigoo:
- Introducing per-account FIFO scheduling to improve fairness. 
- Memory/fs eviction now prefers idle agents but is also capable of evicting agents with non-empty pending queue to avoid starvation


golem-worker-executor/src/services/active_workers/concurrent_agents_scheduler.rs
```diff
@@ -0,0 +1,383 @@
...
+        // from Worker::new before any acquire), but handle gracefully.
+        let raw_semaphore = Arc::new(tokio::sync::Semaphore::new(0));
+        let resource_entry = Arc::new(AtomicResourceEntry::new(
+            u64::MAX,
+            usize::MAX,
+            usize::MAX,
+            u64::MAX,
+            AtomicResourceEntry::UNLIMITED_CONCURRENT_AGENTS,
+        ));
+        self.accounts
```
kmatasfp, line 267: I think we need to also create a corresponding state in underlying `ConcurrentAgentsSemaphore` otherwise I think `aquire` will wait forever.

I think better maybe we blow up with `expect` here as we reached impossible state, or fully register both scheduler and semaphore layers in slow path

golem-worker-executor/src/services/active_workers/tests.rs
```diff
@@ -512,3 +515,174 @@ async fn concurrent_agents_plan_upgrade_does_not_over_add_under_parallel_acquire
...
+
+    // C and D queue.
+    let sched2 = sched.clone();
+    let acc2 = acc;
+    let c_handle = tokio::spawn(async move { sched2.acquire(acc2, agent("C")).await });
+    let sched3 = sched.clone();
+    let d_handle = tokio::spawn(async move { sched3.acquire(acc, agent("D")).await });
+
+    // Give time for C and D to be queued.
+    tokio::time::sleep(Duration::from_millis(50)).await;
```
kmatasfp: probably source of flakyness, maybe better to use  poll `queue_len()` / `running_count()` until expected state or maybe add some barriers/gates like I introduced here - `concurrent_agent_limit_waits_for_running_agent_to_finish`

golem-worker-executor/src/services/active_workers/concurrent_agents_scheduler.rs
```diff
@@ -0,0 +1,383 @@
...
+/// 2. A worker that finishes and re-requests a slot goes to the back of the
+///    queue.
+/// 3. Dropping the [`ConcurrentAgentPermit`] notifies the scheduler to wake the
+///    next queued agent — fully synchronously, no spawned tasks.
+pub struct ConcurrentAgentsScheduler {
+    permits: Arc<ConcurrentAgentsSemaphore>,
+    accounts: scc::HashMap<AccountId, Arc<AccountScheduler>>,
+}
+
+struct AccountScheduler {
```
kmatasfp, line 37: might be worth changing this to:
```
struct AccountScheduler {
    resource_entry: Arc<AtomicResourceEntry>,
    raw_semaphore: Arc<tokio::sync::Semaphore>,
    
   // Fast-path bookkeeping only
    running_count: AtomicUsize,
    waiter_count: AtomicUsize,
    
    // FIFO fairness structure
    ready_queue: parking_lot::Mutex<VecDeque<QueuedAgent>>,
}
```

so mutex only needed when modifying the queue, for fast path can just check the atomics. also note usage of parking_lot crate Mutex, which is fairer than std mutex.

`acquire()` fast path will be something like this then, no locking needed:
```
 // Fast path: no visible queued waiters, try direct permit acquire.
    if account.waiter_count.load(Ordering::Acquire) == 0 {
        match account.raw_semaphore.clone().try_acquire_owned() {
            Ok(raw) => {
                account.running_count.fetch_add(1, Ordering::AcqRel);
                // Fairness recheck: if a waiter appeared concurrently, this
                // worker must not bypass FIFO ordering.
                if account.waiter_count.load(Ordering::Acquire) == 0 {
                    debug!(
                        "ConcurrentAgentsScheduler: fast-path permit for {agent_id} in account {account_id}"
                    );
                    return ConcurrentAgentPermit {
                        raw: Some(raw),
                        account: Some(account),
                        account_id,
                    };
                }
                // Lost fairness race: release optimistic acquire and queue.
                account.running_count.fetch_sub(1, Ordering::AcqRel);
                drop(raw);
            }
            Err(tokio::sync::TryAcquireError::NoPermits) => {
                // Fall through to slow path.
            }
            Err(tokio::sync::TryAcquireError::Closed) => {
                panic!("concurrent agents semaphore for {account_id} has been closed");
            }
        }
    }
```

golem-worker-executor/src/services/active_workers/concurrent_agents_scheduler.rs
vigoo: Removed this fallback path completely, and also made the whole thing more type safe

golem-worker-executor/src/services/active_workers/tests.rs
vigoo: Yes, fixed

golem-worker-executor/src/services/active_workers/concurrent_agents_scheduler.rs
vigoo: I'd leave this for now as is

kmatasfp (approved):

PR #3172, opened 2026-04-13
ts-sdk: compiler driven type detection

### LINKED ISSUE #2829: Avoid matching on type "name" to identify Principal in TS SDK for constructor injection
This was done for constructor injection in Principal.

Rust also has this problem but only in the client generation side (not in agent definition side, because everything there works with Schema typeclass)

This was something I called out, but never became a ticket.
Not describing the solution here, but confirming there are ways out, but essentially delegating the check back to compiler internals through TS Morph but without bringing in any extra dependencies to the final bundle

mschuwalow:
resolves #2829

also cleans up our build / rollup configs a bit

Basic idea is that we generate a synthetic source file into the project (which is never saved), that import our sdk types. We can then use normal compiler isAssignable / symbol comparison checks to determine which type we are dealing with.


test-components/build-components.sh
```diff
@@ -144,7 +144,7 @@ build_rust_apps() {
       "$GOLEM_CLI" build --step check --yes
     elif [ "$clean_only" = false ]; then
       echo "Building $subdir..."
-      "$GOLEM_CLI" --preset release build --yes
+      "$GOLEM_CLI" --preset release build --yes --skip-check
```
mschuwalow, line 147: I don't use rustup, so the check was a bit annoying

sdks/ts/packages/golem-ts-repl/rollup.config.js
```diff
@@ -25,10 +25,7 @@ export default defineConfig([
         tsconfig: './tsconfig.json',
         useTsconfigDeclarationDir: true,
         tsconfigOverride: {
-          compilerOptions: {
-            declaration: true,
-            declarationDir: 'dist',
-          },
+          compilerOptions: { declaration: false },
```
vigoo, line 28: Why this change?

mschuwalow: we were emitting unused d.ts files into the dist folder. We are also generating a single index.d.ts, which is the only one actually getting used

sdks/ts/packages/golem-ts-bridge/package.json
```diff
@@ -18,14 +18,11 @@
   "scripts": {
     "lint": "eslint . --ext .ts,.tsx",
     "lint:fix": "eslint . --ext .ts,.tsx --fix",
-    "build": "tsc --noEmit && rollup -c",
-    "test": "tsc --noEmit",
+    "build": "rollup --failAfterWarnings -c",
```
noise64, line 21: Does this cover type errors properly? AFAIR we needed tsc for many kind of errors.

sdks/ts/packages/golem-ts-repl/package.json
noise64: now that you touching these, could you also please get rid of these tags?

sdks/ts/packages/golem-ts-bridge/package.json
mschuwalow: We are using rollup-plugin-typescript2 for typescript compilation, that one emits errors properly. Together with failAfterWarnings (which covers things like undeclared externals) this should cover everything we want the build to fail on.

```
This version is somewhat slower than the original, but it will print out TypeScript syntactic and semantic diagnostic messages (the main reason for using TypeScript after all).
```

e.g.
```ts
src/index.ts → dist/index.mjs...
[!] (plugin rpt2) RollupError: There was an error during the build:
  [plugin rpt2] src/agentConfig.ts:75:24 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'Type'.

75     WitType.fromTsType(type as unknown, undefined),
                          ~~~~~~~~~~~~~~~
```

sdks/ts/packages/golem-ts-typegen/src/bin/golem-typegen.ts
```diff
@@ -26,6 +26,11 @@ program
     'Exclude methods that override parent class methods',
     true,
   )
+  .option(
```
noise64: Is this only used for tessting insixe the sdk?

mschuwalow: yes. I hid the argument for now.

My initial plan was to move all tests to a third package so that everything can use the normal @golemcloud/golem-ts-sdk import, but that ended up being too many changes. Maybe in a followup later.

sdks/ts/packages/golem-ts-bridge/package.json
noise64: sounds good, just asked, because before we had issues even with that plugin, if it works then i'm happy!

noise64: and a related question, are our tests still typechecked?

noise64 (approved):
added some comments on type checking, otherwise looks good

sdks/ts/packages/golem-ts-bridge/package.json
mschuwalow: They are included in the typescript project, so they are getting transpiled together with the other sources.

(This is something we could revisit, in which case we should ensure that vitest fails on the compilation errors or run tsc separately)

PR #3175, opened 2026-04-14
Flaky RPC test case investigation and fix

vigoo:
Fixes the flaky `agent_await_parallel_rpc_calls`

The problem was that about two weeks ago, while implementing cancellation for RPC, I introduced a forced deletion of child pollables when the `FutureInvokeResult` resource is dropped (because otherwise things were randomly failing due to nondeterministic drop order caused by JS GC).

This "fix" was not a good fix, and caused this other flakyness where we occasionally deleted pollables in the host that were still registered in `wstd`'s internals.

The real fix in this PR makes sure that `FutureInvokeResult` is only dropped when all the child pollables have been dropped, no matter what the drop order is.


.github/workflows/publish-golem-scala.yaml
```diff
@@ -7,7 +7,7 @@ on:
 
 env:
   JDK_JAVA_OPTIONS: -XX:+PrintCommandLineFlags
-  WASM_RQUICKJS_VERSION: v0.2.1
+  WASM_RQUICKJS_VERSION: debug
```
noise64: is this intentional?

vigoo: the other :)

noise64 (approved):

PR #3176, opened 2026-04-14
ts-repl: do not bundle repl

mschuwalow:
We are using the repl as a normal dependency in the packages the cli is generating. Therefore, there is no point in bundling our transitive dependencies, as they will be installed by npm anyway.

(If we do bundle, the dependencies should be declared as devDependencies, not dependencies)

### DIFF
--- a/sdks/ts/packages/golem-ts-repl/rollup.config.js
+++ b/sdks/ts/packages/golem-ts-repl/rollup.config.js
@@ -1,11 +1,7 @@
 // rollup.config.mjs
-import resolve from '@rollup/plugin-node-resolve';
-import commonjs from '@rollup/plugin-commonjs';
 import typescript from 'rollup-plugin-typescript2';
 import dts from 'rollup-plugin-dts';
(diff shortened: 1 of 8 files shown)


sdks/ts/packages/golem-ts-repl/rollup.config.js
```diff
@@ -15,12 +11,18 @@ export default defineConfig([
...
     },
-    external,
+    external: [
+      'node:child_process',
+      'node:fs',
+      'node:process',
+      'node:repl',
+      'node:stream',
+      'node:util',
+      'picocolors',
```
noise64, line 21: why do we need normal deps as external?

mschuwalow: external for rollup means that they are not resolved/inlined during bundling.
This just means that they stay as normal libraries that are resolved from node_modules at runtime (and automatically added there as they are listed in dependencies in the package.json)

noise64 (approved):

PR #3178, opened 2026-04-14
Skill harness 3

vigoo:
Skill changes:
- Many tweaks to the existing skills

Scenarios (all passing with amp):
- `annotate-agent` scenario
- `add-http-endpoint` scenario
- `http-endpoint-parameters` scenario
- `add-cors` scenario
- `configure-api-domain` scenario

Fixes:
- Fixes TS and Scala SDKs to **NOT** close the atomic region in case of an exception
- Fixes the Scala SDK to not capture user exceptions and convert them to AgentErrors (that prevents retries)
- TS SDK was incorrectly setting `auth=false` on endpoints that are not explicitly enabling auth, when the whole agent was `auth=true`
- Server side fixes for our CORS implementation
- Scala SDK had up-to-date check issues (not regenerating some code when it should have)

Harness improvements:
- Tool based skill activation detection for amp, CC and OC
- Adds a new step type to the skill test harness to get agent metadata through the REST API
- Harness now takes per-language outgoing request bodies for http steps
- Harness now allows matching on response headers
- Harness fixes to make skill tracking consistent with agent sessions
- Harness validating regexes on dry-run


Myestery (approved):

golem-api-grpc/proto/golem/customapi/core.proto
```diff
@@ -188,6 +188,14 @@ message RouteBehaviour {
   message CorsPreflight {
     repeated string allowed_origins = 1;
     repeated golem.component.HttpMethod allowed_methods = 2;
+
+    message MethodPolicy {
+      golem.component.HttpMethod method = 1;
+      repeated string allowed_origins = 2;
+      repeated string allowed_headers = 3;
```
mschuwalow, line 192: I do not think allowed_headers is useful for us. There is no meaningful way for users to customize which headers are allowed in cors given our configuration structure, so we will just always allow all declared headers anyway. An easier way to achieve the same thing is to just echo back the Access-Control-Request-Headers during preflight

golem-worker-service/src/custom_api/cors.rs
```diff
@@ -17,56 +17,69 @@ use super::error::RequestHandlerError;
...
-        .any(|pattern| pattern.matches(origin));
-    if !origin_allowed {
-        return Ok(RouteExecutionResult {
-            status: StatusCode::FORBIDDEN,
-            headers: HashMap::new(),
-            body: ResponseBody::NoBody,
-        });
-    }
+    let mut headers = HashMap::new();
+    merge_vary_header(
```
mschuwalow, line 37: merging is not really useful here as the underlying response cannot contain any headers, but does not hurt to add

golem-worker-service/src/custom_api/cors.rs
```diff
@@ -17,56 +17,69 @@ use super::error::RequestHandlerError;
...
+        http::header::ACCESS_CONTROL_ALLOW_METHODS,
+        requested_method.to_string(),
+    );
+    if !requested_headers.is_empty() {
+        headers.insert(
+            http::header::ACCESS_CONTROL_ALLOW_HEADERS,
+            requested_headers.into_iter().collect::<Vec<_>>().join(", "),
+        );
+    }
+    headers.insert(
```
mschuwalow, line 79: After giving it some more thought, I don't think we should ever allow credentials on cors requests until we add a way for users to explicitly enable it on a route. I'm not familiar enough will all the different csrf exploits, but I think a safe default would be to follow spec design and have this be false for now.

golem-worker-service/src/custom_api/cors.rs
```diff
@@ -17,56 +17,69 @@ use super::error::RequestHandlerError;
...
-        })
-        .collect::<Result<Vec<_>, _>>()?
-        .join(", ");
+        .any(|pattern| pattern.matches(origin))
+    {
+        return Ok(forbidden_response(headers));
+    }
 
-    let mut headers = HashMap::new();
+    if !policy.allows_headers(&requested_headers) {
```
mschuwalow: I don't think this adds much value here. Any additional headers are just going to get ignored and the current code just adds all headers used as input anyway

golem-worker-service/src/custom_api/request_handler.rs
```diff
@@ -83,21 +82,19 @@ impl RequestHandler {
...
-
-        if let Some(short_circuit) =
+            short_circuit
+        } else if let Some(short_circuit) =
             apply_session_from_header_security_middleware(request, resolved_route)?
         {
-            return Ok(short_circuit);
-        }
-
-        let mut result = self.execute_route(request, resolved_route).await?;
```
mschuwalow, line 100: I prefer the previous style here to more easily add new middlewares

sdks/ts/packages/golem-ts-sdk/src/decorators/httpEndpoint.ts
```diff
@@ -192,7 +192,8 @@ export function endpoint(opts: EndpointDecoratorOptions) {
       ([headerName, variableName]) => ({ headerName, variableName }),
     );
 
-    const authDetails: AuthDetails = { required: opts.auth ?? false };
+    const authDetails: AuthDetails | undefined =
+      opts.auth !== undefined ? { required: opts.auth } : undefined;
```
mschuwalow:
```suggestion
      opts.auth != undefined ? { required: opts.auth } : undefined;
```

Lets prefer using == for null checks (which covers both null and undefined)

mschuwalow (approved):
minor comments, otherwise lgtm

PR #3184, opened 2026-04-15
Transient error handling and concurrency limit for postgres based indexed storage

vigoo:
The oplog layer was intentionally `.expect` the indexed storage operations because we wanted to the executor to stop running as soon as it stopped being able to access the oplog, to minimize effects of a split-brain scenario, for example.
While checking failed benchmark logs I noticed that with the switch to postgres-based indexed storage implementation, now we can easily overload the sqlx pool with many parallel agents, and that manifests in pool timeout errors that are also treated as such critical issue and crashes the executor.

We should not crash the executor just because it is overloaded; this PR introduces a few things to avoid that:
- error classification for the indexed storage implementations  (transient vs other)
- transient errors such as pool timeout are not causing an immediate panic, just a configurable inline retry
- there is a semaphore in the indexed storage now so parallel writes are controlled by that and not the sqlx layer that has a timeout. this also leaves a configurable amount of free pool slots for non-indexed-storage use


golem-worker-executor/src/services/oplog/compressed.rs
```diff
@@ -85,16 +138,17 @@ impl OplogArchiveService for CompressedOplogArchiveService {
...
-            .await
-            .unwrap_or_else(|err| {
-                panic!("failed to check if compressed oplog exists for worker {owned_agent_id} in indexed storage: {err}")
-            })
+        let is = self.indexed_storage.clone();
+        let agent_id = owned_agent_id.agent_id();
+        let level = self.level;
+        let key = Self::compressed_oplog_key(&owned_agent_id.agent_id);
+        retry_storage_op(&self.retry_config, "compressed_exists", &key, || {
+            let is = is.clone();
```
noise64, line 151: not that it matters much, but do we have to also clone on line 141?

noise64 (approved):

PR #3194, opened 2026-04-15
quota and config test flakyness fixes

mschuwalow:
both tests pass with non_flaky(100) locally


golem-test-framework/src/components/shard_manager/mod.rs
```diff
@@ -106,7 +106,6 @@ async fn env_vars(
             "GOLEM__REGISTRY_SERVICE__PORT",
             registry_service.grpc_port().to_string(),
         )
-        .with_str("GOLEM__QUOTA__LEASE_DURATION", "5s")
```
mschuwalow, line 109: This is to prevent leases being lost. The new setup uses normal duration but more aggressive renewal to make it more responsive during tests

golem-worker-executor/src/services/quota.rs
```diff
@@ -217,6 +221,86 @@ pub trait QuotaService: Send + Sync {
     async fn commit(&self, interest: &mut LeaseInterest, reservation: Reservation, used: u64);
 }
 
+pub struct LazyQuotaService {
```
mschuwalow, line 224: This is for the case of us using grpc_port 0

integration-tests/tests/quota.rs
```diff
@@ -451,15 +451,15 @@ async fn shared_quota_across_two_agents(
     user.start_agent(&component.id, agent_b.clone()).await?;
 
     let mut tasks = JoinSet::new();
-    for agent_id in [agent_a, agent_b] {
+    for (agent_id, expected_use) in [(agent_a, 10u64), (agent_b, 20u64)] {
```
mschuwalow, line 454: This is not a real fix for the underlying issue. The problem is that both agents can get scheduled in such a way that they try to acquire leases with nearly the same priority and get stuck there.

We probably also want to add a mechanism to the shard_manager that guarantees that at least one waiter makes progress, but I need to think about it some more (note that this only happens in quite specific scenarios, I think we are fine for now)

golem-worker-executor/src/services/quota.rs
vigoo: How is this related to the sharding? Registration to the shard manager. Was this related to the flaky test somehow or just an unrelated change?

golem-worker-executor/src/services/registry_event_subscriber.rs
```diff
@@ -117,6 +117,15 @@ impl RegistryInvalidationHandler for WorkerExecutorRegistryInvalidationHandler {
                     "Received resource definition changed event, ignoring"
                 );
             }
+            RegistryInvalidationEvent::AgentSecretChanged { environment_id, .. } => {
```
vigoo, line 120: `EnvironmentState` also has two other things beside secrets:
- `retry_policies`
- `agent_deployment_details`

We should hook events for these to the invalidation path too (definitely for the retry policies, they act the same way as secrets - not sure about the deployment details)

mschuwalow: this should already be the case, no?
https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/services/registry_event_subscriber.rs#L62-L106

vigoo: ah i should have checked. sorry

golem-worker-executor/src/services/quota.rs
mschuwalow: This is an unrelated change. The shard manager internally identifies executors by (IpAddr, Port) for lease owners, so that would misbehave for port 0 (as there could be multiple executors spawned, each getting a new port).

Optimally, I would like to move this to some kind of executor id later, but I think this is fine for now,

vigoo (approved):

PR #3204, opened 2026-04-16
CLI / manifest cleanups and fixes for 1.5

### LINKED ISSUE #3186: CLI tool check: support non-rustup based installations

### LINKED ISSUE #3187: CLI manifest model: add alternative implementation for the dropped "deny_unknown_fields"

noise64:
- resolves https://github.com/golemcloud/golem/issues/3186
- resolves https://github.com/golemcloud/golem/issues/3187
- fixes Makefiles TS SDK tasks: they should work even without globally installed pnpm
- manifest related changes:
  - after experimenting with many options for aligning app_raw and the schema (including proper unknown field handling), the current approach is:
    - in app_raw we manually inline the common field sets, then we use extractor methods to convert them into intermediate common structs (we need this, because serde flatten / deny_unknown / serde_ignore does not work together, and cause "surprising" behaviours)
    - in the JSON schema we also repeat such field sets "manually", which allows expressing properly that we do not allow extra other props (this was quite problematic with other approaches)
    - to prove that these are aligned, and the schema validates our docs, a new property based test was added
   - the above new structure revealed that our retry policy serde format results in such yaml, that is not expressable with a JSON schema (it uses YAML tags); it turned out that there are no really good derivable serde mode that is works in compatible ways with json, yaml, json schema, and also looks okay for DX; so i ended up with custom serialization for these, which mean we can keep the simple rust model, but also can have simple and lenient yaml and json formats
   - related to the above, switched to "UntypedJson" payload in the HTTP API (from  string), and added validations to the CLI
   
Some example of the new format (json and yaml):

```json
{
    "and": [
        true,
        {
            "propEq": {
                "property": "status",
                "value": "transient"
            }
        }
    ]
}
```

```yaml
and:
  - true
  - propEq:
      property: status
      value: transient
```

```json
{
    "andThen": [
        {
            "periodic": {
                "secs": 1,
                "nanos": 0
            }
        },
        {
            "countBox": {
                "maxRetries": 3,
                "inner": "never"
            }
        }
    ]
}
```

```yaml
andThen:
  - periodic:
      secs: 1
      nanos: 0
  - countBox:
      maxRetries: 3
      inner: never
```


vigoo (approved):
Is anything user-facing changes by this that might be already covered by skills? If yes let's list them and I'll adjust in the next batch of skill updates

noise64:
@vigoo the visibe change should be only around retry policy /predicate format, both in the manifest and in the CLI, but I think those do not have yet skills

PR #3209, opened 2026-04-17
move mcp deployments to a dedicated domain

mschuwalow:


golem-registry-service/src/services/domain_registration/mod.rs
```diff
@@ -252,24 +275,48 @@ impl DomainRegistrationService {
...
+}
+
+fn domain_matches_base(domain: &str, base: &str, allow_arbitrary_subdomains: bool) -> bool {
+    let escaped = regex::escape(base);
+    let pattern = if allow_arbitrary_subdomains {
+        format!("^([^\\.]+\\.)+{escaped}$")
+    } else {
+        format!("^[^\\.]+\\.{escaped}$")
+    };
+    Regex::new(&pattern).unwrap().is_match(domain)
```
kmatasfp: Nit: Not sure how big of a perf hit it is but regex compiled on each validation call, prob can use precompiled regex here

kmatasfp (approved):

Issue #3145, opened 2026-04-09
OpenAPI export without deployment


vigoo:
OpenAPI export for code-first routes without deployment as a CLI feature would be useful to integrate it in build flows.

Nanashi-lab:
Swagger UI command can also be re-added here. which opens the openapi, filters out (/openapi.json & /openapi.yaml) and displays the swagger ui (if deployed, with deployment included).

Nanashi-lab:
@vigoo I looked into this. Earlier `pre-1.3` this was simpler as `build` -> `deploy` -> `deploy http api` were 3 different steps, and it was possible to deploy the app without having to deploy the `http api`, and so we could generate openapi and then output it in client between step 2 and 3. (Step two is where compiled httpapi is generated)

Currently this is not the case, `build` -> `deploy` are the only two steps. Deploy is when the agent wit is generated. Currently the openapi generation code sits in `golem-worker-service-base`, there is a stub in client for openapi. Even if the openapi generation is moved to `golem-service-base`, we still need it to be "deployed" or we need a middle step similar to `pre-1.3`. Since generating compiled httpapi and deploying the http_api are the same step, we cannot generate the openapi. 


Do you have any suggestions on how to approach this ?

vigoo:
@Nanashi-lab this ticket needs more planning and just a reminder for a future milestone

PR #3212, opened 2026-04-18
Custom OpenAPI export endpoint

### LINKED ISSUE #3144: Make OpenAPI export configurable

Nanashi-lab:
/closes #3144 

- Added an optional `openapi_endpoint` field in `golem.yaml` for HTTP API deployments (similar to `webhooks_url`) so the generated OpenAPI export can be configured from the root paths to a custom prefix such as `/docs`, while keeping the existing /openapi.json and /openapi.yaml behavior when unset.
- Normalized prefix inputs for both `openapi_endpoint` and `webhooks_url` into canonical forms so equivalent values like docs, /docs, and /docs/ (Though not sure if this was the right call)
- Added another `openapi_endpoint` check, to make sure it does not have any `params` etc
- Moved OpenAPI `Json`/`Yaml` selection into route behavior metadata instead of inferring it from the literal request path
- Added unit/integration coverage for prefix validation, custom OpenAPI export routing, webhook normalization, and deployment update/clear behavior
- Also had some clarification question for https://github.com/golemcloud/golem/issues/3145, left as comment. Please help me out with that as well. Thank you


vigoo (approved):
Looks good to me but @noise64 please check too

mschuwalow (approved):
thanks

PR #3216, opened 2026-04-18
rust and ts sdk - add missing http methods

### LINKED ISSUE #3213: Rust sdk does not support all http method types
The match here is incomplete https://github.com/golemcloud/golem/blob/main/sdks/rust/golem-rust-macro/src/agentic/agent_definition_http_endpoint.rs#L58. Extend this to cover all http methods and add integration tests ensuring all methods are supported.

justcoon:
fixes: #3213


mschuwalow (approved):
thanks

PR #3217, opened 2026-04-20
feat: ephemeral agent perf improvements

kmatasfp:
## Fix: Eliminate blocking S3 operations from ephemeral agent hot path

Ephemeral agent invocations were blocking on S3 round-trips A single `ping` call took ~644ms; under 10 concurrent callers this ballooned to 13+ seconds.

Profiling the hot path revealed four categories of blocking S3 calls and and broken handling of concurrent calls:

**Phase 1 — Archive discovery on open (~300ms, repeated per blob layer)**
`BlobOplogArchive::new` called `exists` (HeadObject) + `list_dir` (ListObjectsV2) to pre-load the historical entry index. `get_last_index` duplicated these same calls to find the previous oplog offset.

**Phase 2 — Synchronous `append` on commit (~50ms × 2–4 writes per invocation)**
`EphemeralOplogState::commit` awaited the S3 `put` before returning, blocking the invocation result.

**Phase 3 — `resume_replay` reading old entries from S3 (~80ms × 3 reads)**
`prepare_instance` called `resume_replay` which fetched the previous `AgentInvocationStarted(initialize)` entries from S3 to replay the constructor on each new invocation.

**Phase 4 — Concurrent request serialization**
With `phantom_agent = false` (shared worker), after each invocation the WASM instance was torn down and the outer loop exited. Requests enqueued while the previous invocation was running were left stranded with no loop to process them, causing all but the first concurrent request to hang indefinitely. The outer loop now continues instead of breaking after each ephemeral invocation, creating a fresh WASM instance for the next pending request.

Per product spec: the ephemeral-agent oplog is written directly to the archive layer (S3) for **observability only** — it is never read back by the executor. Each invocation creates a fresh stateless WASM instance. There is no state to recover, no replay needed.

### Expected hot path after fix

```
open ephemeral oplog  →  0 blocking S3 calls
commit entries        →  0 blocking S3 calls (background task)
prepare_instance      →  0 S3 reads (no replay)
concurrent requests   →  all processed, no requests stranded
```

The only remaining S3 activity is the background `append` — which completes after the response is already returned to the caller.


golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2827,36 +2827,24 @@ impl<Ctx: WorkerCtx> ExternalOperations<Ctx> for DurableWorkerCtx<Ctx> {
...
+            store
+                .as_context_mut()
+                .data_mut()
+                .durable_ctx_mut()
+                .state
+                .replay_state
+                .switch_to_live()
+                .await;
 
-                // Appending a Restart marker
```
vigoo, line 2847: This restart marker is explicitly for marking the (rare) special case of ephemeral non-phantom agents restarting their oplog. We need to keep it

vigoo (commented):
After spending some time understanding both the PR and the slack messages triggering it, I realized that we are trying to solve the wrong problem here and while doing it introducing some new issues.

Ephemeral agents are designed to be always phantom agents. It does not make any sense to use them with non-phantom agents because they would not be able to run in parallel! That's what the original "discovery" shown. So why are we supporting non-phantom ephemeral agents at all? It was just a safety thing to make sure that even if this happens (for example due to two racing requests coming from a retry, pointing to the same agent id for some reason) we guarantee that there is no state preservation between the two. But this is absolutely not what we need to optimize for.

The normal operation is that each ephemeral agent invocation goes to a fresh instance. What the real issue is that in 1.4.2 we enforced this, both in the rib based api gateway and the rib repl, and we don't have this in the new version in all invocation paths:
- for http there is a custom phantom agent property but it is not automatic, leading to the misunderstanding you also had
- the mcp server implementation does not deal with this at all
- some of the rcp and bridge implementations only generate phantom constructors for ephemeral agents, some not

Even though some optimizations in this PR may make sense for the "real path" as well I don't feel them safe enough to include them, as I believe they can lead to corrupt behavior in extreme cases (for example, delayed background writing of the ephemeral oplog vs re-entry of the same instance due to a call-site retry, and can be made even worse with resharding events / crashes).

So I'll close this (and sorry for not properly reviewing it earlier)

PR #3227, opened 2026-04-21
Wasi config unification

### LINKED ISSUE #3188: Unify agent and wasi config

noise64:
Resolves https://github.com/golemcloud/golem/issues/3188


golem-worker-executor/tests/wasi.rs
```diff
@@ -3079,7 +3050,7 @@ async fn wasi_config_initial_worker_config(
 
         assert_eq!(
             result,
-            Value::Option(Some(Box::new(Value::String("v1".to_string()))))
+            Value::Option(Some(Box::new(Value::String("\"v1\"".to_string()))))
```
vigoo: I think this is a bit unfortunate :) and probably it would be better to not use json for primitives. But as our primary way of configuration is not through wasi:config anyway, we can leave it for now

vigoo (approved):

golem-worker-executor/tests/wasi.rs
noise64: oh, sorry, that is horrible, i'm fixing it for strings

mschuwalow: I think this is probably not the behavior we want. If we do want to expose our agent-config as wasi-config (which I still disagree with), imo we should either:

1. Only expose string types and expose them unqouted
2. Expose all types, but unquote string types and stringify everything else

noise64: As i wrote, i'm fixing it (https://github.com/golemcloud/golem/pull/3227/commits/f33eb298bb79271567bb1f5e7f354c2c58eb8a73)

As for the disagree part: we simply have no room for overloading config in multiple ways from DX / UX perspective.

mschuwalow (approved):

PR #3228, opened 2026-04-21
Automatically use phantom agents when an agent type is ephemeral

### LINKED ISSUE #3225: Fix (restore) automatic phantom-agent behavior for ephemeral agents
See https://github.com/golemcloud/golem/pull/3217#pullrequestreview-4146568669

vigoo:
Resolves #3225 

Ephemeral agents were always meant to be phantom as well. In Golem 1.4.2 this was enforced in the rib-based api gateway, the REPL and the generated wit-based RPC. 

This PR fixes this in Golem 1.5 by making it enforced in various levels:

- Ephemeral RPC clients only have `getPhantom/newPhantom` variants (Scala already had this)
- Ephemeral bridge clients only have `getPhantom/newPhantom` variants (TS bridge already had this)
- MCP server uses phantom agents for ephemeral agent types
- HTTP server uses phantom agents for ephemeral agent types, the custom `phantomAgent` property on the mount is left to enforce phantom agent behavior on _durable agents_
- The HTTP agent create/invoke APIs enforce this behavior as well
- The gRPC APIs are not touched, as it is expected that internal calls are always passing the right agent-id

### DIFF
--- a/openapi/golem-service.yaml
+++ b/openapi/golem-service.yaml
@@ -9992,11 +9992,15 @@ components:
       title: AgentInvocationResult
       type: object
       properties:
+        agentId:
+          $ref: '#/components/schemas/AgentId'
         result:
           $ref: '#/components/schemas/UntypedJsonDataValue'
         componentRevision:
           type: integer
           format: uint64
+      required:
+      - agentId
     AgentMetadataDto:
       title: AgentMetadataDto
       type: object
--- a/sdks/ts/packages/golem-ts-sdk/src/decorators/agent.ts
+++ b/sdks/ts/packages/golem-ts-sdk/src/decorators/agent.ts
@@ -349,8 +349,10 @@ export function agent(options?: AgentDecoratorOptions) {
       );
     }
 
-    (ctor as any).get = getRemoteClient(agentClassName, agentType, ctor, false);
-    (ctor as any).getWithConfig = getRemoteClient(agentClassName, agentType, ctor, true);
+    if (agentType.mode === 'durable') {
+      (ctor as any).get = getRemoteClient(agentClassName, agentType, ctor, false);
+      (ctor as any).getWithConfig = getRemoteClient(agentClassName, agentType, ctor, true);
+    }
     (ctor as any).newPhantom = getNewPhantomRemoteClient(agentClassName, agentType, ctor, false);
     (ctor as any).newPhantomWithConfig = getNewPhantomRemoteClient(
       agentClassName,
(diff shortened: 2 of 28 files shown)


openapi/golem-service.yaml
```diff
@@ -9992,11 +9992,15 @@ components:
       title: AgentInvocationResult
       type: object
       properties:
+        agentId:
```
noise64, line 9995: was this added for making testing easier, or we really want to add it to the result?

sdks/ts/packages/golem-ts-sdk/src/decorators/agent.ts
```diff
@@ -349,8 +349,10 @@ export function agent(options?: AgentDecoratorOptions) {
       );
     }
 
-    (ctor as any).get = getRemoteClient(agentClassName, agentType, ctor, false);
-    (ctor as any).getWithConfig = getRemoteClient(agentClassName, agentType, ctor, true);
+    if (agentType.mode === 'durable') {
```
noise64, line 352: I'm a bit confused here, what will get and getWithConfig do on an empheral agent? Shouldn't they not even be available?

sdks/ts/packages/golem-ts-sdk/tests/decorator.test.ts
```diff
@@ -982,6 +982,20 @@ describe('Annotated FooAgent class', () => {
   });
 });
 
+describe('Annotated EphemeralAgent class', () => {
+  it('does not override non-phantom constructors', () => {
+    expect(EphemeralAgent.get).toBe(BaseAgent.get);
```
noise64, line 987: basically i have the same question, as above, what this will do?

sdks/ts/packages/golem-ts-sdk/src/decorators/agent.ts
vigoo: On ephemeral agents there is no `get` just `getPhantom` and `newPhantom`. This was always the intention but it was inconsistently implemented, as i wrote in the pr description

openapi/golem-service.yaml
vigoo: I wanted it in the result, because now the server assigns a random new phantom id for ephemeral agents. Without returning it we have no idea what agent instance performed the invocation

noise64: that is a good reason!

noise64 (approved):
as discussed on chat: approved, with expecting some followup on the get methods

PR #3234, opened 2026-04-21
Fix RPC performance regression by better caching

### LINKED ISSUE #3195: Investigate and fix rpc benchmark regression around 10th of April

vigoo:
Resolves #3195 

Fixes RPC performance regression introduced by https://github.com/golemcloud/golem/pull/3112 - keeps all the relevant chanes of the original PR, just adds caching for the latest component metadata now that we have an invalidation broadcast channel we can subscribe to.

### DIFF
--- a/golem-worker-executor/src/services/registry_event_subscriber.rs
+++ b/golem-worker-executor/src/services/registry_event_subscriber.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use crate::services::agent_types::AgentTypesService;
+use crate::services::component::ComponentService;
 use crate::services::environment_state::EnvironmentStateService;
 use golem_common::model::agent::RegistryInvalidationEvent;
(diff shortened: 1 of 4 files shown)


golem-worker-executor/src/services/registry_event_subscriber.rs
```diff
@@ -59,6 +64,9 @@ impl RegistryInvalidationHandler for WorkerExecutorRegistryInvalidationHandler {
                     environment_id = %environment_id,
                     "Received deployment changed event, invalidating environment caches"
                 );
+                self.component_service
+                    .invalidate_latest_deployed_metadata_for_environment(*environment_id)
```
noise64, line 68: just a comment in general, we keep adding "latest" naming, when it should be called "current" (given it can go back), for deployments. at least that was the agreed naming originally.

noise64 (approved):

PR #3235, opened 2026-04-21
Update test-r and fix integration tests

mschuwalow:


vigoo (commented):
Let's add a `:tag:` run as well (into one of the IT groups, not a separate one as it supposed to be empty) just to be sure that next time if we forget to tag something it runs. I think we are doing this already with worker executor tests.

vigoo (approved):

PR #3238, opened 2026-04-21
Removed binary_wit and exporst from component metadata

### LINKED ISSUE #3160: Check if we can drop the analysed exports / full WIT from the component metadata

vigoo:
Resolves #3160 

We were storing the whole binary WIT and all the analysed exports in every component metadata, and passing it around (storing in DB, through gRPC and REST APIs etc) and we actually don't need it anymore. For every component it is almost the same, but quite big interface.


mschuwalow (approved):
❤️

PR #3241, opened 2026-04-22
Ephemeral agents writing to the first oplog archive layer not the last

### LINKED ISSUE #3236: Ephemeral agents should not write to S3 directly
Depending on what our final prod oplog configuration is, ephemeral agents should either write to:
- postgres-backed compressed oplog archive layer (this is a smaller change because it is already implemented as writing to an archive oplog directly)
- or to the primary oplog (just like durable agents, but with ephemeral-specific bigger buffer sizes etc)

and then the scheduled oplog archive is supposed to eventually move these out to S3 without affecting performance.

vigoo:
Resolves #3236 

Ephemeral agents previously were directly writing to the **last** oplog archive layer (only), which is in production configured to be S3. 

This PR changes this to:
- write to the **first** oplog archive layer which is Postgres
- enable the scheduled (and count based) oplog archive machinery for ephemeral agents as well

by this writing the ephemeral oplog is not slow, but eventually they get archived to the S3 layer just like regular agents.


golem-worker-executor/src/services/oplog/ephemeral.rs
```diff
@@ -161,51 +373,70 @@ impl Oplog for EphemeralOplog {
...
-            let uncommitted_count = last_idx.distance_from(state.last_committed_idx);
-            let buffered_to_take =
-                min(max(0, uncommitted_count), state.buffer.len() as i64) as usize;
-
-            let mut current = state.last_committed_idx;
-            for idx in 0..buffered_to_take {
-                current = current.next();
-                let entry = state.buffer[idx].clone();
-                result.insert(current, entry);
+        // Read committed entries from lower layers (only up to last_committed_idx)
```
kmatasfp: I think we want some kind of short circuiting here, read Postgres first if did not have the indexes we were looking for then look s3.

Reading all layers here I think will hurt the startup latency even more

kmatasfp:
I think we want to introduce `oplog.entry_count_limit_ephemeral` and `oplog.archive_interval_ephemeral`, we want to control the archiving differently for ephemeral agents, on startup we need to look up these early indexes and longer we keep these in postgres the better. Can happen in followup PR. 

Why it matters - on startup we are looking for first replayable `AgentInvocationStarted` entry, to recover just enough to replay first invocation for agent initialization

golem-worker-executor/src/services/oplog/ephemeral.rs
vigoo: will think about this more, thanks!

vigoo: - Added an `exists` check before getting the intial oplog entry, which should speed up the "new, non-existing ephemeral" case (it's a head on s3, should be faster). We cannot skip the existence check because we cannot guarantee that the agent does not exist yet.
- Made `read_many` use the same logic as the non-ephemeral multi-layer oplog (only go deeper levels when needed etc)

mschuwalow (approved):

golem-worker-executor/src/services/oplog/ephemeral.rs
```diff
@@ -96,16 +106,214 @@ impl EphemeralOplog {
...
+                            warn!("No entries to transfer from ephemeral oplog layer {source}");
+                        }
+                    }
+
+                    let _ = keep_alive.take();
+                    if let Some(done) = done {
+                        let _ = done.send(());
+                    }
+                }
+                BackgroundTransferMessage::TransferFromPrimary {
```
mschuwalow, line 249: It would be nice to have a dedicated enum instead of impossible cases. But looks like that will need some adjustment on the WrappedOplogArchive as well, so fine to leave for later

kmatasfp (approved):

PR #3250, opened 2026-04-22
Plugin fixes

### LINKED ISSUE #3247: Built-in plugin grant does not work with local golem
And the CLI crashes with a panic

```
thread 'main' (84207) panicked at cli/golem-cli/src/command_handler/component/staging.rs:370:25:
Plugin grant not found for golem-otlp-exporter/1.5.0
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
   3: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
   4: <alloc::collections::btree::map::BTreeMap<K,V> as core::iter::traits::collect::FromIterator<(K,V)>>::from_iter
   5: golem_cli::command_handler::component::staging::ComponentStager::agent_type_provision_configs
   6: golem_cli::command_handler::app::AppCommandHandler::apply_changes_to_stage::{{closure}}
   7: golem_cli::command_handler::app::AppCommandHandler::deploy::{{closure}}
   8: golem_cli::command_handler::CommandHandler<Hooks>::handle_command::{{closure}}
   9: golem_cli::command_handler::CommandHandler<Hooks>::handle_args::{{closure}}
  10: tokio::runtime::context::runtime::enter_runtime
  11: tokio::runtime::runtime::Runtime::block_on
  12: golem_cli::main_wrapper
```

vigoo:
- builtin plugins have a fix version so server updates are not going to break deployments
- `golem plugin list` lists not only own, but shared plugins too (including builtins)
- plugin diff fix 
- if plugin referenced in the manifest is not found, give a proper error message instead of a panic
- example docker compose for trying out the otlp plugin


golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -955,6 +955,42 @@ impl<Ctx: WorkerCtx> DurableWorkerCtx<Ctx> {
                 ..
             } = &entry
         {
+            // Oplog processor plugin logs are emitted into the server log because
+            // they cannot be easily watched with CLI tools.
+            if self.state.component_metadata.metadata.has_oplog_processor() {
```
noise64, line 960: is this something we want in prod?

vigoo: yes

vigoo: more detailed answer: there is no other way to see what these plugins are doing at the moment. they are not real agents so no cli command works on them, etc. 
i think having them able to log into the server log is a good enough solution for now

noise64: I meant that in cloud user defined code can pollute our service logs.

noise64 (approved):

golem-worker-executor/src/durable_host/mod.rs
vigoo: I know but I think for now that is better than not being able to see anything. We definitely need better observatibility tools for the plugins!

PR #3251, opened 2026-04-22
Stability fixes

### LINKED ISSUE #3219: Check oplog_processor_locality_recovery again
It seems to be still flaky

vigoo:
Trying to resolve #3219 

This very likely fixes one failure case that I observed by looking at many of the flaky failing test runs, when some oplog batches sent to plugins got lost during the "chaos monkey" test. When an executor starts it automatically resumes all the agents that were running before so they continue. This was previously controlled by the agent's _running_ status which is too limited. If we add a pending invocation, it is not guaranteed that an agent is continously "running" while it still have work todo. 

Instead of making big changes to this state management in the invocation loop and around it I just relaxed the condition of what is marked as running. This way if things had pending invocations or updates they also get auto-resumed on restart. 

Note that this generally was not causing very big trouble because if someone "touches" a not-resumed agent by let's say invoking it again, it resumed anyway. But in the oplog processor case we just do a _lookup_ and that detects the last batch is already pending without resuming the agent. Because we see it is pending, we switch to a new local oplog processor and never call the old one again, assuming it will process the pending entry (which it did not because never woke up).

There can be one more oplog processor delivery issue (based on the logs) but much harder to reproduce, so I'm leaving it for now.


mschuwalow (approved):
Looks good. Do you want to keep the warn logs for now?

vigoo:
> Looks good. Do you want to keep the warn logs for now?

no i wanted to undo those, thanks

PR #3254, opened 2026-04-23
Fix App Deletion

kmatasfp:
Fix `INTERNAL_UNKNOWN: Component for id <uuid> not found` errors that surface when a cloud user deletes an application and recreates it with the same name.  

**Reasoning**
Cloud accounts have a hard per-account limit on the number of applications, so deleting and recreating an application with the same name should be supported cleanup path (including in CI). Before this PR after app recreation, invoking any agent fails with `Component not found`, and the failure persists until worker-service is restarted or the cache TTL expires (60 minutes).

**Root cause**
worker-service keeps an in-memory `AgentResolutionCache`  keyed on:
```
struct AgentResolutionCacheKey {
    app_name: String,
    env_name: String,
    agent_type_name: String,
    owner_account_email: Option<String>,
}
```
— strings only, no UUIDs. After delete+recreate the cache key stays identical, but the cached `ResolvedAgentType` still carries the OLD `environment_id` and OLD `component_id`. Invocations route to those orphaned identifiers and the registry returns 404.

**Fix**
Emit a `DeploymentChanged` invalidation event:
- In `environment_repo::delete`, for the deleted environment.
- In `application_repo::delete`, for every non-deleted environment under the application.
The events carry `deployment_revision + 1 / current_deployment_revision + 1` so `AgentResolutionCache::advance_latest_revision` observes a strictly larger value than any currently cached entry. On the next cache hit, is_stale returns true, the entry is removed, and a fresh resolve hits the registry and returns the recreated environment/component.


noise64 (changes requested):
Repo methods that emit events have to return results wrapped in `RequiresNotificationSignal` marker type, which then gets handled in services with the `signal_new_events_available` helper. This is required to handle the differences between how postgres and the (local golem) sqlite backends dispatch events.

golem-registry-service/src/repo/application.rs
```diff
@@ -377,6 +378,63 @@ impl ApplicationRepo for DbApplicationRepo<PostgresPool> {
...
+                for row in env_rows {
+                    let env_id: Uuid = row.try_get("environment_id").map_err(RepoError::from)?;
+                    let current_deployment_revision: i64 = row
+                        .try_get("current_deployment_revision")
+                        .map_err(RepoError::from)?;
+                    let deployment_revision: i64 = row
+                        .try_get("deployment_revision")
+                        .map_err(RepoError::from)?;
+
+                    let change_event = NewRegistryChangeEvent::deployment_changed(
```
noise64: I would rather use a new event, then creating a fake event, especially that it lies and refers to two non-existing revisions

golem-registry-service/src/repo/environment.rs
```diff
@@ -839,6 +840,33 @@ impl EnvironmentRepo for DbEnvironmentRepo<PostgresPool> {
...
+                // the environment was never deployed) so that
+                // AgentResolutionCache::advance_latest_revision observes a strictly
+                // larger revision than any cached entry and marks those entries stale.
+                let bumped_current_deployment_revision =
+                    environment_record.current_deployment_revision.unwrap_or(0) + 1;
+                let bumped_deployment_revision = environment_record
+                    .current_deployment_deployment_revision
+                    .unwrap_or(0)
+                    + 1;
+                let change_event = NewRegistryChangeEvent::deployment_changed(
```
noise64: Like above, I recommend creating a delete event

golem-registry-service/src/repo/application.rs
kmatasfp: my initial idea was to create `ApplicationDeleted` event, but ended up reusing `DeploymentChanged` as it was simpler and smaller fix, but sure let we rework this tomorrow with new event

golem-registry-service/src/repo/environment.rs
kmatasfp: sure will add

golem-worker-service/src/service/registry_event_subscriber.rs
```diff
@@ -136,6 +136,36 @@ impl RegistryInvalidationHandler for WorkerServiceRegistryInvalidationHandler {
...
+                debug!(
+                    application_id = %application_id,
+                    account_id = %account_id,
+                    "Received application deleted event, flushing agent resolution and route caches"
+                );
+                // The AgentResolutionCache is keyed on (app_name, env_name, ...)
+                // strings and has no UUID in its key, so we cannot target entries
+                // for this specific application. Flush the whole cache to ensure a
+                // same-name recreation resolves through a fresh registry lookup.
+                self.agent_resolution_cache.clear().await;
```
mschuwalow: This is going to cause a lot of burst traffic to the registry service. I think we should at least only remove the entries with matching env / app name (which is still going to hit a few unrelated entries).

golem-worker-executor/src/services/registry_event_subscriber.rs
```diff
@@ -134,6 +134,39 @@ impl RegistryInvalidationHandler for WorkerExecutorRegistryInvalidationHandler {
...
+                    application_id = %application_id,
+                    account_id = %account_id,
+                    "Received application deleted event, flushing all caches"
+                );
+                // Worker-executor caches are keyed per-environment/component/agent-type,
+                // none of which carry the application_id. Flush all to guarantee no
+                // cached entries for environments under the deleted application
+                // survive into a same-name recreation cycle.
+                self.component_service.invalidate_all().await;
+                self.environment_state_service.invalidate_all().await;
```
mschuwalow: we should include the owning application_id in the services and invalidation only the matching entries

mschuwalow (approved):
Looks mostly good to me. I would like to make the invalidation more targeted to avoid overloading the registry service due to all caches being invalidated at the same time.

Side note, we probably want the same thing for accounts as well. I believe you can have similar issues with shared environments.

golem-worker-service/src/service/registry_event_subscriber.rs
kmatasfp: makes sense

golem-worker-executor/src/services/registry_event_subscriber.rs
kmatasfp: makes sense

mschuwalow (approved):

noise64:
I have some _comments_ on the _comments_, but will ask for or do a seperate cleanup on that (and some other things i noticed in general around the signals).

So i'm merging this now.

noise64 (approved):

PR #3307, opened 2026-04-29
Big invoke payload

### LINKED ISSUE #3272: Big method payload issues
When a method's payload is bigger than 4mb, the internal gRPC message size limit triggers. An example can be using `UnstructuredBinary` as a parameter and passing a big binary.

- we should probably allow bigger than 4mb payloads (but not unlimited)
- the worker service does not properly handle reaching the limit, and retries it forever:

```
{"timestamp":"2026-04-29T04:51:22.463385Z","level":"WARN","target":"golem_service_base::grpc::client","message":"gRPC call failed: Status { code: ResourceExhausted, message: \"Error decompressing: size limit, of 4194304 bytes, exceeded while decompressing message\", metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\", \"date\": \"Wed, 29 Apr 2026 04:51:22 GMT\"} }, source: None }, not retriable","handle_route":{"domain":"stt-testing.apps.golem.cloud","method":"POST","route":"transcribe/{name}"},"HTTP execute":{"api":"execute","api_type":"http","method":"POST","uri":"/transcribe/my-agent?format=wav"}}
{"timestamp":"2026-04-29T04:51:22.463430Z","level":"INFO","target":"golem_worker_service::service::worker::routing_logic","message":"Retry calling executor after delay","invalidated":false,"error":"FailedToConnectToPod(Status { code: ResourceExhausted, message: \"Error decompressing: size limit, of 4194304 bytes, exceeded while decompressing message\", metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\", \"date\": \"Wed, 29 Apr 2026 04:51:22 GMT\"} }, source: None })","pod":"Some(Pod { ip: 172.18.184.116, port: 9093 })","attempt":972,"delay_ms":"108","op":"invoke_agent","call_on_executor_retry":{"attempt":972,"executor_kind":"AgentId"},"handle_route":{"domain":"stt-testing.apps.golem.cloud","method":"POST","route":"transcribe/{name}"},"HTTP execute":{"api":"execute","api_type":"http","method":"POST","uri":"/transcribe/my-agent?format=wav"}}
```

vigoo:
Resolves #3272 

- Applies max message size to all gRPC clients
- Sets the default to 32Mb (keeps 50Mb in registry service)
- Does not retry resource exhaustion error in worker service -> executor calls
- Return with a specific http status/message when the limit is exceeded


mschuwalow (approved):
Could you adjust the custom_api code to also forward the 413 to the client directly?

PR #3311, opened 2026-04-29
feat: support reader postgres instances

kmatasfp:
Add support for following:

Config example for Aurora
```
[db.config]
host = "my-cluster.cluster-xyz.us-east-1.rds.amazonaws.com"
reader_host = "my-cluster.cluster-ro-xyz.us-east-1.rds.amazonaws.com"
```


vigoo (dismissed):
Looks good, @noise64 please check too

noise64:
Just because something is using with_ro, does not mean it is actually okay for using generally with a read only replica (with_ro _does not_ mean that, "readonly and some lag is okay", and the code was not written with this in mind).

E.g, let's imagine a service level code, where we call repos like this (which can easily happen):
- with_rw: insert x
- with_ro: read x -> not found, even with a small lag

Ofc, usually this is not how we do things, I just wanted to show one simple example, how this can go wrong in subtle ways in one service, or even in one call. If we want to use read replicas, IMHO that should be very explicit, and should be a different helper method / pool altogether, and should come with inspecting the affected code paths, that they are consistent within themselves. Alternatively we should inspect every with_ro usage, and how they are mixed, and make sure we do not break such properties, but that feels far from trivial.

Also, i'm quite unsure, how many races this could introduce for the current CLI service call patterns (even if the lag is small).


Before understanding these, I would not do this, and I think this needs a proper plan to identify places and areas, where we can introduces replicas logically safely (and not just trusting, that the lag is "good enough").

PR #3320, opened 2026-04-30
Do not run scheduled invocations for deleted / recreated agents

### LINKED ISSUE #3293: Deleting an agent should delete scheduled invocations for it

mschuwalow:
resolves #3293


golem-common/src/model/oplog/protobuf.rs
```diff
@@ -2097,405 +2097,6 @@ impl TryFrom<golem_api_grpc::proto::golem::worker::RetryPolicyState> for PublicR
         }
     }
 }
-
-/// Convert a `PublicOplogEntry` to a domain `OplogEntry`.
-///
-/// Some entry types (e.g. `AgentInvocationStarted`) cannot be losslessly
-/// round-tripped because the public representation discards internal detail.
-/// For those cases this conversion returns an error.
-impl TryFrom<PublicOplogEntry> for OplogEntry {
```
vigoo, line 2106: I think this was supposed to replace the similar, older function in debugging-service (`debug_session.rs`), we just failed to delete that one during the oplog refactoring. 

I would like to keep this and remove that so all the conversions stay in `golem-common` where the types are defined

vigoo:
are you going to remove the other one now that the golem-common part was added back?

mschuwalow:
> are you going to remove the other one now that the golem-common part was added back?

Yes, but I would prefer to do that in a separate pr

vigoo (approved):

PR #3322, opened 2026-05-01
CLI - Improved agent-create secret ergonomics

### LINKED ISSUE #3270: CLI `secret create` command does not accept primitive secret values
It should accept strings as is, not JSON encoded into `""`

example:
```
golem agent-secret create apiKey --secret-type String --secret-value sk-abc123
error: Secret value is not valid: expected value at line 1 column 1
```

kmatasfp:
Resolves: #3270


mschuwalow (approved):

cli/golem-cli/src/command_handler/secret.rs
```diff
@@ -318,3 +311,295 @@ impl AgentSecretCommandHandler {
...
+    #[test]
+    fn bare_word_rejected_for_non_string_type() {
+        let result = parse_secret_value_to_json("notANumber", &u32_type(), &lang());
+        assert!(result.is_err());
+        assert!(result.unwrap_err().contains("Secret value is not valid"));
+    }
+
+    // --- CLI argument ordering ---
+
+    fn extract_create_args(cmd: GolemCliCommand) -> (String, Option<String>) {
```
noise64: these are testing if clap works.

noise64 (changes requested):

cli/golem-cli/src/command_handler/secret.rs
```diff
@@ -318,3 +311,295 @@ impl AgentSecretCommandHandler {
...
+        let result = parse_secret_value_to_json(r#""sk-abc123""#, &str_type(), &lang());
+        assert_eq!(
+            result,
+            Ok(serde_json::Value::String("sk-abc123".to_string()))
+        );
+    }
+
+    #[test]
+    fn bare_string_with_spaces_accepted_for_str_type() {
+        let result = parse_secret_value_to_json("hello world", &str_type(), &lang());
```
noise64: Not sure what this test tests, given we treat any input arg together as a string. On the other hand, this does not mean that on the shell level one can use strings with spaces without escaping.

cli/golem-cli/src/command_handler/secret.rs
kmatasfp: my reasoning here was that test are executable spec, and the added tests is behavior we want to support, clap is implementation detail.. Also I follow TDD, I always add red test first then make it green and then refactor. but if you guys feel they are not adding value, can remove.

cli/golem-cli/src/command_handler/secret.rs
kmatasfp: here I wanted to make sure we cover all the input cases to make sure our logic handles this, before this fix one also assumed that following would work: `golem agent-secret create apiKey --secret-type String --secret-value sk-abc123`, or `golem agent-secret create apiKey --secret-type String --secret-value "sk-abc123"` but that was not the case.

But yeah probably property based testing would be better approach here

cli/golem-cli/src/command_handler/secret.rs
kmatasfp: removed

cli/golem-cli/src/command_handler/secret.rs
kmatasfp: introduced property based tests

noise64 (approved):

PR #3323, opened 2026-05-01
feat: support "auto" as update mode on top of "automatic"

### LINKED ISSUE #3281: --update-agents should accept `auto` not just `automatic`
Also every other command accepting update moe

kmatasfp:
Resolves: #3281


mschuwalow (approved):

cli/golem-cli/src/command.rs
```diff
@@ -2132,6 +2133,109 @@ mod test {
         println!("{:?}", builtin_exec_subcommands())
     }
 
+    #[test]
+    fn update_agents_accepts_auto_as_update_mode_alias() {
```
noise64, line 2137: these tests are testing if clap itself works, i do not really see the value of them, especially this many

kmatasfp: The alias annotation I added is legitimate change that is needed, and I would like this to be covered by tests, as this is something we support and want things to fail if someone accidentally removes it and obviously the old way should keep working.

My general reasoning here was that test are executable spec, and the added tests is behavior we want to support, clap is implementation detail,  but if you guys feel they are not adding value, can remove.

kmatasfp: removed some of the tests that also covered the same code path

noise64 (approved):

PR #3336, opened 2026-05-05
cli: do not update ephemeral agents or ones at correct version

### LINKED ISSUE #3282: golem deploy --update-agents behavior when no new deployment happens
when golem deploy --update-agents skips deployment because there were no changes, it tries updating agents that are already on the latest version, and prints errors for them one by one

mschuwalow:
resolves #3282


cli/golem-cli/src/command_handler/worker/mod.rs
```diff
@@ -1672,8 +1672,16 @@ impl WorkerCommandHandler {
         await_update: bool,
         disable_wakeup: bool,
     ) -> anyhow::Result<TryUpdateAllWorkersResult> {
+        let durable_filter = ["mode == durable".to_string()];
```
vigoo: I think the revision check could also be an enumeration filter (`revision < REV` or something)

vigoo (approved):

PR #3345, opened 2026-05-06
registry-service: tie oauth2 logins to github id instead of username

### LINKED ISSUE #2130: oauth2: tie accounts to the GitHub id, not the GitHub username
Users can change their GitHub username, which currently disconnects them from their old Golem account. It will also most likely lead to login issues, as they are now effectively trying to create a new account using the same email when logging in.

Instead of using the login, we should be using the GitHub id.

```

curl -L \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/user

Response with public and private profile information

Status: 200
{
  "login": "octocat", // <- we use this
  "id": 1, // <- we should be using this
  "node_id": "MDQ6VXNlcjE=",
  "avatar_url": "https://github.com/images/error/octocat_happy.gif",
  "gravatar_id": "",
  "url": "https://api.github.com/users/octocat",
...
```

mschuwalow:
resolves #2130


golem-registry-service/src/repo/token.rs
```diff
@@ -216,6 +216,15 @@ impl TokenRepo for DbTokenRepo<PostgresPool> {
         let result: Option<Uuid> = self
             .with_tx_err("delete", |tx| {
                 Box::pin(async move {
+                    // Unlink any OAuth2 identity that points at this token before
```
vigoo, line 219: Do we have to do the same with `oauth2_web_flow_states`?

mschuwalow:
I'm a bit unhappy with how the login is handled as a whole, will clean up as part of this

golem-registry-service/src/repo/token.rs
mschuwalow: done

mschuwalow:
Cleaned up our auth flows a bit:

1. deleted the device flow as it was unused (and untested)
2. split and explicitly encoded how the exchange oauth2 token -> golem token works at the end of the flow. We support either the cli flow (hardcoded redirect to the golem webpage + polling for the token) or a more "usual" browser based flow that redirects with the token and does not allow polling.

vigoo (approved):

PR #3349, opened 2026-05-07
Multi component fix

### LINKED ISSUE #3266: Single->multi component app upgrade does not work with MoonBit (and maybe Scala)

vigoo:
Fixes the Scala and MoonBit multi-component apps

Resolves #3266


cli/golem-cli/src/app/template/plan.rs
```diff
@@ -431,6 +431,39 @@ fn validate_yaml(source: &str) -> anyhow::Result<()> {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum MultiComponentLayoutUpgradePlanStep {
     Move { source: PathBuf, target: PathBuf },
+    Rewrite { path: PathBuf, edit: TextEdit },
+}
+
+/// A targeted in-place edit applied to an existing source file as part of a
+/// multi-component layout upgrade. Edits are designed to preserve any unrelated
+/// user customizations in the file.
+#[derive(Debug, Clone, PartialEq, Eq)]
```
noise64, line 440: we have an edit module for lang / file specific editors, i would prefer to keep these there.
i mean, not the plan step itself, but the actual implementations.

cli/golem-cli/src/model/app.rs
```diff
@@ -1022,12 +1023,22 @@ impl ComponentLayerApplyContext {
         component_dir: Option<String>,
         cargo_target: Option<String>,
     ) -> Self {
+        let component_dir_rel = match (app_root_dir.as_deref(), component_dir.as_deref()) {
+            (Some(app_root), Some(comp_dir)) => Some(
+                std::path::Path::new(comp_dir)
+                    .strip_prefix(app_root)
```
noise64: In other places i replaced many path opertations with the cli crate's fs module ones to get consistent errors and to avoid lossy (i know that we could use camino in more places, but that is a bigger changes), and more importatnly: i hade to replace many places with the "normalize" functions, and "normalize compare", to properly support windows (and to avoid canonicalization on win, which results in different style of paths). Could you review if we could use those?

cli/golem-cli/templates/moonbit/common-on-demand/golem.yaml
```diff
@@ -12,11 +12,11 @@ componentTemplates:
           dir: "{{ appRootDir }}"
         - command: moon build --target wasm
           dir: "{{ appRootDir }}"
-        - command: sh -c "wasm-tools component embed GOLEM_MOONBIT_BUILD_SDK_PATH/wit $(find {{ appRootDir }}/_build/wasm/debug/build -name '*.wasm' -not -name '*.embed.wasm' | head -1) --encoding utf16 --output {{ appRootDir }}/_build/wasm/debug/component.embed.wasm"
+        - command: wasm-tools component embed GOLEM_MOONBIT_BUILD_SDK_PATH/wit {% if componentDirRel == '' %}{{ appRootDir }}/_build/wasm/debug/build/{{ componentName | split(':') | first }}.wasm{% else %}{{ appRootDir }}/_build/wasm/debug/build/{{ componentDirRel }}/{{ componentDirRel }}.wasm{% endif %} --encoding utf16 --output {{ appRootDir }}/_build/wasm/debug/{{ componentName | replace(':', '_') }}.embed.wasm
```
noise64: Could we avoid the conditional logic by exposing a moonbit and component specific variable? I think we already do soemthing similar with rust and `{{ cargoTarget }}`.  I see that this is a bit more complex, but I would happy if the template could be simplified. (But this is just an idea, i'm okay with using this too.)

noise64: And i think we have toSnakeCase and similar in the tempalte context, could that work instead of the custom replace?

noise64 (approved):

PR #3364, opened 2026-05-09
Retry and interrupt related fixes

vigoo:
Fixes discovered during a "final" testing phase before the 1.5.1 release, most by a chaos-monkey backend testing retry policy behavior.

Unfortunately some parts of the new semantic retry logic were incomplete / not properly working - it did not have any way to retry by http status codes (although its data model suggested it can) and it also did not properly apply the policies for non-inline retries. 

Agents being retried repeatedly (trap-based retry failing again) were not interruptible.

**Changes in detail**:

Retry-polcy semantics:
- New helpers (RetryPolicyState::exhausted/is_exhausted, Predicate/RetryPolicy::references_property, evaluate_named_policy_step_resetting_on_invalid_state, lookup_retry_state_with_replay_aggregation) and humantime/env-var support in policy deserialization.
- Worker status now stays Retrying based on the persisted retry_policy_state instead of the legacy retry-config bound.

Non-inline semantic trap retries
- New SemanticTrapRetryOverride carried through the anyhow chain → TrapType::Error → InvokeResult::Failed so the post-trap recovery path honors the policy decision made inside the host call.
- Retry budget is carried across trap+replay boundaries to stop unbounded loops.

HTTP status-code retries
- Status-code policies are filtered/resolved explicitly
- Retries now reuse the connection pool.
- Inject a Host header on reconstructed requests (no-pool path didn't add it, so servers returned 400 and bypassed 5xx retries).
- Wait for the outgoing body to finish before evaluating status-code retries when the response arrives early.

Invocation loop
- Workers can now be interrupted/deleted while sitting in the recovery retry loop or a delayed restart.

Misc
- Duration overflow fix
- Use agent method name (not raw WIT FQ-FN) in logs/events
- Env var substitution for config sections in the manifest

Tests & docs
- Many new retry tests
- Skill updates


noise64 (approved):

vigoo:
Ignoring known flaky sharding test failure

PR #3430, opened 2026-05-13
http-api: better logging on internal errors

mschuwalow:


golem-worker-service/src/custom_api/oidc/handler.rs
```diff
@@ -212,6 +212,8 @@ impl OidcHandler {
             .cloned()
             .collect();
 
+        debug!("getting auth url for oidc");
```
vigoo, line 215: did you want to keep this?

vigoo (approved):

golem-worker-service/src/custom_api/oidc/handler.rs
mschuwalow: Yeah, it's useful to have this as the `identity_provider.get_authorization_url` can fail (unreachable, inconsistent issuer url, ...). This makes it a bit easier to see at a glance that it's that place that failed.

PR #3444, opened 2026-05-16
Http retry fixes

vigoo:
- Idempotency key remains stable when using atomic regions. This allows coupling for example a fetch with a timeout+exception throwing and still reuse the same idempotency key on retries.
- There was an issue in the inline http request retry not re-sending the whole outgoing body
- Idempotency detection is now consistent in the inline/non-inline cases
- Idempotency key was not re-sent on status-based inline retries


mschuwalow (approved):
```
Idempotency key remains stable when using atomic regions. This allows coupling for example a fetch with a timeout+exception throwing and still reuse the same idempotency key on retries.
```

This would feel unintuitive for me. Every request in an atomic region having its own idempotency key feels more consistent with the golem model. But probably this is the way to go, at least until there is more options for users to interact with the idempotency keys directly / we enable them to have either semantics.

vigoo:
> This would feel unintuitive for me. Every request in an atomic region having its own idempotency key feels more consistent with the golem model. But probably this is the way to go, at least until there is more options for users to interact with the idempotency keys directly / we enable them to have either semantics.

I agree and also felt like it's wrong in both ways, but with @jdegoes we decided to do this version for now

PR #3465, opened 2026-05-18
Scheduler rewrite

vigoo:
Better scheduled action implementation in the executor.

The previous version was the original prototype that somehow survived until now and was originally built on Redis sorted sets, abstracted away and implemented now on Postgres/sqlite. 
It had several issues:

- It stored actions in one-hour buckets, these buckets could grow very large
- It queried two whole buckets every time it ran (every 2s)
- It never processed entries that were unprocessed and older than 2h
- Matching entries falling into the same `ms` got deduplicated
- Each executor queried every scheduled entry and only filtered by shard assignment in memory

The new implementation solves all these with a completely new scheduler-specific data table.


noise64 (approved):

golem-service-base/src/db/sqlite.rs
```diff
@@ -420,10 +420,32 @@ impl super::LabelledPoolApi for SqliteLabelledApi {
...
+
+/// Options controlling migration behavior.
+#[derive(Default, Clone, Copy, Debug)]
+pub struct MigrateOptions {
+    /// If true, applied migrations not present in the source are ignored instead
+    /// of being treated as a fatal error. Use this only when multiple modules
+    /// intentionally share a SQLite database file with disjoint migration version
+    /// ranges (e.g. indexed storage running on top of the key-value SQLite pool
+    /// via `KVStoreSqlite`).
+    pub ignore_missing: bool,
```
mschuwalow: This is very dangerous to have imo. It would be cleaner to move these to disjoint databases that can be independently migrated, or include the migrations in both

golem-worker-executor/db/migration/scheduler/postgres/001_init.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TABLE scheduled_actions (
+    schedule_id     UUID NOT NULL,
+    due_at_ms       BIGINT NOT NULL,
+    available_at_ms BIGINT NOT NULL,
+    shard_id        BIGINT NOT NULL,
+    action          BYTEA NOT NULL,
+    lease_owner     UUID NULL,
```
mschuwalow, line 7: I believe it's more common to use SKIP LOCKED instead of explicit lease management when building queues in postgres. 

cc @noise64

vigoo: The side-effects of the scheduled action are happening outside of the db operation and I wanted to protect against double-execution with this in case of resharding / scaling events.

golem-service-base/src/db/sqlite.rs
vigoo: Yeah I agree was just lazy to do it :) doing now

golem-worker-executor/db/migration/scheduler/postgres/001_init.sql
mschuwalow: Yeah, that makes sense. I just meant is that I usually saw this implemented by holding a row level lock on the scheduled actions that you are currently running (and other schedulers pick unlocked rows only using SKIP LOCKED).

But doing it like this will also work, just might delay when the scheduled action becomes visible again for longer.

golem-worker-executor/src/services/scheduler.rs
```diff
@@ -183,227 +198,260 @@ impl SchedulerServiceDefault {
...
+            }
 
-        all.extend(
-            all_from_this_hour
-                .into_iter()
-                .map(|(_score, action)| (current_hour_key.as_str(), action)),
-        );
+            // ! Do not exit early from this loop because of failed actions, as it will cause all other actions to be skipped.
+            // ! Retryable failures are left unacknowledged and retried after lease expiry.
+            for claimed_action in claimed {
```
mschuwalow, line 219: I believe there is a race here. When claiming the due actions, we only assign them to us for 30 seconds until another instance can claim them. If we take more than 30s processing a batch, an action might be handled by 2 instances

vigoo: Yes - can't claim it forever to avoid it getting stuck (if the executor crashes before finishing the processing). And processing such a batch should take milliseconds not 30s ideally.

PR #3466, opened 2026-05-18
Environment deletion

mschuwalow:
Ensure that agents are properly stopped on environment deletion (not just component deletion).

This is a first step in allowing to investigate and shut down misbehaving applications as admins.

### DIFF
--- a/golem-api-grpc/proto/golem/worker/v1/worker_execution_error.proto
+++ b/golem-api-grpc/proto/golem/worker/v1/worker_execution_error.proto
@@ -14,28 +14,29 @@ message WorkerExecutionError {
     AgentAlreadyExists agent_already_exists = 2;
     AgentCreationFailed agent_creation_failed = 3;
     FailedToResumeAgent failed_to_resume_agent = 4;
-    ComponentDownloadFailed component_download_failed = 5;
-    ComponentParseFailed component_parse_failed = 6;
-    GetCurrentVersionOfComponentFailed get_current_version_of_component_failed = 7;
-    PromiseNotFound promise_not_found = 8;
-    PromiseDropped promise_dropped = 9;
-    PromiseAlreadyCompleted promise_already_completed = 10;
-    Interrupted interrupted = 11;
-    ParamTypeMismatch param_type_mismatch = 12;
-    NoValueInMessage no_value_in_message = 13;
-    ValueMismatch value_mismatch = 14;
-    UnexpectedOplogEntry unexpected_oplog_entry = 15;
-    RuntimeError runtime_error = 16;
-    InvalidShardId invalid_shard_id = 17;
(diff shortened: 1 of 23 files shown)


golem-registry-service/src/repo/component.rs
```diff
@@ -647,7 +647,15 @@ impl ComponentRepo for DbComponentRepo<PostgresPool> {
                            cr.object_store_key, cr.binary_hash
                     FROM components c
                     JOIN component_revisions cr ON c.component_id = cr.component_id
-                    WHERE c.component_id = $1 AND cr.revision_id = $2 AND ($3 OR c.deleted_at IS NULL)
+                    JOIN environments e ON e.environment_id = c.environment_id
```
mschuwalow: The semantics here changed such that include_deleted means "include if the entity itself is deleted, but still exclude it if any parent is deleted)

golem-api-grpc/proto/golem/worker/v1/worker_execution_error.proto
```diff
@@ -14,28 +14,29 @@ message WorkerExecutionError {
...
-    PreviousInvocationFailed previous_invocation_failed = 18;
-    UnknownError unknown = 19;
-    PreviousInvocationExited previous_invocation_exited = 20;
-    InvalidAccount invalid_account = 21;
-    AgentNotFound agent_not_found = 22;
-    ShardingNotReady sharding_not_ready = 23;
-    InitialAgentFileDownloadFailed initial_agent_file_download_failed = 24;
-    FileSystemError file_system_error = 25;
-    InvocationFailed invocation_failed = 26;
+    ComponentNotFound component_not_found = 5;
```
vigoo, line 17: The new `WorkerExecutorError::ComponentNotFound` is not handled in:
- `From<WorkerExecutorError>` for `ApiEndpointError` — falls through to INTERNAL_UNKNOWN (HTTP 500).
- `From<WorkerServiceError>` for `agent_error::Error`  for `GolemError(WorkerExecutorError::ComponentNotFound { .. })` — falls through to InternalError.
- `golem-worker-service/src/service/worker/routing_logic.rs` should likely treat it as Expected(...) so it isn't logged/metered as an internal failure.

vigoo:
I noticed that the `EnvironmentDeleted` event handler in the executor only invalidates the `current_component_metadata_cache` and not the `component_metadata_cache` (pre-existing issue), could you fix it in this pr?

golem-api-grpc/proto/golem/worker/v1/worker_execution_error.proto
mschuwalow: Done. This should normally not reach the worker service outside very specific timings (as the auth check will fail before it reaches the executor), but better to clean it up :+1:

vigoo (approved):

PR #3469, opened 2026-05-18
impersonation api

mschuwalow:
Adds a new impersonation endpoint that admins can use to impersonate a user. Typical debugging workflow would be to call this endpoint and then configure the cli to use the impersonation token.

Also removes visibility bypass (e.g. seeing environments they have no relationship with) from admin. These usecases should now be handled via impersonation.

cc @kmatasfp

### DIFF
--- a/golem-registry-service/src/services/auth.rs
+++ b/golem-registry-service/src/services/auth.rs
@@ -15,10 +15,10 @@
 use crate::repo::account::AccountRepo;
 use crate::repo::model::account::{AccountBySecretRecord, AccountRepoError};
 use chrono::Utc;
-use golem_common::model::account::Account;
-use golem_common::model::auth::{AccountRole, TokenSecret};
+use golem_common::model::account::{Account, AccountId};
+use golem_common::model::auth::TokenSecret;
 use golem_common::{SafeDisplay, error_forwarding};
-use golem_service_base::model::auth::{AuthCtx, UserAuthCtx};
+use golem_service_base::model::auth::{AdminImpersonationAuthCtx, AuthCtx, UserAuthCtx};
 use std::collections::BTreeSet;
 use std::sync::Arc;
 use tracing::warn;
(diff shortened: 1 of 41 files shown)


golem-registry-service/src/services/auth.rs
```diff
@@ -64,19 +64,26 @@ impl AuthService {
...
+            None => {
+                let account_roles = BTreeSet::from_iter(target_account.roles.clone());
+                Ok(AuthCtx::User(UserAuthCtx {
+                    account_id: target_account.id,
+                    account_roles,
+                    account_plan_id: target_account.plan_id,
+                }))
+            }
+            Some(admin_uuid) => {
+                let target_account_roles = BTreeSet::from_iter(target_account.roles.clone());
```
vigoo, line 110: Should we check here that the admin is still an admin (and exists)? Or is that somehow implicitly validated here? I'm thinking of having a longer expiration impersonation and the admin being deleted or admin role removed etc (maybe in case of a leak or something)

golem-service-base/src/model/auth/mod.rs
```diff
@@ -315,26 +342,54 @@ impl AuthCtx {
                 account_id: inner.account_id,
             }),
             Self::System => Self::System,
+            Self::AdminImpersonation(inner) => Self::ImpersonatedUser(ImpersonatedUserAuthCtx {
+                account_id: inner.target_account_id,
```
vigoo: Maybe we should add an optional field to this so the fact that it's an impersonated user propagates through the auth context? (Just for logging)

mschuwalow: Done and also improved the naming here a bit. Impersonation became a bit of an overloaded term.

golem-registry-service/src/services/auth.rs
mschuwalow: Done. The impersonation flow should be rare enough that having the extra query here to fetch the minting admin account is a non issue.

vigoo (approved):

PR #3470, opened 2026-05-18
various additional metrics

mschuwalow:


golem-service-base/src/db/sqlite.rs
```diff
@@ -157,6 +157,7 @@ impl SqliteLabelledTransaction {
             api_name = self.api_name,
             "DB transaction rollback",
         );
+        golem_common::metrics::db::record_db_rollback("sqlite", self.svc_name, self.api_name);
```
vigoo: shouldn't we have this for postgres too?

golem-service-base/src/db/postgres.rs
```diff
@@ -64,6 +64,31 @@ impl PostgresPool {
...
+    }
+
+    pub fn num_idle(&self) -> u32 {
+        self.pool.num_idle() as u32
+    }
+
+    /// Runs the periodic pool-metrics reporting loop.
+    /// Records `db_pool_size` and `db_pool_idle` for the given service name every 15 seconds.
+    /// Never returns — intended to be run as a background task.
+    pub async fn run_metrics_loop(&self, svc_name: &'static str) -> anyhow::Result<()> {
```
vigoo, line 79: Can't we set the gauge on each pool size change? Why do it with this background loop?

golem-worker-executor/src/services/active_workers/mod.rs
```diff
@@ -408,4 +412,59 @@ impl<Ctx: WorkerCtx> ActiveWorkers<Ctx> {
         }
         freed >= storage_bytes
     }
+
+    /// Registers a background task in `join_set` that periodically records worker-status
```
vigoo: I don't think we need a background polling task for this.
There is a single point where all status changes are recorded: `update_cached_status` in the `WorkerService` implementation. There we could directly change the gauges, couldn't we?

golem-service-base/src/db/postgres.rs
mschuwalow: I did not see anything in sqlx that allows reacting to pool size changes

golem-worker-executor/src/services/active_workers/mod.rs
mschuwalow: I would still want to have this for the available memory / filesystem semaphore.

We could handle this inline on each acquire / release, but I feel that would impact readability of the code too much. Having it in the background in a loop (with 15 seconds update frequency currently) is not that much overhead while keeping things simpler.

But I can switch to inline metric updates, just wanted a confirmation before proceeding

golem-service-base/src/db/sqlite.rs
mschuwalow: The clankers missed that one :D

golem-worker-executor/src/services/active_workers/mod.rs
vigoo: I'm not 100% against it :) just thought that if we can just set the gauages/counters where they change we don't have to think about whether this loop with the locks involved cause any perf issues or not, and also metrics would be "immediately" up-to-date for scrapers.

mschuwalow: Ok, done https://github.com/golemcloud/golem/pull/3470/commits/652960d50a709cbb98bcabcce9546472110fc206

golem-worker-executor/src/services/active_workers/fs_semaphore.rs
```diff
@@ -30,10 +30,63 @@ pub struct FilesystemStorageSemaphore {
...
+        self.permit
+            .as_mut()
+            .and_then(|permit| permit.split(n))
+            .map(|permit| Self {
+                permit: Some(permit),
+            })
+    }
+
+    pub fn merge(&mut self, mut other: Self) {
+        if let Some(other_permit) = other.permit.take() {
```
mschuwalow, line 65: This is needed so drop of the other permit does not update the metrics

golem-worker-executor/src/services/active_workers/mod.rs
```diff
@@ -70,10 +70,45 @@ pub struct ActiveWorkers<Ctx: WorkerCtx> {
...
+        }
+    }
+
+    pub fn num_permits(&self) -> usize {
+        self.permit
+            .as_ref()
+            .map_or(0, |permit| permit.num_permits())
+    }
+
+    pub fn merge(&mut self, mut other: Self) {
```
mschuwalow, line 92: same pattern here

golem-worker-executor/src/services/active_workers/mod.rs
```diff
@@ -138,14 +179,19 @@ impl<Ctx: WorkerCtx> ActiveWorkers<Ctx> {
     }
 
     pub async fn remove(&self, agent_id: &AgentId) {
-        self.workers.remove(agent_id).await;
+        if let Some(worker) = self.workers.get(agent_id).await {
```
vigoo: I think we should add a return value to the `remove` instead unless it's surely never get double called in some races (which I don't know)

vigoo (approved):

golem-worker-executor/src/services/active_workers/mod.rs
mschuwalow: let me do it using RAII in the worker itself instead.

PR #3471, opened 2026-05-19
No panic on oplog logic errors

vigoo:
Makes several oplog-read / deserialization / unexpected error entries into agent errors instead of panics. The faulty agent can be then examined and fixed with revert.


golem-worker-executor/src/durable_host/golem/v1x.rs
```diff
@@ -1679,21 +1686,41 @@ impl GetPromiseResultEntry {
         }
     }
 
-    pub async fn get_handle(&self) -> &PromiseHandle {
+    pub async fn get_handle(&self) -> Result<&PromiseHandle, &String> {
```
noise64, line 1689: Is this intentionally &String and not &str?

noise64 (approved):

PR #3473, opened 2026-05-19
Remove deleted agent provision configs during component updates

### LINKED ISSUE #3446: Deploy failure when deploying then changing the initial template and deploying again
Sequence (with 1.5.1 release)

- Create a new app with the TS default template
- Deploy
- Remove the `httpApi` section from the manifest
- Remove the `mount` and `endpoint` from the code
- Rename the class to `OtherAgent`
- Deploy

```
Summary
  Failed to stage [ERROR]
    error: Component Service - Error: 400 Bad Request, Agent type 'CounterAgent' is referenced in provision config but not declared in the component's agent types
```

noise64:
Resolves https://github.com/golemcloud/golem/issues/3446 + adds invariant checks before saving metadata

### DIFF
--- a/golem-registry-service/src/services/component/error.rs
+++ b/golem-registry-service/src/services/component/error.rs
@@ -77,6 +77,8 @@ pub enum ComponentError {
     ConflictingEnvironmentPluginGrantId(EnvironmentPluginGrantId),
     #[error("agent type for name {0} not found in environment")]
     AgentTypeForNameNotFound(String),
+    #[error("Multiple agent types with the same name: {0}")]
+    DuplicateAgentTypeName(AgentTypeName),
     #[error(
         "Agent type '{0}' is referenced in provision config but not declared in the component's agent types"
     )]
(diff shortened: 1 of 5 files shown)


golem-registry-service/src/services/component/error.rs
```diff
@@ -77,6 +77,8 @@ pub enum ComponentError {
     ConflictingEnvironmentPluginGrantId(EnvironmentPluginGrantId),
     #[error("agent type for name {0} not found in environment")]
     AgentTypeForNameNotFound(String),
+    #[error("Multiple agent types with the same name: {0}")]
+    DuplicateAgentTypeName(AgentTypeName),
     #[error(
         "Agent type '{0}' is referenced in provision config but not declared in the component's agent types"
```
vigoo, line 83: The reason why this error was so confusing, I think, is that the term "provision config" is never shown to the user anywhere else (or is it)

vigoo (approved):

golem-registry-service/src/services/component/error.rs
noise64: Agree, it should be an internal name only. On the other hand, this is a "should not happen" error, meaning either CLI or server side bug, not fixable by the user.

mschuwalow (approved):

PR #3568, opened 2026-05-23
Fix panic during oplog archive of agent in deleted environment

mschuwalow:


golem-worker-executor/src/services/active_workers/mod.rs
```diff
@@ -183,6 +185,23 @@ impl<Ctx: WorkerCtx> ActiveWorkers<Ctx> {
         self.workers.iter().await
     }
 
+    /// Interrupts and unloads all in-memory workers whose environment matches
+    /// `environment_id`.  Called when the environment is deleted so that
+    /// running workers stop promptly.
+    pub async fn unload_environment(&self, environment_id: EnvironmentId) {
```
mschuwalow, line 191: This was needed to be able to reproduce the panic in tests, but is a useful improvement otherwise as well (so not other actually useful agents are unloaded in favour of these ones)

integration-tests/Cargo.toml
```diff
@@ -76,6 +76,11 @@ name = "agent-config-live-mutation"
 path = "tests/agent_config_live_mutation.rs"
 harness = false
 
+[[test]]
```
mschuwalow, line 79: moved to a separate suite so I can have dedicated configs for it

vigoo (approved):

PR #3574, opened 2026-05-26
Read-only method side-effect violation error

### LINKED ISSUE #3390: Side-effect trap for read-only methods
## Goal

When an agent method is invoked under read-only strictness, **any** host call that would touch the durability layer must trap immediately with a typed `ReadOnlyViolation` error, *before* the host call runs and *before* anything is written to the oplog.

Part of #3388. Depends on #3389 (need the read-only marker on `AgentMethod` to know when to enter strict mode).

### 1. New strictness mode

Add to [`DurableWorkerCtx`](https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/durable_host/mod.rs):

```rust
pub enum InvocationStrictness {
    Normal,
    ReadOnly,
}
```

State lives on the durable host context. Setter pair (`enter_read_only_mode` / `exit_read_only_mode`) modeled after the existing [`begin_call_snapshotting_function`](https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/durable_host/mod.rs) shape, but distinct: snapshotting *drops* writes silently; read-only must *trap*.

### 2. Wire the mode around the invocation

In [`invoke_observed`](https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/worker/invocation.rs) (or the layer just above it that already knows whether the call is read-only), set `InvocationStrictness::ReadOnly` for the duration of an `AgentMethod` invocation whose `AgentMethod.read_only.is_some()`. Restore on exit (incl. on panic / trap).

The strictness is determined from the resolved `AgentMethod` metadata, not by the caller. Lookup happens once at the public entry point (matching the place where the cache lookup of #3391 will sit).

### 3. Trap on every host call

In every host-call path in [`durability.rs`](https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/durable_host/durability.rs), add a check at the very top of `wrap`/`persist_durable_function_invocation`/`read_persisted_durable_function_invocation` (and any other entry point that consults `persistence_level`):

```rust
if state.invocation_strictness == InvocationStrictness::ReadOnly {
    return Err(ReadOnlyViolation { ... }.into());
}
```

This blocks **every** `DurableFunctionType` variant — `ReadLocal`, `ReadRemote`, `WriteLocal`, `WriteRemote`, `WriteRemoteBatched`, `WriteRemoteTransaction`. The reason for blocking the read variants too: they cover clocks, randomness, env reads, RPC, outgoing HTTP, etc. — anything that introduces non-determinism would invalidate the cache contract and break the static "no side effects" guarantee.

### 4. Typed error

Add to [`golem_common::model::oplog::AgentError`](https://github.com/golemcloud/golem/blob/main/golem-common/src/base_model/oplog/raw_types.rs) (and propagate through `AgentInvocationError`):

```rust
ReadOnlyViolation {
    method: String,
    host_function: String,
    function_type: DurableFunctionType,
}
```

This must surface as a typed `agent-error` to the SDK so that user code can distinguish it from other failures. The trap path must produce this error rather than a generic runtime trap.

### 5. Consistency with existing modes

- `PersistenceLevel::PersistNothing` (used by snapshotting) must continue to silently drop writes — the read-only check happens *before* the persistence-level branch, and only when `InvocationStrictness::ReadOnly` is set.
- The strictness must be cleared on every exit path (success, error, panic, restart).

## Acceptance criteria

- [ ] `InvocationStrictness::ReadOnly` exists on `DurableWorkerCtx` with safe enter/exit semantics (no leaked state across invocations).
- [ ] The strictness is set automatically by the worker-executor for any `AgentInvocation::AgentMethod` whose target method has `read_only = Some(_)`.
- [ ] All host calls in [`durability.rs`](https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/durable_host/durability.rs) check the strictness and trap with `ReadOnlyViolation` when set.
- [ ] The trap produces an `AgentError::ReadOnlyViolation` with `method`, `host_function`, and `function_type` populated.
- [ ] No oplog entry is written for the trapped host call.
- [ ] Snapshotting mode behaviour is unchanged.

## Out of scope

- Caching and queue bypass → #3391
- The "AgentInvocationStarted/Finished get written when the read-only call actually runs" behaviour is part of the cache work in #3391.
- SDK-side error mapping for `ReadOnlyViolation` → handled in the SDK marker work in #3389 *if* the SDK already maps `agent-error` variants; otherwise as a small follow-up.

## Related

- Epic: #3388
- Depends on: #3389
- Blocked by tests in: #3393

vigoo:
Stack on https://github.com/golemcloud/golem/pull/3572
Resolves #3390


golem-worker-executor/src/durable_host/websocket/client.rs
```diff
@@ -513,6 +521,10 @@ impl<Ctx: WorkerCtx> HostWebsocketConnection for DurableWorkerCtx<Ctx> {
     ) -> anyhow::Result<Result<(), Error>> {
         self.observe_function_call("golem:websocket/client", "close");
 
+        // Trap immediately if the invocation is restricted to read-only side effects.
```
mschuwalow: Can we include this in our durability structure, relying on DurableFunctionType to detect whether it is read-only?

mschuwalow (approved):

PR #3578, opened 2026-05-26
agent-permissions: pattern-grant types and subsumption engine

### LINKED ISSUE #3492: Pattern-grant types and subsumption engine
Implement the core card algebra types and subsumption checker used at deploy-time
validation and derive-time narrowing checks.

- Implement `PatternGrant` type: (class, owner, recipient, verb, resource_id) as
structured fields; canonical string form `class(owner) @ recipient : verb : resource-id`
- Implement owner path parsing and segment-wise subsumption per §5.5.2
- Implement recipient path parsing: valid depths are 1, 2, or 5; bare leaves rejected;
segment-wise subsumption (wildcards match concrete segments)
- Implement resource-id subsumption per class:
    - Glob: filesystem, config, secret, kv, blob, rdbms
    - Host + port range: network
    - Identifier equality or wildcard: agent, tool, env, card
    - Range containment: oplog
- Implement dual-bound `EffectiveSurface`:
    - Lower: `⋃ E(L_i, H)` — union of lower-bound effectives, recipient-matched against holder H
    - Upper: `⋂ E(U_i, H)` — intersection of upper-bound effectives, recipient-matched;
    starts as `⊤` when no cards with non-trivial upper bounds are held
- Implement multi-parent derivation validation on both bounds:
`E(L_child) ⊆ ⋃ E(L_parent_i)` and `E(U_child) ⊆ ⋃ E(U_parent_i)`
- Implement authorization check:
`(class, owner, recipient, verb, resource_id) ∈ lower ∩ upper`
where recipient is matched against the calling agent's holder identity
- Cover all 34 resource classes (§6.1–§6.34)
- Unit tests: wildcards, truncated owner paths, partial agent-id-cli wildcards
(AgentType(*)), cross-account safety, multi-parent union subsumption, upper bound
intersection (single ceiling card clamps wallet regardless of other grants)

mschuwalow:
resolves #3492

Adds model, parsing and subsumption checking for monomorphic, polymorphic and manifest cards. 
The parsing (parsing_tests.rs) are by far the most mature part for now. Will expand the others as I start using them.
The lifecycle manifest -> polymorphic -> is also still missing here.


golem-common/src/base_model/card/class/filesystem.rs
```diff
@@ -0,0 +1,177 @@
...
+    left: &[FilesystemPathSegmentPattern],
+    right: &[FilesystemPathSegmentPattern],
+) -> bool {
+    if left
+        .first()
+        .is_some_and(|segment| matches!(segment, FilesystemPathSegmentPattern::GlobStar))
+    {
+        return true;
+    }
+    if left.len() != right.len() {
```
vigoo: This (with the `**` check above) will reject things like `/data/**` vs `/data/x/y/z`

golem-common/src/base_model/card/class/network.rs
```diff
@@ -0,0 +1,183 @@
...
+            (
+                Self::HostPort {
+                    host: ah,
+                    ports: ap,
+                },
+                Self::HostPort {
+                    host: bh,
+                    ports: bp,
+                },
+            ) => glob_subsumes(ah, bh) && ap.subsumes(bp),
```
vigoo: this only supports `/` separators, but in a host you may also use `*` in subdomain positions like `*.golem.cloud`, I think that's not supported by it?

golem-common/src/base_model/card/algebra.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct EffectiveSurface {
+    lower: GrantSurface,
+    upper: Vec<GrantSurface>,
+}
+
+impl EffectiveSurface {
+    pub fn from_cards(cards: &[Card], holder: &str) -> Result<Self, CardAlgebraError> {
```
vigoo, line 62: My agent identified a scenario that I think is handled incorrectly with the current logic:

Imagine a card whose upper bound is meant to express "the ceiling is unrestricted, except: never secret:reveal". The natural way to encode that is:

```
Card {
    upper_positive: vec![],                            // ⊤ — no positive narrowing
    upper_negative: vec![secret_reveal_anything()],    // explicit deny
    ...
}
```

What actually happens when that card lands in EffectiveSurface:

1. `upper_positive.is_empty() && upper_negative.is_empty()` is false (negatives are non-empty), so the card is pushed into `self.upper` as a `GrantSurface { positive: [], negative: [secret_reveal_anything] }`.
2. For any incoming request — say `filesystem:read /tmp/foo` — authorize walks the upper list and calls allows on this surface.
3. allows checks `granted = self.positive.iter().any(...)` against an empty `positive → granted = false` → returns `Ok(false)`.
4. authorize returns `Ok(false)`. Filesystem reads are denied — not because of the negative grant, but because there are no positives to gate through.

golem-common/src/base_model/card/class/tool.rs
```diff
@@ -0,0 +1,223 @@
...
+pub struct ToolValueLiteral(pub String);
+
+impl ToolResourcePattern {
+    pub fn any() -> Self {
+        Self::AnyInvocation
+    }
+
+    pub fn command(command: impl Into<String>) -> Self {
+        Self::Invocation(
+            parse_tool_invocation_pattern(&command.into()).expect("invalid tool command"),
```
vigoo: These `expect`s look very unsafe to me - parsing user-provided data, can crash the cli/server, can't they? (Not just in this file, there are several)

golem-common/Cargo.toml
```diff
@@ -92,6 +92,7 @@ itertools = { workspace = true, optional = true }
 lazy_static = { workspace = true, optional = true }
 mac_address = { workspace = true, optional = true }
 nonempty-collections = { workspace = true, optional = true }
+nom = { workspace = true }
```
vigoo: We had `nom` once before and got rid of it in favor of using `combine`. Would it be hard to use that instead? Primary reason just to not have different parsing libraries in different submodules :) (I personally also find `combine` nicer but my primary reason is just consistency)

vigoo (changes requested):

golem-common/Cargo.toml
mschuwalow: Let me switch to combine and also remove nom entirely. It was still left as a workspace dep

golem-common/src/base_model/card/class/tool.rs
mschuwalow: These should not be there, fixed

golem-common/src/base_model/card/algebra.rs
mschuwalow: Fixed

vigoo (approved):

Issue #3493, opened 2026-05-20
Wire accounts, environment shares, and token issuance to cards


mschuwalow:
Replace the existing role-action enum authorization with card-based authorization.
All existing tokens are dropped at deployment; users re-authenticate and receive new
card-backed tokens. This ticket has no agent-facing changes and no WASM enforcement —
it is purely a registry-service and auth-middleware change. After this ticket the card
store is the live authorization substrate for all HTTP API traffic.

**Account and environment → card projection**

- On account creation: mint an account root card (per ticket 4)
- On environment creation: mint an env-default card (per ticket 4)
- On environment share creation: derive a monomorphic card from the env-default card for
the role. Lower bound carries the role's grants; upper bound is `⊤`. Recipient is
anchored to the grantee account's holder path.
    - Admin: all 34 classes wildcarded within the environment owner path
    - Deployer: environment:deploy, environment:rollback, component:update, component:view,
    agent:create, agent:delete, agent:update-revision within the environment
    - Viewer: environment:view, component:view, agent:view, oplog:read within the environment
    Store card_id on the environment share row (add card_id FK to environment_shares)
- On environment share update: hard-delete old card (cascades to descendants), derive
new card from updated role projection
- On environment share deletion: hard-delete the associated card

**Token issuance**

- Truncate tokens table at deployment; drop any legacy columns no longer needed
- New token creation: atomically insert token row and derive a monomorphic card from the
caller's account root card; lower bound filtered to grants scoped to the target account;
upper bound inherited from caller's upper bound; recipient anchored to the token holder.
Store card_id on the token row. Caller cannot mint a token with authority exceeding their
own effective surface on either bound.
- Token deletion: hard-delete the associated card (cascades to descendants); card_id on
the token row becomes null via ON DELETE SET NULL
- Bootstrap path (token_service::create_initial_tokens) derives from the platform root card

**HTTP API authorization cut-over**

- Replace authorize_environment_action and authorize_account_action in golem-service-base
with `check_permission(auth_ctx, class, owner, verb, resource_id)` at each handler.
The recipient is implicitly the calling token's holder identity.
- Token presentation: resolve token_secret → card_id → load card from Postgres; cache
with a short TTL (5 minutes), invalidated when CardRevoked is received for the card_id
- The existing role-action enum is removed; all authorization goes through check_permission
- The (class, owner, verb) triple for each HTTP endpoint is documented inline at the
call site

PR #3582, opened 2026-05-27
agent-permissions: permission share crud

mschuwalow:
part of #3493


golem-registry-service/src/services/permission_share.rs
```diff
@@ -0,0 +1,277 @@
...
+        name: &str,
+        auth: &AuthCtx,
+    ) -> Result<PermissionShare, PermissionShareError> {
+        auth.authorize_account_action(owner_account_id, AccountAction::ViewPermissionShare)?;
+
+        self.permission_share_repo
+            .get_by_owner_and_name(owner_account_id.0, name)
+            .await?
+            .ok_or(PermissionShareError::PermissionShareNotFound(
+                PermissionShareId::new(),
```
vigoo, line 213: I think we shouldn't generate random ids for this error message, could be confusing

vigoo (approved):

golem-registry-service/src/services/permission_share.rs
mschuwalow: Yeah, this should be a dedicated PermissionsShareByNameNotFound error case. Updated

PR #3583, opened 2026-05-28
Handle read-only method cache on the HTTP API level

### LINKED ISSUE #3392: Bind HTTP cache headers for GET on read-only methods
## Goal

When an HTTP request resolves to a read-only agent method, [`golem-worker-service`](https://github.com/golemcloud/golem/tree/main/golem-worker-service) emits cache-friendly response headers and supports `If-None-Match` revalidation, so that browsers, gateways, and CDNs can cache the response.

Part of #3388. Depends on #3391 (need the cache + epoch infrastructure).

### 1. Response headers per cache policy

For `GET` (and `HEAD`) requests resolving to an `AgentMethod` with `read_only.is_some()`:

| `cache-policy` | Response headers |
|----------------|------------------|
| `no-cache`     | `Cache-Control: no-store` |
| `until-write`  | `Cache-Control: <visibility>, no-cache` + `ETag: "<agent-id>:<oplog-index-at-execution>"` |
| `ttl(d)`       | `Cache-Control: <visibility>, max-age=<d>` + same `ETag` |

Where `<visibility>`:

- `read_only.uses_principal == true`  → `private`, plus `Vary: Authorization`
- `read_only.uses_principal == false` → `public` (CDN-friendly)

### 2. `If-None-Match` revalidation (304 path)

On an incoming request with `If-None-Match: "<agent-id>:<index>"` against a read-only `GET` endpoint:

- Fetch the worker's current `oplog_index` (cheaply available via `AgentStatusRecord` without waking the agent).
- If the requested ETag's `<index>` matches the current state (i.e. no non-readonly invocation has happened since the cached response was produced), respond `304 Not Modified` with the same `ETag` and `Cache-Control` headers, **without invoking the executor**.
- Otherwise fall through to the normal invocation path (which itself may then hit the per-worker cache from #3391).

This gives full agent-loading bypass for cold agents whose state hasn't changed.

### 3. Method semantics

Read-only methods bound to non-`GET`/`HEAD` HTTP verbs are allowed but do not get cache headers (warn in the worker-service logs once per (component, method) on first registration).

### 4. Implementation notes

- Header construction sits in the HTTP mapping layer in [`golem-worker-service`](https://github.com/golemcloud/golem/tree/main/golem-worker-service) (the same code path that handles existing endpoint annotations).
- The `<index>` portion of the ETag should be the oplog index at the time the response was *produced*, not the current index when the response is sent — consistent with the cache entry's lifecycle.
- For HEAD, return identical headers with no body.

## Acceptance criteria

- [ ] `GET` to a read-only endpoint returns the correct `Cache-Control`, `ETag`, and (when applicable) `Vary` headers per the table above.
- [ ] `If-None-Match` with a matching ETag returns `304 Not Modified` without invoking the worker-executor (verified by checking that no oplog entries are appended).
- [ ] `If-None-Match` with a non-matching ETag re-runs the request and returns `200` with the new ETag.
- [ ] `uses_principal == false` endpoints emit `public` directive and no `Vary: Authorization`.
- [ ] `uses_principal == true` endpoints emit `private` + `Vary: Authorization`.
- [ ] `no-cache` policy emits `Cache-Control: no-store`.
- [ ] `ttl(d)` policy emits `Cache-Control: ..., max-age=<d>`.
- [ ] Non-`GET`/`HEAD` mappings to read-only methods don't break (no cache headers added; one warning logged per registration).

## Out of scope

- A worker-service-level cache layer (separate from the worker-executor cache) — could be added later if metrics show value.
- Tests → #3393
- Docs → #3394

## Related

- Epic: #3388
- Depends on: #3389, #3391
- Blocked by tests in: #3393

vigoo:
Resolves #3392


golem-common/src/base_model/deployment.rs
```diff
@@ -78,6 +79,13 @@ declare_structs! {
         pub deployment_hash: Hash,
 
         pub current_revision: CurrentDeploymentRevision,
+
+        /// Non-fatal validation warnings produced while compiling the
+        /// deployment (e.g. read-only methods bound to non-`GET` HTTP verbs).
+        /// Empty when the deployment is clean.
+        #[serde(default)]
+        #[cfg_attr(feature = "full", oai(default))]
+        pub validation_warnings: Vec<DeployValidationWarning>,
```
mschuwalow, line 88: 🎉 
Thanks, wanted to do this for a while

golem-worker-service/src/custom_api/call_agent/mod.rs
```diff
@@ -98,19 +125,123 @@ impl CallAgentHandler {
...
+                &mut route_result.headers,
+                &agent_id,
+                read_only_oplog_index,
+                read_only,
+                behaviour,
+                resolved_route,
+            );
+
+            if request_method == Method::HEAD {
+                route_result.body = ResponseBody::NoBody;
```
mschuwalow, line 166: Make sure that this is consistent with the compiled route in the registry service

golem-api-grpc/proto/golem/worker/v1/worker_service.proto
```diff
@@ -161,6 +161,9 @@ message InvokeAgentSuccess {
   optional uint64 fuel_consumed = 2;
   optional uint64 component_revision = 3;
   optional golem.worker.InvocationStatus status = 4;
+  // For read-only agent methods, the oplog index that the representation
+  // returned in `result` was committed at.
+  optional uint64 read_only_oplog_index = 5;
```
mschuwalow: We might want to use the agentfingerprint as part of the etag value. It's going to be very rare, but in principle we could have a deleted, than recreated agent at the same oplog index which would incorrectly not invalidate the cache

mschuwalow (approved):

noise64 (approved):

PR #3584, opened 2026-05-28
agent-permissions: implement permission share logic and validations

mschuwalow:
part of #3493


golem-registry-service/src/repo/card.rs
```diff
@@ -103,23 +104,65 @@ impl<DBP: Pool> DbCardRepo<DBP> {
...
         for parent_id in parent_ids {
             tx.execute(
                 sqlx::query("INSERT INTO card_parents (card_id, parent_id) VALUES ($1, $2)")
-                    .bind(card_id)
+                    .bind(card_id.0)
                     .bind(*parent_id),
             )
             .await
-            .to_error_on_foreign_key_violation(CardRepoError::ParentNotFound(*parent_id))?;
+            .to_error_on_foreign_key_violation(CardRepoError::ParentNotFound(card_id))?;
```
vigoo: Why this change? The error message will be incorrect as it says "Parent card {id} does not exist"

golem-registry-service/src/repo/permission_share.rs
```diff
@@ -256,33 +307,40 @@ impl PermissionShareRepo for DbPermissionShareRepo<PostgresPool> {
...
+                let mut revision = revision;
+                revision.card_id = Some(replacement_card.card_id);
                 let revision = Self::insert_revision(tx, revision).await?;
 
                 let share: PermissionShareRecord = tx.fetch_optional_as(
                     sqlx::query_as(indoc! { r#"
                         UPDATE permission_shares
                         SET name = $1, current_card_id = $2, updated_at = $3, modified_by = $4, current_revision_id = $5
                         WHERE permission_share_id = $6
-                          AND current_revision_id = $7
```
vigoo, line 270: Why don't we need these conditions anymore?

mschuwalow: Concurrent updates / updates based on stale revisions are handled in `insert_revision`. Each update always inserts a revision ad current_revision_id + 1. If there are any concurrent updates, this will fail due to a unique index violation on (permission_share_id, revision_id).

This is also how it is handled in other repos, this previous check was unneeded.

vigoo (approved):

PR #3589, opened 2026-05-29
agent-permissions: account-wide token root cards

mschuwalow:
part of https://github.com/golemcloud/golem/issues/3493

### DIFF
--- a/golem-registry-service/src/repo/permission_share.rs
+++ b/golem-registry-service/src/repo/permission_share.rs
@@ -270,7 +270,7 @@ impl DbPermissionShareRepo<PostgresPool> {
         Ok(inserted)
     }
 
-    async fn invalidate_token_root_card_in_tx(
+    async fn invalidate_token_root_card_and_bump_epoch_in_tx(
         tx: &mut <<PostgresPool as Pool>::LabelledApi as LabelledPoolApi>::LabelledTransaction,
         account_id: Uuid,
     ) -> Result<(), PermissionShareRepoError> {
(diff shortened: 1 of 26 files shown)


golem-registry-service/src/repo/permission_share.rs
```diff
@@ -357,37 +421,48 @@ impl PermissionShareRepo for DbPermissionShareRepo<PostgresPool> {
...
-                    .bind(revision.audit.created_by)
-                    .bind(revision.revision_id)
-                    .bind(revision.permission_share_id),
-                )
-                .await?
-                .ok_or(PermissionShareRepoError::ConcurrentModification)?;
-
-                if let Some(old_card_id) = revision.card_id {
-                    DbCardRepo::<PostgresPool>::delete_tree_in_tx(tx, CardId(old_card_id)).await?;
+        // Share delete removes the share card from the graph.  The FK cascade on
```
vigoo: The agent flagged a potential race condition, please verify:

1. Account has no token root, epoch `E`. Share card `S` exists.
2. Auth thread `A` reads active shares → `[S]`, opens `TX_A`, runs `lock_parent_cards_for_create([S])` `(SELECT … FOR UPDATE on S)`, inserts root `R` with parent `S`, inserts `(R, S)` into `card_parents`, runs the OCC UPDATE on accounts — but has not committed yet.
3. Delete thread `B` opens `TX_B` and runs `delete_tree_in_tx(S)`. The recursive CTE snapshot is taken; because `TX_A` hasn't committed, the CTE does not see `(R, S)` in `card_parents`, so `to_delete = {S}` (no `R`).
4. The `DELETE` blocks on the row lock for `S`. `TX_A` commits. `TX_B` unblocks and deletes `S`. The FK cascade removes `(R, S)` from `card_parents`, but `R` itself is not deleted (it wasn't in `to_delete`).
5. `accounts.token_root_card_id` still points at `R`. Future authentications hit the early-return and use `R`, which is an orphan card with no parents anymore — silently under-granting the user (or, if it had multiple parents, dropping just the deleted share without re-derivation).

openapi/golem-service.yaml
```diff
@@ -14996,13 +14996,20 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/AccountRole'
+        tokenRootCardId:
```
vigoo, line 15016: Are these two fields needed on the public API?

mschuwalow: Probably not needed. I want to keep it for now, but will introduce a public view type for the accounts in the api later if it's still not needed

golem-registry-service/src/repo/permission_share.rs
mschuwalow: Good catch. The issue here is that the card parents have an ON DELETE CASCADE link in both directions (parents and children).

The fix is to make the parent -> child direction non cascading, which will cause a card deletion transaction to fail if the child card tree has changed while the transaction was running. We then need to add service side retries for this error. The actual repo queries already reflect that, just need to adjust the migrations and service side plumbing.

vigoo (approved):

PR #3590, opened 2026-05-29
Add observability metrics for cloud density benchmarks

### LINKED ISSUE #3517: Worker-executor observability metrics for cloud benchmarks
## Parent

#3516

## What to build

Eleven new Prometheus metrics in `golem-worker-executor`, landing as one
observability uplift. None exist today. They serve two audiences: prod operator
dashboards (so operators can diagnose worker-memory pressure, scheduler lag,
and promise-completion load in production), and the cloud density benchmark
sections (which classify breaking points by what was bottlenecked — without
these metrics, density results cannot distinguish memory-bound from
oplog-bound from CPU-bound).

The three metric groups land together because they share `metrics.rs`, the
same wiring style at their call sites, and the same exposition format. The
density buildspec pre-flight checks the presence of all of them via a single
`curl /metrics` call against the deployed executor.

### Worker memory and lifecycle (3 metrics)

- `golem_worker_memory_pool_total_bytes{executor_id}` — gauge, the configured
  worker-memory semaphore size in bytes. Set once at executor startup; matches
  the value reported in the existing `Total memory available for workers`
  startup log line.
- `golem_worker_memory_pool_used_bytes{executor_id}` — gauge, the bytes
  currently acquired from the worker-memory semaphore. The sum of
  `memory_requirement()` across resident agents. Increments on acquire,
  decrements on release.
- `golem_active_workers{executor_id, state}` — gauge, the count of `Worker`s
  in each lifecycle state. The `state` label takes the values `running`,
  `unloaded`, `waiting_for_permit`, and `stopping`. Incremented and
  decremented at the existing `WorkerInstance` state-transition sites.

### Scheduler observability (5 metrics)

- `scheduled_action_lag_seconds{executor_id}` — histogram, the wall-clock
  delay in seconds between a scheduled action's `scheduled_at` and the time
  it actually fires. Observed inside the scheduler's `process()` at fire
  time. Buckets sized to span sub-second to multi-minute lag.
- `scheduler_queue_depth{executor_id}` — gauge, the count of `matching`
  actions to process at the start of each `process()` iteration (i.e., after
  the in-memory shard-ownership filter).
- `scheduler_tick_duration_seconds{executor_id}` — histogram, the wall time
  of a single `process()` iteration.
- `scheduler_actions_dropped_total{executor_id, action_kind}` — counter,
  incremented for each action read from Postgres but filtered out by shard
  ownership (the non-matching branch). The `action_kind` label takes the
  values `invoke`, `complete_promise`, `archive_oplog`, `resume`.
- `scheduled_action_size_bytes{executor_id, action_kind}` — histogram, the
  serialized blob size of a `ScheduledAction` at insert time (observed in
  `add_to_sorted_set`). Buckets at powers of 2 from 1 KB to 64 MB.

### Promise observability (3 metrics)

- `promise_completion_seconds{executor_id, is_cross_pod}` — histogram, the
  wall time of a single `complete_promise` from gRPC entry to write
  committed. The `is_cross_pod` label distinguishes the in-pod path (the
  owning executor handles the completion directly) from the
  `worker_proxy.complete_promise` cross-pod hop. Values `true` or `false`.
- `promise_pending_count{executor_id}` — gauge, the number of distinct
  `PromiseId`s currently in `Pending` state in this executor's
  `PromiseRegistry`.
- `promise_completion_size_bytes{executor_id}` — histogram, the bytes of
  `data` passed to `complete_promise` at call time. Buckets at powers of 2
  from 1 KB to 64 MB.

## Acceptance criteria

- [ ] All eleven metrics are registered in `golem-worker-executor` and exposed
      at the existing `/metrics` endpoint.
- [ ] The three worker-memory/lifecycle gauges update at the
      acquire/release/state-transition sites and are correct in steady state
      (worker-memory `used` matches the sum of `memory_requirement()` across
      resident agents; `active_workers{state=...}` counts match the
      `WorkerInstance` state distribution observable via the worker-executor
      admin gRPC).
- [ ] The five scheduler metrics update inside the scheduler's `process()`
      loop and at scheduled-action insert sites.
- [ ] The three promise metrics update at `complete_promise` entry/exit and at
      `PromiseRegistry` state transitions. The `is_cross_pod` label is set
      correctly via the existing `is_local_worker` shard check.
- [ ] All metrics carry the `executor_id` label sourced from the same value
      already used by existing executor metrics.
- [ ] `curl http://<executor>:<metrics-port>/metrics` returns all eleven
      metric names. The integration suite includes a smoke test that
      `curl`s `/metrics` and asserts each metric name is present.
- [ ] Metric naming follows the existing convention in `metrics.rs` (snake
      case, unit suffix in the name where applicable, descriptive HELP text
      registered alongside each metric).

## Blocked by

None — can start immediately.

kmatasfp:
Resolves: #3517


golem-worker-executor/src/durable_host/golem/v1x.rs
```diff
@@ -223,18 +223,33 @@ impl<Ctx: WorkerCtx> Host for DurableWorkerCtx<Ctx> {
                 Err(other) => Err(other)?,
             };
 
-            let promise_completion_result = if is_local_worker {
+            crate::metrics::promises::record_promise_data_size(data.len());
```
vigoo: I think all these promise metrics should be recorded in the promise service, not here

golem-worker-executor/src/storage/scheduler/memory.rs
```diff
@@ -83,6 +83,19 @@ impl SchedulerStorage for InMemorySchedulerStorage {
         let lease_owner = Uuid::now_v7();
 
         let mut entries = self.entries.lock().unwrap();
+
+        // Count due entries whose shard is not owned by this executor (dropped by shard filter).
+        for (_, entry) in entries.iter() {
```
vigoo: This is the in-memory implementation of the scheduler storage, only used in some unit tests. I don't think we need this

golem-worker-executor/src/metrics.rs
```diff
@@ -227,11 +293,112 @@ pub mod workers {
...
+    }
+
+    pub fn record_worker_kv_cache_value_size(bytes: usize) {
+        WORKER_KV_CACHE_VALUE_SIZE_BYTES
+            .with_label_values(&[crate::metrics::storage::executor_id()])
+            .observe(bytes as f64);
+    }
+
+    #[cfg(test)]
+    mod tests {
```
vigoo: Seems to be useless tests testing that the underlying metrics library works

vigoo (changes requested):

golem-worker-executor/src/metrics.rs
kmatasfp: can remove, just followed the pattern we had in place already, prob should remove the other similar tests as well

vigoo: those were also added by you(r agent) :) just did not notice last time

kmatasfp: done

golem-worker-executor/src/storage/scheduler/memory.rs
kmatasfp: removed

vigoo (approved):

PR #3593, opened 2026-06-01
Value type refactoring 1

### LINKED ISSUE #3395: New value/type model foundations
# Wave 1 — Foundations of the new value/type model

This is the **Wave 1** umbrella for the value/type system refactor. Wave 1 lands every additive piece of the new model — types, WIT package, validators, renderers, derives, canonical encodings, and the adapters/mirror that let later waves migrate consumer subsystems one PR at a time without breaking the build.

## Scope

Wave 1 is **purely additive and parallel-safe**:

- new types live alongside existing ones (`SchemaType` next to `AnalysedType`, `SchemaValue` next to `Value`, `InputSchema`/`OutputSchema` next to `DataSchema`, etc.);
- adapters bridge the two worlds in both directions;
- `golem-wasm` is left in place and functional;
- legacy WIT (`wit-value`/`wit-type`/`value-and-type`/`data-schema`/`element-schema`/etc.) is left in place and functional;
- nothing user-visible changes yet.

After Wave 1, every consumer subsystem (Wave 2) and every wire boundary (Wave 3) can migrate in its own PR while the project stays shippable on every commit.

## Tasks

Each task is a self-contained PR. They can land in any order subject to the dependency sketch at the bottom.

- [ ] **1. WIT v2 package.** New `golem:core/types` v2 with `schema-type` / `schema-value`, plus `fixed-list<T, N>`, `map<K, V>`, and parseable-only stubs for `future<T>` / `stream<T>`. The legacy WIT types stay in place (deleted later in Wave 6). Generate host (`bindgen!`) and guest (`wit-bindgen`) bindings.

- [ ] **2. Rust schema core.** Introduce `SchemaType`, `SchemaValue`, `SchemaGraph`, `TypedSchemaValue`, and `RawValueEnvelope` in `golem-common::schema`. Recursive Rust forms; mechanical conversion to/from the WIT flat form. Structural cases mirroring the WIT type set, first-class `Text` / `Binary` with typed `restrictions`, dedicated capability cases (`Secret`, `QuotaToken`, …), `Path` / `Url` / `Datetime` / `Duration` / `Quantity`, discriminated `Union`, named-type definitions and `Ref(TypeId)`. Carriers transport self-contained `SchemaGraph`s, never bare roots.

- [ ] **3. Validators + placement matrix + subtyping.** Schema well-formedness checks; value-against-schema checks; placement matrix enforcement (constructor / persisted oplog / REST-RPC / public docs / `Custom`); subtyping/assignability module with cycle detection over named/recursive types. First-correct now, polish later.

- [ ] **4. Renderers.** JSON value, JSON Schema, OpenAPI, CLI text, docs renderers — direct from `SchemaType` and **covering every case** of the enum (including `Text`, `Binary`, `Path`, `Url`, `Datetime`, `Duration`, `Quantity`, discriminated `Union`, `Secret`, `QuotaToken`, capability nodes). Splittable per renderer if too big for one PR.

- [ ] **5. Canonical encodings.** Authoritative text / JSON / agent-id / CLI forms for `Text`, `Binary`, `Path`, `Url`, `Datetime`, `Duration`, `Quantity`, `Secret`, `QuotaToken`. These exit-block Wave 1 because Wave 2's agent-id / CLI / secret / REST surfaces depend on them.

- [ ] **6. Derive crate.** New `IntoSchema` / `FromSchema` derive macro crate (replaces `golem-wasm-derive`). **Full attribute surface** (records / variants / primitives plus `text` / `binary` / `role` / `source` / `secret` / `quota_token` / `union(discriminator)` / `constraint` / `rename` / `doc` / `alias` / `example` / `deprecated` / `named` / `transparent`). Recursion handled with a graph-builder reserve/commit step keyed on stable `TypeId`. Handles recursive enums, mutually recursive types, generic types, wrapper types (`Box` / `Rc` / `Arc`), foreign types via manual impls or `transparent` newtypes. `FromSchema` round-trips against a self-contained `SchemaGraph` with structured decode errors. Exposes an **agent-level merge** entry point that combines the per-root `SchemaGraph`s of every constructor / method input / output of an agent into one deduplicated graph per agent, deduplicating by stable `TypeId` and failing on same-`TypeId` body conflicts (§4.22). Backs the per-agent `AgentTypeSchema.schema` field for Rust-SDK-emitted agent types.

- [ ] **7. Extract `component_introspection`.** Pull "read agent declarations / oplog-plugin roles / exports / memories / package / producers from a WASM" out of `golem-wasm` into its own helper (likely `golem-component-introspection` crate or a service-side helper module). Still uses the legacy types underneath; pure isolation. The only surviving WIT-parsing surface in the codebase after Wave 6.

- [ ] **8. Extract `resource_runtime`.** Pull `ResourceStore`, `ResourceTypeId`, and any other helpers needed by the executor outside the value bridge out of `golem-wasm`, likely into a new `golem-executor-runtime-types` module. Required before the value bridge is deleted in Wave 3.

- [ ] **9. New `InputSchema` / `OutputSchema` / `FieldSource` types, agent-layer schema mirrors, and new `ParsedAgentId` shape.** Added to `golem-common`, additive; nothing consumes them yet (consumers migrate in Wave 2).
   - `InputSchema = Parameters(Vec<NamedField>)` with per-field `FieldSource = UserSupplied | AutoInjected(AutoInjectedKind)`.
   - `OutputSchema = Unit | Single(SchemaType)`.
   - Agent-layer mirrors (`AgentTypeSchema`, `AgentConstructorSchema`, `AgentMethodSchema`, `AgentDependencySchema`) carry a per-agent `SchemaGraph` (§4.22). Constructor / method input/output bodies may use `SchemaType::Ref` against that graph — there is no inline-only restriction at the agent layer. `AgentDependencySchema` owns its own graph (independently published).
   - New `ParsedAgentId` in-memory shape carrying parameters as `SchemaValue` against the constructor `InputSchema`. **The CLI text form for agent IDs remains per source language** — handled by the existing `agent_id_display` renderers/parsers, ported in Wave 2.16. The canonical encodings from task 5 feed the individual scalar/capability values (Secret, Quantity, …) and the `SourceLanguage::Other` fallback; there is no single canonical CLI text form replacing the per-language renderers.

- [ ] **10. Adapters in `golem-common`.** The bridge that lets Wave 2 / Wave 3 ship one PR at a time:
   - `AnalysedType → SchemaType` (lossless; `SchemaType` is a strict superset).
   - `Value` / `ValueAndType` ↔ `SchemaValue` / `TypedSchemaValue`.
   - `DataSchema` ↔ `InputSchema` / `OutputSchema`.
   - `ElementSchema` ↔ `SchemaType`.
   - Agent-layer adapters (`AgentType` ↔ `AgentTypeSchema`, `AgentConstructor` ↔ `AgentConstructorSchema`, `AgentMethod` ↔ `AgentMethodSchema`, `AgentDependency` ↔ `AgentDependencySchema`) are **graph-aware**: forward (legacy → schema) produces an empty `SchemaGraph` with bodies fully inlined; reverse resolves refs against the agent's `SchemaGraph` via the existing graph-aware data/element helpers. No inline-only restriction (§4.22).
   - Where conversion is forced lossy, document the loss explicitly.
   Adapters are deleted in Wave 6 once the legacy types disappear.

- [ ] **11. Parallel agent-metadata mirror.** Registry-service emits **both** `AgentType` (legacy) and `AgentTypeSchema` (new — with its per-agent `SchemaGraph` from task 9) during the transition. Migrated consumers (Wave 2) read the new one; unmigrated consumers stay on the old. The mirror itself is deleted in Wave 6 along with the legacy types.

## Exit criteria for Wave 1

Wave 1 is done when all of the following are true:

- the project compiles and all existing tests pass;
- nothing user-visible has changed;
- new types, validators, renderers, derives, canonical encodings, helpers, adapters, and parallel mirror are all in `main`;
- the four open foundational decisions are closed:
  - final shape of `SchemaGraph`,
  - `Quantity` numeric representation,
  - canonical encoding for every Wave-1 scalar / capability,
  - dynamic-walker public API shape (no bare-`SchemaValue` overload anywhere);
- a smoke integration test round-trips a representative typed payload (including a recursive shape) through `IntoSchema` → `SchemaGraph` → WIT `schema-value` → `SchemaGraph` → `FromSchema` and asserts equality.

## Dependency sketch within Wave 1

```
       ╭─── 3 (validators + placement + subtyping)
       ├─── 4 (renderers)
1 ─ 2 ─┤
       ├─── 5 (canonical encodings)
       ╰─── 6 (derive crate)

7   8   9   10   11      (independent of 1–6, parallel-safe)
```

Tasks 1–6 form the schema layer proper; tasks 7–11 are independent extraction / additive-types work that can run in parallel from kickoff.

## Out of scope for Wave 1

- migrating any consumer subsystem (Wave 2);
- changing any wire format or persisted state (Wave 3);
- migrating any SDK (Wave 4);
- consumer-side rich-type sweetening (Wave 5);
- deleting any legacy type, WIT package, crate, or shim (Wave 6).

vigoo:
Resolves #3395 

Note that this first step is mostly just defining all the new data types, but does not start using them yet. We add some parallel views of agent metadata temporarily, so it is possible to step-by-step migrate to the new data models. Also, in this step all the new types are defined in `golem-common` and the derivation in a new crate, but later it is likely that some restructuring will be required, just like we had the core value types in `golem-wasm` before, so the Rust SDK for example can share parts of this. 

Also note that the final step of this epic completely drops `Value`, `WitValue`, `AnalysedType`, the old `AgentType` and all related types, but the process is designed in a way that these intermediate steps are mergeable one by one.


golem-common/src/schema/validation/value.rs
```diff
@@ -0,0 +1,1108 @@
...
+            }
+            match graph.lookup(id) {
+                Some(def) => check(graph, &def.body, v, path, errors, visited_refs),
+                None => errors.push(ValueError::DanglingRef {
+                    path: path.snapshot(),
+                    type_id: id.clone(),
+                }),
+            }
+            visited_refs.remove(id);
+        }
```
noise64: Are resursive values still validated on all levels if we cycle break on the type?

golem-common/src/schema/adapters/data_schema.rs
```diff
@@ -0,0 +1,261 @@
...
+                        tag: e.name.clone(),
+                        body,
+                        // Multimodal unions are not resolved by the generic
+                        // inferred-tag discriminator pipeline: the
+                        // alternative is carried positionally inside the
+                        // outer `list` envelope. The validator special-cases
+                        // `Role::Multimodal` and skips per-branch structural
+                        // discriminator checks, so this slot only needs to
+                        // satisfy the type; pick the cheapest legal rule.
+                        discriminator: DiscriminatorRule::FieldAbsent {
```
noise64, line 145: value validation may fail on these (probe: a multimodal caption: string value → Err([UnionDiscriminatorMismatch])

golem-common/src/schema/render/json_schema.rs
```diff
@@ -0,0 +1,872 @@
...
+            if let Some(t) = inner {
+                collect_union_branch_defs(graph, t, defs, emitted);
+            }
+        }
+        _ => {}
+    }
+}
+
+fn branch_def_key(tag: &str) -> String {
+    format!("union__branch__{}", escape_pointer_token(tag))
```
noise64: Two unions sharing the same tag would collide in the defs

golem-common/src/schema/validation/value.rs
```diff
@@ -0,0 +1,1108 @@
...
+    url: &str,
+    path: &mut ValuePath,
+    errors: &mut Vec<ValueError>,
+) {
+    if url.is_empty() {
+        errors.push(ValueError::UrlEmpty {
+            path: path.snapshot(),
+        });
+    }
+    // Scheme / host validation is parser-level and deferred to the
```
noise64: It seems that these are not enforced in other layers neither. Is that intentional for now?

golem-common/src/schema/render/json_schema.rs
```diff
@@ -0,0 +1,872 @@
...
+        // The def's metadata now lives on `def.body` directly; `render_type`
+        // already attaches inline-node metadata, so no extra `attach_metadata`
+        // call is required here.
+        let mut body = render_type(graph, &def.body, false);
+        if let Some(name) = &def.name
+            && let Some(obj) = body.as_object_mut()
+        {
+            obj.entry("title").or_insert(Value::String(name.clone()));
+        }
+        defs.insert(escape_pointer_token(&def.id.0), body);
```
noise64: this can break on ids and tags with `/` or `~`:
- here it should be raw def.id.0
- in branch_def_key it also should be raw tag
- type_id_with_slash_is_pointer_escaped should be fixed to assert too that $defs key is raw

see spec https://datatracker.ietf.org/doc/html/rfc6901#section-4 where it states that references are unescaped, and then resolved (with the raw value)

golem-common/src/schema/render/json_schema.rs
```diff
@@ -0,0 +1,872 @@
...
+        "contentEncoding".to_string(),
+        Value::String("base64url".to_string()),
+    );
+    if let Some(min) = restrictions.min_bytes {
+        bytes_field.insert("minLength".to_string(), Value::Number(min.into()));
+    }
+    if let Some(max) = restrictions.max_bytes {
+        bytes_field.insert("maxLength".to_string(), Value::Number(max.into()));
+    }
+    let mut mime_field = Map::new();
```
noise64, line 870: Are these still valid with base64urls? (as in, they will be longer with encoding)

golem-common/src/schema/render/json_value.rs
```diff
@@ -0,0 +1,1106 @@
...
+        SchemaType::U16 { .. } => {
+            check_int_range::<u16>(json, path)?;
+            Ok(SchemaValue::U16(json_u64(json, path)? as u16))
+        }
+        SchemaType::U32 { .. } => {
+            check_int_range::<u32>(json, path)?;
+            Ok(SchemaValue::U32(json_u64(json, path)? as u32))
+        }
+        SchemaType::U64 { .. } => Ok(SchemaValue::U64(json_u64(json, path)?)),
+        SchemaType::F32 { .. } => Ok(SchemaValue::F32(json_f64(json, path)? as f32)),
```
noise64: Roundtrip would break on e.g 1e39, as it becomes infinity, should be ranged checked like ints

golem-schema-derive/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+proc-macro2 = { workspace = true }
+quote = { workspace = true }
+syn = { workspace = true, features = ["full"] }
+
+[dev-dependencies]
+chrono = { workspace = true }
+golem-common = { workspace = true, default-features = false, features = ["client"] }
+proptest = { workspace = true }
+test-r = { workspace = true }
+trybuild = "1.0"
```
noise64: is this intentionally not workspace?

golem-common/src/schema/render/json_schema.rs
```diff
@@ -162,8 +177,48 @@ fn collect_union_branch_defs(
...
+/// as a raw `$defs` member name (RFC 6901) and as an OpenAPI
+/// `components.schemas` name (which constrains its key alphabet).
+///
+/// Determinism: `serde_json::to_vec` orders fields by struct definition
+/// order; `UnionBranch` / `SchemaType` / `DiscriminatorRule` / current
+/// `MetadataEnvelope` contain no `HashMap` or other unordered container,
+/// so the serialized bytes — and therefore the hash — are stable across
+/// runs. If future `MetadataEnvelope` extensions introduce unordered
+/// fields, this helper should switch to a canonicalising hasher.
+fn branch_def_key(branch: &UnionBranch) -> String {
```
noise64: I think these will create not-so-great generated type names e.g. when our OpenApi is used to generate a client. Could we use a more tag / name presevering mapping, which handles collisions more gracefully, like we did in bridge TypeNaming?

noise64 (approved):
fixes look good, but added a comment on the union type naming, maybe it's worth to do a followup on that.

PR #3594, opened 2026-06-01
Executor and integration tests and fixes for read-only method caching

### LINKED ISSUE #3393: Read-only method testing
## Goal

Cover the read-only methods feature with new test components and integration tests across all invocation surfaces (executor API, RPC, HTTP). The two key tests are: a slow non-readonly call holds the queue while a read-only call returns immediately from cache, both via the public API and via RPC.

Part of #3388. Depends on #3389, #3390, #3391, #3392.

### 1. New test components

Under [`test-components/`](https://github.com/golemcloud/golem/tree/main/test-components):

**`agent-readonly`** (Rust SDK) — primary test target with a durable agent and the following methods:

- `increment()` — non-readonly write
- `slow_increment(ms: u64)` — non-readonly, sleeps then writes (used to hold the queue)
- `get_count() -> u64` — `#[read_only]` (default = `until_write`, `uses_principal = false`)
- `get_count_for(principal) -> u64` — `#[read_only(uses_principal)]`
- `read_only_with_ttl() -> u64` — `#[read_only(cache = "ttl(2s)")]`
- `pure_compute(x: u32, y: u32) -> u32` — `#[read_only(cache = "no_cache")]`
- `bad_write()` — `#[read_only]` doing a `WriteLocal` → must trap
- `bad_clock_read()` — `#[read_only]` reading wall-clock (`ReadLocal`) → must trap
- `bad_remote_read()` — `#[read_only]` doing an outgoing HTTP call (`ReadRemote`) → must trap
- `bad_random()` — `#[read_only]` calling random (`ReadLocal`) → must trap
- `bad_rpc(target_id)` — `#[read_only]` doing an RPC call (`WriteRemote`) → must trap

Also exposes an HTTP mount with a `GET` mapping for `get_count` (used by H1–H5).

**`agent-readonly-rpc`** (Rust SDK) — RPC caller side, depends on `agent-readonly`. Two agents in one component:

- `Caller` (durable):
  - `read_via_rpc(target_id) -> u64` — calls `agent-readonly`'s `get_count` via RPC
  - `slow_then_read(target_id, ms) -> u64` — fires `slow_increment(ms)` on the target via fire-and-forget, then calls `get_count` via RPC and returns the result
  - `parallel_reads(target_id, n) -> list<u64>` — fires `n` concurrent `get_count` RPCs via `async-invoke-and-await`

**`agent-readonly-bad-ephemeral`** (Rust SDK) — single ephemeral agent with a `#[read_only]` method. Used to assert the ephemeral-conflict error from #3389.

### 2. Worker-executor integration tests

New file `golem-worker-executor/tests/readonly.rs` (or appended to existing test files where natural):

| # | Test | Asserts |
|---|------|---------|
| T1 | `read_only_returns_value_and_writes_no_oplog` | First read on fresh agent → correct value; oplog index unchanged |
| T2 | `read_only_caches_until_write` | Two reads in a row → second is a cache hit, oplog unchanged |
| T3 | `read_only_invalidates_after_write` | read → write → read; second read re-runs, oplog grows by exactly one `Started/Finished` pair |
| T4 | **`read_only_bypasses_queue_during_slow_write`** | Background `slow_increment(2000)` holds the queue; foreground `get_count` returns within < 50ms with the pre-write value |
| T5 | `read_only_bypasses_agent_loading` | Populate cache → evict worker → read again returns cached value without reloading the wasm instance |
| T6 | `principal_aware_caches_per_principal` | Two principals miss separately, then each hits its own; `get_count` (principal-unaware) hits regardless of principal |
| T7 | `ttl_expires_cache_entry` | `read_only_with_ttl` then sleep > TTL then read → cache miss |
| T8 | `no_cache_runs_every_time` | `pure_compute(2,3)` twice → both produce oplog markers, no writes |
| T9 | `read_only_violation_traps_with_typed_error` | Parameterised over `bad_write`, `bad_clock_read`, `bad_remote_read`, `bad_random` → each returns `AgentError::ReadOnlyViolation` with the expected `function_type` |
| T10 | `read_only_oplog_markers_present_when_executed` | Cache miss writes one `Started/Finished` pair; cache hit writes nothing |
| T11 | `concurrent_misses_coalesce` | N concurrent identical first-time reads → exactly one `AgentInvocationStarted` in the oplog |
| T12 | `ephemeral_agent_with_read_only_is_rejected` | Registering `agent-readonly-bad-ephemeral` returns the validation error |

### 3. RPC integration tests

In `golem-worker-executor/tests/readonly.rs` (or extending `tests/rpc.rs`):

| # | Test | Asserts |
|---|------|---------|
| R1 | `rpc_read_only_uses_cache_fast_path` | `Caller.read_via_rpc(target)` twice → target's oplog grows only on first call |
| R2 | **`rpc_read_only_during_slow_target_invocation`** | `Caller.slow_then_read(target, 2000)` → embedded `get_count` returns within < 50ms even though `slow_increment` is still queued on the target |
| R3 | `rpc_parallel_reads_coalesce` | `Caller.parallel_reads(target, 50)` → all results equal; target oplog grows by exactly one `Started/Finished` pair |
| R4 | `rpc_read_only_method_cannot_make_rpc` | Calling `bad_rpc` → traps with `ReadOnlyViolation { function_type: WriteRemote }` |

### 4. HTTP integration tests

New file `golem-worker-service/tests/readonly_http.rs`:

| # | Test | Asserts |
|---|------|---------|
| H1 | `get_returns_etag_and_cache_control` | `GET` returns `Cache-Control: <visibility>, no-cache` and an `ETag` |
| H2 | `if_none_match_returns_304` | Re-`GET` with same `If-None-Match` returns `304` and no executor invocation (oplog unchanged) |
| H3 | `etag_invalidates_after_write` | After a write, the previous `If-None-Match` returns `200` with a new ETag |
| H4 | `principal_unaware_uses_public_cache_directive` | `uses_principal = false` → `Cache-Control: public, ...` and no `Vary: Authorization` |
| H5 | `ttl_method_returns_max_age` | `read_only_with_ttl` HTTP response → `Cache-Control: <visibility>, max-age=2` |

### 5. Test harness additions

- A small helper in [`golem-test-framework`](https://github.com/golemcloud/golem/tree/main/golem-test-framework) to count oplog entries by type between two oplog-index snapshots, used by all "oplog grew by exactly N" assertions.
- A `#[timeout]` wrapper around T4 / R2 so the test deterministically fails if the queue-bypass regresses.

## Acceptance criteria

- [ ] All three test components build via `cargo make rebuild-test-components` (or the targeted component build).
- [ ] All T1–T12 pass on `cargo make worker-executor-tests`.
- [ ] All R1–R4 pass.
- [ ] All H1–H5 pass on `cargo make integration-tests`.
- [ ] T4 and R2 fail (with a clear timeout message) if the queue-bypass regresses.
- [ ] Test components are added to the build manifests so they appear in CI.

## Out of scope

- Per-SDK marker unit tests (smoke tests live with each SDK as part of #3389).

## Related

- Epic: #3388
- Depends on: #3389, #3390, #3391, #3392

vigoo:
Resolves #3393


golem-worker-executor/src/worker/mod.rs
```diff
@@ -885,10 +933,191 @@ impl<Ctx: WorkerCtx> Worker<Ctx> {
...
+            // The spawned closure runs `invoke_and_await_uncoalesced` to
+            // bypass this coalescing path. Returning Err removes the pending
+            // entry so a later caller retries (failures must not poison the
+            // cache).
+            //
+            // The closure is spawned via `tokio::task::spawn` (see
+            // [`Cache::get_or_insert_spawned`]) so that cancellation of any
+            // single Await caller does NOT leave the pending entry stuck
+            // forever — the spawned owner future survives caller drop and
+            // resolves the entry one way or the other.
```
noise64, line 1056: Does this break our idempotency key guarantees? As in, only the first one is recorded, so later repeated read_only request might return different results. If this is true, I'm unsure whether this is something to fix, or document.

noise64 (approved):
Added a question on read_only idempotency key behaviour, otherwise looks good

golem-worker-executor/src/worker/mod.rs
vigoo: I don't think so, this means that if one invocation is already in progress that will satisfy other invocations then they wait for it and not do it multiple times. The code is only executed once.

PR #3595, opened 2026-06-01
agent-permissions: account root card

mschuwalow:
part of https://github.com/golemcloud/golem/issues/3493


golem-common/src/base_model/account.rs
```diff
@@ -128,18 +129,14 @@ declare_structs! {
...
         pub email: AccountEmail,
+        pub roles: Vec<AccountRole>
     }
 
     pub struct AccountUpdate {
         pub current_revision: AccountRevision,
         pub name: Option<String>,
     }
 
-    pub struct AccountSetRoles {
```
mschuwalow, line 138: Removed the option to set roles after creation as that would force rotation of the account root card -> all cards in the account become invalid.

Instead, roles need to be specified during creation and are considered when creating the account root card.

golem-registry-service/db/migration/postgres/025_token_card_id.sql
```diff
@@ -1,3 +0,0 @@
-ALTER TABLE tokens ADD COLUMN card_id UUID NULL REFERENCES cards(card_id) ON DELETE SET NULL;
```
mschuwalow, line 1: Leftover from last pr. I decided to only have a shared token root card (for now). If we ever need token specific cards later we can readd it (they would derive from both the token root card and the account root card)

cli/golem-cli/src/command_handler/account.rs
```diff
@@ -93,6 +93,7 @@ impl AccountCommandHandler {
             .create_account(&AccountCreation {
                 name: account_name,
                 email: AccountEmail::new(account_email),
+                roles: Vec::new(),
```
mschuwalow, line 96: We may need to add something here to enable creation / promotion of accounts to admin / ops accounts that can at least impersonate (so we don't have to use admin for it). Will think about it some more in followups.

golem-registry-service/src/services/auth.rs
```diff
@@ -180,34 +190,83 @@ impl AuthService {
...
             }
 
-            let parent_ids = self
-                .permission_share_service
-                .active_share_cards_for_target(account_id)
-                .await?
-                .into_iter()
-                .map(|card| card.card_id)
-                .collect();
+            EffectiveSurface::from_cards(&parent_cards, &account_holder)
```
mschuwalow, line 235: Just a sanity check to ensure we are deriving legal cards

golem-registry-service/src/services/auth.rs
```diff
@@ -180,34 +190,83 @@ impl AuthService {
...
+            parent_cards.push(account_root_card);
+            parent_cards.extend(share_cards);
 
-                expected_epoch = account.token_root_card_epoch;
+            let mut lower_positive = Vec::new();
+            let mut lower_negative = Vec::new();
+            let mut upper_positive = Vec::new();
+            let mut upper_negative = Vec::new();
+            for card in &parent_cards {
+                add_card_grants(
```
mschuwalow, line 226: I'll try / will have to make this smarter in the future, this is more of a placeholder for now

golem-registry-service/src/services/account/card.rs
```diff
@@ -0,0 +1,338 @@
...
+    EnvironmentMcpDeploymentResourcePattern, EnvironmentPluginGrantResourcePattern,
+    EnvironmentResourceDefinitionResourcePattern, EnvironmentResourcePattern,
+    EnvironmentRetryPolicyResourcePattern, EnvironmentSecuritySchemeResourcePattern,
+    EnvironmentShareResourcePattern, FilesystemResourcePattern, KvResourcePattern,
+    NetworkResourcePattern, OplogResourcePattern, PermissionPattern, PlanResourcePattern,
+    RdbmsResourcePattern, SecretResourcePattern, SystemResourcePattern, SystemVerb,
+    ToolResourcePattern,
+};
+
+pub(super) fn account_root_card_record(account_id: AccountId, roles: &[AccountRole]) -> CardRecord {
```
mschuwalow, line 40: I will see about optimizing this further later, more of a placeholder for now

golem-registry-service/src/repo/card.rs
```diff
@@ -191,9 +204,62 @@ impl DbCardRepo<PostgresPool> {
     }
 }
 
+impl DbCardRepo<SqlitePool> {
```
vigoo, line 207: The new per-card loop:

```
for row in rows {              // rows from recursive CTE
    DELETE FROM card_parents WHERE card_id = $1;   // only this card's incoming-parent edges
    DELETE FROM cards         WHERE card_id = $1;
}
```

SQLite returns the recursive-CTE rows anchor-first. With `PRAGMA foreign_keys = ON`, deleting the root fails because `card_parents` still has rows where `parent_id = root_id` and that FK is non-cascading.

This deterministically breaks:
- `accounts.delete` whenever the account has issued a token (token root card is a child of the account root card)
- the SQLite variant of `permission_share.invalidate_token_root_card_without_epoch_bump_in_tx` if the token root has any descendants (currently it doesn't, but the function is generic)

golem-registry-service/src/services/auth.rs
```diff
@@ -180,34 +190,83 @@ impl AuthService {
...
             }
+
+            // Creating the card failed, refetch account to ensure we are working with up-to-date data on next loop.
+            // This will happen if a new permission share is concurrently created or invalidated
+
+            let account = self
+                .account_service
+                .get(account_id, &AuthCtx::System)
+                .await
+                .map_err(|_| AuthError::CouldNotAuthenticate)?;
```
vigoo: If we drop the error details, we should at least log it as a warning here, I think

golem-registry-service/src/repo/card.rs
mschuwalow: Good catch, I forgot that we are enabling these for cli. Let me adjust our sqlite integration tests to have foreign keys.

vigoo (approved):

PR #3596, opened 2026-06-02
feat: cloud-mode TestMode::Cloud for benchmarks with best-effort cleanup

### LINKED ISSUE #3518: Cloud-mode benchmark client with cleanup helper
## Parent

#3516

## What to build

A new \`TestMode::Cloud\` variant for the benchmark test framework that talks
to a deployed Golem through public Gateway-API hostnames, plus the
best-effort cleanup helper every cloud-perf benchmark needs to avoid
accumulating orphaned accounts, apps, envs, and components across runs.

This slice produces a benchmark binary that can be pointed at any deployed
Golem environment and run the existing seven benchmarks against it,
exercising the same code paths the local mode uses through a different
\`TestDependencies\` implementation. It does not yet ship a workflow or
runbook — those land in a later slice. Demoable on its own by pointing the
binary at a local \`TestMode::Spawned\` cluster via the cloud-mode code path
and observing successful run + cleanup.

### Cloud-mode test dependencies

A new \`TestMode::Cloud\` variant in the benchmark config, wired through
\`BenchmarkTestDependencies::new\`. The variant carries:

- \`registry_service_url\` — the Gateway hostname for registry-service.
- \`worker_service_url\` — the Gateway hostname for worker-service.
- \`apps_base_domain\` — used to compose \`{env_id}.{apps_base_domain}\`.
- Admin account ID, admin account email, admin token, default plan ID,
  builtin-plugin-owner account ID.
- Optional shard-manager gRPC host and port — used only for the
  local/remote RPC labels in the throughput benchmark when a port-forward
  is configured.

Internal-component getters return panic-on-use stubs in cloud mode:
\`rdb()\`, \`redis()\`, \`redis_monitor()\`, \`worker_executor_cluster()\`. The
\`shard_manager()\` getter returns a real \`ProvidedShardManager\` when the
optional host/port are configured, a panic-stub otherwise. The
\`low_*_plan()\` methods on \`CloudRegistryService\` panic with a clear message
naming the calling benchmark so the failure mode is unambiguous.

### Cloud component impls

New impls in the benchmark test-framework's components module:
\`CloudRegistryService\`, \`CloudWorkerService\`, \`CloudWorkerExecutorCluster\`
(panic-stub). \`CloudRegistryService\` and \`CloudWorkerService\` dial their
respective Gateway hostnames via a tuned HTTP client (see below).

### Tuned reqwest client

A single constructor produces the cloud-mode HTTP client with: HTTP/2 prior
knowledge, \`pool_max_idle_per_host(1024)\`, 90-second idle timeout,
\`tcp_nodelay(true)\`, and a 180-second request timeout. The client must not
become the bottleneck before the platform does.

### CLI surface

The benchmark binary takes a new \`cloud\` subcommand whose flags map 1:1 onto
the \`TestMode::Cloud\` config struct. Flags:

\`\`\`
--registry-service-url <url>
--worker-service-url <url>
--apps-base-domain <domain>
--admin-account-id <uuid>
--admin-account-email <email>
--admin-account-token <token>
--builtin-plugin-owner-account-id <uuid>
--default-plan-id <uuid>
--shard-manager-grpc-host <host>     (optional)
--shard-manager-grpc-port <port>     (optional)
--suite <path>
--json                                (existing flag, preserved)
\`\`\`

### Best-effort cleanup helper

A new \`cleanup_user_state(&user, &env_id)\` helper lives next to the existing
\`delete_workers\` helper in the integration-tests benchmarks module. Every
existing cloud-perf benchmark's \`cleanup_iteration\` calls it after
\`delete_workers\`. The helper runs five cascading deletes:

1. List and delete all components in the environment.
2. List and delete any domain registrations created during the run.
3. Delete the environment.
4. Delete the application.
5. Delete the account.

Each step is best-effort: failures are logged with a warning and the run
continues to the next step. A doc comment on the helper references
golemcloud/golem#3291 noting cascading delete is incomplete and orphaned
state may remain.

### Run-id naming convention

Benchmark-created accounts, applications, and environments are prefixed
\`bench-{run-id}-\` where \`run-id\` is a UUID generated at suite start. The
prefix makes external garbage collection of leaked state tractable. The
run-id is also written into the result JSON metadata.

### Throughput benchmark routing-table fetch

When \`shard_manager_grpc_host\` is configured, the throughput benchmark
fetches the routing table at iteration setup and continues labelling RPC
pairs as \`local-\` or \`remote-\`. When unconfigured, the label is dropped
and pairs go in a single bucket — the iteration completes either way.

### Pre-flight warmup

At suite start, the benchmark binary issues approximately 50 GET requests
to \`/healthcheck\` on both registry-service and worker-service, then
performs approximately 5 throwaway \`invoke_and_await_agent\` calls against
a warmup agent (and deletes it). Runs once per suite, not per iteration.
Total cost approximately 20–30 seconds.

## Acceptance criteria

- [ ] \`TestMode::Cloud\` variant exists in the benchmark config and is wired
      through \`BenchmarkTestDependencies::new\`.
- [ ] All four internal-component getters (\`rdb\`, \`redis\`,
      \`redis_monitor\`, \`worker_executor_cluster\`) return panic-on-use stubs
      in cloud mode.
- [ ] \`shard_manager()\` returns a real \`ProvidedShardManager\` when host/port
      are configured and a panic-stub otherwise.
- [ ] \`low_*_plan()\` methods on \`CloudRegistryService\` panic with a message
      naming the calling benchmark.
- [ ] The cloud-mode HTTP client is constructed with HTTP/2 prior knowledge,
      pool size 1024, 90-second idle, \`tcp_nodelay\`, 180-second timeout.
- [ ] The benchmark binary's \`cloud\` subcommand exposes all CLI flags above
      and produces a valid \`TestMode::Cloud\` instance.
- [ ] \`cleanup_user_state\` exists in the integration-tests benchmarks module
      and runs all five cascading deletes, logging each failure as a
      warning and continuing to the next step.
- [ ] Every existing cloud-perf benchmark's \`cleanup_iteration\` calls
      \`cleanup_user_state\` after \`delete_workers\`.
- [ ] Benchmark-created accounts/apps/envs carry the \`bench-{run-id}-\` prefix
      and the run-id appears in the result JSON metadata.
- [ ] The throughput benchmark falls back to the unlabeled single-bucket
      mode when shard-manager host/port are not configured.
- [ ] The pre-flight warmup runs once at suite start and is skipped per
      iteration.
- [ ] Unit tests for \`cleanup_user_state\` exercise selective failure of each
      cascading step with a mock client and assert each failure is logged
      and the next step still runs.
- [ ] An integration smoke test runs at least one existing benchmark
      against a local \`TestMode::Spawned\` cluster via the cloud-mode code
      path end-to-end.

## Blocked by

None — can start immediately.

kmatasfp:
Resolves: #3518


vigoo (changes requested):
We already have a `ProvidedXXX` variant of _all_ test dependencies, created exactly for this. Please don't add duplicated for them with a worse name (`PanicXXX`).

vigoo (approved):

PR #3597, opened 2026-06-02
Merging the docs repo to golem

### LINKED ISSUE #3569: Merge the docs repo to golem
Including it's deployment flow

vigoo:


mschuwalow (approved):
Do we want a github action workflow for the tagged releases as well?

vigoo:
> Do we want a github action workflow for the tagged releases as well?
I would keep that manual for now, but added some bun/makefile/whatever commands that make it simple to copy `next` as a specific version

PR #3605, opened 2026-06-03
Value type refactoring 2

### LINKED ISSUE #3396: Internal migrations to the new value model
# Wave 2 — Internal subsystem schema adoption

This is the umbrella for the Wave 2 **internal** migrations of the value/type system refactor: each subsystem listed here is migrated in place to consume the new schema types (`SchemaType` / `SchemaValue` / `SchemaGraph` / `TypedSchemaValue` / `InputSchema` / `OutputSchema` / new `ParsedAgentId`), bridged by the Wave 1 adapters and the parallel agent-metadata mirror at every outward edge.

Wave 2 covers the **internal** subsystems only. The external-protocol exporters (MCP, OpenAPI) are tracked separately in #3397 and #3398.

## Scope

Each task is a small, self-contained PR that:

- migrates one subsystem to use new schema types as its source of truth internally;
- uses Wave 1 adapters where it talks to subsystems that haven't migrated yet, so the project stays compiling and shipping after every PR;
- does **not** change any external wire format, persisted format, or external rendering;
- does **not** delete any legacy type — that happens in Wave 6.

All five tasks are parallel-safe with each other and can land in any order once Wave 1 (#3395) is in place.

## Tasks

- [ ] **12. Agent secret surface.** Migrate the agent-secret stack off `AnalysedType` / `ValueAndType` and onto `SchemaType` / `SchemaValue` plus the Wave 1 canonical `Secret` encoding. Touches:
  - [`golem-service-base/src/model/agent_secret.rs`](https://github.com/golemcloud/golem/blob/main/golem-service-base/src/model/agent_secret.rs),
  - [`golem-registry-service/src/services/agent_secret.rs`](https://github.com/golemcloud/golem/blob/main/golem-registry-service/src/services/agent_secret.rs),
  - [`cli/golem-cli/src/command_handler/secret.rs`](https://github.com/golemcloud/golem/blob/main/cli/golem-cli/src/command_handler/secret.rs) — the CLI handler keeps calling `parse_type_for_language` / `parse_value_for_language`; only the produced types flip to `SchemaType` / `SchemaValue`, and the canonical `Secret` encoding from Wave 1.5 applies to the `Secret` node specifically.
  REST/gRPC payload shapes are kept legacy at the boundary via Wave 1 adapters; those flip in Wave 3.

- [ ] **13. `AgentError::CustomError`.** Migrate the variant in [`golem-common/src/model/agent/mod.rs`](https://github.com/golemcloud/golem/blob/main/golem-common/src/model/agent/mod.rs) from `CustomError(ValueAndType)` to `CustomError(TypedSchemaValue)`. Update every producer and consumer to use the new carrier; bridge with Wave 1 adapters where the error crosses an unmigrated boundary. Display/render code uses the dynamic-walker rule (`(SchemaGraph, SchemaType, SchemaValue)`).

- [ ] **14. Component metadata.** Switch [`golem-common/src/model/component_metadata.rs`](https://github.com/golemcloud/golem/blob/main/golem-common/src/model/component_metadata.rs) from the in-`golem-wasm` analysis machinery to the extracted `component_introspection` helper from Wave 1. Producers (registry-service) start producing the new shape alongside the legacy one (mirror from Wave 1.11); consumers can read either.

- [ ] **15. Durable host helpers.** Migrate the executor durable host glue off plain-WIT carriers, one file per PR if convenient:
  - [`golem-worker-executor/src/durable_host/golem/agent.rs`](https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/durable_host/golem/agent.rs) drops `AnalysedType` / `WitType` / `WitValue`;
  - [`golem-worker-executor/src/durable_host/wasm_rpc/mod.rs`](https://github.com/golemcloud/golem/blob/main/golem-worker-executor/src/durable_host/wasm_rpc/mod.rs) drops `ValueAndType` / `UntypedDataValue`.
  Wasmtime-side calls are still routed through the legacy bridge until Wave 3.19; this PR only changes the Rust-side data shape consumed/produced by the helpers.

- [ ] **16. CLI rendering stack.** Migrate the CLI display/parse path off `AnalysedType` / `ValueAndType` and onto `SchemaType` / `SchemaValue`. **Per-language parsers/renderers and the `SourceLanguage` dispatch are preserved**; only the type they produce/consume changes:
  - [`cli/golem-cli/src/agent_id_display/`](https://github.com/golemcloud/golem/blob/main/cli/golem-cli/src/agent_id_display) keeps `parse_value_rust` / `parse_value_ts` / `parse_value_scala` / `parse_value_moonbit`, `parse_type_*`, `render_*`, and the `SourceLanguage` enum (Rust / TypeScript / Scala / MoonBit / Other). They are ported in place to produce/consume `(SchemaGraph, SchemaType)` / `SchemaValue` so `SchemaType::Ref` resolves against the agent's `SchemaGraph` while rendering. `ParsedAgentId` from Wave 1.9 uses these renderers for its CLI text form.
  - [`cli/golem-cli/src/bridge_gen/`](https://github.com/golemcloud/golem/blob/main/cli/golem-cli/src/bridge_gen) — per-language code generation (Rust / TypeScript) is inherently language-specific and stays per-language; its input flips from `AnalysedType` to `(SchemaGraph, SchemaType)`. Named types in the agent's `SchemaGraph` are emitted as first-class generated types rather than duplicated inline structs.
  - [`cli/golem-cli/src/model/invoke_result_view.rs`](https://github.com/golemcloud/golem/blob/main/cli/golem-cli/src/model/invoke_result_view.rs) renders typed results via the per-language renderers fed `(SchemaGraph, SchemaType, SchemaValue)`.
  - [`cli/golem-cli/src/model/deploy.rs`](https://github.com/golemcloud/golem/blob/main/cli/golem-cli/src/model/deploy.rs) consumes the new agent-metadata shape.
  Phase-B canonical encodings from Wave 1.5 are used as the underlying JSON form for individual scalar/capability values and as the `SourceLanguage::Other` fallback — **not** a replacement for the per-language CLI surface.

## Exit criteria for Wave 2 (this issue)

Wave 2 is done when all of the following are true:

- all five tasks above are merged;
- the project compiles, all existing tests pass, and behaviour against existing wire formats is byte-identical (legacy adapters at every edge);
- the new types are the **internal** source of truth for every subsystem listed above;
- nothing user-visible has changed yet.

## Dependencies

  - new types and adapters (1.2, 1.10);
  - `InputSchema` / `OutputSchema` / `FieldSource` / new `ParsedAgentId` (1.9) — needed by 16;
  - canonical encodings (1.5) — needed by 12, 13, 16;
  - extracted `component_introspection` (1.7) — needed by 14;
  - parallel agent-metadata mirror (1.11) — needed by 14, 16.
- Independent of MCP migration (#3397) and OpenAPI migration (#3398); those can land in parallel.
- Unblocks Wave 3 boundary cutovers — once an internal subsystem is on the new types, its outward boundary can flip cleanly in Wave 3.

## Out of scope for this issue

- migrating MCP export → see #3397;
- migrating OpenAPI export → see #3398;
- changing any wire format / persisted format → Wave 3;
- migrating any SDK → Wave 4;
- consumer-side rich-type sweetening → Wave 5;
- deleting legacy types or `golem-wasm` → Wave 6.

vigoo:
Resolves #3396

This PR migrates some internal places to the new schema but leaves WIT, gRPC and REST interfaces as they were. For this reason, it temporarily introduces conversions in many places, which are going to be eliminated in the follow-up steps.

- Agent secrets 
- `AgentError` payload
- Secret and RPC host functions in the executor
- Language-specific parsers and renderers in the CLI


golem-worker-executor/src/durable_host/golem/agent.rs
```diff
@@ -368,24 +430,45 @@ impl<Ctx: WorkerCtx> Host for DurableWorkerCtx<Ctx> {
 
         let declaration = agent_type.config.iter().find(|c| c.path == path);
 
+        // Deterministic check: the guest-supplied expected type must
+        // match the type declared on the component. Both are inline
+        // `SchemaType`s, so structural equality is sufficient. Run
+        // before dispatching to the local/secret resolver so a
+        // mismatch never opens a durable function (secret path) nor
+        // touches local state.
+        let declaration_value_type = declaration
```
mschuwalow, line 456: The reason to have expected and declared type was the following:
* declared: the type of the secret in the component at that specific point of replay (not changed after automatic updates due to this being driven by the oplog index)
* expected: the type of the secret requested by the agent at runtime. This does change after automatic updates as the wasm trying to read the secret has changed.

So after an automatic update expected might be != declared. We can handle certain cases like new optional secrets (which is still being done), but other's as well. I didn't get around to ever doing that, but the idea behind that code shape was to do this on the fly transformation of data in  resolve_local_config / resolve_secret_config

mschuwalow (approved):
Only had a rough look, especially for the cli parts.
But the parts I checked look good to me

golem-worker-executor/src/durable_host/golem/agent.rs
vigoo: restored it

cli/golem-cli/src/agent_id_display/mod.rs
```diff
@@ -216,77 +236,112 @@ pub fn parse_value_for_language(
...
+/// Resolve a [`SchemaType::Ref`] against the supplied graph, returning
+/// the def body and the def name when present. Non-ref types pass through
+/// with `None` def name.
+pub(crate) fn resolve_named_ref<'a>(
+    graph: &'a SchemaGraph,
+    ty: &'a SchemaType,
+) -> (&'a SchemaType, Option<String>) {
+    match ty {
+        SchemaType::Ref { id, .. } => match graph.lookup(id) {
+            Some(def) => (&def.body, def.name.clone()),
```
noise64: Is this clone needed?

cli/golem-cli/src/agent_id_display/parse_common.rs
```diff
@@ -50,66 +55,68 @@ impl From<LexError> for ParseError {
...
-/// The shared framework handles the top-level `DataSchema` dispatch,
-/// element dispatch, and common primitives (bool, integers, floats, strings).
-/// Dialect implementations provide language-specific parsing for complex
-/// CM types, unstructured elements, and field-name normalisation.
+/// The shared dispatcher [`parse_cm_value`] handles primitives, list, ref
+/// resolution, and the canonical encoders for the rich semantic types
+/// (text/binary/path/url/datetime/duration/quantity/secret/quota-token).
+/// Dialect implementations cover the structural composites whose syntax
+/// varies between languages.
+#[allow(dead_code)]
```
noise64: Is this intentional?

cli/golem-cli/src/bridge_gen/type_naming/mod.rs
```diff
@@ -88,228 +121,372 @@ impl<TN: TypeName> TypeNaming<TN> {
...
+    }
+
+    pub fn type_name_for_type(&self, typ: &SchemaType) -> Option<&TN> {
+        self.types.iter().find_map(|(t, n)| (t == typ).then_some(n))
     }
 
-    pub fn types(&self) -> impl Iterator<Item = (&AnalysedType, &TN)> {
-        self.types.iter()
+    pub fn types(&self) -> impl Iterator<Item = (&SchemaType, &TN)> {
+        self.types.iter().map(|(t, n)| (t, n))
```
noise64, line 158: is this NOP?

cli/golem-cli/src/bridge_gen/typescript/mod.rs
```diff
@@ -1144,23 +1186,33 @@ impl TypeScriptBridgeGenerator {
...
+                // cast the value at the access point to keep the
+                // generated TS type-correct.
+                writer.write_line(format!("'ok' in {} ?", value));
                 writer.indent();
-                let ok_expr = match &result.ok {
+                let ok_expr = match spec.ok.as_deref() {
                     Some(ok_type) => {
-                        let encoded = self.encode_wit_value(&format!("{}.ok", value), ok_type);
+                        let encoded =
+                            self.encode_schema_value(&format!("({} as any).ok", value), ok_type)?;
```
noise64, line 1205: How is it possible, that we did not need `any` before?

cli/golem-cli/src/bridge_gen/typescript/mod.rs
```diff
@@ -1761,326 +1830,377 @@ impl TypeScriptBridgeGenerator {
...
+                    })
+                    .collect::<Vec<_>>()
+                    .join(", ");
+                let raw_to_js_pairs = flags
+                    .iter()
+                    .map(|name| format!("['{}', '{}']", name, self.to_js_ident(name)))
+                    .collect::<Vec<_>>()
+                    .join(", ");
+                format!(
+                    "((v: unknown) => {{ if (!Array.isArray(v)) {{ throw new Error(`Expected array of flag names, got ${{v}}`); }} const result = {{ {flag_initializers} }}; const __flagMap: Record<string, string> = Object.fromEntries([{raw_to_js_pairs}]); for (const flag of v) {{ if (typeof flag !== 'string') {{ throw new Error(`Expected string flag name, got ${{flag}}`); }} const __js = __flagMap[flag]; if (__js === undefined) {{ throw new Error(`Unknown flag name ${{flag}}`); }} (result as Record<string, boolean>)[__js] = true; }} return result; }})({value})"
```
noise64: this is very hard to read :D (and i know it was like this before) but i _think_ it misses a `;` before `const result`?
(maybe we should have helpers for more coplex snippets?)

cli/golem-cli/src/command_handler/worker/mod.rs
```diff
@@ -2817,16 +2831,156 @@ fn parse_method_argument_value(
...
+        TextType, UnstructuredBinaryElementValue, UnstructuredTextElementValue,
+    };
+    use golem_common::schema::adapters::data_schema_to_input_schema;
+    use golem_common::schema::adapters::value::schema_value_to_value;
+    use golem_common::schema::agent::InputSchema;
+    use golem_common::schema::graph::SchemaGraph;
+    use golem_common::schema::schema_value::{BinaryValuePayload, SchemaValue, TextValuePayload};
+
+    // Multimodal inputs are rejected by data_schema_to_input_schema; only
+    // tuple inputs are accepted here, which mirrors the legacy parser path.
```
noise64, line 2894: So we had no multimodal support previously?

cli/golem-cli/src/bridge_gen/type_naming/mod.rs
vigoo: it's `&(SchemaType, TN) -> (&SchemaType, &TN)`

cli/golem-cli/src/agent_id_display/mod.rs
vigoo: no.

cli/golem-cli/src/agent_id_display/parse_common.rs
vigoo: we previously had some generic code to avoid duplications, with these per-dialect customizations. i'll check if we can restore that (and if not, delete these)

cli/golem-cli/src/bridge_gen/typescript/mod.rs
vigoo: it's not missing a `;` but i'll make it more readable :)

golem-worker-executor/src/durable_host/wasm_rpc/mod.rs
noise64, line 1111: Should this also use output type validation?

cli/golem-cli/src/command_handler/worker/mod.rs
vigoo: Yes, I noticed that multimodal agent constructor parameters were not handled / were disabled in some places before. I don't know why, I definitely added support for it in the old `AgentId` syntax

noise64 (approved):

cli/golem-cli/src/bridge_gen/type_naming/mod.rs
noise64: 🤦 👌

cli/golem-cli/src/bridge_gen/typescript/mod.rs
vigoo: Agent's explanation:

The exact alias is:
```ts
type JsonResult<Ok, Err> = { ok: Ok; err?: undefined } | { ok?: undefined; err: Err };
```
This is the encode path: input `value` has static type `JsonResult<Ok, Err>`.

**Why the old code needed no `any`.** The old discriminator was `value.ok !== undefined`. That's a *value* comparison, and TS treats it as a **narrowing type guard**: in the truthy branch it narrows the union to `{ ok: Ok; err?: undefined }`, so `value.ok` has type `Ok` (not `Ok | undefined`). Passing that straight into the ok-encoder (which expects `Ok`) typechecks — no cast.

**Why it had to change.** `value.ok !== undefined` is wrong at runtime: a valid `Ok` payload can itself be `undefined`. `Ok(())` (unit) and `Ok(None)` both encode to `{ ok: undefined }`. The old test sees `value.ok === undefined` and mis-routes them to the **err** branch, corrupting the value. The fix is `'ok' in value`, a *key-presence* test that correctly separates `{ ok: undefined }` (key present → Ok) from `{ err: … }` (no `ok` key).

**Why the fix costs an `any`.** For `'ok' in value` to narrow a union, `ok` must be present in some arms and absent in others. But this alias deliberately puts `ok` in **both** arms — the err arm has `ok?: undefined`. Since `ok` is a member of every arm, `'ok' in value` does **not** narrow; `value` stays the full `JsonResult`, so `value.ok` keeps type `Ok | undefined`, which isn't assignable to the ok-encoder's `Ok` parameter. Hence `(value as any).ok` to suppress it.

```diagram
discriminator        narrows?   value.ok type   runtime-correct?
─────────────────────────────────────────────────────────────────
value.ok !== undef   yes        Ok              NO  (Ok(())/Ok(None) → err)
'ok' in value        no         Ok | undefined  yes (needs `as any`)
```

PR #3609, opened 2026-06-03
Toon format, stream event cleanups and manifest upgrade handling

### LINKED ISSUE #3373: TOON output format for the CLI

noise64:
- resolves https://github.com/golemcloud/golem/issues/3373
- adds toon format to the CLI
- to support streaming output, it uses a custom pre and post marker, which are documented in the skills: this was tested with an older model (codex-gpt-5.3), and it understood the output, and was able to "translate" it into e.g. a table in the chat
- also bumps the manifest schema version and the manifest version, and adds migration for these versions (this way preparing the place if we need any other migrations during 1.6.0)
- added some skills for the above
- TODO: deploy json schema


cli/golem-cli/src/context.rs
```diff
@@ -124,11 +124,16 @@ impl Context {
             }
         };
 
-        let preloaded_app = ApplicationContext::preload_application(
-            ApplicationContextConfig::app_source_mode_from_global_flags(&global_flags),
-            global_flags.dev_mode,
+        let app_source_mode =
+            ApplicationContextConfig::app_source_mode_from_global_flags(&global_flags);
+        ApplicationContext::plan_and_apply_manifest_upgrades_before_load(
```
vigoo: I think this should not run for `--help` (and we have the `log_output_for_help` already to decide that here). I think if a help run starts updating stuff that's a weird experience

cli/golem-cli/src/command_handler/log.rs
```diff
@@ -134,3 +95,56 @@ impl LogHandler {
         DecoratedIndent::new_secondary(self.ctx.format())
     }
 }
+
+pub fn render_toon_document<S: Serialize>(value: &S) -> String {
```
vigoo: Are these encoders (toon and serde below) never really fail? Just to be sure about the unwraps

vigoo (approved):
Two comments, otherwise good

cli/golem-cli/src/command_handler/log.rs
noise64: in practice, not, but switched to error propagation instead

cli/golem-cli/src/context.rs
noise64: fixed

PR #3610, opened 2026-06-03
agent-permissions: use cards for existing permission checks (2)

mschuwalow:
More migration of existing checks. There is still some temporary plumbing code and a few paths like list_visible_environments are unoptimized. That will come in followups


golem-registry-service/src/services/deployment/write.rs
```diff
@@ -185,7 +186,7 @@ impl DeploymentWriteService {
             });
 
         let latest_deployment = self
-            .get_latest_deployment_for_environment(&environment, auth)
+            .get_latest_deployment_for_environment(&environment, &AuthCtx::System)
```
mschuwalow, line 189: So that shares do not need a million view permissions to perform deployments. The deployment action itself is still guarded.

I imagine there are other places like this now that the permission model is more finegrained, will adjust as I encounter them.

vigoo (approved):

PR #3612, opened 2026-06-04
agent-permissions: use cards for existing permission checks (3)

mschuwalow:
Migrates the remaining apis and removes the now unused parts of the AuthCtx


golem-registry-service/src/services/agent_secret.rs
```diff
@@ -59,6 +59,32 @@ pub enum AgentSecretError {
...
+            owner: EnvironmentOwnerPattern::Environment {
+                account: environment.owner_account_id.to_string(),
+                application: environment.application_name.0.clone(),
+                environment: environment.name.0.clone(),
+            },
+            resource: key
+                .map(|key| {
+                    EnvironmentAgentSecretResourcePattern::Key(
+                        EnvironmentAgentSecretKeyPathPattern::parse(&key.join("."))
+                            .expect("agent secret keys are valid card resources"),
```
vigoo: why is this an `expect`?

mschuwalow: Fixed, this was a leftover.

Generally services should construct the PermissionTargets directly instead of using parse. Ping me if you notice others during the reviews :pray:

vigoo (approved):

PR #3613, opened 2026-06-04
agent-permissions: remove old environment-shares logic & add permission shares to cli

mschuwalow:


cli/golem-cli/src/command_handler/account.rs
```diff
@@ -151,3 +372,24 @@ impl AccountCommandHandler {
         }
     }
 }
+
+fn permission_share_data(grants: PermissionShareGrantArgs) -> PermissionShareData {
+    PermissionShareData {
+        lower_positive: grants.lower_positive.unwrap_or_default(),
+        lower_negative: grants.lower_negative.unwrap_or_default(),
+        upper_positive: Vec::new(),
```
mschuwalow, line 380: I'll most likely remove upper entirely on the server side as well for permission shares (as it's weird to be able to constrain another account), but will keep it like this until I have decided

vigoo (approved):

PR #3615, opened 2026-06-05
1.5.x cloud benchmarks

kmatasfp:
Also contains replacement for memory semaphore plus fixes eviction deadlock.

for memory semaphore replacement see:
* golem-worker-executor/src/services/active_workers/admission/mod.rs
* golem-worker-executor/src/services/active_workers/component_charge/mod.rs
* golem-worker-executor/src/services/active_workers/memory_probe.rs


mschuwalow (approved):

golem-worker-executor-test-utils/src/lib.rs
```diff
@@ -1358,6 +1379,31 @@ impl InvocationContextManagement for TestWorkerCtx {
...
+        // framework and other services, so a process-RSS probe cannot isolate
+        // this executor's footprint. When a test pins a memory limit via
+        // system_memory_override, give the gate a fixed probe reporting that
+        // limit with zero current usage, so admission is decided solely on the
+        // granted accounting (exact and process-isolated) against the pinned
+        // limit. The usable_ratio (worker_memory_ratio) still applies, matching
+        // the pre-gate semaphore pool size of system_memory_override * ratio.
+        match golem_config.memory.system_memory_override {
+            Some(limit) => Arc::new(ActiveWorkers::new_with_probe(
+                Box::new(FixedProbe::new(limit, 0)),
```
vigoo, line 1400: I think with this we make the test cases explicitly testing the memory-based scheduling under pressure unusable (and not testing anything meaningful). 
These are:
- `initial_large_memory_allocation`
- `dynamic_large_memory_allocation`
- `eviction_prefers_idle_workers_over_warm_runnable`

I'd like to keep these tests, these were the proof that the original scheduler was working (before writing them these tests did not pass at all)

kmatasfp: these should still work, see `TestContext.start_customized` what the `scalability.rs` tests use and this commit https://github.com/golemcloud/golem/pull/3615/changes/a8fcf52168155fa2d3f33291ecbf86727a348087

vigoo: I thought by using the fixed probe we are not being able to detect the actual memory use coming from our agents? Maybe I misunderstand

kmatasfp: also some of what they were testing are now also covered unit test level like `evicts_idle_before_warm` and `usage_never_exceeds_limit` etc

kmatasfp: So the new AdmissionController if wired for prod/live, it checks 2 things - current RSS (contains the rust heap + wasm mmap) + memory grants (these are similar to the permits the old impl was using, but less "optimistic" how much virtual memory is assigned to agents, it can be because we also check how much mem we actually have free now, so less estimating, ceiling is the total ram we have assigned for the workers) - so basically to admit an agent it needs to fit into free RSS plus the virtual memory grants cannot go over the available ram we have.

So in case of `initial_large_memory_allocation`, `dynamic_large_memory_allocation` and `eviction_prefers_idle_workers_over_warm_runnable` using `FixedProbe` we limit the mem cealing (what is the max memory available for the test) and use only the grant based approach, so basically works the same way as before where we only had the semaphore

kmatasfp: @vigoo tldr there is no acutal mem reporting for the agents when it comes to AdmissionController, only will be used for billing, first of all it is quite slow, in case of billing we can do sampling to make it better,  also after the virtual memory has been granted to the agent there is no way to limit how much of it can use, so if it has been granted 1gb of virtual memory, it can use all of it at moments notice - no way to take it away either if it does not need that much virtual memory anymore. 

So the current impl is based on old one with bunch of tweaks

kmatasfp: One more thing to note, it is still not 100% OOM proof, because there is no  limit how much memory the host uses (rust heap). Lets say we have 16gb host, we logically "assign" 12gb for the wasm and rest for the host. There is currently no mechanism to make sure that the host stays withing the logical 4gb, it can easily use 8gb for some reason or another at one point, and then agents try to use all their virtual memory and boom we oom, because this spiked the wasm part of memory to 12gb which makes us go over the total limit of 16gb and then crash. It was also the case with the previous implementation but now using hint from the actual free RSS it should be better. 

In prod it all depends how fast the auto scaling kicks in and how big of a burst of mem usage we see

we probably should set MI_MAX_HEAP_SIZE, limiting the rust heap, so we have true separation? Hard to figure out what the value should be

golem-worker-executor/src/worker/mod.rs
```diff
@@ -2349,10 +2398,41 @@ impl WaitingWorker {
...
+                // returns the reservation to the gate). This is what makes a
+                // start cancelled mid-flight — e.g. the worker being deleted while
+                // still waiting for its remaining permits — release rather than
+                // leak its grant.
+                let memory_grant = parent.active_workers().acquire(memory_requirement).await;
+                // Reserve the component's compiled module size once per resident
+                // component (shared by all its workers). Held for as long as this
+                // worker is resident; the module faults into RAM when the first
+                // worker loads, so reserving it keeps later admissions honest.
+                let component_charge = match parent.component_charge_requirement().await {
```
vigoo: By doing this after `acquire` above, we could over-commit if this is the first instance of a component, can't we? Let's say we had a capacity of 1000, memory requirement is 900 and the component size is 200

golem-worker-executor/src/services/active_workers/mod.rs
```diff
@@ -471,11 +405,111 @@ impl<Ctx: WorkerCtx> ActiveWorkers<Ctx> {
...
+    priority: EvictionPriority,
+    needed_bytes: u64,
+) -> u64 {
+    let target_class: crate::worker::EvictionClass = priority.into();
+
+    let mut candidates = Vec::new();
+    for (agent_id, worker) in workers.iter().await {
+        if let Some(class) = worker.eviction_class().await
+            && class == target_class
+            && let Ok(mem) = worker.memory_requirement().await
```
vigoo, line 555: Not sure if we could handle this safely here but this will not take into account if the instance is the last instance of a component, evicting that will evict the space taken by the component as well.

golem-worker-executor-test-utils/src/lib.rs
```diff
@@ -533,6 +534,16 @@ fn make_base_test_config(deps: &WorkerExecutorTestDependencies) -> GolemConfig {
...
         // not exist in this test setup.
         resource_limits: ResourceLimitsConfig::Disabled(ResourceLimitsDisabledConfig {}),
+        // The measured-headroom admission gate requires the executor to own its
+        // memory environment (cgroup/process). The in-process test harness runs
+        // the executor alongside the test framework and other services, so the
+        // probe cannot isolate this executor's footprint. Disable the gate so
+        // admission always proceeds and tests are not subject to a memory limit
+        // derived from the shared host.
+        memory: MemoryConfig {
+            enable_measured_admission: false,
```
vigoo, line 544: `start_with_overrides` resets this to true (by `MemoryConfig::default()`), that may not be intended

golem-worker-executor/src/worker/mod.rs
kmatasfp: you are probably right need to double check

golem-worker-executor/src/services/active_workers/mod.rs
kmatasfp: Yeah currently when last instance of the component is dropped then the "charge" will also be dropped, at least it should. Let me make sure we have tests for that case

golem-worker-executor-test-utils/src/lib.rs
kmatasfp: let me check

golem-worker-executor/src/worker/mod.rs
kmatasfp: addressed

golem-worker-executor/src/services/active_workers/mod.rs
kmatasfp: addressed

golem-worker-executor-test-utils/src/lib.rs
kmatasfp: done

golem-worker-executor/tests/observability.rs
```diff
@@ -216,7 +216,7 @@ async fn search_oplog_1(
     }
 
     assert_eq!(result1.len(), 2, "G1002"); // TODO: this is temporarily not working because of using the dynamic invoke API and not having structured information in the oplog
-    assert_eq!(result2.len(), 2, "imported-function");
+    assert_eq!(result2.len(), 1, "imported-function");
```
kmatasfp: this and other similar test changes are because we do not call host function to get agent metadata anymore

sdks/rust/golem-rust-macro/src/agentic/agent_implementation_impl.rs
```diff
@@ -107,30 +107,21 @@ pub fn agent_implementation_impl(_attrs: TokenStream, item: TokenStream) -> Toke
...
-                golem_rust::agentic::register_agent_instance(
-                    golem_rust::agentic::ResolvedAgent::new(agent_instance)
-                );
-                Ok(())
-            }
-        }
+    if matches!(get_asyncness(&constructor_method.sig), Asyncness::Future) {
+        return syn::Error::new_spanned(
+            &constructor_method.sig.ident,
+            "Agent constructors must be synchronous. Async constructors can call host APIs during snapshot restore and make recovery fall back to full replay.",
```
kmatasfp: prob need a better message here, regardless Agent constructor should not be async imho, adds more complexity than value, unless I am missing something

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -34,6 +34,46 @@ mod sockets;
 pub mod wasm_rpc;
 pub mod websocket;
 
+#[cfg(any(test, feature = "test-support"))]
+pub mod snapshot_recovery_test_support {
```
kmatasfp: could not think of a better way how to do this, I need some hook inside the snapshot recovery code that actually records that we recovered from snapshot or not. Rust makes it easier than other languages to use test Spy with conditional compilation

sdks/rust/golem-rust-macro/src/agentic/agent_implementation_impl.rs
vigoo: I disagree, I think that would be too limiting (no awaitable rpc, no http calls etc) - and whether it does side effects or not (it has to be able to!) does not actually depend on the "asyncness", in p2 all host functions are sync and in p3 we get async ones beside them.

vigoo: But I see how this interferes with load-snapshot initializing the agent. I need to think more about it :)

vigoo: One real-world example: it has been a useful pattern to spawn child agents and trigger some background operation in the constructor. During recovery, of course this does not get triggered again because it's persisted in the oplog. But if load-snapshot calls the same constructor, it's going to be re-trigger the same background operation with a fresh idempotency key (or crash if we detect it as an unallowed side-effect). Both are wrong. The assumption was so far that "it's the user's responsibility" but it's probably too easy to do it wrong.

PR #3616, opened 2026-06-05
agent-permissions: make cards more typesafe & consistently use account email everywhere

mschuwalow:


golem-worker-executor/src/model/public_oplog/wit.rs
```diff
@@ -59,6 +59,7 @@ impl From<PublicOplogEntry> for oplog::PublicOplogEntry {
                 local_agent_config,
                 original_phantom_id,
                 instance_id,
+                ..
```
vigoo: Let's not do this, so we not forget updating this code when needed

golem-common/src/base_model/card/parsing.rs
```diff
@@ -179,44 +177,18 @@ struct PermissionParts {
...
-    let owner = owner.trim();
-    let owner = owner
-        .strip_suffix(')')
-        .ok_or_else(|| "missing class owner close before recipient separator".to_string())?
-        .trim();
+    let (class, rest) = value
+        .split_once('(')
+        .ok_or_else(|| "missing class owner open".to_string())?;
+    let (owner, rest) = rest
+        .split_once(") @ ")
```
vigoo, line 184: Is the syntax intentionally not more whitespace-friendly?

vigoo (approved):
Added one small request and a question, otherwise looks good

golem-common/src/base_model/card/parsing.rs
mschuwalow: I plan to do a pass over the parser later and improve it. There is probably also still a few cases in which it is ambiguous with the resource grammars.

PR #3618, opened 2026-06-06
Agent status record optimization

vigoo:
Optimizes how the agent status record is cached (the information that can be reconstructed by reading the oplog, but is cached in redis or sqlite for faster execution).

Changes:

- Split the `agent_mode` to a separate key to support the fast path routing between ephemeral and durable agents
- Previously the agent status record contained some user payload (pending invocation inputs and enqueued manual update snapshots). For other payloads, such as invocation results, it already correctly only stored an oplog index reference to the entry that owns the data. From now on we only store oplog indices for the pending invocations and updates as well.
- A few fields of the status record could grow unbounded - invocation results per idempotency key, update outcomes, deleted and skipped regions, etc. These are now stored in different sub-keys and only updated when needed. We no longer send all this growing data on every status record update.
- Previously we were updating this record together with every oplog commit, in a synchrounous way, with an old comment to make this a background update later. Now that's done too - updates are performed in the background, not slowing down the execution. 
- We had (and still have) some cases where the cached status record gets invalidated. The most important one that affects hot path is retries and other cases of _jumps_ where we add a skipped oplog region and if the cached status record was for an index within the newly skipped region, we recalculated the whole thing from scratch (lazily). This is now better by periodically saving a _second_ version of the agent status record into the cache, but only at safe points (between invocations, possibly can be extended to some within-invocation cases based on the retry-point tracking). This means even when we have to drop the status because of a jump, we only recalculate it from the last "stable" cached entry if it exists (there is always at most one of it).


golem-common/src/model/mod.rs
```diff
@@ -759,9 +757,10 @@ pub struct AgentStatusRecord {
     /// Timestamp of the last automatic snapshot entry in the oplog.
     pub last_automatic_snapshot_timestamp: Option<Timestamp>,
     /// The agent mode the worker was created with. Decided at create time and persisted in the
-    /// `Create` oplog entry; immutable for the life of the worker. Recorded here so that out-of-band
-    /// callers (e.g. `WorkerService::get_agent_mode`) can resolve the worker's oplog namespace
-    /// directly from the cached status record without a separate KV lookup.
+    /// `Create` oplog entry; immutable for the life of the worker. `#[transient]`: it is not part
+    /// of the serialized record (it is persisted separately) and defaults to `Durable` on
+    /// deserialization, so readers must restore it from its own source.
+    #[transient(AgentMode::Durable)]
```
mschuwalow, line 763: I don't think this is worth changing

golem-common/src/model/mod.rs
```diff
@@ -723,13 +721,13 @@ impl SafeDisplay for RetryConfig {
...
     pub status: AgentStatus,
     pub skipped_regions: DeletedRegions,
     pub overridden_retry_config: Option<RetryConfig>,
-    pub pending_invocations: Vec<TimestampedAgentInvocation>,
-    pub pending_updates: VecDeque<TimestampedUpdateDescription>,
+    pub pending_invocations: Vec<PendingInvocationRef>,
+    pub pending_updates: VecDeque<PendingUpdateRef>,
     pub failed_updates: Vec<FailedUpdateRecord>,
     pub successful_updates: Vec<SuccessfulUpdateRecord>,
     pub invocation_results: HashMap<IdempotencyKey, OplogIndex>,
```
mschuwalow, line 733: invocation_results was by far the most problematic part during the hackathon

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -3367,13 +3407,15 @@ impl<Ctx: WorkerCtx> ExternalOperations<Ctx> for DurableWorkerCtx<Ctx> {
...
+            let latest_worker_status = calculate_last_known_status_with_checkpoint(
                 this,
                 &owned_agent_id,
-                worker.initial_worker_metadata.agent_mode,
+                agent_mode,
                 worker.last_known_status,
             )
-            .await;
+            .await
+            .expect("Failed to calculate worker status for existing worker");
```
noise64: Is this intentionally panic instead of returning error?

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -4135,6 +4172,25 @@ struct PrivateDurableWorkerState {
...
+    /// left open by a trap is cleared on restart.
+    open_rollback_regions: HashSet<OplogIndex>,
+
+    /// The minimum oplog index handed to the guest via `get_oplog_index` during the current
+    /// invocation (the `NoOp` marker it plants). It is the only realistic `set_oplog_index` target,
+    /// which deletes `(M.next()..source]` and preserves `M`, so a checkpoint at an index `<= M`
+    /// survives such a jump. Mid-invocation checkpoints are not advanced past this watermark. Reset
+    /// at the start of every invocation (a marker held across invocations only costs graceful
+    /// fallback, never correctness).
+    min_exposed_marker: Option<OplogIndex>,
```
mschuwalow, line 4212: Can getting the oplog index via api (not host function) and reverting the agent break things?

golem-worker-executor/src/worker/mod.rs
```diff
@@ -1174,14 +1206,82 @@ impl<Ctx: WorkerCtx> Worker<Ctx> {
...
             .await
             .pending_invocations
             .clone()
     }
 
+    /// Reads the `PendingAgentInvocation` oplog entry referenced by `pending` and reconstructs the
+    /// full invocation, downloading its payload from external storage if needed. The status record
+    /// only keeps a lightweight reference, so callers that need to execute the invocation hydrate
+    /// it on demand.
+    async fn hydrate_pending_invocation(
```
mschuwalow, line 1221: I'm not sure doing this for pending invocations / updates. In non-degenerate cases these should be small and bounded (unlike invocation results). I would rather just enforce a (high) max limit of pending updates / invocations and reject incoming ones when that limit is reached.

It might have benefits for full replays though, but these should be rarer after these changes

mschuwalow (approved):
I'm not sure whether redis prefers one large key or multiple independent reads / writes.

golem-worker-executor/src/services/worker.rs
```diff
@@ -122,40 +369,265 @@ impl DefaultWorkerService {
...
+    /// in a single atomic `set_many` (one `HMSET` on Redis, one transaction on SQL). This preserves
+    /// the invariant that each persisted field's content matches `core.oplog_idx`, which the oplog
+    /// fold relies on. When stale fields must be removed (e.g. invocation results dropped by a
+    /// revert), they are deleted *together with* `core` before the `set_many`. Dropping `core`
+    /// first is what makes the two-step delete-then-write crash-safe: with `core` absent, any crash
+    /// or read in the gap before the final write is treated as a cache miss and recomputed from the
+    /// oplog, rather than reassembling a torn record whose remaining fields no longer match the
+    /// (stale) marker. The final `set_many` always re-establishes `core` atomically with the
+    /// changed parts.
+    async fn write_status_fields(
```
mschuwalow, line 471: Do we need to invalidate invocation results / other parts of the agent status on a revert?

noise64 (approved):

golem-common/src/model/mod.rs
noise64: I almost ocmmented the same, as @mschuwalow, but I think keeping is simply wasteful, as it cannot change ever, and there is not point in adding it to every core update

golem-worker-executor/src/durable_host/mod.rs
vigoo: It was panicking before as well (just with an `assert!` inside), as this case can only happen if there is an oplog corruption (= we have cache but the oplog is missing). But now we can actually handle it better

golem-worker-executor/src/durable_host/mod.rs
vigoo: It's not going to break anything, just fallback to the original "recalc from scratch"

golem-common/src/model/mod.rs
vigoo: I don't want anything in the 'core' part of that status record that has "user-dependent" size. 
Manual updates had their whole serialized snapshot included in this queue before, also invocations their full serialized input parameters (think about passing large binary data etc!). So that explains the switch to he `xxxRef` entries that are just pointing to the oplog.
But I also wanted to separate them in the underlying data store because we do a lot of writes, and it does not make any sense to continuously send these queues on each status record update when their size is not static. Maybe on the hackathon noone overloaded an agent's invocation/update queue, but that's definitely something that can easily happen.

golem-worker-executor/src/worker/mod.rs
vigoo: For me this change feels important, tried to explain more in https://github.com/golemcloud/golem/pull/3618#discussion_r3372564876

golem-common/src/model/mod.rs
vigoo: Maybe not with all the other changes, without them it was definitely worthy (and I did this first) - the whole record was read just to decide whether it's an ephemeral or durable agent. I think storing it separately does not have any disadvantage even with all the other changes it is not _that_ big anymore (still a full record vs a bool)

golem-worker-executor/src/services/worker.rs
vigoo: I think it's handled correctly here already

Issue #3611, opened 2026-06-03
mcp: resolve registered agent type at deployment time, not at runtime


mschuwalow:
[This step](https://github.com/golemcloud/golem/blob/f44e13f56eed18d3a4e3f2a2e2c0c904c30a9129/golem-registry-service/src/services/deployment/mcp.rs#L86) should not exist, instead the data should be resolved at deployment time an included in the blob.

That both ensures that the data is consistent within one deployment and improves performance.

PR #3621, opened 2026-06-08
Migrated MCP export

### LINKED ISSUE #3397: MCP export migration to the new schema model
# Wave 2 — MCP export migration to the new schema model

This is the Wave 2 task for migrating the **MCP exporter** to the new value/type model. It is split out from the other Wave 2 internal migrations (#3396) because MCP is an external-protocol exporter — its output is user/tool-visible, the migration interacts with the structural-vs-intent split for multimodal, and it deletes the legacy top-level multimodal wrapper from this consumer.

## Scope

This issue covers:

- rewriting the MCP exporter on top of `SchemaType` / `SchemaValue` / `SchemaGraph` / `TypedSchemaValue` / `InputSchema` / `OutputSchema`;
- replacing the top-level `Multimodal` wrapper with the structural form (`list<union<…>>` carrying `Role::Multimodal` in the typed metadata envelope);
- preserving every existing MCP-specific behaviour:
  - input multimodal lowered as `{ parts: [...] }`,
  - output multimodal omits `outputSchema`,
  - constructor multimodal rejected,
  - prompt and resource paths branch on multimodal role,
  - content-block rendering driven by `Role`.

## Tasks

- [ ] Rewrite the MCP exporter against the new schema types. Adapters from Wave 1 bridge to producers that haven't migrated yet (registry-service mirror from Wave 1.11). The exporter reads the agent's per-agent `SchemaGraph` (§4.22) once and emits its named definitions as content-block `$defs` referenced from every method's input/output schema, rather than inlining the same type once per method.
- [ ] Implement the structural-vs-intent multimodal handling: detect `Role::Multimodal` on `list<union<…>>` schemas and apply the existing MCP-specific rendering (parts array, omitted outputSchema, prompt/resource branches, constructor rejection).
- [ ] Migrate input-schema rendering to read from `InputSchema::Parameters`, including correct handling of `FieldSource::AutoInjected` (auto-injected fields are not surfaced to MCP clients).
- [ ] Migrate output-schema rendering to read from `OutputSchema = Unit | Single(SchemaType)`.
- [ ] Migrate content-block rendering for `Text`, `Binary`, `Path`, `Url`, capability-bearing nodes (with redaction where applicable per `HostManagedKind`).
- [ ] Update MCP exporter tests so each MCP-specific behaviour above has a regression test against the new types.
- [ ] Drop the top-level multimodal wrapper from this consumer (the type is still exported elsewhere until Wave 6 deletion).

## Exit criteria

- MCP exporter consumes only the new schema types internally;
- behaviour against MCP clients is unchanged for every existing case (parts array, omitted outputSchema, prompt/resource handling, constructor rejection);
- regression tests cover input multimodal, output multimodal, prompt/resource multimodal, constructor multimodal rejection, and content-block rendering for `Text` / `Binary` / `Path` / `Url`;
- top-level multimodal wrapper is no longer referenced inside the MCP exporter.

## Dependencies

  - new types and adapters (1.2, 1.10),
  - renderers covering all `SchemaType` cases (1.4),
  - `InputSchema` / `OutputSchema` / `FieldSource` (1.9),
  - canonical encodings (1.5),
  - parallel agent-metadata mirror (1.11).
- Independent of the internal subsystem migrations (#3396) and the OpenAPI migration (#3398); all three can land in parallel.

## Out of scope

- migrating internal subsystems (agent secrets, custom error, component metadata, durable host helpers, CLI rendering) → #3396;
- migrating OpenAPI export → #3398;
- deleting the top-level multimodal wrapper from `golem-common` and from other consumers → Wave 6;
- capability redaction policy hooks across other subsystems → Wave 5;
- any wire-format change at the REST / gRPC / oplog boundary → Wave 3.

vigoo:
Resolves #3397


golem-worker-service/src/mcp/schema/mcp_tool_schema.rs
```diff
@@ -12,68 +12,220 @@
...
+    let required = if resolves_to_option(graph, ty) {
+        Vec::new()
+    } else {
+        vec![Value::String(FALLBACK_OUTPUT_FIELD_NAME.to_string())]
+    };
+    Some(rmcp::model::object(json!({
+        "type": "object",
+        "properties": { FALLBACK_OUTPUT_FIELD_NAME: inner },
+        "required": required,
+    })))
```
noise64: This seems to be adding defs under properties, while the refs pointing to the document root (unlike input_schema_to_json_schema, where things are directly under the root)

golem-worker-service/src/mcp/agent_mcp_server.rs
```diff
@@ -165,18 +171,31 @@ pub async fn get_agent_capabilities(
...
+                    // model, so invoking it would always fail. Skip advertising
+                    // it instead of exposing a broken tool/resource.
+                    tracing::warn!(
+                        "Skipping method {} of agent type {} for domain {}: {:#}",
+                        method.name,
+                        agent_type.type_name.0,
+                        domain.0,
+                        e
+                    );
+                }
```
noise64: prompts are pushed before this validation, shouldn't the validation happen first?

golem-worker-service/src/mcp/schema/mcp_tool_schema.rs
```diff
@@ -12,68 +12,220 @@
...
+    graph: &SchemaGraph,
+    constructor: &AgentConstructorSchema,
+    method: &AgentMethodSchema,
+) -> JsonObject {
+    let mut fields: Vec<NamedField> = Vec::new();
+    fields.extend(constructor.input_schema.fields().iter().cloned());
+    fields.extend(method.input_schema.fields().iter().cloned());
+    let combined = InputSchema::Parameters(fields);
+    input_schema_to_mcp(graph, &combined)
+}
```
noise64, line 91: If there are duplicated names between constructor and method fields names, during rendering them into a record / map will only keep the last one. Or is this validated / checked elsewhere?

golem-common/src/schema/render/tests/json_schema_props.rs
```diff
@@ -0,0 +1,119 @@
...
+//
+//     http://license.golem.cloud/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Property-based tests asserting that every JSON Schema document produced by
```
noise64, line 15: 👏

noise64 (approved):

golem-api-grpc/proto/golem/schema/schema.proto
```diff
@@ -0,0 +1,254 @@
...
+  string field_name = 1;
+  optional string literal = 2;
+}
+
+message SecretSpec {
+  optional string category = 1;
+}
+
+message QuotaTokenSpec {
+  optional string resource_name = 1;
```
mschuwalow, line 248: This will require some more data like the accumulated credit, expected use, the owning environment, ...

But let's adjust this when we adjust the quotas to use this for rpc.

golem-registry-service/src/services/deployment/mcp.rs
```diff
@@ -84,45 +86,48 @@ impl DeployedMcpService {
 
                 let mut registered_agent_types = Vec::new();
                 for agent_type_name in compiled_mcp.agent_type_implementers.keys() {
-                    match self
+                    // Fail loudly on any inconsistency: an active MCP deployment
```
mschuwalow, line 89: This is broken in my opinion and the resolution should happen at deployment time https://github.com/golemcloud/golem/issues/3611 (similar to http)

mschuwalow (approved):

PR #3622, opened 2026-06-08
Migrated OpenAPI exporter

### LINKED ISSUE #3398: OpenAPI export migration to the new schema model
# Wave 2 — OpenAPI export migration to the new schema model

This is the Wave 2 task for migrating the **OpenAPI emitter** to render directly from the new schema model. It is split out from the other Wave 2 internal migrations (#3396) because OpenAPI is an external-protocol exporter — its output is user/tool-visible.

## Scope

This issue covers:

- rewriting the OpenAPI emitter to read `SchemaType` / `SchemaValue` / `SchemaGraph` / `InputSchema` / `OutputSchema` directly, with no projection or intermediate model;
- emitting `components/schemas` for named types from the `SchemaGraph` definition registry;
- correct rendering of every `SchemaType` case the renderer supports (per Wave 1 task 4): structural cases, `Text`, `Binary`, `Path`, `Url`, `Datetime`, `Duration`, `Quantity`, discriminated `Union`, capability nodes (`Secret`, `QuotaToken`) with schema-only / no-value placement;
- preserving the current behaviour against OpenAPI consumers for every existing endpoint.

## Tasks

- [ ] Rewrite the OpenAPI emitter against `SchemaType` / `SchemaValue` / `SchemaGraph` / `InputSchema` / `OutputSchema`.
- [ ] Emit `components/schemas` from the agent's per-agent `SchemaGraph::defs` (§4.22) for every named type, once per agent, with `$ref` for `Ref(TypeId)` use sites and inline expansion only for anonymous types. The same named type referenced from multiple methods/constructors of one agent appears exactly once in `components/schemas`.
- [ ] Cover every `SchemaType` case: structural primitives, `Record` / `Variant` / `Enum` / `Flags` / `Tuple` / `List` / `FixedList` / `Map` / `Option`, semantic scalars (`Text` / `Binary` / `Path` / `Url` / `Datetime` / `Duration` / `Quantity`), discriminated `Union`, capability nodes.
- [ ] Map endpoint protocol policies correctly (these stay in the OpenAPI emitter, not in `SchemaType`):
  - `option<T>` body → 404 / present-body discriminated by HTTP status;
  - `result<ok, err>` body → 200 / 500 (or configured pair);
  - `Content-Language` headers driven by `Text` semantics;
  - capability nodes appear as schema-only in responses (no value rendering for `Secret`, redacted form for `QuotaToken`).
- [ ] Map `InputSchema::Parameters` to OpenAPI parameters, hiding `FieldSource::AutoInjected` from the public surface.
- [ ] Map `OutputSchema = Unit | Single(SchemaType)` to OpenAPI response shapes.
- [ ] Bring across docs / aliases / examples / deprecation from the typed metadata envelope into the matching OpenAPI fields.
- [ ] Update OpenAPI emitter tests so each protocol-policy mapping above has a regression test against the new types.

## Exit criteria

- OpenAPI emitter renders directly from `SchemaType` with no intermediate projection;
- generated OpenAPI documents are byte-identical (or behaviourally equivalent under the documented protocol mappings) for every existing endpoint;
- regression tests cover each `SchemaType` case end-to-end, the named-type / `$ref` emission path, and the protocol mappings above;
- nothing in the emitter still references legacy `AnalysedType` / `Value` / `DataSchema` / `ElementSchema`.

## Dependencies

  - new types and adapters (1.2, 1.10),
  - renderers covering all `SchemaType` cases (1.4),
  - `InputSchema` / `OutputSchema` / `FieldSource` (1.9),
  - canonical encodings (1.5),
  - parallel agent-metadata mirror (1.11).
- Independent of the internal subsystem migrations (#3396) and the MCP migration (#3397); all three can land in parallel.

## Out of scope

- migrating internal subsystems → #3396;
- migrating MCP export → #3397;
- consumer-side rich-type sweetening (e.g. richer OpenAPI `format` hints, unit-bearing literals) → Wave 5;
- any wire-format change at the REST / gRPC / oplog boundary → Wave 3.

vigoo:
Resolves #3398


golem-common/src/schema/render/json_schema.rs
noise64: this seems to push every field as required

noise64: or do we encode optionality stricly by always requiring fields with explicit nulls?

golem-common/src/schema/render/openapi.rs
```diff
@@ -81,6 +96,85 @@ pub fn to_openapi_components(graph: &SchemaGraph, ty: &SchemaType) -> Value {
...
+/// `{ "components": { "schemas": { … } }, "root": { … } }` and every `$ref`
+/// pointer targets `#/components/schemas/<TypeId>`.
+///
+/// The `root` is an `object` schema whose `properties` are the input's
+/// **user-supplied** parameters (`FieldSource::AutoInjected` fields are host
+/// provided and never surfaced, so they are omitted) and whose `required`
+/// array lists every user-supplied parameter whose schema is not an
+/// `option<…>` — the exact behaviour of
+/// [`input_schema_to_json_schema`](super::json_schema::input_schema_to_json_schema).
+pub fn input_schema_to_openapi_components(graph: &SchemaGraph, input: &InputSchema) -> Value {
```
noise64: this and the output variant seems to be only used in tests

golem-worker-service/src/custom_api/openapi/response_schema.rs
```diff
@@ -282,57 +125,176 @@ pub fn get_route_response_schema(route: &RichCompiledRoute) -> RouteResponseOpen
...
+                    responses.insert(
+                        200,
+                        ResponseBodyOpenApiSchema::Known {
+                            schema,
+                            content_type: "application/json".to_string(),
+                        },
+                    );
+                }
+                None => {
+                    responses.insert(204, ResponseBodyOpenApiSchema::Unknown);
```
noise64: This should be NoBody

noise64 (approved):

PR #3627, opened 2026-06-10
agent-permissions: initial agent wallet

mschuwalow:


golem-common/src/model/component_metadata.rs
```diff
@@ -432,10 +472,52 @@ impl RawComponentMetadata {
             root_package_version: self.root_package_version,
             agent_types,
             agent_type_provision_configs,
+            agent_type_initial_permissions: BTreeMap::new(),
+        }
+    }
+}
+
+impl AgentInitialPermissionTemplate {
```
vigoo, line 480: Is the meaning of this is that by default an agent can only do RPC to instances of the same component? I think the default should be to be able to call any component in the same environment, as it was before. (but maybe I misunderstand this default template)

golem-worker-executor/src/services/worker_proxy.rs
```diff
@@ -133,6 +136,18 @@ pub trait WorkerProxy: Send + Sync {
         environment_id: Option<EnvironmentId>,
     ) -> Result<InvocationStatus, WorkerProxyError>;
 
+    async fn lookup_invocation_status_as_system(
```
vigoo: Why do we have an erroring default implementation for this?

golem-worker-executor/src/services/oplog/plugin.rs
```diff
@@ -445,16 +445,14 @@ impl<Ctx: WorkerCtx> OplogProcessorPlugin for PerExecutorOplogProcessorPlugin<Ct
...
         target_agent_id: &AgentId,
-        caller_account_id: AccountId,
-        caller_account_email: &golem_common::model::account::AccountEmail,
+        _caller_account_id: AccountId,
+        _caller_account_email: &golem_common::model::account::AccountEmail,
         idempotency_key: &IdempotencyKey,
     ) -> Result<InvocationStatus, WorkerExecutorError> {
         self.worker_proxy
-            .lookup_invocation_status(
+            .lookup_invocation_status_as_system(
```
vigoo: Why is this "as system" now?

golem-worker-executor/src/durable_host/wasm_rpc/mod.rs
```diff
@@ -687,12 +693,14 @@ impl<Ctx: WorkerCtx> HostWasmRpc for DurableWorkerCtx<Ctx> {
             })?;
             Ok(fut)
         } else {
+            let self_effective_surface = self.agent_effective_surface();
```
vigoo: Do we have calculate this for each RPC call? I'm afraid of the perf cost

golem-common/src/model/component_metadata.rs
mschuwalow: This is just a placeholder. The proper defaults will come when I do the cli integration / expose it in the api (though iirc one of the later prs in this series already adjusts this set as you suggets)

golem-worker-executor/src/durable_host/wasm_rpc/mod.rs
mschuwalow: Just a placeholder. The next pr changes the agent to hold a materialized effective surface that is updated on card add / removal. This function is changed to just return a handle to that.

golem-worker-executor/src/services/oplog/plugin.rs
mschuwalow: cleaned up here https://github.com/golemcloud/golem/pull/3627/commits/471c2e1e83ead8d27eada32668dea5eeb0cb2def

golem-worker-executor/src/services/worker_proxy.rs
mschuwalow: cleaned up here https://github.com/golemcloud/golem/pull/3627/commits/471c2e1e83ead8d27eada32668dea5eeb0cb2def

vigoo (approved):

PR #3628, opened 2026-06-11
Concurrent durability

### LINKED ISSUE #3407: Concurrent durability helper
## Goal

Make the durability/oplog layer concurrency-aware so multiple p3 host
async calls issued by the same guest task can overlap and still be
durably recorded and deterministically replayed. Define the
`Drop`/cancellation contract that this requires.

## Why

Under p3 the guest can issue multiple async host calls concurrently
(parallel RPC, parallel HTTP, parallel keyvalue, …). With the
Accessor-based host shape no `&mut store` is held across awaits, so
those host futures actually overlap. Our oplog assumes a strictly
linear sequence of durable ops per worker, and we cannot fix that by
serializing — we would lose the parallelism that p2 expressed via
pollables.

## Main steps

- New oplog schema with per-call `call_id`:
  `Start { call_id, … }`, `End { call_id, result }`,
  `Cancelled { call_id, partial? }`.
- Per-worker monotonic `call_id` allocator, incremented synchronously
  at host-call entry (initiation order is deterministic; completion
  order is not).
- New `Durability` API
  (`start_call(...) -> CallHandle`, `CallHandle::finish(result)`,
  `CallHandle::cancel(partial?)`); port all callers.
- Replay engine resolves results by `call_id` in recorded completion
  order (driving suspended replay futures via oneshot channels).
- Re-express idempotency keys / Begin-End nesting per `call_id`.
- Bump the oplog format version and migrate old entries (treat each
  legacy entry as Start immediately followed by matching End).
- Implement the `Drop` / cancellation contract for every durable
  host fn: drop = cancellation; cleanup state must be captured in
  the future (no `Accessor` use in `Drop`); default-deny — untagged
  durable host imports are treated as not-durably-cancellable.
- Re-implement the host fns currently left as `unimplemented!()`
  `TODO(p3)` against the new API:
  `HostFutureInvokeResultWithStore::get`,
  `HostGetPromiseResultWithStore::get`, keyvalue cache
  `Future*ResultWithStore::get`, blobstore async stream-returning
  methods, keyvalue async stream-returning methods.
- Tests covering: two concurrent durable calls completing in reverse
  initiation order; durable call cancelled mid-flight via `select!` /
  drop; replay reproduces both completion order and cancellation.

The shared low-level "durable, externally-resolvable host future /
replayable oneshot" primitive that this work consumes is owned by
the suspendable-wait sub-task.

vigoo:
Resolves #3407


golem-worker-executor/src/durable_host/concurrent.rs
```diff
@@ -0,0 +1,1126 @@
...
+    /// Consumes the handle into its parts (used by the replay-state driver).
+    pub fn into_parts(self) -> (OplogIndex, ReplayableOneshotReceiver<Resolution>) {
+        (self.start_idx, self.receiver)
+    }
+}
+
+/// Event emitted when a [`CallHandle`] is dropped without being finished or cancelled.
+///
+/// There is no recorder actor in production yet, so an unfinished drop only logs. The drop policy
+/// is exercised in unit tests by attaching a sink that records these events.
```
noise64, line 221: Just to be sure: this means that full cancellation handling will come in the next batch?

golem-worker-executor/src/durable_host/concurrent.rs
```diff
@@ -0,0 +1,1126 @@
...
+
+impl DropPolicy for NotCancellable {
+    fn unfinished_drop(start_idx: OplogIndex, sink: Option<&UnboundedSender<DropEvent>>) {
+        if let Some(sink) = sink {
+            let _ = sink.send(DropEvent::UnfinishedNotCancellable { start_idx });
+        } else if cfg!(debug_assertions) && !std::thread::panicking() {
+            panic!("non-cancellable durable call {start_idx} dropped without finish/cancel");
+        } else {
+            tracing::error!(
+                "non-cancellable durable call {start_idx} dropped without finish/cancel"
```
noise64, line 270: Is using different tactics locally and in prod justified? Will this change in later phases?

golem-worker-executor/src/durable_host/replay_state.rs
```diff
@@ -631,6 +670,212 @@ impl ReplayState {
...
+    /// Drives the replay cursor forward, feeding the committed-consume hook, until the call
+    /// identified by `handle` resolves.
+    ///
+    /// The replay cursor is shared with legacy positional readers, and this driver only commits
+    /// `End`/`Cancelled` entries. Any other non-hint entry between the claimed `Start` and its
+    /// resolution (an unclaimed `Start`, a scope marker, a persistence-level change, ...) means the
+    /// cursor would be driven past something a legacy positional reader still expects, so it
+    /// returns an error instead of corrupting the cursor. With serialized host calls a call's oplog
+    /// is its `Start` followed by its own `End`/`Cancelled` (hint entries aside), so this never
+    /// triggers.
```
noise64, line 852: Just want to clarify, what legacy means here, is this part of a transional step, and "legacy" will be gone in next steps?

golem-worker-executor/src/durable_host/http/types.rs
noise64: Does this reasoning still good with the async changes? (As in, will the end markers added?)

golem-worker-executor/src/durable_host/wasm_rpc/mod.rs
noise64, line 1386: Similar question, as with the HTTP, are ENDs handled with drop?

golem-api-grpc/proto/golem/worker/raw_oplog.proto
```diff
@@ -15,11 +15,17 @@ import "golem/worker/worker_id.proto";
 import "google/protobuf/timestamp.proto";
 
 // An oplog entry in its raw/internal representation.
+//
+// The variants `host_call = 3`, `begin_remote_write = 14`, `end_remote_write = 15` were
+// removed and replaced by `start`, `end`, `cancelled`. Their tag numbers are reserved to
+// prevent accidental reuse should we ever need rolling-version compatibility in the future.
```
mschuwalow: Let's remove the reservations as we are breaking backwards compat anyway (for protobuf especially)

golem-worker-executor/src/durable_host/concurrent.rs
vigoo: The cancellation is prepared for P3 futures - there is nothing that can use it yet

golem-worker-executor/src/durable_host/concurrent.rs
vigoo: I believe this is one of those things that should never happen when everything is perfect, but if it does we don't want the whole server to go down because of it. I think having the panic in tests will help in the follow-up tasks for now so I'd leave this as is

golem-worker-executor/src/durable_host/replay_state.rs
vigoo: While checking this realized that 3 code locations were not migrated to the new "reader" yet, fixing that

mschuwalow (approved):
Looks good to me except for the wit_bindgen::spawn handling as discussed.

PR #3631, opened 2026-06-11
agent-permissions: store agent cards in registry service

mschuwalow:


golem-registry-service/src/services/component/write.rs
```diff
@@ -216,12 +248,16 @@ impl ComponentWriteService {
             provision_configs,
         )
         .await?;
-        let component_metadata =
-            component_metadata.with_agent_initial_permissions(default_initial_permissions(
+        let component_metadata = component_metadata.with_agent_initial_permissions(
+            self.create_default_initial_permissions(
```
vigoo, line 254: What happens with this if the validate step below fails?

golem-worker-executor/src/services/card.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+#[async_trait]
+impl CardService for CardServiceDefault {
+    fn record_revoked_cards(&self, card_ids: &[CardId]) {
+        self.cache_revoked_cards(card_ids);
+    }
+
+    async fn check_cards(
```
vigoo, line 96: How efficient this batched registry service call is? As we do this on every agent instantiation. Can we cache it?

golem-common/src/model/component_metadata.rs
```diff
@@ -798,6 +812,15 @@ mod protobuf {
                         )
                     })
                     .collect(),
+                agent_type_initial_permissions: value
+                    .agent_type_initial_permissions
+                    .into_iter()
+                    .map(|(k, v)| {
+                        crate::serialization::serialize(&v)
+                            .map(|template| (k.0, template))
+                            .expect("failed to serialize agent initial permission template")
```
vigoo, line 836: is this acceptable here?

mschuwalow: The component-metadata cards are just placeholders until they are properly passing in through the api and put into the AgentProvisionConfig. So I would like to leave it like this here

golem-worker-executor/src/services/card.rs
mschuwalow: We cannot really cache it as we can otherwise miss invalidations. The correct behaviour relies on us starting the subscription first and then checking all live cards explicitly.
We can potentially improve this by skipping the fetch for a card if another agent already loaded it and we can guarantee that it is still valid. Let me do this in a followup, as the shape of this service changes quite a bit in the following prs.

golem-registry-service/src/services/component/write.rs
mschuwalow: Right now this will leak the card (which is not great, but not a correctness problem). I will switch this to be part of the component transaction once we are actually getting the cards from the api instead of using a placeholder

vigoo (approved):

golem-registry-service/src/services/component/write.rs
mschuwalow: Fixed here https://github.com/golemcloud/golem/pull/3649

golem-worker-executor/src/services/card.rs
mschuwalow: Fixed here https://github.com/golemcloud/golem/pull/3649

PR #3632, opened 2026-06-12
agent-permissions: per agent revoked card queue 

mschuwalow:


golem-worker-executor/src/durable_host/replay_state.rs
```diff
@@ -655,31 +669,47 @@ impl ReplayState {
...
+                    .await;
+                let mut internal = self.internal.write().await;
                 internal.next_skipped_region = internal
                     .skipped_regions
                     .find_next_deleted_region(self.last_replayed_index.get());
             }
         }
     }
 
+    async fn record_card_revoked_events_in_region(&mut self, region: &OplogRegion) {
```
vigoo, line 704: There is a small and a big issue with this. The small is that it reads the whole region in memory, instead it should read in batches, as the skipped region can be quite big.

The bigger problem is that the snapshot based recovery (and the manual update) is also implemented with skipped regions, so this will basically read the whole oplog _before_ the snapshot in memory (or go through it, even if the batching is done) which kind of against the idea of recovering from a snapshot. Maybe the revoke queue need to be included in the snapshot when it gets saved?

golem-worker-executor/src/durable_host/replay_state.rs
```diff
@@ -325,6 +327,12 @@ impl ReplayState {
                         let hash = Self::hash_log_entry(*level, context, message);
                         logs.insert(hash);
                     }
+                    if let OplogEntry::CardRevoked { card_id, .. } = &entry {
```
vigoo: I think this is already recorded by the `internal_get_next_oplog_entry` above

golem-worker-executor/src/durable_host/replay_state.rs
mschuwalow: I was mistakenly assuming that the replay would only touch the entries after the snapshot load. Adjusted.

The code should only need to consider the entries after the last snapshot, as all prior card events should have been included in the snapshot

golem-worker-executor/src/durable_host/replay_state.rs
mschuwalow: record_card_revoked_events_in_region is only used for skipped regions. This handles the events in non-skipped regions

vigoo (approved):

PR #3633, opened 2026-06-15
CLI structured output cleanups

### LINKED ISSUE #3374: Verify that every CLI command properly supports all output formats

noise64:
- resolves https://github.com/golemcloud/golem/issues/3374
- unifies CLI outputs:
 - all meaningful comand now has specific output shape
 - every output has `$type` meta field
 - normalizes casing
 - cleaned up  / reviewed how we log outputs
- introduces JSON schema for the union of the outputs, which can be access with a new CLI command (`golem output-schema`), which can also be used to list all the types, and to return a focused subschema
- to help keeping the schema and the DTOs in shapes:
  - a property based test was added, to generate CLI output DTOs, and match it against the schema
  - there is unit test that find "unregistered" types for testing
  - added relevant skill for this work, and updated related ones


cli/golem-cli/src/command_handler/environment.rs
```diff
@@ -85,15 +92,17 @@ impl EnvironmentCommandHandler {
...
+        if env_summaries.is_empty() && !self.ctx.format().is_structured() {
             logln(format!(
                 "No application environments are available. Use '{}' to create one.",
                 "golem deploy".log_color_highlight()
             ));
-        } else {
-            self.ctx.log_handler().log_view(&env_summaries)?;
         }
 
+        self.ctx.log_handler().log_view(&EnvironmentListView {
```
vigoo: This will print an empty table after we already logged that "No environments are available", is that the intention?

cli/golem-cli/src/model/text/secret.rs
```diff
@@ -39,7 +41,12 @@ impl MessageWithFields for SecretCreateView {
     }
 }
 
+impl CliOutput for SecretCreateView {
```
vigoo: If I understand correctly, these Secret* views hold the secret value itself in `AgentSecretDto` and the `show_sensitive` flag used for text output. If we serialize this to JSON with serde, we are leaking out the secret values regardless of the show-sensitive flag.

cli/golem-cli/src/command_handler/environment.rs
noise64: Yes, the intention is to always return a structured response, in listing commands too.

cli/golem-cli/src/model/text/secret.rs
noise64: I will recheck all the places where we do sensitive checks, most probably we had mixed filtering for these even before, in non-text modes; but thanks for pointing it out, let's "normalize" handling these.

cli/golem-cli/src/command_handler/environment.rs
vigoo: But I'm asking about the human output

noise64: Ah, sorry, will recheck.

noise64: fixed

cli/golem-cli/src/model/text/secret.rs
noise64: reworked secret masking

vigoo (approved):

PR #3637, opened 2026-06-15
 agent-permissions: card installation host function 

mschuwalow:
The wit is a placeholder for now, just need something to use for testing


golem-common/src/base_model/card/monomorphization.rs
```diff
@@ -32,25 +35,52 @@ pub struct AgentPermissionMonomorphizationContext {
...
+) -> EffectiveSurface {
+    let holder = agent_recipient_pattern(context);
+    let cards = wallet_cards
+        .values()
+        .map(|card| match card {
+            StoredCard::Concrete(card) => Ok(card.clone()),
+            StoredCard::Polymorphic(_) => monomorphize_stored_card(card, context),
+        })
+        .collect::<Result<Vec<_>, _>>()
+        .unwrap_or_default();
```
vigoo: This is silently ignoring errors, is that intentional?

golem-worker-executor/src/durable_host/golem/v1x.rs
```diff
@@ -578,6 +604,82 @@ impl<Ctx: WorkerCtx> Host for DurableWorkerCtx<Ctx> {
         Ok(result.uuid.into())
     }
 
+    async fn self_card(&mut self) -> anyhow::Result<Option<golem_api_1_x::host::Card>> {
+        self.observe_function_call("golem::api", "self-card");
+        Ok(self
+            .state
+            .agent_wallet_cards
+            .keys()
```
vigoo, line 612: This is non-deterministic (`HashSet`); we either have to make it deterministic or add to the oplog. Probably just using a different data structure is enough?

golem-common/src/base_model/card/monomorphization.rs
mschuwalow: The entire Result here is a leftover. Updated so it always succeeds :+1:

golem-worker-executor/src/durable_host/golem/v1x.rs
mschuwalow: Done, switched to a btreemap.

Note that this host function is a placeholder for tests and will be removed when I switch to the resource-based representation, but this change is needed anyway

vigoo (approved):

PR #3641, opened 2026-06-17
feat(worker-executor): expose tokio runtime metrics via tokio-metrics crate

kmatasfp:


mschuwalow (commented):
I would prefer to use https://docs.rs/metrics-prometheus/latest/metrics_prometheus/ to emit the metrics-rs metrics directly to the prometheus registry instead of using metrics-exporter-prometheus and concatenating at runtime ourselves.

vigoo:
Should we add these to all services not just the executor?

kmatasfp:
@vigoo now all the services that use observability.rs expose their tokio metrics

vigoo (approved):

PR #3642, opened 2026-06-17
perf(worker-executor): use mimalloc with metrics, enable thin LTO and pin target-cpu

kmatasfp:
Inspired by https://nnethercote.github.io/perf-book/build-configuration.html#maximizing-runtime-speed


mschuwalow (approved):

.github/workflows/ci.yaml
```diff
@@ -860,10 +860,14 @@ jobs:
...
             name: linux/amd64
             target: x86_64-unknown-linux-gnu
+            target_cpu_env: CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTFLAGS
+            target_cpu: "-C target-cpu=x86-64-v3"
           - platform: linux/arm64
             name: linux/arm64
             target: aarch64-unknown-linux-gnu
             cross: true
+            target_cpu_env: CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUSTFLAGS
+            target_cpu: "-C target-cpu=neoverse-n1"
```
mschuwalow: This seems a bit specific for generic docker images we are publishing for users to use. If we want to use specialized images it should be in addition, not instead

kmatasfp: its the same as me saying this x86-64-v3 for amd64 cpus, it says we want at least ARMv8.2-A instruction set. It is not a "specific cpu". This is so SMID, better atomic, better memcopy etc will be enabled, i think we limit ourselves if we also target ancient CPUs, so for amd64 here I am saying we need cpu newer than something from 2013 and for arm 2019, does not seem unreasonable?

All azure, google cloud and aws arm offerings are covered by neoverse-n1, and self hosters with Ampere Altra arm cpus - so basically all datacenter arm cpus

vigoo: If these are "not very special" as you say, both in user's own computers and cloud environments, then I think it's ok to have only one version that requires them, and just document this fact - if we think it's very unusual that someone does not have these, they can just build their own.

Cargo.toml
```diff
@@ -339,6 +340,7 @@ debug = "line-tables-only"
 
 [profile.release]
 panic = "abort"
+lto = "thin"
```
vigoo, line 344: Let's add `lto = false` to the `dev-release` profile above so we don't make the regular CI slower

vigoo (changes requested):

.github/workflows/ci.yaml
mschuwalow: I think using --target-feature is a better fit for us then.

The arm target is also much more specific than the x86 one. From what I understand that really targets a specific microarchitecture (similar to using something like skylake as target cpu for x86). The x86-64-v3 is more of a broad feature target, which I am more fine with shipping by default.

kmatasfp: done

Cargo.toml
kmatasfp: done

vigoo (approved):

PR #3644, opened 2026-06-17
test(benchmark): add cloud benchmark suites and CloudMode fixes for running them

kmatasfp:


golem-test-framework/src/benchmark/results.rs
```diff
@@ -484,13 +484,106 @@ impl Display for BenchmarkResultView {
...
+
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct BenchmarkSuiteResultCollection {
     pub runs: Vec<BenchmarkSuiteResult>,
 }
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct BenchmarkSuiteResult {
+    /// Result format version. Always `1` for results produced by this binary.
+    pub schema_version: u32,
```
vigoo, line 593: This needs a default, because on the daily CI benchmark run we append results to an old result file (that is rendered at https://golemcloud.github.io/benchmark-results/)

golem-test-framework/src/benchmark/results.rs
```diff
@@ -484,13 +484,106 @@ impl Display for BenchmarkResultView {
...
+
+impl RunMetadata {
+    /// Reads all `GOLEM_BENCH_*` environment variables and returns a populated
+    /// `RunMetadata`.  Missing variables produce `None` for that field.
+    pub fn from_env() -> Self {
+        fn env_str(key: &str) -> Option<String> {
+            std::env::var(key).ok().filter(|v| !v.is_empty())
+        }
+        fn env_u32(key: &str) -> Option<u32> {
+            env_str(key).and_then(|v| v.parse().ok())
```
vigoo: I would panic if the env var is set but not a number, to avoid confusion

vigoo (changes requested):

golem-test-framework/src/benchmark/results.rs
kmatasfp: done

golem-test-framework/src/benchmark/results.rs
kmatasfp: done

vigoo (approved):

PR #3652, opened 2026-06-22
Fix: ephemeral worker cleanup after invocation completion

kmatasfp:


golem-worker-executor/src/worker/invocation_loop.rs
```diff
@@ -270,6 +276,13 @@ impl<Ctx: WorkerCtx> InvocationLoop<Ctx> {
             .await;
     }
 
+    async fn archive_ephemeral_oplog(&self) {
```
vigoo: We should not blockingly do this before we remove the ephemeral agent. This should be strictly a background operation.

Doing it like this will hold locks until its done (the per-account max concurrent agent one, for example)

I think the correct fix is to not cancel the scheduled background transfer when the ephemeral oplog is dropped, or something along that line.

kmatasfp: good point, should be addressed now

vigoo (approved):

PR #3657, opened 2026-06-23
Value type refactoring 5 fix 1

### LINKED ISSUE #3654: Value type refactoring followup: restore REST API and OpenSchema conventions
# Missing camel case  / using snake case currently
This should checked in general, examples: AgentConstructorSchema, AgentMethodSchema, AgentDependencySchema, AgentTypeSchema, and AgentConfigDeclarationSchema SchemaType.

This was found on the CLI ouptut schema branch, which has some checks around how the json schema looks like; so I think we could also add a few simple checks in "generate OpenAPI", that could e.g. check that field names in the spec are all camelCase (or if we have exceptions, those should be documented).

# SchemaValue: missing OpenAPI representation
SchemaValue do have serde  tagged represenatation, but in OpenAPI it is "type: object" only

vigoo:
Resolves #3654


golem-schema-derive/src/codegen/poem.rs
```diff
@@ -524,6 +632,178 @@ fn type_schema_ref_expr(ty: &Type) -> TokenStream {
...
+                        let first_str = quote! { #first }.to_string();
+                        iter.all(|e| quote! { #e }.to_string() == first_str)
+                    }
+                    None => true,
+                }
+            };
+            let items_ref = if homogeneous {
+                synth_schema_ref(&t.elems[0])
+            } else {
+                let elem_refs = t.elems.iter().map(synth_schema_ref);
```
noise64: Is it okay that this accepts more then the actual tuple type? As in, (A, B) becomes "array of A or B".

noise64 (approved):

Issue #3529, opened 2026-05-21
New secret host API with opaque secret transfer over RPC


vigoo:
Implements the `golem:secrets@0.1.0` WIT package referenced by `tool-common` (§5.2.4). Builds on the core value-type refactor epic.

Scope:

- Three sibling interfaces a component imports separately so capability is import-gated:
  - `golem:secrets/types` — `secret` resource (handle only; no plaintext getters), `secret-id`, `secret-metadata`. Receive-and-pass capability.
  - `golem:secrets/create` — `secret-from-string`, `secret-from-bytes`. Mint a secret from arbitrary bytes (no versioned source).
  - `golem:secrets/reveal` — `reveal-as-string`, `reveal-as-bytes`. Plaintext extraction; capability-gated by the import.
- First-class secret in the value-tree: `secret-t` arm in `type-node`, `v-secret(secret)` arm in `value-node`.
- Cross-component flow: wire form serializes only `secret-id`; receiver resurrects the resource in its own resource table. WAC `with`-substitution unifies the type at composition time.
- Pin-by-default versioning at resolve time (recorded in `secret-metadata.version`). Required by durability/replay.
- Oplog rules:
  - Reveal: one entry per call recording `(calling-agent, secret-id, config-key, timestamp)`. **Plaintext bytes never logged.**
  - Config-fetch resolve: structured event with resolved version, oplogged for replay determinism.
  - `secret-from-*`: not necessarily oplogged.
- Host-mediated substitution at the syscall boundary (e.g., `golem:secrets-http` accepting `secret` directly in `Authorization` headers) is the recommended non-reveal consumption path; carve out the extension point even if individual extensions follow later.
- No conveniences: no `clone`, no `into-string`, no `equals(secret, string)`, no implicit display.

Out of scope here but related: manifest-side narrowing (`secretKeysReadable` / `secretKeysRevealable`) — that lives in #3550 / the manifest-binding work.

PR #3662, opened 2026-06-23
Value type refactoring 6

### LINKED ISSUE #3404: Finalize the rich types
# Wave 5 — Finalize the rich types

This is the **Wave 5** umbrella for the value/type system refactor. The rich types themselves — cases, validators, canonical encodings, renderers, derive support — already exist after Wave 1 (#3395). Wave 4 (#3400 / #3401 / #3402 / #3403) migrated each SDK's derives over those types. **Wave 5 only contains consumer-specific behaviour that needs both the type to exist AND a real consumer subsystem already migrated.**

Mostly small, mostly per-consumer, parallel-safe. Any sub-task that turns out empty after Waves 1–4 (because a renderer / derive already covered the case adequately) can be dropped — nothing here is required by definition.

## Scope

Three clusters:

- **Capability observability + policy hooks** — wire `HostManagedKind`-driven redaction / audit / no-display behaviour across already-migrated subsystems.
- **Bespoke quota-token derive removed (Rust SDK)** — drop the hand-written `IntoValue` / `FromValueAndType` for `QuotaToken` in the Rust SDK now that the schema-based derive can carry it. (Other SDKs do the equivalent removal in their respective Wave 4 PR — they don't need a Wave 5 step.)
- **Consumer-specific rich-type sweetening** — small per-consumer touches that aren't required for correctness but light up the rich types in idiomatic ways, plus the cross-cutting integration tests.

### 28. Capability observability + policy hooks

- [ ] Log-redaction filter: any `SchemaValue::Secret` / `SchemaValue::QuotaToken` (and any future capability node tagged via `HostManagedKind`) is redacted in tracing output. Display sites consult `HostManagedKind` rather than hard-coding per-case logic.
- [ ] Audit-log writer: capability-bearing values produce structured audit entries (kind, identity, redacted value); use `HostManagedKind` classification.
- [ ] CLI display formatter: capability values render as `<redacted: kind>` rather than the raw value.
- [ ] REST response shaping: capability values appear as schema-only / placeholder in responses where the placement matrix says so; never leak raw secret material.
- [ ] Error rendering (`AgentError::CustomError`): capability fields inside custom errors redact correctly.
- [ ] Decide before merge whether `HostManagedKind` justifies its existence — at least 2 named consumers from the above bullets must remain. If we cannot name 2, drop the marker and enumerate cases inline at each consumer.

### 29. Bespoke quota-token derive removed (Rust SDK)

- [ ] Drop the hand-written `IntoValue` / `FromValueAndType` for `QuotaToken` in the Rust SDK (`sdks/rust/`).
- [ ] Replace with the schema-based derive landed in #3400.
- [ ] Receiver-side lazy lease re-acquire path stays untouched — it operates on the reconstructed `QuotaToken` value regardless of how it was decoded.
- [ ] Integration test: a quota token produced by the Rust SDK round-trips through RPC and through oplog `Custom` and reconstructs correctly on the receiver, using only the schema-based derive.

### 30. Consumer-specific rich-type sweetening

- [ ] **Per-language CLI value parsers** — each of `parse_value_rust` / `parse_value_ts` / `parse_value_scala` / `parse_value_moonbit` gains recognition for the language's native literal forms of the rich semantic types and produces typed `SchemaValue`s instead of strings:
  - `Quantity` literals (e.g. `5.kg()` / `Duration::from_secs(30)` in Rust, `5n * kg` / `Duration.seconds(30)` in TypeScript, equivalent idioms in Scala / MoonBit);
  - `Duration` literals in each language's native form;
  - `Datetime` literals in each language's native form;
  - `Url` and `Path` literals.
  Each language picks what's idiomatic; no shared cross-language literal grammar. The `SourceLanguage::Other` fallback uses the canonical encodings from Wave 1.5.
- [ ] **MCP exporter** maps `Path` and `Url` to MCP resource references where semantically appropriate (file paths → resource URIs, URLs → link blocks).
- [ ] **OpenAPI emitter** polish — format hints for `Datetime` (`format: date-time`), `Duration` (`format: duration`), `Quantity` (with unit annotation), `Path` (`format: file-path`), `Url` (`format: uri`).
- [ ] End-to-end integration test that round-trips each rich type (`Path`, `Url`, `Datetime`, `Duration`, `Quantity`) through SDK → REST → executor → oplog → SDK and asserts canonical-encoding preservation.
- [ ] End-to-end integration test that round-trips a `Secret` and a `QuotaToken` and asserts redaction in tracing output and CLI display, plus correct value reconstruction on the receiver side.

## Exit criteria

- redaction / audit / no-display behaviour is wired across the listed consumers and verified by the integration tests;
- Rust SDK no longer carries the bespoke quota-token derive;
- the integration tests for rich-type round-trip and capability redaction pass end-to-end;
- the `HostManagedKind` marker is either justified by ≥ 2 named consumers or has been dropped per the decision in 28.

## Dependencies

- Depends on **Wave 1** (#3395): rich-type cases, validators, canonical encodings, renderers, derive support.
- Depends on **Wave 2** (#3396 / #3397 / #3398): the consumer subsystems being polished must already be on the new types.
- Depends on **Wave 3** (#3399): wire formats already on the new shape so the integration tests can run end-to-end.
- Depends on **Wave 4** (#3400 / #3401 / #3402 / #3403): SDK derives in place; the bespoke quota-token derive removal in particular needs #3400.
- Independent of **Wave 6** (#3405) cleanup; both can run in parallel after the above.

## Out of scope

- deleting any legacy type, WIT package, crate, or shim → Wave 6 (#3405);
- per-language CLI parser/renderer polish / error-message polish / docs / performance pass → Wave 6 (#3405).

vigoo:
Resolves #3404 

The quota token implementation is the first among multiple similar ones.
Opaque secrets are coming in a separate follow-up PR (#3529) and permission cards can also be added in the same way.

I did not try to generalise the code for the opaque quota tokens yet; I'm sure once we have multiple such examples there will be patterns allowing simplifications.


noise64 (approved):

mschuwalow (approved):
Reviewed the quota token lifting/lowering specifically. Makes sense to me 👍

PR #3663, opened 2026-06-24
Retries and increaesd pool size

vigoo:


golem-worker-executor/src/services/golem_config.rs
```diff
@@ -42,6 +42,11 @@ pub struct GolemConfig {
     pub tracing: TracingConfig,
     pub tracing_file_name_with_port: bool,
     pub key_value_storage: KeyValueStorageConfig,
+    /// Retry policy applied to SQL-backed key-value storage operations when the connection pool
+    /// is briefly exhausted (a pool acquisition timeout). Retrying these transient failures keeps
+    /// hot paths such as promise and worker status updates from crashing the executor under load.
+    #[serde(default = "default_key_value_storage_retry")]
+    pub key_value_storage_retry: RetryConfig,
```
noise64, line 49: Should we (or do we) have similar for all the backends / DBs?

noise64: I just saw that this PR also adds scheduler retry, nested into the Scheduler config. If these retries are tied to the storage, shouldn't they be part of the XxxStorageConfig uniformly?

noise64 (approved):

PR #3669, opened 2026-06-26
Avoid reactivating rpc targets during replay

### LINKED ISSUE #3567: Do not fetch worker metadata on HostWasmRpc::new
Followup to #3563 

We want to avoid fetching worker-metadata and activating the remote agent of a replayed HostWasmRpc::new. 

Couple of options:
1. Make HostWasmRpc::new durable and never activate the remote agent on replay
2. Make HostWasmRpc::new durable and scan forward in the oplog during replay. If the rpc resource is not yet dropeed activate the remote worker, otherwise skip.

vigoo:
Resolves #3567

### DIFF
--- a/golem-worker-executor/src/durable_host/wasm_rpc/mod.rs
+++ b/golem-worker-executor/src/durable_host/wasm_rpc/mod.rs
@@ -356,7 +356,10 @@ impl<Ctx: WorkerCtx> HostWasmRpc for DurableWorkerCtx<Ctx> {
                 }
             }
 
-            ensure_rpc_target_activated(self, self_).await?;
+            if let Err(err) = ensure_rpc_target_activated(self, self_).await {
+                handle.abandon_for_trap();
+                return Err(err);
+            }
 
             let retry_properties =
                 RetryContext::rpc("invoke-and-await", &remote_agent_id, &method_name);
@@ -515,7 +518,10 @@ impl<Ctx: WorkerCtx> HostWasmRpc for DurableWorkerCtx<Ctx> {
                 }
             }
 
-            ensure_rpc_target_activated(self, self_).await?;
+            if let Err(err) = ensure_rpc_target_activated(self, self_).await {
+                handle.abandon_for_trap();
+                return Err(err);
+            }
 
             let retry_properties = RetryContext::rpc("invoke", &remote_agent_id, &method_name);
             let result = loop {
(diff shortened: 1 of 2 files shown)


golem-worker-executor/src/durable_host/wasm_rpc/mod.rs
```diff
@@ -341,6 +356,8 @@ impl<Ctx: WorkerCtx> HostWasmRpc for DurableWorkerCtx<Ctx> {
                 }
             }
 
+            ensure_rpc_target_activated(self, self_).await?;
```
noise64: If this fails, then handle is not completed or abandonded. Is that okay?

golem-worker-executor/src/durable_host/wasm_rpc/mod.rs
```diff
@@ -498,6 +515,8 @@ impl<Ctx: WorkerCtx> HostWasmRpc for DurableWorkerCtx<Ctx> {
...
```
noise64: Same question, as above for the handle.

golem-worker-executor/src/durable_host/wasm_rpc/mod.rs
```diff
@@ -851,6 +873,7 @@ impl<Ctx: WorkerCtx> HostWasmRpc for DurableWorkerCtx<Ctx> {
         }
 
         let schedule_id = ScheduleId::from_idempotency_key(&idempotency_key);
+        let target_worker_fingerprint = ensure_rpc_target_activated(self, this).await?;
```
noise64: Same question, as above for the handle.

golem-worker-executor/src/durable_host/wasm_rpc/mod.rs
```diff
@@ -1533,24 +1565,82 @@ async fn reconstruct_wasm_rpc_resource<Ctx: WorkerCtx>(
...
+                payload.span_id.clone(),
+                env.clone(),
+                config.clone(),
+                *target_fingerprint,
+            ),
+        }
+    };
+
+    let stack = ctx.clone_as_inherited_stack(&span_id);
+    let demand = ctx
```
noise64: spawn_rpc_task_with_retry does classification on create_demand, while here there is no error mapping. it seems that creates different behaviour with different call patterns / replays. shouldn't the classification be common?

golem-worker-executor/src/durable_host/wasm_rpc/mod.rs
vigoo: If these fail that's a panic in the guest and the agent fails and unloads

noise64 (approved):

PR #3671, opened 2026-06-26
Ephemeral agent listing fixes

### LINKED ISSUE #3431: REPL :agent-list --mode ephemeral does not seem to work
Did not try if the corresponding CLI command does not work either - but assuming so.

vigoo:
Resolves #3431 

Adds several tests and fixes two problems:

- `--mode all` was not listing durable+ephemeral (only durable)
- ephemeral agent listing did not work with S3 backend


cli/golem-cli/src/command_handler/worker/mod.rs
```diff
@@ -2186,6 +2206,65 @@ impl WorkerCommandHandler {
         let mut workers = Vec::<AgentMetadata>::new();
         let mut final_result_cursor = Option::<ScanCursor>::None;
 
+        // When a structured mode overlay is provided (--mode all), the string
```
noise64: Can we unify the with and without mode_overlay cases (or always use just the more powerful)? It feels a bit duplicated this way with the repeated loop / cursor handling

noise64 (approved):
Look good, added a minor comment.

PR #3673, opened 2026-06-27
Rust tool definition macros

### LINKED ISSUE #3532: Rust tool definition macros (testable through `discover-tools`)
`#[agent_definition]`-style + fluent / derive entry points for authoring a tool in Rust (§5.3.1, §5.3.5.1).

Scope:

- Annotation-driven definition: command tree, globals, options/flags/positionals, body method, error projection, stream slots, annotations.
- **Full authoring vocabulary from §5.3 / §5.3.1** (lines 1300–1302, 1618–1622): multi-level subcommands, pure dispatchers, aliases, `env-var`, defaults, count / negatable flags, all repeatable modes, multiple constraint kinds, extra type constructors, multiple result formatters, `optional-scalar`, `tail.separator` / `verbatim`, record-typed input params, union discriminators. The macro must cover the full surface, not just the `grep`-level subset — downstream steps (3, 4) depend on it. See the product spec for details.
- Build-time metadata extraction → `discover-tools` registration is a side effect of macro expansion.
- Implicit-body convention enforced (§5.8.1): tool's body method name must equal `tool.commands[0].name`, equal to the tool's metadata name. Divergence is a compile error.
- Doc records (`summary` / `description` / `examples`) preserved in metadata.
- **Help-text rendering** (§2 line 23, §5.1 line 359): SDK-side projection of `doc` records into formatted help text accessible at any depth in the command tree — root command, named subcommand at arbitrary depth, individual argument. The Rust SDK ships its own renderer (idiomatic Rust output). Each subsequent SDK definition step (5, 8, 11, 14) ships its own. A shared metadata-rendering library is acceptable if it doesn't get in the way of language-idiomatic formatting.
- **Cross-language metadata equivalence** (§4.7 lines 336–343): the canonicalized metadata this macro produces must be byte-equivalent (modulo allowed exceptions) to the equivalent definition in TS / Scala / MoonBit. This is a normative invariant, not just a testing nice-to-have — Steps #3536, #3539, #3542 inherit the same requirement and #3560 verifies it end-to-end. See the product spec for details.
- Tested through the exported `discover-tools` function (round-trip authoring → metadata).

vigoo:
Resolves #3532


sdks/rust/golem-rust/tests/tool_canonical.rs
```diff
@@ -0,0 +1,1137 @@
...
+
+        #[derive(ToolError)]
+        enum SetUrlError {
+            #[tool_error(kind = "runtime-error", exit_code = 1)]
+            Failed(String),
+        }
+
+        /// Stupid content tracker.
+        #[tool_definition]
+        trait Git {
```
jdegoes, line 534: I don't see an example of an implicit command (syntax TBD).

noise64 (approved):

PR #3674, opened 2026-06-28
Opaque secrets

### LINKED ISSUE #3529: New secret host API with opaque secret transfer over RPC

vigoo:
Resolves #3529


golem-worker-executor/src/durable_host/secrets/mod.rs
```diff
@@ -0,0 +1,436 @@
...
+                    });
+                    if result.is_ok() {
+                        live_secret = Some(secret);
+                    }
+                    result
+                }
+                Ok(None) => Err(SecretRevealError::VersionNotFound(
+                    entry.pinned_revision.get(),
+                )),
+                Err(error) => Err(SecretRevealError::Internal(error.to_string())),
```
noise64, line 322: Is it okay to persist here potentially transient errors? Can these be retried later?

golem-registry-service/src/repo/agent_secret.rs
```diff
@@ -376,4 +398,31 @@ impl AgentSecretRepo for DbAgentSecretRepo<PostgresPool> {
 
         Ok(results)
     }
+
+    async fn get_revision(
```
noise64, line 413: Is it intentional (probably yes), that this does not filter on deleted apps / envs (AFAIR we usually check those too)?
If yes, maybe it worth mentioning this, or check what naming pattern we use in similar cases where we allow seeing deleted entities.

sdks/ts/packages/golem-ts-sdk/scripts/patch-agent-guest-dts.mjs
```diff
@@ -0,0 +1,68 @@
+import fs from 'node:fs';
```
noise64: Why do we need patching? Is this a wasm-rquickjs missing things, or something else?

golem-registry-service/src/repo/agent_secret.rs
vigoo: Yes because we are not storing any secrets now in the oplog - so we have to be able to resolve arbitrary old ones during replay.

sdks/ts/packages/golem-ts-sdk/scripts/patch-agent-guest-dts.mjs
vigoo: i have no idea what this is and definitely should not exist 😱

golem-worker-executor/src/durable_host/secrets/mod.rs
vigoo: it should be retried with the active retry policy and eventually fail permanently. let me check if this is the case.

golem-registry-service/src/repo/agent_secret.rs
```diff
@@ -405,21 +416,27 @@ impl AgentSecretRepo for DbAgentSecretRepo<PostgresPool> {
...
             .fetch_optional_as(
                 sqlx::query_as(indoc! {r#"
                     SELECT sec.environment_id, sec.path, sec.agent_secret_data, sec.created_at AS entity_created_at, secr.agent_secret_id, secr.revision_id, secr.agent_secret_revision_data, secr.created_at, secr.created_by, secr.deleted
                     FROM agent_secrets sec
+                    JOIN environments e ON e.environment_id = sec.environment_id
+                    JOIN applications ap ON ap.application_id = e.application_id
+                    JOIN accounts a ON a.account_id = ap.account_id
                     JOIN agent_secret_revisions secr ON secr.agent_secret_id = sec.agent_secret_id AND secr.revision_id = $4
                     WHERE sec.environment_id = $1 AND sec.agent_secret_id = $2 AND sec.path = $3 AND secr.deleted = FALSE
+                        AND ($5 OR (sec.deleted_at IS NULL AND e.deleted_at IS NULL AND ap.deleted_at IS NULL AND a.deleted_at IS NULL))
```
noise64: in general we switched from this style to use two static selects, or a query builder, to avoid accidental planner complexities

noise64 (approved):
added minor comment on the sql fix, otherwise looks good

PR #3677, opened 2026-06-29
Add Kotlin SDK for building Golem agents (MVP)

JohnSColeman:
## Summary

This adds a **Kotlin language binding for Golem** so `golem new --template kotlin` works
end-to-end. It is modelled one layer down on the Scala SDK and reaches Wasm the same way:
**Kotlin/JS (IR) → rollup bundle → `wasm-rquickjs inject-js` into a prebuilt QuickJS guest**.
No new engine is introduced — the JS path, WASI host imports, and WIT are reused unchanged.

Opening this as an **initial PR for feedback** — it is an MVP focused on the durable-counter
path (parity with the Quickstart TS/Scala counter), not a feature-complete SDK. I'd welcome
direction on the integration approach before expanding host-API coverage.

> **Rebased onto current `main` and migrated to `golem:agent@2.0.0`** (the `schema-value-tree`
> value model). Verified end-to-end against a current-`main` host — see below.

### Architecture (load-bearing)

The SDK is a **normal Kotlin/JS library** (`cloud.golem:golem-kotlin-sdk`) that Kotlin/JS
**bundles into the agent's own JS bundle** — exactly like Scala's `scala.js`. The embedded
`agent_guest.wasm` is therefore **generic** (a `user=@slot` only, no baked-in SDK), so it is
SDK-version-independent and only regenerated when the Golem WIT surface changes.

## What's included

- **`sdks/kotlin/`** — the SDK toolchain:
  - `sdk/` — `@Agent` / `@Endpoint` / `@Prompt` / `@Description` annotations, `BaseAgent`, the
    agent runtime (`initialize`/`invoke`/`getDefinition`/`discoverAgentTypes`), the 2.0.0
    `schema-value-tree` value marshalling, and the `golem:api/host` binding backing `agentId`.
  - `ksp/` — a KSP processor (the sbt-macro analogue) that reads `@Agent` classes and generates
    the registration glue, `main()`, and agent metadata at compile time.
  - `gradle-plugin/` — the `cloud.golem.wasm-component` Gradle plugin (the `sbt-wasm-component`
    analogue): bundles the agent JS and ships the prebuilt `agent_guest.wasm` as a resource.
  - `wit/` (synced to `golem:agent@2.0.0`), `scripts/generate-agent-guest-wasm.sh`, `example/`,
    `AGENTS.md`, `README.md`.
- **`cli/golem-cli/templates/kotlin/`** — the `golem new --template kotlin` project template
  (mirrors the Scala template: `common/`, `common-on-demand/`, `component/`, `default/`).
- **`golem-skills/skills/kotlin/`** — 7 core Kotlin agent skills (annotate, create-instance,
  invoke, multi-instance, http-params, stateless, repl).
- **`GuestLanguage::Kotlin`** across `golem-cli` (mirrors Scala) + `build.rs` `SUPPORTED_LANGUAGES`,
  including a Kotlin tool-requirements check (Java 17+, Gradle).

## Verified end-to-end (against `golem:agent@2.0.0` / current `main`)

Driven live through the real scaffolding flow:

```
golem new --template kotlin --component-name example:counter --yes app
golem build            # gradle bundle → injectToPrebuiltQuickjs → wizer preinitialize
golem deploy --yes
golem agent invoke 'CounterAgent("c1")' increment      # → 1, 2, 3 (durable state)
# HTTP gateway (:9006, Host app.localhost:9006):
POST /counters/c1/increment → 200, GET /counters/c1/value → 200; a fresh instance c2 → 1
```

SDK unit tests (`jsTest`) and KSP processor tests pass.

## What is NOT done yet (MVP scope)

This is a deliberate counter-grade MVP (~10–15% of the Scala SDK surface). None of the gaps are
regressions — they are roadmap items, tracked for follow-ups:

- **Host APIs** (largest gap, mostly unimplemented): durability control, transactions, promises,
  agent-to-agent RPC, RDBMS, LLM, quota, oplog, retry. Only `getSelfMetadata` is wired (for `agentId`).
- **`schema-value-tree` value model**: only `s32` / `s64` / `string` / `unit` round-trip today —
  records, variants, enums, lists, options, results, floats, bools, and u-widths are not yet done.
- **`agentType` / `agentName`**: best-effort (`""` fallback); faithful values need
  `golem:agent/host` `parse-agent-id` bridged into the wasm template.
- **Extended annotations**: `@Agent` mode/auth/cors/snapshotting/phantom, `@Endpoint` auth/cors,
  `@header`, multimodal typing — not yet threaded through KSP.
- **`suspend` → durable-invocation** mapping (the counter is synchronous).
- **First-class WIT functions for methods** (currently emitted as WIT metadata comments).
- Only 7 Kotlin skills so far (Scala ships ~37); the rest follow as the underlying SDK features land.

`sdks/kotlin/README.md` carries this scope note as well.

## Notes for reviewers

- **Build prerequisites for a scaffolded project**: a JDK 17+, a **system `gradle` 8.11+**
  (Kotlin 2.4.0 compatible) on `PATH`, and `rollup`. The template intentionally ships **no Gradle
  wrapper** — a binary jar can't pass through `golem new`'s text-based file generation, so the
  build template invokes a system `gradle`, mirroring how the Scala template invokes a system `sbt`.
- The SDK toolchain is **not built by `cargo make build`** (it has its own Gradle build); for local
  dev / CI it publishes to mavenLocal at `0.0.0-SNAPSHOT` (the Scala `golemPublishLocal` analogue).
- `GuestLanguage::Kotlin` mirrors Scala in every match arm (no bridge gen, REPL → TypeScript).
- Happy to split this into smaller PRs (e.g. SDK toolchain vs. golem-cli template/skills) if that's
  easier to review.


vigoo:
Do you think it would be possible to make this a "proper" compile-to-wasm SDK instead of taking the compile-to-js + wasm-rquickjs path? With Scala we had no choice but it adds a lot of overhead. The Kotlin docs says there is "experimental" component model support, did you explore that?

JohnSColeman:
> Do you think it would be possible to make this a "proper" compile-to-wasm SDK instead of taking the compile-to-js + wasm-rquickjs path? With Scala we had no choice but it adds a lot of overhead. The Kotlin docs says there is "experimental" component model support, did you explore that?

Hi, was not aware of this option so modelled on Scala as this was a proven route. I will investigate that option.

JohnSColeman:
@vigoo This is what I got after investigation, Kotlin/Wasm is mandatorily WasmGC and Golems engine has WasmGC off, Claude says...

"The Component Model's canonical ABI lowers WIT types (strings, records, lists) over **linear
memory**. Kotlin's values live in the **WasmGC heap**. Bridging the two requires a bindgen that
copies GC objects to/from linear memory at the export boundary — the exact role `wit-bindgen` plays
for Rust/C. No such generator exists for Kotlin, and WasmGC-inside-components is itself nascent in
the toolchain. This is *also* why Scala (and even TypeScript) take the JS path: it is currently the
only ABI-complete route for these languages onto Golem."

Will dig in further and see if there are workarounds.

vigoo:
I think we can turn on wasm-gc, if that's the only blocker (if I remember correctly I tried it once).

JohnSColeman:
> I think we can turn on wasm-gc, if that's the only blocker (if I remember correctly I tried it once).

it seems this is the full requirement @vigoo :
Golem's `Config` would need `wasm_exceptions(true)` in addition to  `wasm_gc(true)` and `wasm_function_references(true)`. Enable all 3 and we can do as you propose.

PR #3679, opened 2026-06-30
Fix snapshot recovery

kmatasfp:


golem-worker-executor/src/worker/invocation_loop.rs
```diff
@@ -1154,12 +1154,13 @@ impl<Ctx: WorkerCtx> Invocation<'_, Ctx> {
                 .await;
         }
 
-        self.store.data_mut().begin_call_snapshotting_function();
```
vigoo, line 1157: I think this should not have been removed - without it we will write side-effects to the oplog during invocation below. Also keeping the "end" below, but fixing it to work without this removed begin seems to be weird to me.

golem-worker-executor/src/worker/invocation_loop.rs
```diff
@@ -1378,12 +1379,13 @@ impl<Ctx: WorkerCtx> Invocation<'_, Ctx> {
             return CommandOutcome::Continue;
         }
 
-        self.store.data_mut().begin_call_snapshotting_function();
```
vigoo, line 1381: Same concern as above

vigoo (changes requested):
During review I've found a couple of (pre-existing) places where the `snapshotting_mode.is_none()` check is missing, only `is_live()` is checked. Those should be adjusted to work the same as the rest of the places already properly handling the special snapshotting-function mode:

| API | Location |
|---|---|
| **RDBMS transaction begin** — `if self.is_live()` then replay-cursor `get_oplog_entry!` in else | `golem-worker-executor/src/durable_host/mod.rs:1703` (fn), `:1710` (raw is_live), else branch `:1744-1785` |
| RDBMS **rollback** uses *effective* `durable_execution_state().is_live` → split-brain vs begin | `golem-worker-executor/src/durable_host/rdbms/mod.rs:939`, `:952`, `:963` |
| RDBMS **commit** uses *effective* is_live → split-brain vs begin | `golem-worker-executor/src/durable_host/rdbms/mod.rs:987`, `:999`, `:1009`, `:1035` |
| `get_oplog_index` | `golem-worker-executor/src/durable_host/golem/v1x.rs:319` (fn), `:321` |
| `set_oplog_index` | `golem-worker-executor/src/durable_host/golem/v1x.rs:346` (fn), `:367`, `:390` |
| `mark_begin_operation` (atomic region) | `golem-worker-executor/src/durable_host/golem/v1x.rs:411` (fn), `:414` |
| `mark_end_operation` / persistence-level branch | `golem-worker-executor/src/durable_host/golem/v1x.rs:483` (fn), `:488`, `:525` |
| `set_retry_policy` / `remove_retry_policy` | `golem-worker-executor/src/durable_host/golem/retry_api.rs:131`, `:147` |

vigoo (approved):

PR #3680, opened 2026-07-01
TS SDK redesign

### LINKED ISSUE #3449: Fluent TS SDK: Basic agent features
Create a temporary local-only, non-published fluent/declarative SDK by porting the core effect-golem adjusted to use Standard Schema.

Checkpoint: a minimal counter-style agent builds, deploys, and can be invoked.

Context: #3401

### LINKED ISSUE #3450: Fluent TS SDK: Feature parity, finalized naming
Cover and test all current TS SDK features, finalize naming questions. If needed, also align naming changes with TS bridge and REPL.

### LINKED ISSUE #3451: Fluent TS SDK: Switch to fluent API
Replace the old SDK with the new fluent SDK as the official TypeScript SDK. This also includes the adjusted templates and skills.

Templates could / should also showcase using multiple schema libs.

### LINKED ISSUE #3453: Fluent TS SDK: Docs
Update docs.

noise64:
- resolves https://github.com/golemcloud/golem/issues/3449
- resolves https://github.com/golemcloud/golem/issues/3450
- resolves https://github.com/golemcloud/golem/issues/3451
- resolves https://github.com/golemcloud/golem/issues/3453
- adds skills for migration from old SDK, but in a followup card will implement hinting on using that skill when detecting upgrades or compile errors around using the old SDK
- we should also do a follow-up later on TS bridges, if there is any naming that we should align


vigoo (commented):
A few gaps around compile-time safety and client ergonomics, left as inline comments.

sdks/ts/packages/golem-ts-sdk/src/fluent/defineAgent.ts
```diff
@@ -0,0 +1,307 @@
...
+   *   every mount `{var}` must be an id field — enforced via
+   *   {@link MountSpecCovering}.
+   * - Every `{var}` in the optional `webhookSuffix` must be an id field —
+   *   enforced via {@link WebhookVarsValid}.
+   *
+   * Plain object-literal / segment-array forms (which carry no phantom brand)
+   * skip these gates and are validated only by the runtime checks in
+   * `runtime.ts`.
+   */
+  http?: MountSpecCovering<Id, MV, WV> & WebhookVarsValid<Id, WV>;
```
vigoo: `http` stays optional even when one of the methods declares HTTP endpoints, so forgetting the mount only fails at runtime (the `'declares HTTP endpoint(s) but the agent has no HTTP mount'` throw in `runtime.ts`). Since `MethodSpec` already carries its endpoints, we could track a "has HTTP" flag at the type level (e.g. a boolean phantom on `MethodSpec`, aggregated over `Methods`) and make `http` a *required* field of `AgentSpec` whenever any method declares endpoints. That turns a deploy-time error into a compile error, and we already do similar type-level work here with `MountSpecCovering` / `WebhookVarsValid`.

sdks/ts/packages/golem-ts-sdk/src/fluent/schema/zod.ts
```diff
@@ -0,0 +1,482 @@
...
+      getter: def.typeName === 'ZodLazy' ? def.getter : undefined,
+    };
+  }
+  throw new Error('Unrecognised Zod schema shape (expected Zod v3 or v4 internals)');
+}
+
+/** Wrappers that pass through their inner schema unchanged. */
+const TRANSPARENT = new Set(['default', 'readonly', 'catch', 'effects', 'nonoptional', 'pipe']);
+/** Wrappers that make the value optional. */
+const OPTIONAL = new Set(['optional', 'nullable']);
```
vigoo, line 185: `optional` and `nullable` are both collapsed into the same WIT `option` codec, and `fromValue` decodes the none case to `undefined` unconditionally. That breaks round-tripping for nullable fields: a field declared `z.number().nullable()` infers as `number | null`, but after a round trip the receiver gets `undefined` — a value that isn't even assignable to the schema's own inferred type. Same issue in the valibot walker (`nullable` / `nullish` / `undefinedable` all in one `OPTIONAL` set).

Suggest recording which wrapper produced the option and decoding none accordingly: `undefined` for `optional`, `null` for `nullable`, and picking one (or `undefined`) for `nullish`. The wire format doesn't change; only the decode side needs the distinction.

sdks/ts/packages/golem-ts-sdk/src/fluent/defineAgent.ts
```diff
@@ -0,0 +1,307 @@
...
+/**
+ * Snapshotting configuration. Either a bare {@link SnapshotPolicy} (the SDK
+ * snapshots all of `this` by reflection — back-compat default), or `{ policy,
+ * state }` where `state` is a Standard Schema: only the schema-declared fields of
+ * `this` are serialized (typed + scoped), fixing over-broad snapshots. For fully
+ * custom serialization supply `snapshot: { save, load }` on `implement(...)`.
+ */
+export type SnapshottingSpec =
+  | SnapshotPolicy
+  | { policy?: SnapshotPolicy; state: StandardSchemaV1 };
```
vigoo: In `{ policy, state }`, `state` is an untyped `StandardSchemaV1`, so nothing connects it to the `State` object that `init` actually returns. A typo'd field name, or drift after a refactor of `init`, compiles fine and only surfaces when `validateSnapshotState` rejects the instance at the first snapshot save — on a deployed agent.

Since `State` is only known at `implement(...)` time, one way to get compile-time checking: thread the snapshot schema type through `AgentSpec` into `AgentImplementation` and constrain `State` against the schema's output type (e.g. `State extends StandardSchemaV1.InferOutput<StateSchema>`), so the schema-declared fields must exist on the state with matching types.

sdks/ts/packages/golem-ts-sdk/src/fluent/client.ts
```diff
@@ -0,0 +1,260 @@
...
+ * Build a typed RPC client factory for a remote agent definition.
+ *
+ * ```ts
+ * const counter = clientFor(CounterDef);
+ * const c1 = counter({ name: 'c1' });
+ * const next = await c1.increment({ by: 5 });
+ * c1.increment.trigger({ by: 1 }); // fire-and-forget
+ * ```
+ */
+export function clientFor<Id extends IdRecord, Methods extends MethodsRecord>(
```
vigoo: A few ergonomics gaps on the client surface:

1. **Fresh phantom clients.** `clientFor(def)(id, phantomId?)` supports *existing* phantom ids, but there's no helper for "create a new phantom instance and give me its id" — callers must generate a `Uuid` themselves and thread it through. Consider something like `client.newPhantom(id)` returning the client with the generated `phantomId` exposed on it (mirroring how `fork()` returns `ForkDetails`).
2. **Cancellation on awaited calls.** The plain call `method(input)` can't be cancelled; you must switch to `abortable(signal, input)`, which also reorders the arguments. An optional trailing options argument on the main call shape (`method(input, { signal })`) would keep one call shape and make cancellation compositional with existing `AbortSignal`-based code.
3. **`schedule` vs `scheduleCancelable`.** `schedule()` returns `void` while `scheduleCancelable()` returns a `CancellationToken`. Unless the non-cancelable host path is meaningfully cheaper, consider having `schedule` always return the token — one method, and cancellation doesn't require deciding upfront.

sdks/ts/packages/golem-ts-sdk/src/index.ts
```diff
@@ -174,13 +166,8 @@ async function invokeTool(
...
 async function discoverAgentTypes(): Promise<AgentType[]> {
   try {
-    // Check if there were any validation errors during agent registration
-    const validationError = getAgentValidationError();
-    if (validationError) {
-      // Don't return any agent types if there was a validation error
-      throw createCustomError(validationError.message);
-    }
-
+    // The fluent surface validates eagerly (throws during `defineAgent(...)` at
```
vigoo: `defineAgent(...)` throwing at module load means a single invalid agent definition takes down the whole component at instantiation: the guest never gets far enough for `discover-agent-types` to run, so the CLI/registry see a generic wasm trap instead of a structured, per-agent diagnostic. This also means one bad agent hides all the valid ones in the same component.

Consider capturing definition-validation failures at `defineAgent`/`implement` time and re-emitting them as typed `AgentError`s from `discoverAgentTypes` (and from `initialize` for the affected agent type), so tooling can render "agent X is invalid because Y" while other agents in the component keep working. Eager throwing could stay as-is under a dev/test flag where fail-fast is preferable.

vigoo (approved):

PR #3683, opened 2026-07-02
Rust guest bridge generation

### LINKED ISSUE #3535: Rust tool bridge (CLI tool discovery + tool bridge manifest section)
The bridge piece is the cross-language code generator path (§5.8.3) — produces a Rust client for a tool authored in some other language. Includes the manifest surface that drives generation, plus the minimum tool-deployment / -registry support needed to make `from: registry` meaningful.

Scope:

- **`toolClients:` manifest block** (§5.5 lines 2158–2188): per-language `tools:` list with `as:` rename and `from: registry | local` source pin.
- CLI subcommand emitting Rust clients for selected tools from selected metadata sources (§5.8.3.1).
- Metadata-source resolution (§5.8.3.2): local component build wins by default; `from: registry` forces a fresh fetch; `from: local` forbids fallback.
- **Path 2 codegen contract** (§5.8.3.2 lines 2820–2836): control-plane metadata endpoint (`GET /v1/tools/{tool-name}/metadata`), CI-safe non-interactive auth, version / WIT compatibility checks, response header carrying WIT version, hard failure on unknown constructors / record-shape drift. See the product spec for details.
- Output layout per language (§5.8.3.3), build orchestration (§5.8.3.4).
- **Local CLI discovery**: walk local app components, run `discover-tools`, list available tool metadata.
- **Minimal deployed-tool registry support** for `from: registry`: a per-environment list of registered tool types plus their metadata, sufficient for the bridge generator's fetch path. The full per-tool binding shape (versions, parameters, secret-key narrowing, middleware bindings, dispatch table generation) is #3550 — this step delivers only what's needed to *list and fetch metadata*.
- Symmetry property (§5.8.7): a tool authored in language A and called from language B via the generated client behaves identically to the same-language case.

Subsequent bridge tickets (Steps #3538, #3541, #3544) reuse the manifest block and CLI machinery, adding per-language output emitters.

vigoo:
Resolves #3535


cli/golem-cli/src/app/build/check/rust.rs
```diff
@@ -41,13 +41,17 @@ pub(super) fn plan_rust_cargo_fix_steps(
     let mut steps = Vec::new();
     for component_name in ctx.application_context().selected_component_names() {
         let component = ctx.application().component(component_name);
-        if component.guess_language() != Some(GuestLanguage::Rust) {
+        let is_rust = component.guess_language() == Some(GuestLanguage::Rust);
```
noise64: I think this change is not needed

cli/golem-cli/src/app/build/check/rust.rs
```diff
@@ -95,7 +99,17 @@ pub(super) fn plan_rust_cargo_fix_steps(
     }
 
     if let Some((workspace_cargo_toml_path, workspace_source)) = cargo_workspace_manifest {
-        let mut working = workspace_source.clone();
+        // A root Cargo.toml can be both a package manifest and the workspace manifest, so the
+        // component loop above may already have produced a fix step for this same path. Continue
+        // from that step's edited content and merge into it, otherwise the two steps would be
+        // applied in sequence and the later one would overwrite the earlier one's changes.
+        let existing_step_index = steps
```
noise64: this solution feels fragile, cannot we store instead step by path or something cleaner?

cli/golem-cli/src/app/build/add_metadata.rs
```diff
@@ -65,10 +65,23 @@ fn add_metadata(
...
+    let component_names = ctx
+        .application_context()
+        .selected_component_names()
+        .iter()
+        .cloned()
+        .collect::<Vec<_>>();
+    add_metadata_to_components(ctx, &component_names).await
+}
+
+pub async fn add_metadata_to_components(
```
noise64, line 67: The naming of add_metadata_to_selected_components, add_metadata_to_components, build_components, build_selected_components is quite random, and not following the same patterns.

But my bigger problems is that there should not be a parallel selection process on top of the BuildContext, as that can lead two "double" building and logic. If i understand correctly, we need these to enfocre dep ordering. If that is true I would prefer having logic expressed through the selection, and make that ordered, to avoid having multiple ways to reason about what to build and when.

cli/golem-cli/src/app/build/gen_bridge.rs
```diff
@@ -1,141 +1,750 @@
...
 use anyhow::bail;
 use camino::Utf8PathBuf;
+use golem_common::model::agent::extraction::ExtractedComponentMetadata;
+use golem_common::model::component::ComponentName;
 use itertools::Itertools;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
+
+#[derive(Debug, Default)]
+pub(crate) struct ComponentMetadataCache {
```
noise64: Why do we need another cache? And it feels out of place here in the gen_bridge.

cli/golem-cli/src/app/edit/cargo_toml.rs
```diff
@@ -116,6 +116,88 @@ pub fn collect_dependency_specs(
     Ok(result)
 }
 
+pub fn collect_package_dependency_specs(source: &str) -> anyhow::Result<Vec<DependencySpec>> {
```
noise64: These are only used in tests

cli/golem-cli/src/app/build/task_result_marker.rs
```diff
@@ -331,20 +333,23 @@ impl TaskResultMarkerHashSource for GenerateBridgeSdkMarkerHash<'_> {
...
-            self.component_name, self.agent_type_name, self.language
-        )))
+        Ok(HashFromString(serde_json::to_string(&serde_json::json!({
+            "componentName": self.component_name.to_string(),
+            "targetName": self.target_name,
+            "kind": self.kind,
+            "language": self.language.id(),
+            "bridgeMode": self.bridge_mode.id(),
+        }))?))
     }
```
noise64, line 345: Why the inline json? The other markers us serde annotations and simply self, for json hashes

cli/golem-cli/src/app/component_metadata.rs
```diff
@@ -16,15 +16,22 @@ use crate::log::LogColorize;
...
 use std::collections::{BTreeMap, BTreeSet};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
-pub struct AgentTypeRegistry {
-    cache: Cache<ComponentName, (), Vec<AgentTypeSchema>, Arc<anyhow::Error>>,
+/// Returns the name of a tool: the name of the root node of its command tree.
+pub fn tool_name(tool: &Tool) -> Option<&str> {
+    tool.commands.nodes.first().map(|node| node.name.as_str())
+}
```
noise64: Feels like something that should be on Tool itself

cli/golem-cli/src/model/app_raw.rs
```diff
@@ -257,6 +258,8 @@ pub struct ComponentTemplate {
     pub component_wasm: Option<String>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub output_wasm: Option<String>,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub dependencies: Vec<String>,
```
noise64: Are these AgentTypeNames or Components?

cli/golem-cli/src/model/app_raw.rs
```diff
@@ -958,6 +1011,26 @@ pub struct BridgeSdkLanguageTargets {
     pub agents: LenientTokenList,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub output_dir: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub guest: Option<BridgeSdkModeTargets>,
```
noise64: I would like to revisit this shape, as it feels "half-nested"; but probably separetely from this PR.

I think:
- either we should have two root fields for bridges and guest bridges (this seems to be simple, and could be also done as part of this PR)
- or have "sibling" nesting here, but that would require manifest migration too

cli/golem-cli/src/model/app_raw.rs
noise64: And how these interact with the guest bridge manifest parts?

cli/schema.golem.cloud/app/golem/1.6.0-dev.3/golem.schema.json
```diff
@@ -1213,6 +1213,25 @@
       "description": "Bridge SDK language targets",
       "type": "object",
       "additionalProperties": false,
+      "properties": {
```
noise64: older schemas should not be changed

cli/schema.golem.cloud/app/golem/1.6.0-dev.4/golem.schema.json
```diff
@@ -178,6 +178,13 @@
           "type": "string",
           "description": "File path for the output WASM component which is ready to be uploaded to Golem."
         },
+        "dependencies": {
```
noise64: older schemas should not be changed

cli/golem-cli/src/app/build/mod.rs
```diff
@@ -33,6 +53,11 @@ pub async fn build_app(ctx: &BuildContext<'_>) -> anyhow::Result<()> {
         check_build_tool_requirements(ctx).await?;
     }
 
+    if should_run_bridge_scheduler(ctx) {
```
noise64: Connected to the nodes on the "build" and "metadata" steps: I do not think we should have two separated build steps (expecially not after each other), rather we should have one unified build, where we calculate the (potenitially extended) staged selection and order based on deps, and then use that in the substeps. or if we need to mix the stages, then create a proper build plan as part of the build context. we did similar sthings when we had generated wasm-rpc deps.

Also the scheduler naming is a bit odd choice for me, this is not scheduling, but dependency ordering,

cli/golem-cli/src/bridge_gen/rust/mod.rs
```diff
@@ -516,29 +909,300 @@ impl RustBridgeGenerator {
...
+            .methods
+            .iter()
+            .map(|method| self.to_rust_ident(&method.name))
+            .collect::<Vec<_>>();
+
+        let await_names = user_bases
+            .iter()
+            .map(|base| Self::ident_from_name(impl_names.fresh(base)))
+            .collect::<Vec<_>>();
+        let trigger_names = user_bases
```
noise64: Are these the user visible names?

cli/golem-cli/src/bridge_gen/rust/mod.rs
```diff
@@ -422,6 +598,223 @@ impl RustBridgeGenerator {
...
+                        #phantom_id_param.map(Into::into),
+                    ).map_err(|__e| crate::__golem_bridge_runtime::ClientError::RpcFailed { message: format!("{__e:?}") })?;
+
+                    let wasm_rpc = golem_rust::golem_agentic::golem::agent::host::WasmRpc::new(
+                        #agent_type_name,
+                        golem_rust::encode_schema_value(&constructor_value)
+                            .map_err(|__e| crate::__golem_bridge_runtime::ClientError::SchemaEncodeFailed { message: __e.to_string() })?,
+                        #phantom_id_param.map(Into::into),
+                        #agent_config_param,
+                    );
```
noise64: It seems that the agent_id is not (yet?) used at all, and this part is duplicated:

```
golem_rust::encode_schema_value(&constructor_value)
                            .map_err(|__e| crate::__golem_bridge_runtime::ClientError::SchemaEncodeFailed { message: __e.to_string() })?,
```

cli/schema.golem.cloud/app/golem/1.6.0-dev.1/golem.schema.json
```diff
@@ -128,6 +128,37 @@
         }
       }
     },
+    "initialCard": {
```
noise64, line 131: dev.1 is still modified, which it should not be, and there is two new schema now in the PR (dev.5 and dev.6), one should be enough per PR.

cli/golem-cli/src/app/build/add_metadata.rs
noise64: I see that we are now having planning on the higher level, and not driven by the selections only, but then let's do the cleanup fully and only:
- add_metadata_to_components and
- build_components

both working on component_names: &[ComponentName], and let's remove the "selected" naming variants (note that the naming is still the opposite in these two, but after only hav ing 1 variants, that won't matter anymore)

cli/golem-cli/src/app/build/task_result_marker.rs
```diff
@@ -315,10 +317,39 @@ impl TaskResultMarkerHashSource for GetServerIfsFileHash<'_> {
...
 }
 
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
 pub struct GenerateBridgeSdkMarkerHash<'a> {
+    #[serde(serialize_with = "serialize_display")]
```
noise64: Do we really need these 3 custom serialize_with?

cli/golem-cli/src/model/app.rs
```diff
@@ -833,26 +909,103 @@ impl Application {
...
+                .and_then(|sdk| sdk.output_dir.as_ref()),
+        };
+
+        match output_dir {
+            Some(output_dir) => self
+                .bridge_sdks
+                .source
+                .join(output_dir)
+                .join(bridge_client_directory_name_for_mode(agent_type_name, mode)),
+            None => match mode {
```
noise64, line 942: I think we should not have both `bridge_client_directory_name` and `bridge_client_directory_name_for_mode`, only bridge_client_directory_name with mode always requird.

cli/golem-cli/src/model/app.rs
```diff
@@ -833,26 +909,103 @@ impl Application {
...
+                    .temp_dir()
+                    .join("bridge-sdk")
+                    .join(language.id())
+                    .join(mode.id())
+                    .join(bridge_client_directory_name_for_mode(agent_type_name, mode)),
+            },
+        }
+    }
+
+    pub fn dependency_bridge_sdk_dir(
```
noise64, line 953: What is the relation between bridge_sdk_dir and dependency_bridge_sdk_dir?

cli/golem-cli/src/app/build/add_metadata.rs
noise64: After checking in more detail, how the planning works, now we only call these with exactly one component (both build and add metadata). This will make the build logging unnecceseraly nested kinda like:

```
Building components
  Building component a
Building components
  Building component b
```

So that should be fixed, probably by adding a root nesting log at build_components_with_dependency_ordering itself, and switch these two work on exactly one component.

cli/golem-cli/src/model/app_raw.rs
```diff
@@ -946,20 +986,94 @@ impl BridgeSdks {
...
+pub struct BridgeSdkExternalTargets {
+    #[serde(default, skip_serializing_if = "LenientTokenList::is_empty")]
+    pub agents: LenientTokenList,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub output_dir: Option<String>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct BridgeSdkModeTargets {
```
noise64: Should this be called BridgeSdkInternalTargets?
Also I'm still a but confused on how the dependencies and this part plays together.
I would understand customizing just the the output dir for internal guests, so one could make them part of the application source code (e.g. put them into ts/guests/src or similar), but having agents and tools here feels duplicated or consfusing. Both of these places drive build dependecy ordering? Or how do these relate? Maybe I'm missing some obvious usecase.

noise64 (approved):
Added a few more notes, but otherwise looks good

PR #3697, opened 2026-07-13
Port batch oplog replay

kmatasfp:
This PR adds batched replay.

Summary of changes:
- Replay oplog entries in buffered batches of up to 1,024 entries instead of issuing a storage read for every entry.
- Preserve correct replay behavior through speculative rollback, hint skipping, skipped-region jumps, sparse cross-layer reads, and replay-target changes.
- Bound each batch to the current replay target and invalidate prefetched entries when the target moves backward, which is needed for debug replay/rewind overrides.
- Fence archive transfers during worker deletion: abort the in-flight transfer before deleting oplog layers, preventing deleted oplogs from being recreated.
- Clean up transfer-fiber registrations when oplogs close, guarded by pointer identity so a prior instance cannot remove a newer reopened instance’s registration.
- Add regression coverage for sparse batch reads, replay-target lowering, and deletion during an in-flight archive transfer.


golem-worker-executor/src/services/oplog/multilayer.rs
```diff
@@ -267,6 +273,39 @@ impl MultiLayerOplogService {
...
+            .transfer_fibers
+            .lock()
+            .unwrap()
+            .remove(agent_id)
+            .and_then(|transfer_fiber| transfer_fiber.upgrade());
+
+        if let Some(transfer_fiber) = transfer_fiber
+            && let Some(transfer_fiber) = transfer_fiber.lock().unwrap().take()
+        {
+            transfer_fiber.abort();
```
vigoo, line 317: `abort` is just a request, not awaiting the fiber to stop so we still have a chance for the original race condition here. we should wait until it stops.

golem-worker-executor/src/services/oplog/multilayer.rs
```diff
@@ -267,6 +273,39 @@ impl MultiLayerOplogService {
         }
         Ok(ids)
     }
+
+    fn register_transfer(&self, agent_id: AgentId, transfer_fiber: &TransferFiber) {
```
vigoo: Possible other race condition, please check: 
> The transfer is registered while its `Option<JoinHandle>` is still `None`. Concurrent deletion can remove that registration, abort nothing, and then construction installs a live unregistered task.

kmatasfp: addressed

vigoo: I think the race is still there, it has just moved before registration:

> The transfer is spawned and its handle is stored before it is registered. Concurrent deletion can find no registration, delete the oplog layers, and then construction registers a live transfer that was not fenced.

We need to either synchronize the whole transfer construction/registration lifecycle with the deletion, or let the deletion mark an in-progress construction as cancelled somehow.

kmatasfp: this should now be addressed

vigoo (approved):

PR #3700, opened 2026-07-16
feat(cli): add `agent shell` to drive an agent as an interactive shell

Aditya1404Sal:
This PR adds a new subcommand:

```
golem agent shell '<AgentType("instance")>'
```

Each line you type is sent to the agent's `eval` method and the result (stdout/stderr/exit code) is printed back. If the agent pauses on a question, the CLI prompts you (a select menu when there are choices) and delivers the answer as a separate invocation; agents can't block on a human, since invocations are serialized per instance. A question left dangling by a dead session re-surfaces on the next connect.

Works with **any agent type** exposing this surface, validated against the reflected schema on connect:

| Method | Signature |
|---|---|
| `eval` | `(string) -> eval-result` |
| `answer_prompt` | `(string) -> eval-result` |
| `abort_prompt` | `() -> eval-result` |

`eval-result` = `{ stdout: string, stderr: string, exit-code: u8, pending-prompt: option<{ question: string, choices: option<list<string>> }> }`

**Commits:**
1. The feature : clap wiring, connect-time validation (method shapes; ephemeral types rejected because no session state, no answerable prompts), REPL loop, positional decode; 18 unit tests.
2. `test-components/agent-shell`, a 150-line reference implementer in plain golem-rust, + two worker-executor integration tests covering the reflected-schema contract and the full pause round-trip.

**Motivation:** Original [clank.sh](https://github.com/clank-sh/clank.sh) : a bash-compatible shell running as a durable Golem agent; this command is what makes a deployed shell instance feel like a shell. But the contract is generic: a SQL console, an ops console, or an approval flow fit the same three methods.
Current wip for clank.sh is [here](https://github.com/Aditya1404Sal/clank.sh)

**Known limitations (deliberate):**
- return shape validated at decode time, not connect
- Ctrl-C mid-invoke doesn't cancel the in-flight invocation 
- `--format` flag is ignored (like `golem repl`)
- method names matched verbatim (snake_case by default)


Aditya1404Sal:

PR #3703, opened 2026-07-17
Port scheduler improvements

kmatasfp:


golem-worker-executor/src/services/scheduler.rs
```diff
@@ -271,56 +276,123 @@ impl SchedulerServiceDefault {
...
+            self.process_claimed_actions(claimed, now).await;
 
             if claimed_count < self.claim_batch_size as usize {
                 break;
             }
         }
 
+        match self
+            .scheduler_storage
+            .count_due(Utc::now(), &assignment)
```
vigoo, line 300: Will these two COUNTs that are only used for metrics OK if usage grows?

golem-worker-executor/src/services/scheduler.rs
```diff
@@ -271,56 +276,123 @@ impl SchedulerServiceDefault {
...
+                crate::metrics::scheduler::set_scheduler_due_action_backlog_after_tick(backlog)
+            }
+            Err(error) => warn!(error, "Failed to count due scheduled actions after tick"),
+        }
+
         crate::metrics::scheduler::record_scheduler_tick_duration(tick_start.elapsed());
         Ok(())
     }
 
+    async fn process_claimed_actions(
```
vigoo, line 313: Please check the following identified scenario:

`process_and_ack_claimed_action` returns `false` on an acknowledgement error, but `process_claimed_actions` consumes that result only to stop the affected lane. It then collects all lanes and returns `()`. If the batch was full, `process` proceeds to claim another batch.

A concrete failure sequence:

1. A full batch contains same-agent actions A1 and A2.
2. A1 performs its side effect, but its acknowledgement returns an error.
3. The lane stops, leaving A2 leased but unprocessed.
4. Because the batch was full, the outer loop claims another batch.
5. A3 for the same agent can be claimed and processed while A2 remains unavailable behind its lease.

This is a regression from `main`, where `ack(...).await?` aborted the tick on an acknowledgement error. It also contradicts the new per-agent ordering intent.

golem-worker-executor/src/services/scheduler.rs
kmatasfp: Good question. The query is backed by `(shard_id, due_at_ms)` index and expected due backlog is only a few thousand actions, so exact counts should be cheap. A significantly larger backlog would already indicate scheduler/database issues. I removed the redundant pre-tick count and retained post-tick backlog for alerting.

So if this becomes a bottleneck we can remove it in the future or introduce some sampling, but I think it is useful to have, as I mentioned above I would set up alert based on the backlog size, if this grows it means we are having issues and should take some steps to mitigate

golem-worker-executor/src/services/scheduler.rs
kmatasfp: should be addressed now

vigoo (approved):

PR #3706, opened 2026-07-20
website: add Google Ads and Contentsquare tracking tags

mdegoes:
## Summary
- Adds the Google gtag.js snippet (AW-18172281472) to `BaseLayout.astro`
- Adds the Contentsquare tracking snippet (ed8fdf824ed4d) to `BaseLayout.astro`
- Both load site-wide since `BlogPostLayout` and `ContentPage` wrap `BaseLayout`
- Uses `is:inline` on all script tags so Astro doesn't bundle them as ES modules

## Test plan
- [x] `npm run build` succeeds
- [x] Verified both tags render unmangled in built HTML output


mdegoes:

vigoo (approved):

PR #3709, opened 2026-07-21
CLI cleanups

noise64:
- resolves GOL-330: command / args / ux
  - explicit `--auth oauth2|static` profile auth mode, masked static-token prompt, clearer help/docs
  - unified structured output for bulk agent ops (update / redeploy / delete): each agent with its component and its revision/version transition
  - misc arg/help/error-context cleanups (`parse_cursor` errors, `bail` instead of `unimplemented`)
  - fix `agent update` result polling to aggregate across all update records and report the most recent outcome, instead of deciding per-record (which could spuriously error, or report a stale success/failure, when the update history had multiple or non-target-revision records)
- resolves GOL-331: text / output rendering
  - structural agent-id formatting for terminal display, integrated with the table rendering
- resolves GOL-332: dead code & structural moves
  - unify the agent identifier as `agentId` in the output and rename `agent_name` -> `agent_id` throughout the cli internals (worker<->agent naming)
  - remove dead code (`OverwriteSafeAction`), streamline `guess_language`, make `WorkerCreateView` agent id non-optional
  - dissolve `model/text/`: command-output views now live next to their domain entity in `model/<domain>`; only the generic formatting toolkit remains, as `model/text_format.rs`
  - rename the minijinja `Template` trait/module to `TemplateRender` / `model/template_render.rs`; split the MCP deployment model into its own `model/mcp`
- resolves GOL-333: validation & infra hardening
  - validate component preset references, with proper errors during environment / runtime selection
  - validation guards: template cycles, duplicate initial-file targets, duplicate component matches
  - healthcheck cleanups, configurable agent stream ping interval, staging approval prompt context

### Deletion-output unification (output-schema change)

- every delete / revoke / unregister renders a consistent `Deleted <entity> <id>` view showing id-like fields only (id + cheap on-hand handles like name / version / domain / environment), and never secret / credential / config-body values (e.g. `secret delete` no longer emits the secret value)
- bulk ops (agent delete-all, card revoke) render a table + count summary
- renamed `*Result` table views to `*View` to match the `*View` = renders / `*Result` = structured-only convention

### CI / it-cli

- the stricter preset validation surfaced two latent config bugs, now fixed:
  - dropped the global `GOLEM_PRESET=quick` from CI — `quick` is a TS/Scala wizer-skip and undefined for rust/moonbit apps, so applying it as a global default now hard-errors
  - aligned stale test manifests: TS-only test apps carried rust preset names (`debug`/`release`); a TS app generated today sets no `componentPresets`, so those were dropped
- added `buildMergeMode: replace` to 15 guest-bridge planner components: they synthesize their component wasm with `cp` / `cargo metadata` proxy build steps, but without `replace` the rust template's `cargo build` was appended and ran incidentally — cuts it-cli test-work ~56% (146 -> 65 min) with no change to what the tests cover
- split the `it-cli` job into 3 parallel shards by module-level test-r tag (`:tag:agents` / `:tag:bridge_gen` / untagged `:tag:` catch-all): it was the workflow's sole tail (~24m), each shard now runs ~9-11m so it is off the critical path, and the shards' test counts sum back to the full set (nothing dropped)
- moved the `agent_guest.wasm` build + Scala SDK `publishLocal` into `build-golem-scala` (which already does that work and finishes before it-cli starts) and cache/restore them, trimming ~2.5 min of per-shard setup

> NOTE: the split renames the required status check from `it-cli` to the three matrix checks (`it-cli (agents)`, `it-cli (bridge_gen)`, `it-cli (core)`) — branch-protection required checks need updating.

TODO: follow-ups tracked separately — GOL-411 (redeploy fail-fast vs best-effort), GOL-412 (first-class on-demand templates), GOL-413 (non-manifest server flags / help->STDERR).


cli/golem-cli/src/command_handler/agent/mod.rs
```diff
@@ -1840,61 +1804,65 @@ impl WorkerCommandHandler {
...
+                    .component_version_at(component_id, agent.component_revision)
+                    .await,
+                version: version.clone(),
+            });
         }
 
         if await_update {
-            for worker in workers_to_update {
+            for agent in agents_to_update {
                 let _ = self
```
vigoo: This is not new code but we are ignoring errors here

cli/golem-cli/src/command_handler/agent/mod.rs
```diff
@@ -1992,92 +1963,88 @@ impl WorkerCommandHandler {
...
-                        "is not pending anymore, but no outcome has been found",
-                    );
-                    return Err(anyhow!(
-                        "Unexpected agent state: update is not pending anymore, but no outcome has been found"
-                    ));
-                }
+            if pending_count > 0 {
+                log_action("Agent update", "is still pending");
+                tokio::time::sleep(Duration::from_secs(2)).await;
+            } else if let Some(success) = latest_success {
```
vigoo: This will work most of the time in practice but in theory there can be multiple updates to the same version and could be that we updated to X first successfully, then to something else, then again back to X and it failed. So picking the latest result should happen from the union of the successes and failures, simply using the one with the latest timeout

cli/golem-cli/src/command_handler/profile/mod.rs
```diff
@@ -142,6 +140,35 @@ impl ProfileCommandHandler {
...
+        static_token: Option<String>,
+    ) -> anyhow::Result<AuthenticationConfig> {
+        match (mode, static_token) {
+            (Some(ProfileAuthMode::OAuth2), Some(_)) => {
+                log_error(
+                    "--static-token cannot be combined with --auth oauth2. A static token implies --auth static.",
+                );
+                bail!(NonSuccessfulExit);
+            }
+            (Some(ProfileAuthMode::Static) | None, Some(token)) => {
```
vigoo, line 165: We should check for empty-string `token` - for example a `--static-token ${GOLEM_API_TOKEN}` in a script may resolve it to `--static-token ""` accidentally if the env var is missing.

cli/golem-cli/src/model/cascade/store.rs
```diff
@@ -62,33 +61,44 @@ impl<L: Layer> Store<L> {
...
-            layer: &L,
+            layer: &'a L,
             value: &mut L::Value,
+            path: &mut Vec<&'a L::Id>,
         ) -> Result<(), StoreGetValueError<L>> {
-            for layer_id in layer.parent_layers() {
-                let Some(layer) = store.layers.get(layer_id) else {
-                    return Err(StoreGetValueError::LayerNotFound(layer_id.clone()));
+            let layer_id = layer.id();
+            if path.contains(&layer_id) {
```
vigoo, line 73: Did you want to do the cycle check only resolve-time? This way we don't detect cycles in ununsed templates etc. (I'm fine with keeping it like this if that was the intention)

noise64: I think it is fine to be lenient here, and not stop on unused parts, so for now I would keep it like this

PR #3714, opened 2026-07-22
Follow-up fixes for the ephemeral hot-path optimisation

vigoo:
Resolves GOL-341


golem-worker-service/src/service/worker/service.rs
```diff
@@ -133,6 +133,16 @@ fn normalize_agent_invocation_identity(
...
+        // A phantom generated for this invocation must be the one derived from
+        // the invocation's idempotency key; anything else indicates a caller
+        // bug and is rejected instead of being silently replaced.
+        if parsed_agent_id.phantom_id != Some(ephemeral_invocation_phantom_id(&idempotency_key)) {
+            crate::metrics::record_ephemeral_explicit_phantom_invocation_rejection();
+            return Err(WorkerServiceError::TypeChecker(
+                "The ephemeral invocation phantom ID does not match the identity derived from the invocation's idempotency key"
+                    .to_string(),
+            ));
+        }
```
noise64: Please check if this is true:

> The strict equality is right, but the block it lives in rests on an invariant that's neither named nor observable.
>
> **The guard flag doesn't say what it means.** `phantom_was_generated_for_invocation` reads as "we generated this phantom", but the only non-`false` call site passes `trusted_internal_caller && !is_lookup` (grpcapi/worker.rs:541) — i.e. "caller is System/Agent". It's safe today only because the one trusted caller derives the phantom from the same key it sends (`invocation_target_agent_id`, wasm_rpc/mod.rs:1875) — an implicit invariant spanning four files. And `idempotency_key` is defaulted to `fresh()` at :106, so a future caller forwarding a phantom without its key hard-fails here.
>
> **And when that happens, the metric can't show it.** :130 and :140 share one unlabeled `IntCounter`, but they're different signals: :130 is routine client misuse (expected, ignorable), :140 means a trusted internal caller sent an inconsistent phantom/key pair — exactly the bug above, an the one you'd want to alert on. It can't be, without :130 drowning it. The name and help text also only describe :130 now ("rejected for selecting a phantom ID"), but the new branch isn't an *explicit* phantom.
>
> Three concrete changes:
>
> 1. Rename the parameter to `caller_is_trusted_internal`, doc-commented with the contract it relies on: *a trusted caller that supplies a phantom on an ephemeral agent id must supply the idempotency key it was derived from.*
> 2. Drop `&& !is_lookup` at the call site — `normalize` already early-returns on `observation_only` at :120 before this flag is read, and `observation_only` is `mode == Lookup` ([service.rs:1017](<http://service.rs:1017>)), so it's redundant.
> 3. Give the counter a `reason` label (`explicit-phantom` / `derived-phantom-mismatch`) and a generalised help string. The executor already has this pattern: `record_non_suspending_failure(reason)` over a `CounterVec` (golem-worker-executor/src/metrics.rs:966

golem-worker-service/src/custom_api/call_agent/mod.rs
```diff
@@ -59,7 +61,22 @@ impl CallAgentHandler {
...
-        let agent_id = self.build_agent_id(resolved_route, behaviour)?;
+        // Phantom routes address a fresh agent instance on every call. For
+        // durable agents this requires generating a random phantom id here;
+        // for ephemeral agents the final per-invocation phantom identity is
+        // derived downstream from the invocation's idempotency key, so the
+        // target is addressed by its logical agent id.
+        let phantom_id = if behaviour.phantom {
+            let agent_mode = self
+                .worker_service
+                .agent_mode(behaviour.component_id, &behaviour.agent_type)
```
noise64: Can the `CallAgentBehaviour` contain the agent_mode?

> (`route_compilation.rs:205` already reads `agent.mode` to compute `phantom`, so the compiler has it. Would also drop the duplicate `get_current_by_id` — `invoke_agent` fetches the same component again at [service.rs:955](<http://service.rs:955>) — and remove the impossible `None` arm.)

golem-worker-executor/src/services/oplog/primary.rs
```diff
@@ -312,10 +312,38 @@ impl OplogService for PrimaryOplogService {
...
         execution_status: read_only_lock::std::ReadOnlyLock<ExecutionStatus>,
     ) -> Arc<dyn Oplog> {
-        self.create(
+        record_oplog_call("create_fresh");
+
+        let key = Self::oplog_key(&owned_agent_id.agent_id);
+
+        // The caller guarantees the agent id is freshly derived and unused, so
+        // the existence probe performed by `create` is skipped: the initial
+        // entry is appended directly without any prior read.
```
noise64, line 321: > `PrimaryOplogService::create_fresh` has only two possible callers, and both look closed in any configuration we ship — so this body appears to be dead code today:
>
> 1. **Directly**, when the wired `OplogService` is a bare `PrimaryOplogService`. [lib.rs:784](<http://lib.rs:784>) selects that only when no archive layers are configured (`indexed_storage_layers <= 1 && blob_storage_layers == 0`); defaults are 2/1 (golem_config.rs:1658), both shipped configs set 2/1, and
>
> local-run / integration-tests / the test framework don't override them.
>
> 2. **Via** `MultiLayerOplogService`, which delegates to the primary only from the `AgentMode::Durable` arm (the new `if self.fresh` at [multilayer.rs:443](<http://multilayer.rs:443>)) — the `Ephemeral` arm appends to `lower.first()` ([multilayer.rs:501](<http://multilayer.rs:501>)) and never touches the primary. But durable + `KnownFresh` is
>
> rejected by `validate_resolved_invocation_identity` (worker/mod.rs:4702), and validation runs ahead of creation on all three paths (grpc/mod.rs:2010, [rpc.rs:923/1023](<http://rpc.rs:923/1023>)).
>
> The two new tests don't reach it either: both build a multilayer service *with* lower layers and pass `AgentMode::Ephemeral`, landing in the Ephemeral arm.

sdks/ts/packages/golem-ts-sdk/src/fluent/client.ts
```diff
@@ -282,7 +282,8 @@ export function clientFor<
     config?: Record<string, unknown>,
   ): RemoteClient<Methods, Mode> => {
     const constructorTree = encodeRecord(idCodecs, id as Record<string, unknown>);
-    const agentId = makeAgentId(def.name, constructorTree, phantomId);
+    const agentId =
+      def.mode === 'ephemeral' ? def.name : makeAgentId(def.name, constructorTree, phantomId);
```
noise64, line 302: Given this is for error reporting only, could we also drop makeAgentId for durable too, and construct the error from name and constructor tree (and if yes, do it lazily)?

golem-worker-executor/src/services/oplog/primary.rs
vigoo: Agreed to add more tests to cover this - but it's not dead code, could be used by setting up (or probably even by configuration) the oplog topology differently

sdks/ts/packages/golem-ts-sdk/src/fluent/client.ts
vigoo: I ended up not doing this - did not want 'fake' agent-ids (type name + json, for example), and did not want to deal with weird edge cases of potentially mutated stored constructor args.

golem-api-grpc/proto/golem/customapi/core.proto
```diff
@@ -221,6 +221,7 @@ message RouteBehaviour {
     optional golem.component.ReadOnlyConfig read_only = 10;
     CompiledInputSchema constructor_input = 11;
     CompiledInputSchema method_input = 12;
+    optional golem.component.AgentMode agent_mode = 13;
```
noise64: i think this should not be optional

noise64 (approved):
added 1 minor comment on the agent mode optionality, otherwise looks good

PR #3715, opened 2026-07-23
Fix promise completion wakeup races

kmatasfp:
## Summary
- synchronously update local promise handles after completion, preserving first completion payload
- reload and broadcast the stored winning payload when a second executor completes an already-completed promise
- add deterministic missed-wakeup coverage, multi-thread completion contention coverage, and duplicate-payload service coverage
- reduce successful and expected API request log volume to debug


vigoo:
There seem to be a (pre-existing) issue with the sqlite version of `set_if_not_exists`:

The SQLite implementation of `set_if_not_exists` performs:

1. `SELECT` to determine whether the key exists.
2. `INSERT OR IGNORE`.
3. Returns the result of the earlier `SELECT`, ignoring the insert’s affected-row count.

With two concurrent callers:

```
Caller A                    Caller B
────────                    ────────
SELECT → absent             SELECT → absent
INSERT payload A → wins
                            INSERT payload B → ignored
return true                 return true  ← incorrect
```

Under this PR, both callers then treat their own payload as the winner. Whichever reaches the local handle first can become its value, even if the other payload is the one stored durably.

Redis, Postgres, and in-memory storage implement this atomically and are safe. The SQLite bug predates this PR, so the PR does not regress it, but it leaves the stated invariant incomplete. The SQLite method should return `execute(...).rows_affected() == 1`; a concurrent backend-level test should accompany that fix.

vigoo (approved):
Otherwise approved

PR #3716, opened 2026-07-24
TS guest rpc/tool bridge

vigoo:
Resolves GOL-13

Note: Created GOL-343 as a follow-up, as multiple bridge generators are embedding schemas as json - for now followed that practice in this one too, but want to change later.


cli/golem-cli/src/app/build/gen_bridge.rs
```diff
@@ -527,11 +528,16 @@ async fn collect_tool_manifest_targets_for_entry(
     if bridge_mode != BridgeMode::Guest
         || !matches!(
             target_language,
-            GuestLanguage::Rust | GuestLanguage::Scala | GuestLanguage::MoonBit
+            GuestLanguage::Rust
```
noise64: Don't we have a method on the enum for this? Seems like we keep repeating this match.

noise64 (approved):

PR #3728, opened 2026-08-07
Unify filesystem storage accounting

kmatasfp:
Implements GOL-416.

## Summary

- replace logical filesystem byte deltas, storage permits, and custom hard-link accounting with one `AgentFilesystems` / `AgentFilesystem` lifecycle
- use managed XFS project quotas as the authority for per-agent allocated-byte usage, filesystem-object usage, and limits, while using fresh `statvfs` observations for physical node capacity
- route WASI P2 and P3 direct operations and asynchronous streams through shared lifecycle admission and failure classification
- move filesystem creation, initial-file materialization, replay settlement, and cleanup under the `AgentFilesystem` lifecycle
- meter authoritative allocated-byte observations over the same `ConcurrentAgentPermit` window as linear memory, with compute, memory, and filesystem metering configured independently

## Behavior

`AgentFilesystem` now owns each resident agent's ephemeral filesystem across creation, initial-file materialization, oplog replay, execution, and unload. Startup removes and verifies stale local resources before creating an unmanaged directory or managed XFS project. Replay-owned filesystem effects settle before startup publishes the runtime. Unload seals new filesystem work, tears down runtime resources, deletes the filesystem, and verifies cleanup before publishing `Unloaded`.

Managed XFS validates project-quota support, project identity and inheritance, reflink support, permissions, and exclusive ownership of the managed root. Project observations supply authoritative allocated bytes and filesystem-object counts. The executor installs the public allocated-byte limit together with an internally derived object limit. Read-only and read-write initial files both count toward quota and billing. Initial-file revisions are staged, validated, and installed atomically so a failed update does not publish partial files. Reducing either limit below current usage suspends and unloads the agent until reconstruction fits the current policy.

Unmanaged filesystems remain available when project quotas are not configured, but reject finite filesystem limits because they cannot provide authoritative enforcement. Unsupported allocation observation disables only the filesystem metering dimension; compute and memory metering continue independently.

Physical capacity remains separate from project usage. The executor obtains available bytes and filesystem objects from fresh `statvfs` observations. On physical exhaustion it may unload only eligible loaded-idle agents, and recognizes reclaimed capacity only after deletion and another observation. Agent quota exhaustion never triggers pressure eviction.

P2 and P3 preserve errno, completed prefixes, and effect evidence until classification. Retries are bounded and apply only to a proven unwritten suffix or an operation whose postcondition proves no effect. Unknown effects and terminal infrastructure failures invalidate the disposable local runtime while leaving durable work available for reconstruction. XFS reports `ENOSPC` for both project-quota and physical exhaustion, so classification uses fresh quota, capacity, and backend-health observations instead of errno alone.

Enabled storage billing integrates ordered absolute `allocated_bytes` observations prospectively. Storage and memory start and stop together at one monotonic timestamp while the worker owns its concurrent-agent permit. Active execution, host calls, host I/O waits, and replay are billable while retaining the permit. Loaded-idle, unloaded, and durably sleeping workers accrue nothing. Unload and filesystem deletion happen outside the billing window.

Compute, memory, and filesystem usage metering are independent switches and all default to disabled for OSS and self-hosted deployments. The worker executor and standalone `golem` server honor `GOLEM__RESOURCE_USAGE_METERING__COMPUTE`, `GOLEM__RESOURCE_USAGE_METERING__MEMORY`, and `GOLEM__RESOURCE_USAGE_METERING__FILESYSTEM`. These switches affect measurement and export, not quota enforcement, memory admission, or physical-pressure recovery.

The superseded logical accounting path has been removed across oplog types, protobuf, WIT, CLI schemas, SDK bindings, documentation, and the built-in OTLP exporter. `ConcurrentAgentPermit` remains because it defines the execution and billing window; it is unrelated to the removed filesystem byte permits.


vigoo:
You need to rebuild the built-in plugins (`cargo make build-plugins`) because of the WIT change

vigoo:
Both filesystem APIs (p2 and p3) expose hard-link creation without charging another copy (`link_at`), which is correct because both names refer to one inode, but `unlink` releases the file's entire size after any successful unlink without tracking reference counts.

vigoo:
One more review finding to check:

Read-write provisioning writes directly to the destination:

- Cache hits use `tokio::fs::copy`
- Cache misses `File::create` the final path and stream into it

Either operation can fail after leaving a partial regular file. On a `revision-update` retry, `update_filesystem` accepts any existing regular file as `NoChanges`, without validating its expected size or content.

The resulting sequence is:

1. A revision update reserves N bytes for a new read-write file.
2. The download writes K < N bytes and fails.
3. The reservation is rolled back because no successful growth was reported.
4. The partial K-byte destination remains.
5. Retry sees a regular file and reports NoChanges.
6. The update can succeed with those K bytes absent from quota, capacity, billing, and recovery accounting.

Required fix: write/copy into a sibling temporary file, delete it on failure, and atomically rename it only after complete success. Do not accept an arbitrary pre-existing regular destination without validating that it represents a completed provisioning operation.

kmatasfp:
All of the above you mentioned @vigoo  plus some other things I ran into myself reviewing this are the reason I changed the direction and abandoned the golem internal quota/disk usage tracking/management and now building out a proper FS support I mentioned during the standup. GOL-416 issue description contains details of the new approach

golem-worker-executor/src/sandbox_filesystem/xfs.rs
```diff
@@ -0,0 +1,1871 @@
...
+}
+
+impl ManagedProvisioning {
+    pub(super) fn new(
+        root: &Path,
+        cleanup_retry: &RetryConfig,
+    ) -> Result<Self, FilesystemStorageError> {
+        let root_fd = File::open(root)
+            .map_err(|error| FilesystemStorageError::io("open managed XFS root", root, error))?;
+        flock(&root_fd, FlockOperation::NonBlockingLockExclusive).map_err(|error| {
```
vigoo: This is probably not an issue in practice but sharing to be sure:

> XFS project IDs and quota records are filesystem-wide, but this lock protects only the configured root inode, while the allocator below is process-local and starts at ID 1. Two executors using different managed roots on the same XFS filesystem can therefore reserve the same project ID—even sequentially while the first project has zero usage—and then share or overwrite a quota record. Could we either verify that the managed root exclusively owns the whole mount, or use a filesystem-wide inter-process project-ID lease/registry held for the project’s lifetime?

golem-worker-executor/src/filesystem_pressure.rs
```diff
@@ -0,0 +1,1182 @@
...
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) struct FilesystemWritePressurePolicy {
+    minimum_available_bytes: u64,
+    target: FilesystemCapacityTarget,
+}
+
+impl FilesystemWritePressurePolicy {
+    pub(crate) fn from_config(config: &FilesystemPressureConfig) -> Self {
+        Self {
```
vigoo, line 145: Should we also deal with minimum/target `available_filesystem_objects` in this?

golem-worker-executor/src/sandbox_filesystem/mod.rs
```diff
@@ -0,0 +1,1839 @@
...
+    }
+
+    pub(crate) async fn delete_and_verify(&self) -> Result<(), FilesystemStorageError> {
+        self.root.close();
+        let Some(mut state) = self
+            .lease
+            .state
+            .lock()
+            .expect("sandbox filesystem lease lock poisoned")
+            .take()
```
vigoo: This one seems valid:

> Taking `LeaseState` out before the following async deletion makes this future cancellation-unsafe. If the task is aborted during `state.cleanup.delete().await`, the local state is dropped; when `SandboxFilesystem` subsequently drops, its fallback cleanup sees `lease.state == None` and does nothing. That can leave the directory, project assignment, limits, or allocator reservation behind. Could the state remain owned by an armed RAII guard that performs blocking cleanup or restores it when the future is cancelled?

golem-worker-executor/src/wasi_filesystem/p2/types.rs
```diff
@@ -0,0 +1,1492 @@
...
+            ))),
+            ErrorCode::CrossDevice
+        );
+        assert_eq!(
+            error_code::<()>(Err(p2_link_access_error(
+                agent_filesystem::AccessError::NotPermitted
+            ))),
+            ErrorCode::NotPermitted
+        );
+        #[cfg(unix)]
```
vigoo: This block is enabled on macOS, but `golem-worker-executor` declares `libc` only under `cfg(target_os = "linux")`, so library tests fail here with unresolved `libc::EXDEV`. The analogous P3 test has the same problem. Could these assertions be Linux-gated, use a portable error construction, or make `libc` available as a Unix test dependency?

golem-worker-executor/src/sandbox_filesystem/xfs.rs
kmatasfp: yeah In practice it should not be an issue - virtually all production grade CSI drivers provision a block device per PV and do fresh mkfs, meaning each executor has their own xfs filesystem, no sharing.  Let me add a check on startup to make sure noone is using some custom local-path-provisioner which creates sub-directories on a single pool XFS or something - which is not a recommended deployment pattern anyway for HA reasons, losing a node here would mean you would lose multiple executors at once - also this would break the solutions anyway because in this case there would be nested quotas which would be problematic on its own for xfs

golem-worker-executor/src/filesystem_pressure.rs
kmatasfp: Implemented in c2cbf9567. Write pressure now activates when either available bytes or filesystem objects reach their minimum. Reclamation requires both target values, and startup validates both targets against total managed capacity.

golem-worker-executor/src/sandbox_filesystem/xfs.rs
kmatasfp: Implemented in c2cbf9567. Startup now verifies that the managed root is both the XFS filesystem root inode and a mount root before acquiring the exclusive lock. Subdirectories and bind-mounted subdirectories are rejected, so the lock and project-ID allocator cover the whole XFS filesystem.

golem-worker-executor/src/sandbox_filesystem/mod.rs
kmatasfp: Implemented in c2cbf9567. An armed lease-state guard now owns cleanup state across the async deletion. Errors or cancellation restore the state for retry or drop fallback; successful verified deletion disarms it. A regression test covers cancellation while the state is armed.

golem-worker-executor/src/wasi_filesystem/p2/types.rs
kmatasfp: Implemented in c2cbf9567. The native EXDEV assertions in both P2 and P3 tests are now gated to Linux, matching the target-specific libc dependency. The portable cross-project mapping assertions still run on every platform.

vigoo (approved):
Good to go once the conflict is resolved. Make sure you use forked wit-bindgen for regenreating the moonbit bindings (there is a proper regen script for it somewhere)

PR #3729, opened 2026-08-09
Upgrade Ignite RDBMS backend to ignite-v2-client 1.0.1

JohnSColeman:
## Summary

Upgrades `golem-worker-executor`'s Ignite RDBMS backend from `ignite-v2-client` 0.2.0 to **1.0.1**. All changes are confined to the service layer — the `golem:rdbms/ignite2@1.5.0` WIT, its SDK bindings, the `DbValue` types, and the oplog format are **unchanged**.

## Changes

- **Config:** `IgniteClientConfig` is now built via the builder API (1.0.0 added `addresses`, `partition_awareness`, `endpoint_discovery`, `data_center_id`), so the old struct literal no longer compiles.
- **Errors:** handle the new `IgniteError::PoisonedLock` and `IgniteError::UnsupportedOnTransaction` variants.
- **Values:** `IgniteValue` is now `#[non_exhaustive]` and adds binary-object value kinds (`Object`, `IntArray`, `StringArray`, `Collection`, `Map`, `Enum`). `ignite_to_service` becomes fallible; the fallibility threads through `convert_query_result` and the result-stream loop.

All of the above are the 0.2.0 → 1.0.0 adaptation. The subsequent 1.0.1 bump needed **no source changes** (see below).

## Note on 1.0.1

1.0.1 is a **packaging-only** upstream release: the `src/` trees of both `ignite-v2-client` and `ignite-client-derive` are byte-identical to 1.0.0. The manifest changes are:

- Exact pins (`=1.50.0`, `=0.7.18`, …) relaxed to caret ranges across all 18 dependencies — the meaningful fix, since the `=` pins previously forced exact transitive versions on every consumer.
- `tokio` features narrowed from `"full"` to `["rt", "net", "sync", "io-util", "time"]` — no effect here, as golem enables tokio `full` elsewhere and Cargo unifies features.

## Design note — why the binary-object kinds are not in `db-value`

The binary-object value kinds are **deliberately not** represented in the SQL `db-value` contract. Apache Ignite SQL columns are restricted to scalar types (`org.apache.ignite` `CommonUtils.isSqlType`: boxed primitives, `BigDecimal`, `String`, `UUID`, temporal types, `byte[]`, Geometry); a complex cache field is projected as its scalar leaf columns via QueryEntity property paths, and the binary-object kinds belong to Ignite's **key-value binary-object API**, not the SQL layer. So a binary-object value cannot appear as a column value for a well-formed schema; if one ever does (e.g. an explicit `SELECT _VAL`), `ignite_to_service` returns a clear `QueryResponseFailure` rather than guessing a scalar mapping or losing data.

## Testing

Re-run in full against 1.0.1:

- `cargo check -p golem-worker-executor` ✅
- Unit tests (2): scalar round-trip; binary-object kinds rejected ✅
- Integration tests (6, against a live Apache Ignite 2.17 container via testcontainers): connection, connection-error, query-stream, create/insert/select, transaction commit, transaction rollback ✅

> Note for local runs on colima: testcontainers probes `/var/run/docker.sock` directly and does not read docker contexts, so the integration tests need `DOCKER_HOST` pointed at the colima socket (e.g. `unix://$HOME/.colima/default/docker.sock`).


kmatasfp:
We ended up forking your repo here https://github.com/golemcloud/ignite-v2-client, having dependency on fixed version of tokio like you have, `tokio = { version = "=1.50.0", features = ["full"] }` is causing issues for us

JohnSColeman:
> We ended up forking your repo here https://github.com/golemcloud/ignite-v2-client, having dependency on fixed version of tokio like you have, `tokio = { version = "=1.50.0", features = ["full"] }` is causing issues for us

@kmatasfp do I need to take anymore action? I will bump to a more compatible version 1.0.1.

kmatasfp (approved):

PR #3737, opened 2026-08-12
Fix P3 response bodies created without persistence

vigoo:
## Summary

A P3 HTTP send begun under `PersistNothing` must pass that initiation-time policy to the response body. Without it, `response::consume-body` starts a long-lived durable host call after the ambient persistence level has returned to `Smart`; restoring `Smart` between golem-ai stream polls then traps with:

> Cannot change oplog persistence level: durable host calls are still in flight

This change records the send's initiation-time persistence policy on the response. Durable sends retain durable response-body behavior, while `PersistNothing` sends use pass-through body handling. This preserves golem-ai's logical stream-event replay and restart semantics.

Motivating integration case: golemcloud/golem-ai#128.

## Verification

Previously verified in the source orb (not rerun during publication):

- Focused regression `p3_stream_created_in_persist_nothing_stays_non_durable` passed.
- Existing `oplog_replay_after_streaming_http_read` passed.
- `cargo check -p golem-worker-executor --lib` passed.
- golem-ai LLM/Ollama checks passed.
- Formatting and diff checks passed.

### DIFF
--- a/golem-worker-executor/src/durable_host/concurrent/call.rs
+++ b/golem-worker-executor/src/durable_host/concurrent/call.rs
@@ -1332,6 +1332,17 @@ impl<Pair: HostPayloadPair, P: DropPolicy> CallHandle<Pair, P> {
         self.start_idx
     }
 
+    /// Whether resources produced by this call must retain their own durable lifecycle after the
+    /// call returns. This uses the call's initiation-time state rather than ambient worker state,
+    /// which may change while a returned resource remains live.
+    pub fn persistence_enabled(&self) -> bool {
+        self.retry
+            .durable_execution_state()
+            .snapshotting_mode
+            .is_none()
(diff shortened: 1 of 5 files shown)


vigoo:
Closing because this is not the right durability model.

`PersistNothing` operations are still recorded for observability; replay skips them. Making an escaped P3 response body fully pass-through avoids the in-flight-call trap, but incorrectly removes the response-body parent/chunk/terminal records from the oplog. Conversely, recording ordinary durable body entries after the skipped region would leave replay-visible continuations for a physical HTTP send that replay never reconstructs.

The required design must separate recording from replay participation and preserve an observational-only disposition across the complete long-lived operation, including continuations that outlive the lexical `PersistNothing` region.

The findings, affected async/resource paths, design alternatives, and acceptance criteria are tracked in [GOL-428](https://linear.app/golem-cloud/issue/GOL-428/preserve-observability-for-long-lived-operations-initiated-in).

PR #3739, opened 2026-08-13
Tool definition host functions

vigoo:
Resolves GOL-30

### DIFF
--- a/golem-worker-executor/src/durable_host/tool/mod.rs
+++ b/golem-worker-executor/src/durable_host/tool/mod.rs
@@ -14,10 +14,9 @@
 
 //! Host implementation of `golem:tool/host@0.1.0`.
 //!
-//! The interface is wired into the linker for every component shape so that any
-//! agent or tool component may import it and instantiate. Tool discovery is
-//! implemented as a durable environment-state read; invocation operations
-//! remain placeholders until the tool runtime is implemented.
+//! The interface is wired into the linker for every agent component. Tool
+//! discovery is implemented as a durable environment-state read; tool
+//! invocation operations are currently unsupported.
 
 use crate::durable_host::concurrent::{CallHandle, CallReplayOutcome, NotCancellable};
 use crate::durable_host::durability::{ClassifiedHostError, HostFailureKind};
@@ -38,17 +37,17 @@ use golem_common::model::oplog::{
 use golem_common::model::tool::ToolName;
 use golem_common::schema::tool::DiscoveredTool;
 use golem_common::schema::tool::wit::wire::Tool as WitTool;
+use std::sync::Arc;
 use wasmtime::component::{Accessor, HasSelf, Resource, StreamReader};
 
-const NOT_IMPLEMENTED: &str = "golem:tool/host is not yet implemented";
+const NOT_IMPLEMENTED: &str = "golem:tool/host tool invocation is not yet implemented";
 
 /// Host-side resource table entry backing the `golem:tool/host.tool-rpc`
-/// resource. A placeholder until the tool runtime is implemented.
+/// resource.
 pub struct ToolRpcEntry;
 
 /// Host-side resource table entry backing the
-/// `golem:tool/host.future-invoke-result` resource. A placeholder until the tool
-/// runtime is implemented.
+/// `golem:tool/host.future-invoke-result` resource.
 pub struct FutureInvokeResultEntry;
 
 impl TryFrom<&DiscoveredTool> for WitRegisteredTool {
@@ -73,7 +72,8 @@ impl TryFrom<&DiscoveredTool> for WitRegisteredTool {
 fn classify_tool_discovery_error(error: &ToolDiscoveryError) -> HostFailureKind {
     match error {
         ToolDiscoveryError::Retrieval(_) => HostFailureKind::Transient,
-        ToolDiscoveryError::InconsistentSnapshot { .. } => HostFailureKind::Permanent,
+        ToolDiscoveryError::AgentContextRequired
+        | ToolDiscoveryError::InconsistentSnapshot { .. } => HostFailureKind::Permanent,
     }
 }
 
@@ -85,9 +85,11 @@ fn terminal_tool_discovery_error(message: String) -> anyhow::Error {
 }
 
 impl<Ctx: WorkerCtx> DurableWorkerCtx<Ctx> {
-    pub(crate) async fn get_all_tools_model(&mut self) -> anyhow::Result<Vec<DiscoveredTool>> {
+    pub(crate) async fn get_all_tools_model(&mut self) -> anyhow::Result<Vec<Arc<DiscoveredTool>>> {
         let agent_type = self.parsed_agent_id().map(|agent_id| agent_id.agent_type);
         let environment_id = self.state.owned_agent_id.environment_id;
+        let component_id = self.state.owned_agent_id.agent_id.component_id;
+        let component_revision = self.state.component_metadata.revision;
 
         let mut handle = CallHandle::<GolemToolGetAllTools, NotCancellable>::start(
             self,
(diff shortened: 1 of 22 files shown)


golem-worker-executor/src/durable_host/tool/mod.rs
```diff
@@ -40,15 +51,189 @@ pub struct ToolRpcEntry;
...
+                    match handle
+                        .try_trigger_retry_or_loop(self, &result, classify_tool_discovery_error)
+                        .await?
+                    {
+                        InternalRetryResult::Persist => break result,
+                        InternalRetryResult::RetryInternally => continue,
+                    }
+                }
+            } else {
+                Ok(Vec::new())
```
kmatasfp: If `parsed_agent_id()` up at line 89 comes back `None`, we land here and return an empty list. That empty list doesn't just go back to the caller — it goes through `handle.complete(...)` and gets written to the oplog, so every replay from then on sees "no tools" as well, and there's no way to tell it apart from an environment that genuinely has none.

GOL-30 says to filter by the calling agent's accessible set. If we can't work out who the caller is, we don't have an accessible set — that's a bug on our side rather than an empty answer, so I think this should return an error. `classify_tool_discovery_error` and `terminal_tool_discovery_error` are already set up in this file

Same thing happens at line 197.

golem-worker-executor/src/durable_host/tool/mod.rs
```diff
@@ -40,15 +51,189 @@ pub struct ToolRpcEntry;
...
+                    match handle
+                        .try_trigger_retry_or_loop(self, &result, classify_tool_discovery_error)
+                        .await?
+                    {
+                        InternalRetryResult::Persist => break result,
+                        InternalRetryResult::RetryInternally => continue,
+                    }
+                }
+            } else {
+                Ok(None)
```
kmatasfp: Same as line 129 — when `agent_type` is `None` we save an empty answer.

If `valid_tool_name` is `None` because someone passed a name that isn't a valid tool name, returning `None` looks right to me — GOL-30 wants `get-tool` to answer unknown names with none. It's only the missing agent id that I think should be an error.

golem-worker-executor/src/durable_host/tool/mod.rs
```diff
@@ -40,15 +51,189 @@ pub struct ToolRpcEntry;
...
+            }
+
+            let result = if let (Some(agent_type), Some(valid_tool_name)) =
+                (&agent_type, &valid_tool_name)
+            {
+                loop {
+                    let result = self
+                        .state
+                        .environment_state_service
+                        .get_accessible_tool(environment_id, agent_type, valid_tool_name)
```
kmatasfp: Each of these two functions goes and fetches the environment state on its own, so a `get-all-tools` and a `get-tool` in the same invocation can end up looking at different deployments. An agent could list a tool and then fail to find it a moment later.

GOL-30 asks to use the same environment-state snapshot that GOL-29 produces, which reads like it wants one snapshot held for the whole invocation. But the live test here deliberately checks that `get-all-tools` picks up a new deployment part way through, so I'm guessing seeing changes as they happen is what you actually wanted. Which of the two is it?

golem-worker-executor/src/services/environment_state.rs
```diff
@@ -20,16 +20,108 @@ use golem_common::model::agent_secret::{
...
+    };
+
+    bindings
+        .keys()
+        .map(|tool_name| {
+            deployment
+                .registered_tools
+                .get(tool_name)
+                .cloned()
+                .ok_or_else(|| ToolDiscoveryError::InconsistentSnapshot {
```
kmatasfp: This lookup, and the one at line 114, match a binding to a tool on the name alone. Neither the binding's `deployment_revision` nor the tool's gets checked against the deployment's.

Worth flagging that the test at line 483 doesn't cover this, even though it reads like it might — `deployment_state()` builds every tool at revision 1, so asserting they all come back at revision 1 passes regardless of what this function does with revisions. It'd pass just as happily if the join ignored them entirely, which is what it does.

In practice it's fine as long as GOL-29 always hands over a snapshot from a single revision, which I'd expect it to. So this is a question rather than a complaint: is leaving the revision check entirely to the code that builds the snapshot the split you meant? Happy for the answer to be yes — just want it to be a decision rather than an accident.

golem-worker-executor/src/services/environment_state.rs
```diff
@@ -20,16 +20,108 @@ use golem_common::model::agent_secret::{
...
+    if !bindings.contains_key(tool_name) {
+        return Ok(None);
+    }
+
+    deployment
+        .registered_tools
+        .get(tool_name)
+        .cloned()
+        .map(Some)
+        .ok_or_else(|| ToolDiscoveryError::InconsistentSnapshot {
```
kmatasfp: Small thing: this error is word for word the same as the one at line 87. Creating a little `dangling_binding(agent_type, tool_name)` or similar helper would make sure these 2 stay in sync, unless in the future it is ok for them to drift?

golem-worker-executor/tests/tool_discovery.rs
```diff
@@ -0,0 +1,361 @@
...
+        name.to_string(),
+        "1.0.0".to_string(),
+        format!("Summary for {name}"),
+        vec![format!("{name}-alias")],
+        high_bits,
+        low_bits,
+    )
+}
+
+#[test]
```
kmatasfp, line 167: Couple of gaps against the coverage the ticket asks for, on access filtering and revision isolation:

Nothing here uses a second environment — every test runs against `context.default_environment_id`, so we never actually check that a tool in one environment stays hidden from another.

Nothing covers a stale revision either. Revision isolation only shows up indirectly, through the replay test. A case where a binding or a `RegisteredTool` carries an older `deployment_revision` than the deployment would also address the comment I raised over on `environment_state.rs:87`.

kmatasfp (commented):
There's one thing we should probably change before merge (comment on `durable_host/tool/mod.rs`), and two things I wasn't sure about and would rather ask than assume. All inline.

On tests — the ticket asks for coverage of enumeration, lookup, empty environments, unknown names, access filtering and revision isolation. The first four are there. Revision isolation only really shows up indirectly through the replay test, and nothing sets up a second environment, so the "don't expose tools from unrelated environments" part isn't checked anywhere.

Tiny one: the comment at the top of `durable_host/tool/mod.rs` (lines 15-17) ends with "until the tool runtime is implemented". `AGENTS.md` asks that comments say what the code does now rather than what's coming later. The "layered on in later steps" bit was already dropped — this is just the leftover.

kmatasfp (commented):
One follow-up on performance, inline.

golem-worker-executor/src/services/environment_state.rs
```diff
@@ -20,16 +20,108 @@ use golem_common::model::agent_secret::{
...
+        return Ok(Vec::new());
+    };
+
+    bindings
+        .keys()
+        .map(|tool_name| {
+            deployment
+                .registered_tools
+                .get(tool_name)
+                .cloned()
```
kmatasfp, line 198: Each call here deep-clones every tool it returns, and the source is already behind an `Arc<EnvironmentState>` from the cache — so a shared, immutable tool definition gets copied in full, schema graph and all, every time. `DiscoveredTool::from` then destructures it and drops everything from `source` bar the component id. There's a second `Vec` on top of that, allocated in `get_all_tools_model` for the `.map(DiscoveredTool::from).collect()`.

The same `.cloned()` is in `get_accessible_tool_from_snapshot` at line 115. That one only copies a single tool rather than the whole accessible set, so it matters less, but it's the same wasted copy and the same fix.

One agent doing this occasionally is nothing, but with a lot of agents each calling `get-all-tools` per model turn it could start to add up — and it looks fairly easy to avoid. If both functions built `DiscoveredTool` directly they'd only need to clone `definition` and the component id, and the full copies and the second `Vec` would go away.

golem-worker-executor/src/durable_host/tool/mod.rs
vigoo: This is an error now (in practice, if it's None it means that an oplog processor plugin called this host function)

golem-worker-executor/src/durable_host/tool/mod.rs
vigoo: This now works the same as the agent type related host functions - an agent sees the deployment it's originated from, not the latest.

golem-worker-executor/src/services/environment_state.rs
vigoo: Fixed

PR #3748, opened 2026-08-19
Executor client connect timeout

kmatasfp:
**The symptom**

Killing a worker-executor pod stalled every request already queued against it for about two minutes, even though the shard-manager recovered in under nine seconds.

Chaos run S5 on golem-dev (2026-08-19, run 32272077341) killed one executor. It left the routing table 1.3s later and its shards were reassigned at 8.4s. Despite that, 496 operations submitted in a 32-second window each stalled around 120s. One traced request spent 119,781ms on a single attempt, then succeeded in 86ms on retry against the replacement.

Correctness held throughout: retries under the same idempotency key succeeded, read-back was clean, 250 of 250 agents consistent. This is an availability bug.

**What was broken**

All of it lives in `golem-service-base/src/grpc/client.rs`, so all of it applies to every gRPC client: shard-manager, registry, compilation and debugging services, not only worker-service.

- `requires_reconnect` matched only `Code::Unavailable`. A dead transport arrives as `Unknown`, and a connection that closes with requests still on it arrives as `Cancelled`, the commonest of the three at 277 of 400 concurrent calls against a peer killed mid-burst. Broken channels stayed cached indefinitely, so every later request queued onto a connection that could never work again.
- None of tonic's keep-alive options were set; they all default to `None`. A peer that accepts TCP but never finishes the HTTP/2 handshake, which is what a pod mid-teardown looks like, hung with no bound at all. I gave up and stopped the test at 180s. `connect_timeout` covers only the TCP connect, and `request_timeout` stays unset because agent invocations run arbitrarily long.
- Every waiter paid for its own connect. tonic's `Channel` is a tower `Buffer` and serves requests one at a time, so each queued request waited for all those ahead of it to burn a full `connect_timeout` before its own connect began. Worst case `queue_depth x connect_timeout`. Divide 119,781ms by the 10s `connect_timeout` and that request sat about twelfth in line. Sharing one attempt per target is not enough on its own: tonic reconnects below anything we can reach, serially inside the one `Buffer` worker, which is the incident's own path.
- Eviction was keyed on the target alone, so a failure arriving late discarded the replacement an earlier failure's retry had already established.
- The cache was pruned only on failure. A pod that died and was never called again held its `Channel`, and the `Buffer` worker task tower spawned for it, for the life of the process. Measured at one leaked task per pod.
- `internal_grpc_success_seconds` timed only the RPC, and only the final attempt of it, so a call that spent ten seconds connecting and 80ms on the wire recorded 80ms.

**What it does now**

One connect attempt per target, shared by every waiter through a `Shared` future bounded by `connect_timeout` and driven by a task of its own, so losing every waiter cannot strand it. A successful attempt is kept and reused; a failed one is never handed on.

One cache, holding the client rather than the channel underneath it, so the client is built with the connection instead of after it. `GrpcClient` is `MultiTargetGrpcClient` with its target filled in rather than a second copy of the same logic; the two had drifted while they were separate, and the single-target call loop the stall lived in had no test reaching it.

Each connection carries an identity and a retirement. Eviction matches on the identity, so only the connection that actually failed is dropped. The retirement records two things separately: that the connection is retired, which any failure sets, and that it is gone, which only a failed transport sets. Being gone also releases every request still riding the connection as soon as any one of them proves it dead, which is what reaches the callers tonic would otherwise re-dial for one at a time. A status the peer sent back leaves those requests running.

A caller handed a connection a sibling retired in the meantime goes back for another, up to three times, rather than passing that sibling's failure on. Measured with 16 callers sharing a peer that retires one call in ten, going back takes callers failing outright from 9.5% to 1.5%, 0.25% and 0.03% over the three turns, for 6%, 9% and 8% more connections opened: what the caller sees decays geometrically while what it costs saturates after the second turn.

A connection counts as dead on `Unavailable`, `Unknown` or `Cancelled` carrying a transport error: respectively a blackholed connect or an expired keep-alive ping, a connection the kernel gave up on, and a connection that closed with requests still on it. A request timeout is excluded by its `tonic::TimeoutExpired` source rather than by its code, and the codes tonic derives from an HTTP/2 reset are excluded by their code, since a reset ends one stream and leaves the connection carrying everyone else. Pinned against a peer that resets every stream with `ENHANCE_YOUR_CALM`, which arrives as `ResourceExhausted` with a transport error attached.

HTTP/2 keep-alive is on by default at 10s interval and 10s timeout, pinging while idle, which bounds detection at roughly 20s without capping how long a request may run.

A target nothing has called for ten minutes has its connection dropped, and built again on the next call. The sweep runs on any call once a minute has passed since the last one, rather than on a call that missed the cache: a miss means a target nobody has reached before, and a cluster whose targets are all connected stops producing them. A connection counts as in use while a request is riding it, however long ago it was handed out, so a long invocation is not swept out from under itself.

**Measured**

With `connect_timeout = 1s`, eight concurrent calls to an unreachable peer went from 8.008s to 1.002s, and the same eight queued on a cached channel whose peer had died went from 8.008s to 1.089s. The healthy path is unchanged, 16 concurrent calls over a single connection in 43ms. A connection already open when its peer went quiet is bounded at 3.0s, against 13.0s falling through to the kernel.

On golem-dev, against an earlier build of this branch: the S5 rerun went from 340 retries during the fault to 0, and worst fault p99 from 143,447ms to 11,912ms. S13's five rolling restarts produced 0 retries across 109,716 operations, with 300 of 300 agents consistent. The current head has not had a chaos run.

**Behaviour changes worth knowing**

- Connections are established when a client is asked for rather than lazily on first RPC, so connection failures surface from establishment. Both run through the same retry path and `retries_on_unavailable` still governs both. The public `call()` is unchanged.
- A dead connection now ends the other requests in flight on it, rather than leaving each to wait its turn to re-dial. A status the peer sent back does not, since a peer that answered at all proves the transport works.
- A request that never returns keeps its connection out of the idle sweep for as long as it runs. Nothing else ends such a request either, since `request_timeout` stays unset for agent invocations.
- `internal_grpc_success_seconds` covers the whole call: establishing the connection, every attempt, and the backoff between them. Cold and reconnecting calls read higher than they did before.
- `GrpcClient` delegates to `MultiTargetGrpcClient`. Its one connection is still exempt from the idle sweep, and its tracing span now carries the endpoint, which it did not before.
- Carries `133039638`, which raises the recursion limit in `golem-worker-executor-test-utils`. Not this PR's defect: `1.5.x` does not build under rustc 1.98 without it, and it is cherry-picked here so CI can run. It drops out on rebase once the branch it came from lands.
- Three config fields are added to every gRPC client section across all six services: `http2_keep_alive_interval`, `http2_keep_alive_timeout`, `http2_keep_alive_while_idle`. All are overridable per service in the usual way.

**Known imprecision**

A status the peer sent back still rebuilds the connection, because `requires_reconnect` matches any `Unavailable` whatever its source. That predates this change, but the retirement now sticks, so callers holding that connection go and build their own instead of reusing it. Measured with 64 concurrent callers and 5% of calls answered `Unavailable`: 15% of them run out of turns and fall back to the outer retry, and connections opened go from 0.014 to 0.04 per call. Narrowing the predicate to require a transport source would remove it, at the cost of keeping a connection the peer is draining until the transport actually closes, so it belongs in its own change with its own chaos run.

`REFUSED_STREAM` resets a single stream but arrives as `Unavailable`, which is also where an expired keep-alive ping arrives, so no status code separates the two. A server that sends GOAWAY and then refuses streams above `last_stream_id` will have its still-valid streams cut short. That costs a reconnect and a retry rather than a wrong answer, which is the cheaper way to be wrong.


vigoo (changes requested):
Please check the following three items the agentic review found:

### [P1] Single-flight does not cover callers already queued in tonic's cached channel

On a cache hit (`golem-service-base/src/grpc/client.rs:398-408`), callers retain clones of the existing tonic `Channel` and its shared Tower buffer. Once that channel is dead, tonic's `Reconnect` exposes a failed reconnect to one buffered request, consumes that error, and starts another reconnect for the next buffered request. Removing the map entry after the first result (`client.rs:373-375`) cannot rescue requests that already own clones of the old buffer, while `Connections` only participates on a cache miss. Those requests can therefore still pay `queue_depth × connect_timeout`, which is the incident behavior this PR is intended to eliminate.

The regression test does not cover this sequence: `grpc_client.rs:715-731` waits for a separate call to retire the stale channel before launching the concurrent cohort. The cohort consequently starts after eviction and tests only fresh callers. Please add coverage where callers are queued on the dead cached channel and ensure they are all invalidated or moved onto the shared replacement attempt.

### [P2] A late failure from an old connection can evict its replacement

Eviction is keyed only by URI (`client.rs:373-375`, with the analogous single-target `take()` at `client.rs:266-268`). If two callers retain connection A, the first failure can remove A and install replacement B; when the second caller's failure arrives later, it unconditionally removes B and its settled `SharedConnect`, causing connection C to be opened.

A local reproducer with two staggered `Unavailable` results opened three TCP connections instead of the expected initial connection plus one replacement. Cached typed clients need a generation/identity, and eviction must remove an entry only when it is still the connection used by the failing call.

### [P2] Successful-call latency now excludes connection establishment

Both call loops await `connected_client()` before starting the success timer (`client.rs:252-260` and `client.rs:359-368`). Previously `get()` returned a lazy channel, so establishment occurred inside the timed RPC future. A cold attempt that spends 10 seconds connecting and 80 ms in the RPC now records only 80 ms in `internal_grpc_success_seconds`, masking the latency this change is intended to improve.

Please start the per-attempt timer before `connected_client()` (or record establishment separately) so successful cold/reconnect attempts retain observable caller latency.

vigoo:
One more:

```
Peer-originated Unavailable can resurrect an evicted generation.

1. Caller B obtains generation G.
2. Caller A receives peer-returned Unavailable.
3. retire evicts G but intentionally does not cancel it (client.rs:469–479, 687–694).
4. B reaches install; it only checks cancellation (435–450, 641–656), so it caches G again.
5. New callers receive the generation that was just evicted.
```

vigoo (approved):

PR #3749, opened 2026-08-20
Fix replay of durable calls completed in the crash window

vigoo:
Fixes the crash-window replay divergence originally observed as the flaky Go outgoing-HTTP replay test (`go_outgoing_http_replayed_without_network` on the `go-sdk` branch).

Linear: [GOL-443](https://linear.app/golem-cloud/issue/GOL-443/fix-flaky-mixed-p2p3-durable-replay-in-go-outgoing-http-test)

## Problem

The executor recorded a durable call's `End` entry when the **host** completed the operation, before the **guest** observed the result. A crash between host completion and guest observation (e.g. an HTTP response fully received by the host but not yet polled by the guest) made replay hand the guest a completion it never saw, causing divergent replay.

## Fix

The oplog now separates host completion from guest consumption:

- `End` keeps recording host completion and persists the result.
- `CompletionDelivered` / `CompletionDiscarded` record whether the guest actually consumed the completion (delivered through waitable-set wait/poll, a guest callback, or sync lowering — or consumed without observing the result via `subtask.cancel` / cancellation).
- During replay, an `End` with a marker waits for its exact marker before releasing the result to the guest; a markerless `End` (the crash window) waits for natural replay-tail exhaustion and then becomes live-armed, so the guest re-observes the completion in live mode.
- Cancelled partial results remain immediately deliverable.

The delivery point is captured with terminal observers added to the golem-wasmtime fork; the required changes are already merged into `golem-wasmtime-v46.0.1-p3` (the branch `main` pins), and `Cargo.lock` is updated to that revision.

The oplog format change is not backward compatible; acceptable for the unreleased format.

## Validation

- `cargo test -p golem-worker-executor --lib -- durable_host::replay_state durable_host::concurrent` — 139/139 passed on this branch.
- `cargo test -p golem-worker-executor --lib -- tail_gated` — 3/3 passed.
- The original flaky Go integration test only exists on the `go-sdk` branch; the same executor fix passed it there 40/40 across repeated runs.


vigoo:
Note: the huge diff is due to WIT changes causing regenerated bindings

PR #3682, opened 2026-07-01
Matrix dependency for sqlite/psql integ tests

vigoo:

PR #3760, opened 2026-08-24
Update test-r to 3.0.12

vigoo:
## Summary

- update test-r from 3.0.11 to 3.0.12
- refresh test-r-core, test-r-macro, and quick-xml lockfile entries
- pick up the upstream IPC worker lifetime race fix

## Verification

- `cargo make unit-tests`
- `cargo make fix`
- `CARGO_INCREMENTAL=0 CARGO_BUILD_JOBS=2 cargo make build`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,7 +33,7 @@ permissions:
 env:
   BUILD_TARGET: "x86_64-unknown-linux-gnu"
   WASM_RQUICKJS_VERSION: "0.4.1"
-  MOONBIT_INSTALL_VERSION: "0.10.1+a46be2066"
+  MOONBIT_INSTALL_VERSION: "0.10.2+1bb3e16cf"
   WASI_SDK_VERSION: "25"
   WASI_SDK_PATH: /opt/wasi-sdk
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7886,9 +7886,9 @@ dependencies = [
 
 [[package]]
 name = "quick-xml"
-version = "0.38.4"
+version = "0.41.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b66c2058c55a409d601666cffe35f04333cf1013010882cec174a7467cd4e21c"
+checksum = "e660451e55124f798a69a5af3f49ccfbefbd41910eefd25caf2393e1f3473ec1"
 dependencies = [
  "memchr",
 ]
@@ -10084,9 +10084,9 @@ dependencies = [
 
 [[package]]
 name = "test-r"
-version = "3.0.11"
+version = "3.0.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c6a37b0a7d3976dccffa7df6aae2dba4127093484ec6376b658d6902e2b44d9f"
+checksum = "1dfad2a3c321239151d31a82d84e9ba64ce954c1248e3e945cec27f580b54040"
 dependencies = [
  "ctor",
  "test-r-core",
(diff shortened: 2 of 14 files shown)


vigoo:

vigoo:
recheck

vigoo:
## CI investigation summary

The 9 failing checks reduce to 7 root causes. The scoped fixes below were reviewed against the fixes already developed for #3758, implemented locally, and validated; they have **not** been pushed to this PR.

1. **`build-golem-moonbit` — unavailable MoonBit release**
   - Update `MOONBIT_INSTALL_VERSION` from `0.10.1+a46be2066` to `0.10.2+1bb3e16cf`.
   - The new binary and core asset URLs both return HTTP 200; the old download returned 403.

2. **`scala-sdk-integration-tests` — RDBMS check never connects**
   - `Rdbms.Postgres.open` and `Rdbms.Mysql.open` return lazy handles, so the explorer never exercised the unreachable endpoints.
   - Follow each `open(...)` with `.flatMap(_.query("SELECT 1"))` and report the query result.

3. **`worker-tests-group1` — resource-limit tests exhaust CI memory**
   - Add `sequential_suite!(resource_limits)`.
   - This is test-resource isolation rather than masking a product bug: each test starts a full production-context executor, and parallel execution reached roughly 24 GiB RSS on a 31 GiB runner. Concurrency exercised inside each test is unchanged.

4. **CLI agent shards / `worker-tests-group3` — injected environment metadata is denied by the no-default-permissions helper**
   - `without_default_host_permissions` is a test helper. Keep no-card behavior allow-by-default; when cards are present, explicitly grant account-scoped reads for all Golem-injected metadata variables:
     - `GOLEM_AGENT_ID`
     - `GOLEM_AGENT_TYPE`
     - `GOLEM_WORKER_NAME`
     - `GOLEM_COMPONENT_ID`
     - `GOLEM_COMPONENT_REVISION`
   - Use exact names rather than `GOLEM_*`, so unrelated user variables are not exposed.
   - P2/P3 revocation and replay tests were extended to cover all five variables.

5. **Filesystem stream tests — nonblocking zero write is not completed/flushed**
   - Rename `stream_to_file` to `write_zeroes_to_file_via_stream` and update all callers.
   - The helper intentionally generates zero bytes; there is no source stream. These tests need deterministic file size/quota/replay behavior without allocating a large payload.
   - Replace `write_zeroes(len)` with `blocking_write_zeroes_and_flush(len)` so backpressure is observed and all bytes are flushed before the agent method returns.

6. **`integration-tests-group8` — DB matrix pairs fixtures from different owners**
   - Split the Redis control smoke test into PostgreSQL and SQLite cases, and explicitly tag both `EnvBasedTestDependencies` and `WorkerExecutorClusterControlStub` with the same backend.
   - This is incorrect use of test-r's explicit fixture graph, not a test-r bug: only `deps` was matrix-selected while the companion stub remained attached to the default owner.
   - The mismatch was introduced by #3682 (merged July 2, 2026); #3760 exposed it.

7. **`integration-tests-group9` — `get_config_value` drops an unfinished durable call on resolver failure**
   - Run the re-executable `ReadRemote` operation through `CallHandle::run`.
   - Permission denial remains a persisted normal response mapped to typed `PermissionDenied`.
   - Missing/deleted secrets and resolver/storage/schema failures abandon the unfinished call and propagate as retryable invocation failures, rather than dropping a live `NotCancellable` handle.
   - This matches Agent Permissions §5.7 (denials do not trap) and the WASI P3/concurrent-durability incomplete-replay model. A deleted required secret now produces the expected service error instead of killing the executor.

### Local validation

Passed:

- Rust formatting, `git diff --check`, relevant crate/test-binary compilation
- Host API release component rebuild
- Scala formatting and `testAgents/compile`
- 9 resource-limit tests
- stream/quota/replay tests
- P2/P3 permission revocation and replay tests
- SQLite DB matrix smoke test
- deleted-secret Rust/SQLite test
- recipient-isolation SQLite test
- workflow YAML parsing and MoonBit asset checks

PostgreSQL variants could not start locally because `/var/run/docker.sock` was unavailable. The TS deleted-secret variant was blocked by a pre-existing stale local WASM ABI mismatch; the freshly exercised Rust variant passed.

noise64 (approved):

PR #3761, opened 2026-08-24
Chaos testing phase 3

kmatasfp:


integration-tests/src/chaos/scenarios/s9.rs
```diff
@@ -0,0 +1,722 @@
...
+
+/// What the running code reports on each of the two builds.
+///
+/// Compiled into the WASM rather than read from metadata, which is the whole
+/// reason these numbers can be trusted: metadata says what the platform
+/// believes, `component_version` says what is executing.
+const VERSION_ON_THE_NEW_BUILD: u32 = 2;
+const VERSION_AFTER_ROLLBACK: u32 = 1;
+
+pub async fn run(
```
vigoo, line 118: There seem to be some confusion in this scenario between component _revision_ and component _version_. These are not guaranteed to be the same thing with rollbacks

integration-tests/src/chaos/scenarios/s11.rs
```diff
@@ -0,0 +1,661 @@
...
+
+/// How many waiters to sample after the baseline to prove wakeups happen at all.
+///
+/// A handful, because this is a smoke test rather than a measurement: if the
+/// completion path is broken, every waiter is equally broken, and the point is to
+/// fail before spending the fault window on a run that would report a clean
+/// account of nothing.
+const WAKE_PROOF_SAMPLE: usize = 5;
+
+pub async fn run(
```
vigoo, line 120: To be checked:

> stood_down is a global count. Once any waiter stalls, every unreadable waiter missing a confirmed wakeup is classified as NeverWoke, even if that particular waiter never stood down. One genuine stall plus an unrelated read timeout can therefore produce false violations and fail the scenario.

> Fix: retain stalled waiter identities and require both unreadable.contains(agent) and stalled_waiters.contains(agent).

integration-tests/src/chaos/scenarios/s10.rs
```diff
@@ -0,0 +1,725 @@
...
+/// How many targets to sample after the baseline to prove actions are firing at
+/// all.
+///
+/// A handful, because this is a smoke test rather than a measurement: if the
+/// scheduling path is broken, every target is equally broken, and the point is
+/// to fail before spending the fault window on a run that would report a clean
+/// account of nothing.
+const FIRE_PROOF_SAMPLE: usize = 5;
+
+pub async fn run(
```
vigoo, line 130: To check:

> needs_attention() checks only total == 0. If pending actions exist exclusively on surviving executors, on_killed_executor == 0, but the result is classified as normal context even though the kill exercised no lease recovery on the killed executor. S11 correctly handles the analogous condition.

> Fix: use self.total == 0 || self.on_killed_executor == 0, add the corresponding warning in describe(), and cover it with a test.

vigoo (approved):
Added three findings, other than that looks good

PR #3775, opened 2026-08-26
TS tool middleware support

vigoo:
Resolves GOL-26

### DIFF
--- a/cli/golem-cli/templates/ts/common-on-demand/golem.yaml
+++ b/cli/golem-cli/templates/ts/common-on-demand/golem.yaml
@@ -12,6 +12,7 @@ componentTemplates:
             GOLEM_APP_ROOT: "{{ appRootDir }}"
             GOLEM_TEMP: "{{ golemTempDir }}"
             GOLEM_COMPONENT_NAME: "{{ component_name | to_kebab_case }}"
+            GOLEM_COMPONENT_ROLE: agent
           sources:
           - "{{ componentDir }}/src"
           - "{{ golemTempDir }}/common/ts/rollup.config.component.mjs"
@@ -34,6 +35,7 @@ componentTemplates:
             GOLEM_APP_ROOT: "{{ appRootDir }}"
             GOLEM_TEMP: "{{ golemTempDir }}"
             GOLEM_COMPONENT_NAME: "{{ component_name | to_kebab_case }}"
+            GOLEM_COMPONENT_ROLE: agent
           sources:
           - "{{ componentDir }}/src"
           - "{{ golemTempDir }}/common/ts/rollup.config.component.mjs"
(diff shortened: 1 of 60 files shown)


sdks/ts/packages/golem-ts-sdk/src/internal/tool/middlewareGuest.ts
```diff
@@ -0,0 +1,95 @@
...
+function throwRegistrationError(): void {
+  const error = middlewareRegistrationError();
+  if (error) throw error;
+}
+
+function invalidToolName(name: string): ToolError {
+  return { tag: 'invalid-tool-name', val: name };
+}
+
+export const golemTool010ToolMiddlewareGuest: GolemToolMiddlewareGuest = {
```
noise64: > **Blocker: this export name doesn't match what the generated wrapper looks up.**
>
> `wasm-rquickjs` resolves guest exports off `globalThis.userModule` by the interface's **short** lower-camel name, package-qualifying only when two exports collide (`exported_interface_js_names`, 0.4.2 `src/lib.rs:388`). `tool-middleware-guest` collides with nothing in either new world, so it stays `toolMiddlewareGuest`.
>
> Generating the wrapper crates with `wasm-rquickjs-cli 0.4.2` (the version this PR pins) confirms it. For `--world agent-tool-middleware-guest`, `src/lib.rs` contains:
>
> ```rust
> &["golemAgent200Guest", "initialize"],          // qualified: `guest` collides
> &["golemTool010Guest",  "discoverTools"],       // qualified: `guest` collides
> &["saveSnapshot",       "save"],                // short: no collision
> &["toolMiddlewareGuest", "discoverToolMiddlewares"],
> &["toolMiddlewareGuest", "getToolMiddleware"],
> &["toolMiddlewareGuest", "invokeToolMiddleware"],
> ```
>
> `--world tool-middleware-guest` emits the same three `toolMiddlewareGuest` paths. Both crates set `JS_EXPORT_MODULE_NAME` to the SDK entry for their role (`@golemcloud/golem-ts-sdk/middleware` and `@golemcloud/golem-ts-sdk` respectively), so `userModule` is the module these lines resolve against.
>
> Since `src/middlewareRuntime.ts:23` and `src/index.ts:166` export only `golemTool010ToolMiddlewareGuest`, `get_cached_js_export` misses and panics via `dump_cannot_find_export`. Both `ts-tool-middleware` and `ts-agent-tool-middleware` components build clean and then fail on the first host call into the guest, `discover-tool-middlewares` included. The CLI role test only decodes the component's WIT world, so it can't catch this.
>
> Suggested fix, two parts:
>
> 1. Add the short-name alias to both entries, matching what `index.ts:902-912` already does for the agent guest:
>    ```ts
>    export const toolMiddlewareGuest = golemTool010ToolMiddlewareGuest;
>    ```
>    Nothing else in the repo references the name, so this is purely additive.
>
> 2. Type this object against the generated `tool-middleware-guest` module rather than the hand-written `GolemToolMiddlewareGuest` interface above, so `tsc` enforces the WIT contract. The absence of that link is what made this invisible. Worth extending to the hand-declared `GolemAgentGuest` / `GolemToolGuest` in `index.ts`.
>
> Also: `cli/golem-cli/test-data/ts-tool-middleware-roles/middleware/src/main.ts:18` already reaches into the SDK module to assert the guest is wired — it just asserts the qualified name. Switching it to `toolMiddlewareGuest` turns that fixture into a build-time regression test.
>
> *(Aside, pre-existing: the comment at `index.ts:902-912` says the wrapper looks up guests "by the WIT interface short name" and exports both `guest` and `tool`. Per the generated output above, both agent-guest interfaces are in fact qualified — so `guest` is dead, and `tool` was never any interface's short name. That stale comment is plausibly why middleware got qualified-only treatment.)*

sdks/ts/packages/golem-ts-sdk/src/internal/tool/middlewareRuntime.ts
```diff
@@ -0,0 +1,627 @@
...
+  const underlying = InvocationScopedUnderlying.create(raw, ownership);
+  try {
+    const result = await invoke(underlying);
+    try {
+      validateInvocationResult(result);
+    } catch (error) {
+      await closeAsyncIterable(invocationStdout(result));
+      throw new ToolInvokeError({ tag: 'invalid-result', val: errorMessage(error) });
+    }
+    ownership.forwardStdout(result.stdout);
```
noise64: > **Middleware that transforms `stdout` silently emits nothing.**
>
> `forwardStdout` transfers by exact object identity (`InvocationOwnership.transfer`, L347). Returning the tracked stream verbatim works; returning a generator that *wraps* it leaves the source untransferred, so `dispose()` in the `finally` below closes it before the host iterates:
>
> ```ts
> const res = await underlying.invoke(commandPath, input, stdin);
> return { stdout: (async function* () {
>   for await (const b of res.stdout!) yield redact(b);
> })() };
> ```
>
> And it fails silently: `TrackedOutputStream.next()` (L364) never checks `this.closed`, so it delegates to an already-`return()`ed iterator and yields `{done: true}`. The host sees a well-formed, empty stream — no bytes, no error. Redacting or filtering tool output seems like one of the main reasons to reach for a middleware.
>
> Two things I checked before raising this: post-return lazy iteration does work, since the pass-through case already hands the host a `TrackedOutputStream` — a JS wrapper this file builds, not the raw bridged reader (`tool-middleware-runtime.test.ts:447`) — so the sweep is the only obstacle. And Rust doesn't hit it: `InputStream` is moved by value with no sweep (`tool_middleware_impl.rs:120`), so holding the reader and pumping into a new stream works there.
>
> Is streaming `stdout` transformation meant to be supported? If yes, middleware needs some way to claim ownership so the sweep skips the stream. If not, that's a fair scope call — but `TrackedOutputStream.next()` should then throw after disposal rather than report `done`, so this fails loudly instead of returning empty output.

cli/golem-cli/templates/ts/common-on-demand/rollup.config.component.mjs
```diff
@@ -51,6 +51,147 @@ function loadComponentTsConfig(componentDir) {
...
+    if (roleName === "agent") {
+        const sdkDeclarationFiles = new Set();
+        for (const sourceFile of userSources) {
+            const resolved = ts.resolveModuleName(
+                sdkPackage,
+                sourceFile.fileName,
+                parsedTsConfig.options,
+                ts.sys,
+            ).resolvedModule;
+            if (resolved) sdkDeclarationFiles.add(normalizedFileName(resolved.resolvedFileName));
```
noise64: I think this should not be silent

cli/golem-cli/templates/ts/common-on-demand/rollup.config.component.mjs
```diff
@@ -51,6 +51,147 @@ function loadComponentTsConfig(componentDir) {
...
+    const name = declaration?.name;
+    return name && (ts.isIdentifier(name) || ts.isStringLiteralLike(name)) ? name.text : undefined;
+}
+
+function isSdkModule(moduleName) {
+    return moduleName === sdkPackage || moduleName.startsWith(`${sdkPackage}/`);
+}
+
+function validateComponentRole(componentDir, parsedTsConfig, roleName) {
+    const role = componentRoles[roleName];
```
noise64, line 139: componentDir is unused

cli/golem-cli/templates/ts/common-on-demand/golem.yaml
```diff
@@ -46,6 +48,100 @@ componentTemplates:
           into: "{{ golemTempDir }}/agents/{{ component_name | to_snake_case }}.dynamic.wasm"
         componentWasm: "{{ golemTempDir }}/agents/{{ component_name | to_snake_case }}.dynamic.wasm"
         outputWasm: "{{ golemTempDir }}/agents/{{ component_name | to_snake_case }}.wasm"
+  ts-tool-middleware:
```
noise64, line 51: Just to check: does this mean that tool middlewares need dedicated components?

sdks/ts/packages/golem-ts-sdk/src/internal/tool/middlewareRuntime.ts
vigoo: Note that there is a redesign of how stdin/out streaming works in GOL-337, coming together with GOL-35 in a separate branch. (But applying some fixes to this)

sdks/ts/packages/golem-ts-sdk/src/internal/tool/middlewareGuest.ts
vigoo: Fixed

cli/golem-cli/templates/ts/common-on-demand/golem.yaml
vigoo: They don't need, but if they do it's safer (wasm level guarantee that they can only call the underlying tools and not other ones). 
I'm not sure this feature will survive until the end of the agent tools epic, its benefits feel less than the complexity; in the rust SDK the same thing is achieved with feature flags. 
So I wanted to go with it for now, and review in the end (when everything is wired and working) if we want to simplify.

PR #3777, opened 2026-08-26
Add generic agent reflection and TypeScript clients

noise64:
- resolves GOL-44
- resolves GOL-45
- hardens reflected lifecycle creation cleanup and requires agent view permission for full-ID type lookup
- makes reflection identity environment-scoped with component-free ParsedAgentId; the runtime resolves and authorizes the deployed implementation while AgentType.implementedBy remains metadata
- exposes reflection host identities as direct WIT `string` values without a generated host `AgentId` alias; structured core/management `AgentId` and public `ParsedAgentId` remain unchanged
- moves generic schema renderers into golem-schema and keeps only platform renderers in golem-common
- replaces TypeScript clientFor(...) with definition-owned .client factories across the SDK, bridge, REPL, templates, tests, skills, and docs
- completes immutable shared graph decoding, strict/optional lookup documentation, ComponentId SDK exposure, required-output validation, integer bounds/precision behavior, and reusable fatal UTF-8 decoding
- adds the common and TypeScript reflection skills and regenerates skill-derived documentation and WIT declarations

### Verification

- focused lifecycle, authorization, payload, renderer, reflection, decoder, and reflected RPC integration tests
- TypeScript build, declaration/type tests, full unit suite, lint, formatting, and 4 GB agent-template build
- Scala guest declarations and all three runtime role contracts
- MoonBit binding regeneration plus SDK-wide `moon check --target wasm`
- reflection playground type-check and component build
- WIT and generated-document drift checks
- merged latest `main` at `d43c34ddb` and reran local WIT, Rust, TypeScript, Scala, MoonBit, runtime-contract, and template validation
- refreshed CI for head `461105d70` pending


sdks/ts/packages/golem-ts-sdk/src/client.ts
```diff
@@ -223,124 +334,213 @@ function encodeConfigOverrides(
...
-      : ({ tag: 'single', codec: retCodec } as const);
-    return { name, inputCodecs, output };
-  });
+  const methodCodecs = compileRemoteMethods(def.methods);
 
   const configDecls: ConfigDeclaration[] = compileConfig(def.config);
 
+  const createAgentId = (id: InferRecord<CallerInput<Id>>, phantomId?: Uuid): AgentId =>
+    AgentId.create({
+      componentId: getSelfMetadata().agentId.componentId,
```
vigoo: This `componentId` is not necessarily true - the invocation API only takes type name and resolves the implementer component on the host side. It can be implemented in another component, that's actually quite possible if someone is using reflection.

My guess is that `AgentId` (which contains `componentId`) is used in some places where (a TS equivalent of) golem-common's `ParsedAgentId` (only the type/params/phantom-id part) would be more appropriate.

This can lead to actual problems where the returned agent id is used in other host APIs which are actually using the component-id. One example:

```
// Running in component A, while `Counter` is implemented by compoennt B
const Counter = defineAgentClient({
  name: 'Counter',
  id: { name: z.string() },
  methods: {
    increment: method({ input: {}, returns: z.number() }),
  },
});

const id = Counter.agentId({ name: 'orders' });

const reflected = getAgentTypeByAgentId(id);
```

wit/deps/golem-agent/host.wit
```diff
@@ -11,6 +11,9 @@ interface host {
     /// Get a specific registered agent type by name
     get-agent-type: func(agent-type-name: string) -> option<registered-agent-type>;
 
+    /// Gets the registered agent type used by an existing agent, identified by its agent ID.
+    get-agent-type-by-agent-id: func(agent-id: agent-id) -> option<registered-agent-type>;
```
vigoo: I guess we need this (in addition to `get-agent-type`) to get the actually implemented agent metadata of a given living agent instance (and not the latest deployed one for its type name).

I don't like that it's re-introducing component-id being part of the API (as part of `agent-id`) when we (tried to) eliminated that from all public APIs already. The "agent-id" itself is `string` in WIT (at the moment, unfortunately), as you can see in `make-agent-id` below etc. So I think the correct thing would be to use `string` here too,  better to introduce at least a type alias for it, even better but probably out of scope to have its own wrapper WIT type.

sdks/ts/packages/golem-ts-sdk/src/client.ts
noise64: Fixed in 059b27811. Reflection now uses a component-free ParsedAgentId across definitions, dynamic clients, reflected clients, metadata, and getAgentTypeByAgentId. No caller component is manufactured. The host parses the environment-scoped identity, resolves the deployed implementation in the caller environment, and performs lookup and authorization against that actual full target. Component-bearing AgentId remains only in the lower-level host-management API.

wit/deps/golem-agent/host.wit
noise64: Fixed in 059b27811. The WIT host interface now defines parsed-agent-id as a string alias and uses it consistently for get-agent-type-by-agent-id, make-agent-id, parse-agent-id, and invocation metadata. The executor resolves the actual implementing component from the caller environment before retrieving the instance schema and requires AgentVerb::View on the resolved target.

sdks/ts/packages/golem-ts-sdk/src/client.ts
noise64: Fixed in 059b27811. Reflection now uses a component-free ParsedAgentId across definitions, dynamic clients, reflected clients, metadata, and getAgentTypeByAgentId. No caller component is manufactured. The host parses the environment-scoped identity, resolves the deployed implementation in the caller environment, and performs lookup and authorization against that actual full target. Component-bearing AgentId remains only in the lower-level host-management API.

cli/golem-cli/wit/deps/golem-agent/host.wit
```diff
@@ -11,22 +11,25 @@ interface host {
     /// Get a specific registered agent type by name
     get-agent-type: func(agent-type-name: string) -> option<registered-agent-type>;
 
+    /// An environment-scoped agent identity created by `make-agent-id`.
+    type parsed-agent-id = string;
```
vigoo: This thing is called **parsed** agent id in Rust because it's parsed and not a string. We should find a better name here :)

wit/deps/golem-agent/host.wit
noise64: Fixed in 059b27811. The WIT host interface now defines parsed-agent-id as a string alias and uses it consistently for get-agent-type-by-agent-id, make-agent-id, parse-agent-id, and invocation metadata. The executor resolves the actual implementing component from the caller environment before retrieving the instance schema and requires AgentVerb::View on the resolved target.

sdks/moonbit/golem_sdk/wit/deps/golem-agent/host.wit
```diff
@@ -11,19 +11,25 @@ interface host {
     /// Get a specific registered agent type by name
     get-agent-type: func(agent-type-name: string) -> option<registered-agent-type>;
 
+    /// An environment-scoped agent identity created by `make-agent-id`.
+    type parsed-agent-id = string;
+
+    /// Gets the registered agent type used by an existing agent, identified by its agent ID.
+    get-agent-type-by-agent-id: func(agent-id: parsed-agent-id) -> option<registered-agent-type>;
```
vigoo: > **P2 — Regenerate the MoonBit bindings for this signature change.** The WIT now accepts a string, but `sdks/moonbit/golem_sdk/interface/golem/agent/host/top.mbt:8334–8338` still accepts `@types.AgentId` and passes its component UUID. The corresponding import in `ffi.mbt:19` still takes `(Int64, Int64, Int, Int, Int)`, whereas the new string argument plus return-area pointer requires `(Int, Int, Int)`. MoonBit components using this function therefore have an incompatible import when linked against the updated WIT. Please regenerate the MoonBit bindings and `.mbti` declarations with the pinned generator.

cli/golem-cli/wit/deps/golem-agent/host.wit
noise64: Fixed in 26dfbc4a0. The host WIT now uses a local `agent-id = string` alias for the serialized environment-scoped wire identity. `ParsedAgentId` remains the name of SDK-side values that expose parsed type/constructor/phantom parts, so the wire alias no longer claims to be parsed.

sdks/moonbit/golem_sdk/wit/deps/golem-agent/host.wit
noise64: Fixed in 26dfbc4a0. I regenerated the MoonBit bindings with the pinned Golem wit-bindgen fork. `get_agent_type_by_agent_id` now takes `String`, and its import ABI is `(Int, Int, Int)` for pointer/length/return-area. The generated public interface and dependent FFI helper numbering were updated together; `moon check --target wasm` passes.

vigoo (approved):

PR #3783, opened 2026-08-28
Resolve MCP agent types during deployment

vigoo:
## Summary

- compile and persist full MCP registered-agent schemas with each deployment
- derive implementer mappings from the persisted schemas as a single source of truth
- remove per-agent database lookups from active MCP resolution

## Testing

- `cargo test -p golem-registry-service --lib -- deployment_context::tests --report-time`
- `cargo check -p golem-registry-service --all-targets`
- `cargo clippy -p golem-registry-service --all-targets -- --no-deps -Dwarnings`
- `cargo fmt -p golem-registry-service -- --check`

Resolves GOL-174


golem-registry-service/src/services/deployment/deployment_context.rs
```diff
@@ -484,6 +480,19 @@ impl DeploymentContext {
                 None
             };
 
+            let agent_type_implementers = registered_agent_types
```
noise64: > agent_type_implementers has no reader left in the repo once the loop in mcp.rs is gone — CompiledMcp::agent_types() (golem-service-base/src/mcp/mod.rs:47), its only other consumer, is itself uncalled. This adds a second and third copy of the same name→(component_id, revision) derivation (here, repo/model/deployment.rs:906, and golem-service-base/src/mcp/protobuf.rs:10), plus two more in the new tests, to keep a field alive that is now write-only.                                                                                                                                                                              
>
> Since registered_agent_types is the source of truth now, could we drop the field and expose fn agent_type_implementers(&self) -> AgentTypeImplementers on CompiledMcp instead? That collapses all five copies into one. The proto still carries the map (mcp/core.proto:15) — is there an out-of-repo consumer, or can that go too?

golem-registry-service/src/repo/model/deployment.rs
```diff
@@ -867,9 +867,9 @@ impl DeploymentRevisionCreationRecord {
 #[derive(Debug, Clone, BinaryCodec)]
 #[desert(evolution())]
 pub struct CompiledMcpData {
-    pub implementers: golem_service_base::mcp::AgentTypeImplementers,
     #[desert(default)]
```
noise64, line 870: Is this default has effect while we are breaking the schema? (there are a few other places like this)

vigoo: `None` is represented by a single `0` byte. this attribute tells the generated codec that when it encounters that, use `T::default()` to set the field.

noise64 (approved):

PR #3787, opened 2026-08-29
Tool invoke implementation

vigoo:
Resolves GOL-35


golem-worker-executor/src/durable_host/tool/attachment.rs
```diff
@@ -0,0 +1,2150 @@
...
+        }
+        let attachment = self
+            .consumer
+            .as_ref()
+            .expect("attachment stream producer lost its consumer")
+            .attachment
+            .clone();
+        let remaining = dst.remaining(store.as_context_mut());
+        if finish {
+            attachment.consumer_cancel();
```
noise64: > **`finish = true` means the read was cancelled, not that the stream closed**
>
> The fork documents `StreamResult::Cancelled` as "the operation was interrupted … and the producer **may be able to produce more items**" (`futures_and_streams.rs:588-595`), and retains the producer afterwards (`:2414`). The flag comes only from a guest `stream.cancel-read` (`:4172`) — reader *drop* is a separate path, already handled by `Drop for AttachmentConsumer`.
>
> Here it's treated as terminal: `consumer_cancel()` clears the buffer and selects the immutable `ConsumerCancelled` terminal, and `finished = true` makes every later poll return `Dropped`. So one cancelled read silently discards buffered stdout and kills the attachment for good.
>
> Same at `:1087`; the stdin side (`tool/mod.rs:226`, `:305`) has a worse variant, noted separately.

golem-worker-executor/src/durable_host/tool/mod.rs
```diff
@@ -99,6 +148,285 @@ pub struct FutureInvokeResultEntry {
...
+                Poll::Ready(Ok(())) => self.pending_acknowledgement = None,
+                Poll::Ready(Err(_)) => {
+                    self.items.take();
+                    return Poll::Ready(Ok(StreamResult::Dropped));
+                }
+            }
+        }
+
+        if finish {
+            self.items.take();
```
noise64: > **Same `finish = true` misreading on stdin — here it becomes silent truncation**
>
> Dropping the sender ends the pump's `recv()` loop (`:349`), which falls through to `producer.finish()` (`:364`) — `ByteStreamCloseCause::Finished`, a clean EOF. So a guest that cancels one pending `stream.write` into its stdin source hands the tool a **truncated input that looks like normal end-of-stream**, and the invocation completes and records a success terminal.
>
> Same at `:305` in `UnderlyingToolStdinStreamConsumer`. See the note on `attachment.rs:979` for why `Cancelled` isn't a close.

golem-worker-executor/src/durable_host/tool/mod.rs
```diff
@@ -457,232 +1233,2301 @@ fn tool_owner<Ctx: WorkerCtx>(
...
+                        &mut deferred_cleanup,
+                        &parent,
+                        start,
+                        operation::BodyAdmissionState::Staging,
+                        stdin_controller.as_ref(),
+                        stdout_controller.as_ref(),
+                    )
+                    .await;
+                }
+                let host_resource_exhausted = stdin
```
noise64, line 2913: > **Replay can re-derive stdin exhaustion and hard-diverge from the recorded terminal**
>
> The preflight at `:2410` only short-circuits when the *recorded* terminal is `Skipped`. A call recorded as `Executed` falls through to here — and `host_resource_exhausted` traces back to `ActiveAgents::try_acquire`, which is global node headroom after eviction, not a function of the invocation's inputs.
>
> So recovering a previously-successful capable call on a loaded node synthesizes a `Skipped`/`ResourceExhausted` terminal, which then fails the equality check in `complete_without_body_access` (`entity.rs:571`) with `unexpected_oplog_entry` — non-retriable. Co-tenant memory pressure at recovery time permanently breaks recovery of already-durable work.
>
> Tests can't catch it: `ActiveAgents::admission` is `None` under test, so `try_acquire` always grants.
>
> Suggestion: skip this branch when `scope().mode() == ReplayingCompleted`.

golem-worker-executor/src/durable_host/tool/mod.rs
```diff
@@ -457,232 +1233,2301 @@ fn tool_owner<Ctx: WorkerCtx>(
...
+    {
+        if !operation.transition_admission(
+            operation::BodyAdmissionState::Staging,
+            operation::BodyAdmissionState::SettledWithoutBody,
+        ) {
+            return Err(anyhow!(
+                "tool invocation was fenced before pre-dispatch cancellation"
+            ));
+        }
+        return cancel_registered_tool_before_body(
```
noise64, line 2760: > **Pre-dispatch cancellation runs before the attachments are configured, so publication no-ops**
>
> This returns at `:2517`, above the `match filesystem` block at `:2527` that calls `configure_live()` / `configure_completion()` — so both attachments are still `AttachmentMode::Pending`. `cancel_registered_tool_before_body` then does `cancel()` + `publish_completion()` (`:2208`), but `publish_completion` bails unless the mode is already `Completion { published: false }` (`attachment.rs:264`).
>
> Net effect: the terminal is selected but never made visible, so the caller's `stream<byte-stream-item>` for stdout never yields an item and never closes — `poll_produce` sees `chunks_visible = false`, `terminal_visible = false` and parks. Trigger: `async-invoke-and-await` with stdout against an incapable tool, cancelled before dispatch.
>
> `SkippedToolAttachmentEndpoints::publish_failure` (`:1968`) has the right shape — `configure_completion()` first, then fail, then publish.
>
> The existing cancellation coverage reads a stdout chunk before cancelling (`rust-caller:687`), so it only exercises the post-dispatch path where the mode is already `Live`.

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -9683,6 +9816,7 @@ impl PrivateDurableWorkerState {
             (OwnerRuntime::Agent, None) => golem_common::model::card::EffectiveSurface::default(),
             (OwnerRuntime::Entity(_), _) => configured_agent_effective_surface,
         };
+        let local_live_tail = matches!(runtime, OwnerRuntime::Entity(_)) && replay_state.is_live();
```
noise64: > **`local_live_tail` is seeded from the cursor's *position*, not from this store's execution mode**
>
> `replay_state.is_live()` here is the positional check (`last_replayed_index == replay_target`), and the flag is a one-way latch — 11 `store(true)` sites, no `store(false)`. So an entity Store constructed while the shared cursor is transiently clamped (e.g. a `NonPrimary` repair calling `CursorTx::switch_to_live`) latches `true` even when its own `scope.mode()` is `ReplayingCompleted`, and `is_live()` (`:10379`) then returns true for the rest of its life.
>
> Such a body takes live paths everywhere, never claims its recorded descendants, and re-appends them at the oplog tail under the old parent Start.
>
> The mode is known exactly at construction — is there a reason not to seed from `scope.mode()` (live ⇒ true, either replay mode ⇒ false) rather than from the shared cursor?

golem-worker-executor/src/durable_host/replay_state/claims.rs
```diff
@@ -850,27 +1005,163 @@ pub(super) fn parent_start_index_of(function_type: &DurableFunctionType) -> Opti
...
+            let HostRequest::EntityInvocation(request) = value else {
+                return Ok(false);
+            };
+            let metadata = desert_rust::deserialize::<EntityInvocationRequest>(&request.metadata)
+                .map_err(|error| {
+                format!("failed to decode entity invocation request metadata: {error}")
+            })?;
+            Ok(expected.matches(&metadata, &request.input))
+        }
+        RequestClaimIdentity::ToolInvocation(expected) => match value {
```
noise64, line 1128: > **Accept/reject claim identity has no task ordinal, so two identical concurrent calls can swap outcomes on replay**
>
> This arm matches either the accepted entity `Start` or the predispatch rejection, and `claimed_starts` only prevents double-claiming — nothing binds a claimed `Start` back to the future that produced it. The identity fields (`ToolInvocationRejectedIdentity`, `entity.rs:595-603`) are tool name / command path / input / decode failure / has_stdin / has_stdout / call_mode. No ordinal.
>
> The accept-vs-reject decision is made *after* awaits — `get_tool_activation` and `authorize_live_permissions_at_serialized_access` (`tool/mod.rs:1430-1546`) — so which of two identical concurrent calls records its `Start` first is IO/scheduling order, not deterministic task-initiation order. If one was accepted and one rejected (a single-use permission card, or a tool deployment changing mid-invocation), replay can hand the rejection to the other future. Both claims succeed, so nothing detects the swap.

sdks/rust/golem-rust/src/agentic/tool_client.rs
```diff
@@ -345,12 +385,159 @@ fn map_infallible_rpc_error(error: WitRpcError) -> ToolError<Infallible> {
...
+/// attachment. The host boundary is required for error-bearing byte streams,
+/// which cannot rendezvous between sibling futures in one Component Model task.
+pub fn pump_tool_stdin(source: InputStream) -> agentic_host_api::ToolStdin {
+    agentic_host_api::create_stdin_from_stream(source)
+}
+
+/// A started stdout-bearing tool call. Output, structured completion, and
+/// cancellation are independent capabilities.
+pub struct ToolInvocation<T, E> {
+    pub stdout: InputStream,
```
noise64: > **`stdout` is a public field, but consuming it without also driving `result()` deadlocks for a capable tool**
>
> The doc above says output and completion are "independent capabilities", and `:458` says stdout "may be moved into a concurrent consumer". For a filesystem-**incapable** tool that holds — the body runs off-lane and the attachment is `Live`. For a **capable** tool it doesn't: the body isn't launched until result demand forms a lane cohort (`release_capable_tool_cohort`), and its stdout is completion-staged, so a consumer that only reads `stdout` blocks forever — and the caller's invocation can't end, so the `ParentEnd` cohort never fires either.
>
> `collect()` (`:478`) is safe because it joins both. But the same generated client code deadlocks or not depending on a property of the *tool* that the caller can't see.
>
> Worth either documenting the constraint on this field or making the shape enforce it (e.g. hand out stdout only via something that also drives the result).

sdks/rust/golem-rust/tests/tool.rs
```diff
@@ -1300,10 +1386,19 @@ impl EchoMiddleware for Policy {
...
+
+fn stdout(invocation: ToolInvocation<(), Infallible>) -> InputStream {
+    invocation.stdout
+}
+
+async fn consume(_stdout: InputStream) {}
 
-fn stdout(result: InvocationResult) -> Option<InputStream> {
-    result.stdout
+fn consume_result_and_stdout_concurrently(invocation: ToolInvocation<(), Infallible>) {
```
noise64, line 1398: > **This fixture never polls, so it can't catch the deadlock the deleted test covered**
>
> `result` and `output` are constructed and immediately dropped at `:1401`, so this only type-checks that the two can coexist — no future is ever polled. The runtime test it replaced (`async_stdout_tool_writes_do_not_deadlock`) drove a real wasmtime harness and asserted `reader_drop_is_observed`.
>
> Given `ToolInvocation::stdout` is public and consuming it without driving `result()` deadlocks for a capable tool, this is the one place that behaviour was covered.

sdks/rust/golem-rust/src/agentic/tool_client.rs
```diff
@@ -345,12 +385,159 @@ fn map_infallible_rpc_error(error: WitRpcError) -> ToolError<Infallible> {
...
+    future: Rc<agentic_host_api::FutureInvokeResult>,
+    decode: Rc<dyn Fn(InvocationResult) -> Result<T, ToolError<E>>>,
+    decode_error: Rc<dyn Fn(TypedSchemaValue) -> Result<E, String>>,
+}
+
+impl<T, E> ToolInvocation<T, E> {
+    /// Returns an independently owned structured-completion future. The
+    /// stdout field may be moved into a concurrent consumer after this call.
+    pub fn result(&self) -> impl Future<Output = Result<T, ToolError<E>>> + use<T, E> {
```
noise64, line 616: > **`result(&self)` can produce two concurrent `get` calls, which the host rejects**
>
> Takes `&self` and clones the `Rc<FutureInvokeResult>`, so calling it twice and polling both issues two overlapping `get`s. The WIT says only one may be outstanding (`host.wit`, `future-invoke-result`), and the host turns the second into a `ProtocolError` ("tool invocation future already has an outstanding get"). Nothing in the signature or docs prevents it — `&mut self`, or a consumed `self`, would.

sdks/ts/packages/golem-ts-sdk/src/bridge/tool.ts
```diff
@@ -1,96 +1,168 @@
...
 }
 
 export function createToolClientTransport(toolName: string): ToolClientTransport {
   let rpc: ToolRpc | undefined;
   return {
-    invokeAndAwait(commandPath, input, stdin) {
+    start(commandPath, input, stdin, withStdout) {
       rpc ??= new ToolRpc(toolName);
-      return rpc.invokeAndAwait([...commandPath], input, stdin);
+      const inputEndpoints = stdin === undefined ? undefined : createStdin();
```
noise64, line 65: > **Endpoints are created before the invoke, and orphaned if it throws**
>
> `createStdin()` (`:53`) and `createStdout()` (`:54`) run before `asyncInvokeAndAwait` (`:55`). If that throws, five handles leak and — because the pump is only started afterwards at `:61` — the stdin writer never selects a terminal, so anything waiting on `tool-stdin-closed` hangs.
>
> The Rust SDK avoids this by encoding the input before `create_stdout()` (`tool_client.rs:517`). Same shape in Scala at `ToolRpcClient.scala:72`.

sdks/ts/packages/golem-ts-sdk/src/tool.ts
```diff
@@ -1604,6 +1670,161 @@ function createToolClientMethod(
...
+      }
+
+      let invocation: ToolClientInvocationResult;
+      try {
+        invocation = transport.start(commandPath, input, stdin, commandBody.stdout !== undefined);
+      } catch (error) {
+        throw mapFailure(error, { phase: 'invoke', body: commandBody, callName });
+      }
+
+      const result = invocation.result
```
noise64: > **`result` is live before the two `throw` paths below, so a failure here is an unhandled rejection**
>
> `invocation.result` is already in flight (`bridge/tool.ts:64` calls `future.get()` eagerly), and this `.then().catch()` chain rethrows. The guards at `:1724` and `:1732` then `throw` synchronously without ever attaching a handler to `result`, so if the invocation fails the rejection escapes.
>
> Broader form: `startedToolInvocation` stores `result` unhandled, so any caller that reads only `.stdout` and never touches `.result`/`.collect()` gets the same. Not covered — `tool-streams.test.ts:73` always awaits `.result` first.

sdks/scala/core/js/src/main/scala/golem/runtime/tool/client/ToolRpcClient.scala
```diff
@@ -54,24 +53,38 @@ private[golem] final class JsToolRpcTransport(rpc: ToolHostApi.RawToolRpc) exten
...
-      case Right((jsInput, jsStdin)) =>
-        try awaitFutureResult(rpc.asyncInvokeAndAwait(commandPath.toJSArray, jsInput, jsStdin))
-        catch {
+        Left(failure)
+      case Right(jsInput) =>
+        try {
+          val stdinEndpoints  = stdin.map(_ => ToolHostApi.createStdin())
+          val stdoutEndpoints = if (stdout) Some(ToolHostApi.createStdout()) else None
+          stdinEndpoints.foreach { case (writer, _, closed) => pump(stdin.get, writer, closed) }
+          val observer = rpc.asyncInvokeAndAwait(
```
noise64, line 72: > **Endpoints created before the invoke are never released when it throws**
>
> `createStdin()`/`createStdout()` run at `:69-70` and the pump starts at `:71`; the `catch` at `:85` exists precisely because `asyncInvokeAndAwait` can throw. But that handler only maps the failure — the stdin/stdout handles are never dropped and the pump keeps running against a writer whose invocation never happened.
>
> (Unlike the TS equivalent at `bridge/tool.ts:53`, the pump ordering here is fine; it's only the cleanup that's missing.)

sdks/scala/model/src/main/scala/golem/tool/ToolStreams.scala
```diff
@@ -25,15 +25,80 @@ import scala.concurrent.Future
...
+    def drain(chunks: Vector[Array[Byte]]): Future[Array[Byte]] =
+      stdout.read().flatMap {
+        case Right(Some(chunk)) => drain(chunks :+ chunk)
+        case Right(None)        => Future.successful(chunks.flatten.toArray)
+        case Left(failure)      => Future.failed(new ToolStreamException(failure))
+      }
+    val terminal = result.map(Right(_): Either[Throwable, Either[ToolError[E], A]]).recover { case t => Left(t) }
+    val output   = drain(Vector.empty).map(Right(_): Either[Throwable, Array[Byte]]).recover { case t => Left(t) }
+    terminal.zip(output).flatMap {
+      case (Left(error), _)              => Future.failed(error)
```
noise64: > **A failed stdout can swallow the real tool error**
>
> `case (Left(error), _)` only fires when the *result future itself* failed. A normal completion carrying a declared tool error is `Right(Left(toolError))`, so if stdout also ended in a failure item, `case (_, Left(error))` at `:97` wins and surfaces `ToolStreamException` — the tool error in `result` is discarded.
>
> Since a guest that fails an invocation often fails its stdout too, this is the common ordering rather than the rare one. Existing coverage only exercises the reverse.

sdks/scala/core/js/src/main/scala/golem/runtime/tool/ToolImplementationRuntime.scala
```diff
@@ -18,28 +18,122 @@ package golem.runtime.tool
...
+  def decodeFailure(value: js.Dynamic): ByteStreamFailure = value.tag.asInstanceOf[String] match {
+    case "cancelled"          => ByteStreamFailure.Cancelled
+    case "abandoned"          => ByteStreamFailure.Abandoned
+    case "resource-exhausted" => ByteStreamFailure.ResourceExhausted
+    case "failed"             => ByteStreamFailure.Failed(value.`val`.asInstanceOf[String])
+  }
+}
+
+/** Adapts the middleware ABI's legacy `stream<u8>` stdin. */
+final class JsMiddlewareInputStream(val underlying: JsWasiInputStream) extends ToolInputStream {
```
noise64: > **Middleware stream wrappers override only `close()`, so any read or write traps**
>
> Both `JsMiddlewareInputStream` and `JsMiddlewareOutputStream` (`:130`) implement `close()` and nothing else, so the `read`/`write` defaults on `ToolInputStream`/`ToolOutputStream` (`ToolStreams.scala:30-51`) apply — and those return `Future.failed(UnsupportedOperationException)`, which `Guest.scala:192-194` turns into a trap.
>
> So a middleware that touches its stdin or the underlying stdout traps rather than failing cleanly. `ToolStreamLifecycleSpec` only exercises `close()`.

sdks/scala/core/js/src/main/scala/golem/runtime/guest/Guest.scala
```diff
@@ -192,9 +193,9 @@ object Guest {
             // the wire-encoded `tool-error`; a failed Future (user code error)
             // propagates as an unhandled rejection so it becomes a WASM trap.
             FutureInterop.toPromise(
-              invoker(commandPath.toList, in, stdin.toOption, scalaPrincipal).map {
+              invoker(commandPath.toList, in, stdin.toOption, stdout.toOption, scalaPrincipal).map {
```
noise64: > **Early rejections skip the invoker, leaving both attachments without a terminal**
>
> `rejectToolError` at `:180` (`InvalidToolName`) and `:189` (`InvalidInput`) return before this call, which is the only place `stdin`/`stdout` are handed on. So the stdout writer is never finished or failed and stdin is never cancelled — the caller's reader sees `Abandoned` from the dropped writer instead of the actual `invalid-tool-name` / `invalid-input`.
>
> `adaptHandler` always terminates stdout on its error paths and is tested; these two paths don't.

golem-worker-executor/src/durable_host/tool/mod.rs
```diff
@@ -457,232 +1233,2301 @@ fn tool_owner<Ctx: WorkerCtx>(
...
+            ctx.owner_execution.tool_operations(),
+            ctx.owner_execution.deferred_tool_admission(),
+        ))
+    })?;
+    let stdin_controller = stdin.as_ref().map(ToolStdinEntry::controller);
+    let stdout_controller = stdout.as_ref().map(ToolStdoutWriterEntry::controller);
+    let stdout_completion_only = stdout
+        .as_ref()
+        .is_some_and(ToolStdoutWriterEntry::completion_only);
+    if !operation.attach(stdin_controller.clone(), stdout_controller.clone()) {
```
noise64, line 2712: > **Cancelling here reaches controllers that no fencing path can see**
>
> When `attach()` fails, the controllers were never registered on the `RegisteredOperation`, so neither `select_owner_failure` nor `close_failed_attachments` can reach them — and both are still `Pending`, so `cancel()` selects a terminal that no reader can observe (same mechanism as the pre-dispatch cancel at `:2517`).
>
> Latent rather than live: `attach()` only fails when the owner is already failing, and `interrupt_current_execution` tears the store down. But if that ordering ever changes, this parks a reader.

golem-worker-executor/src/durable_host/tool/operation.rs
```diff
@@ -0,0 +1,2968 @@
...
+    id: u64,
+    context: Arc<OwnerToolOperationContext>,
+    owner: Arc<OwnerToolOperations>,
+    _winner: watch::Sender<ToolOperationWinner>,
+    lease: Arc<()>,
+}
+
+impl Drop for OwnerToolOperation {
+    fn drop(&mut self) {
+        if Arc::strong_count(&self.lease) != 2 {
```
noise64: > **`Arc::strong_count` is read before the lock, so two concurrent final drops can both bail**
>
> The count is sampled before this handle's own `lease` is decremented, so if the last two clones drop on different threads both observe `3` and both return, and the entry is never removed. Clones live on the sidecar task, the finalize closure, and the `Arc<ToolExecution>` in the guest resource table, so concurrent drops are reachable.
>
> Backstop only — every terminal path calls `settle()`, which removes unconditionally — so I couldn't construct a live leak. Still, `drain_owner_failure_lanes` already makes this decision under the lock; doing the same here would remove the question.

golem-worker-executor/src/durable_host/replay_state/claims.rs
```diff
@@ -241,24 +374,67 @@ impl ReplayState {
...
                     state
-                        .with_tx(async |tx| {
-                            let claimed = tx.claim_start(&owned_claim).await?;
-                            Ok((claimed, tx.blocked_on_completion_delivery))
+                        .with_tx(async |tx| match tx.claim_start(&owned_claim).await {
+                            Ok(claimed) => {
+                                Ok((claimed, tx.blocked_on_completion_delivery, false, false))
+                            }
+                            Err(_) if tx.cursor.is_live() => Ok((None, false, true, false)),
+                            Err(_) if tx.deleted_region_contains_start(&owned_claim).await? => {
```
noise64: > **This guard also swallows payload-download failures**
>
> `claim_start_matching_request` can fail with `runtime("failed to load durable call request payload…")` when an external `HostRequest` payload can't be downloaded. `Err(_)` catches that too, so if any deleted region happens to contain an identity-matching `Start`, a transient blob-store hiccup is reported as `DeletedRegion` — and the caller then continues live and re-appends the call instead of failing retriably.
>
> Matching the specific "no matching Start" kind rather than `Err(_)` would keep the intent and let I/O errors stay retriable. (`:418` is fine — when the cursor is live the scan range is empty, so only the no-match error is reachable.)

wit/deps/golem-tool/host.wit
```diff
@@ -50,34 +50,133 @@ interface host {
...
   resource tool-rpc {
     constructor(tool-name: string);
 
+    /// Waits for the structured terminal. Callers that supplied stdout must
+    /// drive this wait and the already-created reader concurrently.
     invoke-and-await: async func(
       command-path: list<string>,
       input:        typed-schema-value,
-      stdin:        option<stream<u8>>,
+      stdin:        option<own<tool-stdin>>,
```
noise64, line 149: > **A manually created, unfinished stdin deadlocks a synchronous `invoke-and-await`**
>
> The host stages capable stdin to its terminal before admitting the body, so if a caller passes a `tool-stdin` from `create-stdin` and intends to write *after* the call, it blocks: the host waits for the stdin terminal while the guest is blocked inside this function, and there's no cancellation path for the synchronous mode.
>
> The SDKs avoid it by always routing through `create-stdin-from-stream`, but the raw contract permits the shape. Worth saying so here, next to `create-stdin`.

sdks/moonbit/golem_sdk/tool/attachments.mbt
```diff
@@ -0,0 +1,217 @@
...
+fn stdin_byte_stream(
+  source : @async-core.Stream[Byte],
+) -> @async-core.Stream[Result[FixedArray[Byte], @toolHost.ByteStreamFailure]] {
+  @async-core.Stream::produce(
+    async fn(sink) {
+      for ;; {
+        let chunk = source.read(65536) catch {
+          error => {
+            let failure : Result[FixedArray[Byte], @toolHost.ByteStreamFailure] = Err(
+              @toolHost.ByteStreamFailure::Failed(repr(error)),
```
noise64, line 188: > **Every source failure is re-encoded as `failed`, losing the variant**
>
> `Cancelled`, `Abandoned` and `ResourceExhausted` all become `ByteStreamFailure::Failed(repr(error))`, so the caller can't distinguish a cancelled source from an exhausted one. This is the only place across the four SDKs where a `byte-stream-failure` identity is dropped — Rust, TS (`isRpcError`) and Scala (both directions) all map the variants exhaustively.

sdks/moonbit/golem_sdk/tool/attachments.mbt
```diff
@@ -0,0 +1,217 @@
...
+      }
+    },
+    fn() { state.val.open && !state.val.dropped },
+    None,
+  )
+  { state, output, }
+}
+
+///|
+pub fn ProviderStdout::output(self : ProviderStdout) -> @async-core.Sink[Byte] {
```
noise64: > **A MoonBit tool can't fail its stdout**
>
> `ProviderStdout` exposes `output` (`:89`), `finish` (`:94`) and `drop` (`:113`), but nothing wrapping `tool-stdout-writer.fail`. So a MoonBit tool can only ever select `finished` — or `abandoned` by dropping — and can never report `failed(msg)`, `resource-exhausted` or `cancelled` on its output, even though the WIT and the host both support it. Rust exposes this as `OutputStream::fail` (`tool_impl.rs:80-86`).

sdks/moonbit/golem_sdk_tools/lib/tools_emit.mbt
```diff
@@ -2024,25 +2000,38 @@ fn emit_invoker(
...
-          "      let (__golem_stdout, " + source + ") = @tool.new_stdout()\n",
-        )
-      None => ()
+    for arg in args {
+      match arg.placement {
+        Hidden(InputStream) =>
+          buf.write_string(
+            "      let " +
+            arg.source +
+            " = match __golem_stdin {\n        Some(__golem_stream) => __golem_stream\n        None => return Err(@toolCommon.ToolError::InvalidInput(\"required stdin stream is missing\"))\n      }\n",
```
noise64: > **These emitted error returns skip the resource-release helper**
>
> The `None => return Err(...)` emitted here (and at `:2017` for stdout) returns directly, bypassing `@tool.reject_decoded_invocation`, which every other error path in the generated function calls (see the emissions at `:2001` and `:2044`). So canonical input resources are never released and a live `__golem_stdin` is never dropped. The pre-PR stdin branch did call it.

test-components/tool-streaming-ts/caller/src/main.ts
```diff
@@ -0,0 +1,87 @@
+import {
+  client,
+  defineAgent,
+  method,
+  s,
+  toolDefinition,
+} from "@golemcloud/golem-ts-sdk";
+import { z } from "zod/v4";
+
+const streamingTool = toolDefinition("ts-streaming").body((body) =>
```
noise64: > **`typescript_generated_client_streams_live` doesn't use the generated client**
>
> `golem.yaml:20-22` declares `dependencies.tools: golem-it:tool-streaming-ts-provider/ts-streaming`, which is what drives the TS bridge generator. But this re-declares the tool shape by hand (`:10-16`) and calls `client(streamingTool)` at `:49`, so the invocation goes through a locally-built definition rather than the generated bridge.
>
> The Rust counterpart does the opposite — `rust-caller/src/lib.rs:1,14` import `StreamingClient` / `CapableStreamingClient` from the generated crates. So despite the test name, `bridge_gen/typescript/tool.rs` — which this PR reworks substantially — has no end-to-end coverage.
>
> Coverage is also much narrower than the Rust pair: one method, one command, required stdin + required stdout, one mode. No fire-and-forget, cancel, optional-stream combinations, mid-stream failure, or capable tool.

sdks/ts/packages/golem-ts-sdk/src/bridge/tool.ts
vigoo: My agent's analysis says this is not true:

The source ordering is correct, but the claimed five-handle leak and tool-stdin-closed hang are contradicted by endpoint ownership and executor cleanup.

Detailed explanation
- asyncInvokeAndAwait consumes the stdin source and stdout target. Ordinary rejection closes/rejects them; exceptional dispatch calls cleanup_failed_tool_dispatch for both.
- Deleting the stdin source drops its attachment consumer, selects ConsumerCancelled, and makes a closure watcher resolvable. The stdout target is rejected, so its reader receives a terminal.
- The stdin pump and its closed.wait() are created only after asyncInvokeAndAwait returns. If the JS call synchronously throws, no pump is waiting and therefore no closure wait can hang.
- Caller-owned JS wrappers may rely on ordinary resource/GC cleanup after a low-level pre-host trap, but that weaker concern does not establish five leaked handles or the stated hang. Input encoding also occurs before endpoint creation, just as in Rust.

sdks/scala/core/js/src/main/scala/golem/runtime/tool/client/ToolRpcClient.scala
vigoo: This also does not seem to be true:

Actual host invoke failures clean both consumed endpoints, wake the stdin-closed observer, and terminate/cancel the Scala pump. The broad leak claim does not hold.

Detailed explanation
- The executor records both endpoint representations before dispatch. Dispatch errors run cleanup_failed_tool_dispatch; normal rejection also closes stdin and rejects stdout before returning a ready result observer.
- Deleting stdin drops its consumer and selects a cancellation terminal. The Scala pump races each read against tool-stdin-closed.wait(); closure winning cancels the source and terminates the pump.
- Caller-owned wrapper reclamation can still depend on generated binding/GC behavior after termination, and a hypothetical JavaScript exception before host ownership transfer lacks explicit rollback. Neither supports “the pump keeps running against a writer whose invocation never happened.”

noise64 (approved):

PR #3799, opened 2026-08-31
Fix standalone server not catching SIGINT/SIGTERM

### LINKED ISSUE #1603: Golem standalone server doesn't catch SIGINT
A running instance of the Golem standalone server doesn't seem to catch SIGINT termination signal (e.g. pressing `Ctrl+C` in the running terminal window or using `kill -INT <pid>`).

baltasarblanco:
Fixes #1603

## Problem

`golem server run` never installs a signal handler — the run path just awaits the services' `JoinSet` until completion. Running natively that is invisible: with no handler installed, the kernel applies SIGINT's default disposition and the process dies. But when the standalone server runs as **PID 1** in a container (the Docker repro in the issue), the kernel only delivers a signal to a PID namespace's init process if init has installed a handler for it — default dispositions are never applied ([pid_namespaces(7)](https://man7.org/linux/man-pages/man7/pid_namespaces.7.html)). So SIGINT (Ctrl+C, `kill -INT 1`) and SIGTERM (`docker stop`) are silently discarded, which matches every observation in the issue thread.

## Fix

Wrap the `join_next()` loop in `ServerSubcommand::Run` in a `tokio::select!` against a shutdown signal (SIGINT + SIGTERM via `tokio::signal::unix` on Unix, `ctrl_c()` elsewhere). On signal: log, `join_set.shutdown().await`, and return. Aborting the tasks drops the router task, which owns the worker executor's `RunDetails`; its `Drop` cancels the graph-wide shutdown token and stops the epoch thread — the same mechanism already documented in `cli/golem/src/router.rs` ("Dropping its RunDetails cancels the shutdown token that drives background services").

One file changed, no new dependencies (`tokio` already has the `signal` feature enabled workspace-wide).

## Verification

Reproduced the PID 1 condition without Docker, since the bug's trigger is being the init of a PID namespace: `sudo unshare --fork --pid --mount-proc ./target/debug/golem server run -vv`.

**Before (main):**
- Normal process: `kill -INT <pid>` kills it instantly (kernel default action, no shutdown of any kind).
- As PID 1: the same `kill -INT` is silently ignored; the server keeps running and can only be stopped with SIGKILL.

**After (this branch):**
- Both cases shut down cleanly on SIGINT and SIGTERM, logging `Received shutdown signal, stopping Golem server`, with exit code 0.
- The shutdown token propagates to the background services; depending on timing you can see the scheduler log `Shutdown requested, stopping scheduler background loop` before the process exits.

`cargo build -p golem` and `cargo make fix` pass.

Note: on unpatched versions, `docker run --init` works around the issue, since tini becomes PID 1 and forwards signals.


baltasarblanco:

cli/golem/src/command_handler.rs
```diff
@@ -56,11 +56,27 @@ impl CommandHandlerHooks for ServerCommandHandler {
                     .await
                     .map_err(|err| map_local_server_startup_error(err, &data_dir))?;
 
-                while let Some(res) = join_set.join_next().await {
-                    res??;
-                }
+                let run_result = tokio::select! {
```
vigoo, line 74: We should probably race the shutdown signal with `launch_golem_services` too so the service startup can be also interrupted

baltasarblanco: You're right. I covered it by sharing the same signal future across both phases, so both are covered and no signal gets lost in between.

One more design note I'd rather mention: interrupting mid-startup drops whatever was built up to that point.

PR #3805, opened 2026-09-01
tool-invoke review nits

noise64:
- mechanical review nits on top of #3787, one commit each so any can be dropped
- restores the five sdk-path helper tests dropped from `helpers.rs`, and deletes the unreachable stdout arms in `decode_client_result` (the `(Some, true)` arm called the two-arg `agentic::decode_result_with_stdout` with one argument)
- skips empty stdout/stdin chunks in the ts, scala and moonbit guest writers, which the wit forbids and the client readers reject
- releases resources on paths that returned early without them: moonbit `ToolClient::start` on encode failure, moonbit `byte_stream` dropped unstarted, scala `start` when declared stdout is missing
- publishes staged stdout on the cancelled no-body tool terminal, and notifies waiters when an unaccepted tool operation is removed, both mirroring the adjacent path
- makes the rust test caller fail on a stdout failure item instead of skipping it, so 31 assertions stop passing regardless of how stdout closed
- verified locally: `cargo check -p golem-worker-executor`, `cargo test -p golem-rust-macro` (121 pass), `cargo fmt --check`; ts/scala/moonbit rely on CI
- discussion-level findings are separate review comments rather than commits here

### DIFF
--- a/golem-schema/wit/deps/golem-tool/common.wit
+++ b/golem-schema/wit/deps/golem-tool/common.wit
@@ -375,7 +375,7 @@ interface common {
     body:  string,
   }
 
-  // Invocation contract — shared between guest and host
+  // Invocation contract — shared between guest and host.
   variant tool-error {
     invalid-tool-name(string),
     invalid-command-path(list<string>),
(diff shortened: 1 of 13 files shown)


vigoo:
These have been cherry-picked to the tool invocation PR

PR #3806, opened 2026-09-01
Fix MoonBit multi-component embed paths

vigoo:
## Summary

- run every MoonBit `wasm-tools component embed` step from the application root
- cover ordinary, tool-middleware, and combined components in debug and release presets
- add an embedded-template regression test for all six commands

## Why

Published MoonBit applications resolve the SDK WIT directory through the app-relative `.mooncakes/golemcloud/golem_sdk/wit` path. External commands without an explicit `dir` run from each component directory, so multi-component applications looked for `.mooncakes` beneath `moonbit-main`, `moonbit-second`, and similar directories.

Setting the embed commands' working directory to `{{ appRootDir }}` makes the published relative path resolve correctly. Local SDK overrides remain valid because they are absolute, single-component behavior is unchanged, and the subsequent `component new` commands already use app-rooted input and output paths.

The issue was reproduced with the real official MoonBit compiler: both MoonBit packages compiled successfully, after which `wasm-tools component embed` failed to find the root `.mooncakes` path from a component directory. Running that same embed command from the application root succeeded.

## Verification

- `cargo test -p golem-cli moonbit_embed_commands_run_from_app_root -- --report-time`
- `cargo fmt -p golem-cli -- --check`
- `cargo clippy -p golem-cli --all-targets --no-deps -- -D warnings`
- `cargo check -p golem-cli --all-targets`

Resolves GOL-176

### DIFF
--- a/cli/golem-cli/src/app/template/repo.rs
+++ b/cli/golem-cli/src/app/template/repo.rs
@@ -329,7 +329,8 @@ impl AppTemplateRepo {
 
 #[cfg(test)]
 mod tests {
-    use super::AppTemplateRepo;
+    use super::{AppTemplateRepo, TEMPLATES_DIR};
+    use crate::model::app_raw::{Application, BuildCommand};
     use crate::model::language::GuestLanguage;
     use std::fs as stdfs;
     use std::path::{Path, PathBuf};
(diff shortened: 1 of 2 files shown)


noise64 (approved):
checked locally both with flat and multi-component layouts 👍

vigoo:
thanks!

PR #3808, opened 2026-09-01
MoonBit related fixes

noise64:
- fixes the MoonBit build artifact path for nested component dirs: `moonbitBuildPackagePath` was built as `<dir>/<dir>.wasm`, but `moon build` names the artifact after the last segment of the package dir, so `dir: nested/second` looked for `.../build/nested/second/nested/second.wasm` while moon wrote `nested/second/second.wasm`; only reachable by hand-writing a nested `dir:`, since for single-segment dirs the last segment is the whole path
- checked the other language templates for the same problem: Rust, Scala and TypeScript name artifacts after the component rather than the directory, and all build clean with a nested `dir:`
- makes `golem_sdk_tools` skip byte-identical writes of the generated `.mbt` files, which previously got a fresh mtime on every run and, since `moon` is mtime-driven, forced a re-link on every build
- declares `sources`/`targets` on the MoonBit `wasm-tools` embed/new steps, the only language template that had none, so they no longer rewrite the component wasm every build and re-trigger `add-metadata` and `gen-bridge`; also makes `golem clean` remove the intermediate `.embed.wasm`
- keeps `reexports`, `agents` and `moon build` always-running on purpose: the first two read and write the same `moon.pkg`, and `moon build`'s command text is identical across components, so they would share one task-result marker
- adds tests for `moonbitBuildPackagePath` (root/single/nested), codegen content-stability and stale-file removal, a guard that every MoonBit `wasm-tools` step declares both `sources` and `targets`, and an end-to-end incremental build check that runs without `--force-build`


vigoo (changes requested):
`scripts/release/main.mbt` need to be updated to handle the new paths in the example yaml
(if we still use it for switching between local vs published, i'm not sure)

vigoo (approved):

PR #3816, opened 2026-09-02
Expose entity-aware owner oplog attribution

vigoo:
## Summary

- attribute every public owner-oplog entry to either the agent or an entity invocation, including the entity kind, name, start index, call mode, safe operation metadata, and root-to-parent ancestry
- persist and validate initiation-time entity anchors for durable asynchronous work, retries, card workflows, streams, and sessions while preserving physical oplog order and page/search independence
- expose attribution through protobuf, REST, and the existing CLI text/structured output, with regenerated schemas and documentation
- render safe secret audit metadata without plaintext and keep tool stdout as a live, non-durable attachment

This PR is stacked on #3787 while the tool invocation implementation is still open. Once #3787 merges, this branch can be retargeted to `main` without changing the GOL-38 diff.

## Validation

- `cargo clippy` for all affected packages and targets with `-D warnings`
- `cargo test -p golem-worker-executor --lib` — 1,369 passed
- focused owner-oplog attribution, debugging-service, and card recovery integration tests
- CLI renderer, oplog, structured-output, and schema suites
- cross-package `cargo check`
- OpenAPI drift, docs lint/format, Rust format, JSON schema, and diff checks

Resolves GOL-38


golem-common/src/model/oplog/mod.rs
```diff
@@ -356,7 +490,32 @@ impl OplogScopeProjection {
...
+            | OplogEntry::CardTransferred { .. }
+            | OplogEntry::CardRevokedCascade { .. }
+            | OplogEntry::CardTransferConfirmed { .. }
+            | OplogEntry::StreamRegistered { .. }
+            | OplogEntry::StreamItems { .. }
+            | OplogEntry::StreamEnd { .. }
+            | OplogEntry::StreamCancel { .. }
+            | OplogEntry::StreamSession { .. } => entry
+                .entity_parent_start_index()
+                .is_some_and(|parent| self.starts.contains(&parent)),
```
noise64, line 518: > These entries are now included in the scope through `entity_parent_start_index`, but `scope_entry_owner` still returns no owner for most of the newly included variants (and continues to use `retry_from` for `Error`). When a nested entity body is still active, `unconsumed_scope_head` can therefore treat that body's anchored entry as an unowned descendant and report permanent structural divergence.
> Could we make `scope_entry_owner` prefer `entry.entity_parent_start_index()` and add replay coverage for at least an anchored `NoOp` and `Error` while their owning nested entity body remains active?

golem-worker-executor/src/model/public_oplog/mod.rs
```diff
@@ -65,16 +72,451 @@ use golem_common::schema::{
...
+        entity,
+        start_index,
+        call_mode,
+        operation,
+    }
+}
+
 /// Projects one entity invocation's transitive durable-call tree from its owner's raw oplog.
 /// Entity histories remain owner records; this is a filtered view, not a child oplog or status.
 pub fn project_entity_oplog_entries(
```
noise64: > `project_entity_oplog_entries` filters `Start` records using the physical `parent_start_index`, while public attribution prefers `observational_owner`. If this helper is intended to back an entity-history view, membership and the displayed attribution can therefore describe different trees.
> There are no in-tree callers today, so this is non-blocking, but could we either align the projection with public attribution or document that this helper deliberately projects the physical replay tree?

noise64 (approved):

PR #3819, opened 2026-09-03
Redesign snapshot restoration for agents

vigoo:
## Summary

- Execute `load-snapshot` as a specially supported read-only lifecycle operation using the existing `read_only_method: Some("load-snapshot")` path, with no oplog writes and with mutating host operations plus outgoing HTTP/RPC rejected before they take effect.
- Replace mutation-based snapshot loaders with restoration factories across the Rust, TypeScript, Scala, and MoonBit SDKs. Restoration skips normal initialization, receives identity/principal/config context as supported by each SDK, and installs only a complete successfully restored state or instance.
- Generate automatic loading only where a JSON-deserialization contract exists, including requiring an explicit TypeScript state schema, while preserving the existing snapshot envelope, MIME, and WIT contracts.
- Make failed automatic loads discard partial state and recreate/replay without trying older automatic snapshots; preserve the previous component for failed manual updates.
- Migrate in-tree consumers and templates, add executor and SDK regression coverage, and update source and generated documentation.

## Validation

- `cargo fmt -p golem-worker-executor -- --check`
- `cargo clippy -p golem-worker-executor --all-targets --no-deps -- -D warnings`
- `cargo check -p golem-worker-executor --all-targets`
- Targeted worker-executor snapshot/update acceptance tests: 4 passed
- Rust SDK tests: 132 passed, 4 ignored
- TypeScript SDK tests: 17 passed
- Scala formatting and model JVM/JS, macro, and core tests: 1,346 passed
- MoonBit SDK tests: 259 passed; SDK tools tests: 304 passed
- Migrated snapshot fixtures built successfully
- Docs lint, formatting, typecheck, links, version checks, and `cargo make check-docs-skills`

Resolves GOL-487


sdks/moonbit/golem_sdk_tools/lib/agents.mbt
```diff
@@ -1201,6 +1260,13 @@ pub fn parse_agents(
   // Validate
   let agents : Array[AgentInfo] = []
   for name, description in agent_names {
+    let has_snapshot_saver = agent_snapshot_savers.get(name).unwrap_or(false)
+    let has_snapshot_loader = agent_snapshot_loaders.get(name).unwrap_or(false)
+    if has_snapshot_saver != has_snapshot_loader {
+      raise AgentParseError(
+        "Agent '\{name}' must define both an @agents.Snapshottable save_snapshot implementation and a static load_snapshot factory",
+      )
+    }
```
noise64, line 1269: > Could we reject enabled snapshotting here unless the agent has either both JSON derives or the complete custom saver/loader pair? With neither mechanism, this validation passes, `emit_load_snapshot` registers `load_snapshot: None`, and the save stub emits an empty payload. An agent using `snapshotting="every_n(5)"` can therefore compile as snapshot-enabled but can never restore its snapshots. A generator error would keep MoonBit aligned with the other SDKs.

noise64 (approved):

PR #3820, opened 2026-09-03
Bound invocation result state in worker status

vigoo:
## Summary

- keep the complete invocation-result map in `AgentStatusRecord` for small agents, then promote one-way to a bounded recent exact map plus a fixed-size Bloom filter only when the exact map would use more memory
- maintain the complete idempotency-key-to-oplog-index mapping in a dedicated physical KV namespace for every durable agent, with bounded chunked oplog catch-up and branch invalidation across reverts
- resolve invocation results through hydrated payloads, adaptive status membership, the physical index, and a bounded oplog fallback while serializing duplicate admission through the status actor
- bound the hydrated invocation-result payload cache and add configuration, metrics, storage coverage, and an idempotency-key lookup benchmark
- upgrade to `desert_rust` 0.1.11 and derive codecs directly for the `im::OrdMap`/`im::Vector`-backed status types

## Design

Small agents preserve the original exact in-memory lookup behavior and do not allocate or serialize a Bloom filter. When the estimated exact-map footprint exceeds the 128 KiB Bloom filter plus the bounded recent maps, status folding promotes the membership representation one-way. Recent results and definitive Bloom-filter misses stay in memory; Bloom positives consult the complete physical index.

The physical index is maintained for every durable agent during status persistence, so it is ready before promotion. It records coverage and revert generation, and both hits and misses are authoritative only when physical coverage overlaps the recent in-memory window. This prevents stale results from a lagging index and obsolete results from a reverted branch. Cold reconstruction and old uncached lookups use bounded oplog chunks rather than retaining an unbounded intermediate map.

## Verification

- `cargo test -p golem-worker-executor --lib worker::status::test -- --report-time`
- `cargo test -p golem-worker-executor --lib services::worker::tests -- --report-time`
- `cargo test -p golem-worker-executor --lib invocation_result_cache -- --report-time`
- `cargo test -p golem-common invocation_result_membership -- --report-time`
- `cargo test -p golem-worker-executor --test key_value_storage agent_invocation_result_index -- --report-time`
- `cargo test -p golem-worker-executor --test integration -- api::invoking_with_same_idempotency_key_is_idempotent_after_restart rpc::durable_streaming_output_recovers_after_executor_restart --report-time`
- `cargo check -p golem-worker-executor --lib`
- `cargo check -p integration-tests --bin benchmarks`
- `cargo make generate-configs`

The new benchmark is compiled but was not run against benchmark infrastructure.

Resolves GOL-175


golem-common/src/model/mod.rs
```diff
@@ -692,6 +692,203 @@ impl SafeDisplay for RetryConfig {
...
+/// without consulting the physical invocation-result index. False positives are allowed; false
+/// negatives are not.
+#[derive(Clone, Debug, PartialEq, Eq, BinaryCodec)]
+pub struct InvocationResultBloom {
+    words: Vector<u64>,
+    bit_count: usize,
+    hash_count: u8,
+}
+
+impl InvocationResultBloom {
```
noise64, line 710: Is using a custom bloom filter justified? (Can be, just interested)

golem-worker-executor/src/worker/mod.rs
```diff
@@ -4349,11 +4382,11 @@ impl<Ctx: WorkerCtx> Worker<Ctx> {
         let status = self.get_last_known_status().await;
         for idempotency_key in unfinished {
             let mut invocation_result = {
-                self.invocation_results
+                self.hydrated_invocation_results
```
noise64: > hydrated_invocation_results is initialized empty for every reconstructed worker, so this lookup always misses during startup. If a crash happened after committing AgentInvocationFinished but before recording the stream
  > session’s Finished entry, recovery silently skips that persisted result and leaves the session unfinished indefinitely; the periodic reconciler does not retry this method.

golem-worker-executor/src/worker/status.rs
```diff
@@ -195,17 +200,143 @@ where
...
+        let count = (last_oplog_index.as_u64() - first.as_u64() + 1).min(chunk_size);
+        let entries = this
+            .oplog_service()
+            .read_exact(owned_agent_id, agent_mode, first, count)
+            .await;
+        if entries.is_empty() {
+            return None;
+        }
+        deleted_regions = calculate_deleted_regions(deleted_regions, &entries);
+        region_entries.extend(entries.iter().filter_map(|(index, entry)| {
```
noise64, line 285: Is it okay, the region etries are not bounded? Should we have a solution for that too, or is that acceptable for now?

golem-common/src/model/mod.rs
```diff
@@ -692,6 +692,203 @@ impl SafeDisplay for RetryConfig {
     }
 }
 
+pub const DEFAULT_RECENT_INVOCATION_RESULTS_CAPACITY: usize = 1024;
+pub const DEFAULT_INVOCATION_RESULT_BLOOM_BITS: usize = 1 << 16;
+pub const DEFAULT_INVOCATION_RESULT_BLOOM_HASHES: u8 = 4;
+
```
noise64, line 699: > GOL-175 promises that new invocations normally require no KV request and asks for validation through N/10N/100N histories. With the current defaults (N = 1024, 65,536 Bloom bits, 4 hashes), the expected false-positive rate at  
  > 100N is approximately 99.2%, so nearly every genuinely new key takes the physical-index KV path. Since the benchmark was not run, the PR has not demonstrated the promised behavior. Please tune or revise the filter design based 
  > on the required measurements before resolving GOL-175.

golem-worker-executor/src/worker/status.rs
vigoo: Let's not touch that in this PR. I think it's much harder to solve

noise64 (approved):

PR #3828, opened 2026-09-03
Refactor drop method in GrantedSlot struct

shailendra-codes:
Refactor drop implementation to use async for granting next slot.


shailendra-codes:
I have read the CLA Document and I hereby sign the CLA Document and I hereby sign the CLA

PR #3832, opened 2026-09-04
Avoid cloning durable host call responses

vigoo:
## Summary

- move typed durable host-call responses into `HostResponse` instead of cloning them before persistence
- return the same owned response through the `HostPayloadPair` self-roundtrip invariant in both direct and accessor completion paths
- remove the unnecessary `Clone` bound while keeping replay conversion fallible and classified as `unexpected_oplog_entry`

## Testing

- `cargo check -p golem-common -p golem-worker-executor --all-targets`
- `cargo test -p golem-worker-executor --lib -- access_terminal_end_is_appended_before_cleanup_and_permit_release --report-time`
- `cargo test -p golem-common --lib -- p3_http_client_consume_body_chunk_host_payload_pairs_roundtrip --report-time`

Resolves GOL-481


Aditya1404Sal (approved):

golem-worker-executor/src/durable_host/concurrent/call.rs
```diff
@@ -4507,9 +4507,11 @@ async fn prepare_end_entry<Pair: HostPayloadPair>(
     // cache so this changes neither the oplog representation nor same-process payload-read behavior.
     let response = Box::new(response);
     let prepared = tokio::task::spawn_blocking(move || {
-        let host_response: HostResponse = response.as_ref().clone().into();
+        let host_response: HostResponse = (*response).into();
         let bytes = golem_common::serialization::serialize(&host_response)?;
-        Ok::<_, String>(Box::new((response, bytes, Arc::new(host_response))))
+        let cached = Arc::new(host_response.clone());
```
noise64: Maybe I misunderstand this, but aren't we still cloning with this the full response? Isn't the issue's goal was to eliminate this?

noise64 (approved):

PR #3833, opened 2026-09-04
Fix indeterminate oplog writes and batch persistence

vigoo:
## Summary

- classify indexed-storage write failures so primary oplog retries distinguish safe transient failures, indeterminate outcomes, conflicts, and permanent errors
- reconcile a possibly committed append by reading back the expected range and accepting only an exact ID and serialized-byte match
- serialize oplog batches once and reuse the same `Arc<[(u64, Bytes)]>` across retries and reconciliation
- remove PostgreSQL per-row namespace/key allocations and add transactional, zero-explicit-copy SQLite batch writes with single-lookup MultiSqlite routing
- cover initial and batched appends, conflict safety, partial/different read-back, backend error classification, atomic SQLite batches, and per-agent MultiSqlite routing

## Verification

- `cargo fmt -p golem-worker-executor -- --check`
- `cargo check -p golem-worker-executor --all-targets`
- `cargo clippy -p golem-worker-executor --all-targets --no-deps -- -D warnings`
- 19 focused `golem-worker-executor` tests covering GOL-535 and GOL-480
- `git diff --check`

Resolves GOL-535

Resolves GOL-480


golem-worker-executor/src/storage/indexed/redis.rs
```diff
@@ -111,6 +113,20 @@ impl RedisIndexedStorage {
         }
     }
 
+    fn classify_append_error(error: RedisError, primary_oplog_insert: bool) -> IndexedStorageError {
+        if primary_oplog_insert
+            && error
+                .details()
+                .contains("ID specified in XADD is equal or smaller than")
+        {
+            IndexedStorageError::Conflict(error.to_string())
```
noise64, line 123: > Fred automatically retries in-flight commands after reconnect (RedisPool::configured retains the default max_command_attempts = 3). If Redis applies this XADD but the response is lost, Fred can replay it and expose only the    
  > replay’s duplicate-ID error here. The outer oplog logic therefore sees Conflict without a preceding Indeterminate and panics as if there were a concurrent writer. This leaves the Redis form of GOL-535 unresolved. Please prevent
  > Fred from replaying primary-oplog appends, or reconcile the committed write inside the Redis adapter, with a committed-then-disconnected regression test.

golem-worker-executor/src/services/oplog/primary.rs
```diff
@@ -91,6 +92,163 @@ where
...
+            .await
+        {
+            Ok(()) => return,
+            Err(error) => error,
+        };
+
+        let retryable = match &error {
+            IndexedStorageError::Indeterminate(_) => {
+                write_may_have_committed = true;
+                true
```
noise64, line 193: > Indeterminate is treated as retryable here, so the same non-idempotent append is submitted again before its outcome is established. For PostgreSQL, a committed first attempt therefore produces the unique violation that GOL-535 
  > explicitly says to avoid; the tests currently require this two-write sequence. Please determine the previous attempt’s outcome before resubmitting it—for example, by reconciling the exact stored batch or making the backend     
  > append idempotent. If the batch is absent it can be retried; an exact match is success; partial or different contents must remain a fail-stop condition.   

I'm not sure if reconciliation worth it (as in if it is cheap enough), but we might want to check the above in general.

golem-worker-executor/src/services/oplog/primary.rs
```diff
@@ -91,6 +92,163 @@ where
...
+    expected: &[(u64, Bytes)],
+) -> bool {
+    let first_id = expected.first().expect("non-empty oplog batch").0;
+    let last_id = expected.last().expect("non-empty oplog batch").0;
+    retry_storage_op(retry_config, "append_reconcile", key, || {
+        let namespace = namespace.clone();
+        async move {
+            indexed_storage
+                .with_entity("oplog", "append_reconcile", "entry")
+                .read_raw(namespace, key, first_id, last_id)
```
noise64, line 119: > This reconciliation read is executed through retry_storage_op, but RedisIndexedStorage::read currently maps every XRANGE failure—including transient connection failures—to IndexedStorageError::Other. A temporary Redis failure  
  > while reconciling an indeterminate append therefore immediately panics instead of retrying. Since the read is idempotent, please classify retryable Redis errors as Transient, or handle them explicitly here, and add coverage for
  > an I/O failure during reconciliation.

noise64 (approved):

PR #3836, opened 2026-09-05
Prevent external capability forgery and support guest RPC bridges

vigoo:
Rersolves GOL-181

## Summary

Prevent external callers from forging Secret, QuotaToken, and PermissionCard schema values while preserving their use in trusted persistence, replay, and agent-to-agent RPC, including generated guest bridges.

### Separate external JSON from trusted capability transport

- Add shared traversal of capability-bearing schema types and value trees, with paths for actionable rejection messages.
- Introduce validated external schema-value DTOs and a dedicated untrusted JSON decoder. External REST, custom HTTP, MCP, CLI, configuration, and secret-value inputs cannot admit authoritative capability snapshots. Absent optional values and unselected branches remain valid when no capability value is supplied.
- Keep canonical SchemaValue serialization and internal protobuf/WIT transport lossless for trusted storage, replay, and RPC. Capability type metadata remains discoverable; only authoritative values are restricted.
- Redact capability snapshots on external output paths, including metadata, configuration, operation logs, and CLI output. Generate directional input/output schemas: external capability inputs are unsatisfiable, while outputs describe redacted values.
- Update API DTO consumers and regenerate OpenAPI definitions together with the boundary change.

### Support capabilities in generated guest bridges

- Map all three capability types to native guest SDK handles in Rust, TypeScript, Scala, and MoonBit, including nested named types and multimodal method IO.
- Reject capability-bearing generated external APIs, constructors, and local configuration before generation has side effects. Host-injected fields and secret-sourced configuration remain supported.
- Avoid unsupported data-like derives on capability-containing generated composites. Allow generated Scala clients to access the package-restricted cancellation-token factory without changing cancellation behavior.

### Preserve affine ownership

- Use the guest SDK's safe capability codecs rather than ordinary JSON conversion.
- Harden TypeScript holders with private ownership tracking, exact-carrier adoption, and take-once transfer. Reject forged holders, aliases, duplicate adoption, and consumed handles.
- Make failed composite conversions roll back their own partial adoptions without invalidating values returned by successful reentrant conversions.

## Verification

Implementation validation before rebasing:

- TypeScript SDK build/typecheck and full suite: **719 passed, 20 skipped**.
- Generated guest compiler checks for Rust, TypeScript, Scala, and MoonBit, including nested capabilities.
- Real agent-to-agent round trip of all three capability types, usability after return, and duplicate/consumed-handle rejection: **1 passed**.
- Scope-card cleanup, revocation, replay, and authorization regressions: **4 passed**.
- Shared-schema/rendering tests: **110 passed**.
- CLI structured-output tests: **22 passed**.
- Focused HTTP/MCP external boundary tests: **5 passed**.
- Affected-package and consumer compile checks; scoped formatting and whitespace checks.
- OpenAPI documentation generation, formatting, and drift checks passed with no MDX changes.

Oracle reviewed the implementation milestones and corrective changes. The bounded bug-finder review finished with no open findings.

Post-rebase validation against the latest fetched `origin/main`:

- `CARGO_INCREMENTAL=0 cargo check -p golem-cli -p integration-tests --all-targets` passed.
- `CARGO_INCREMENTAL=0 cargo clippy -p golem-cli -p integration-tests --all-targets --no-deps -- -D warnings` passed.
- `cargo fmt -p golem-cli -p integration-tests --check` and `git diff origin/main --check` passed.
- Resolved overlapping test imports/helpers while retaining upstream tool-release functionality. The implementation test runs above were not repeated after the rebase.

## Validation limitations

- Repository-wide CI remains the broad regression check; the complete integration suites were not run locally.
- The aggregate docs wrapper encountered an existing `concurrently` lookup issue. Direct documentation generation, formatting, and drift checks passed.
- An unrelated pre-existing formatting difference in the Rust test fixture's `config.rs` was left untouched; the modified fixture source passes its scoped formatting check.


golem-worker-service/src/service/worker/service.rs
```diff
@@ -2682,6 +2682,11 @@ impl WorkerService {
                     .cloned()
                     .unwrap_or_else(|| SchemaType::tuple(Vec::new()));
                 let typed_output = TypedSchemaValue::new(output_graph, output_value);
+                let typed_output = typed_output.try_into().map_err(|error| {
+                    WorkerServiceError::Internal(format!(
+                        "Agent method result cannot cross the external JSON boundary: {error}"
+                    ))
+                })?;
```
noise64, line 2717: > Could we reject capability-bearing output schemas before dispatching an external REST invocation? This conversion happens only after the awaited method has completed, so a guest-RPC method that legitimately returns a secret, quota token, or permission card can commit durable side effects and then turn the external response into an internal error. A hand-written caller that omitted an idempotency key also never receives the generated key, making a retry unsafe. Moving the policy check before dispatch would fail without executing the method; if external redaction is intended instead, the result needs an explicit conversion matching that contract.

noise64 (approved):

PR #3837, opened 2026-09-06
Restore invocation performance and optimize durable streaming (GOL-514)

vigoo:
## GOL-514: invocation and streaming RPC performance

The Oracle-approved hot-path plan was implemented and measured at [dd3b838b9](https://github.com/golemcloud/golem/commit/dd3b838b9a4111746683b51a464946f5b65cdc61), after merging main at [9b18a58a4](https://github.com/golemcloud/golem/commit/9b18a58a4). That revision's [CI](https://github.com/golemcloud/golem/actions/runs/34297197133) is green on attempt 2. The first approved noise64 fixes at [ac02c310f](https://github.com/golemcloud/golem/commit/ac02c310f301f5ff2769ac0e085ad867b58c41c5) passed [CI](https://github.com/golemcloud/golem/actions/runs/34321142969) on the first attempt. The latest metadata follow-up is [368ed1ab4](https://github.com/golemcloud/golem/commit/368ed1ab44ba2e0544ef2fb48aea5b7f201189e6); [its CI](https://github.com/golemcloud/golem/actions/runs/34335507326) is green on attempt 3: 52 passed and seven normally skipped jobs. Docs passed. A fresh fetch confirms this branch contains latest main; there are no merge conflicts. Merging still requires the repository's mandatory approving review (the current review is COMMENTED). No merge or release was performed.

### Noise64 review follow-up

- Fixed cold finished-session validation: hydrate the session metadata before deciding whether new events are allowed. Regression coverage includes cold load, eviction, unrelated open sessions, and no storage reads for warm checks.
- Fixed CLI runtime scope: parse once before creating a runtime; only a successfully parsed `server run` gets executor-sized stacks. Existing parsing, environment overrides, help/errors and dispatch are retained; ordinary commands and `server clean` use the ordinary runtime.
- Local verification: 14 focused executor tests, 283 broader executor tests, real-parser runtime-selection test, scoped all-target Clippy, formatting, and focused bug-finder review passed.
- Approved and implemented: derive input/output stream classification once into immutable component-revision metadata after registry schema validation. Persist and propagate it through binary/gRPC/REST metadata; use the exact executing/replayed/remote revision for admission, routing and dispatch. Input/output value validation remains intact. Input-only streaming does not select streaming-output dispatch. Missing classification never defaults to scalar. Oracle reviewed the implementation and the output-dispatch follow-up without blockers.
- Metadata follow-up checks: 57 common metadata/schema tests, 36 worker-service tests, 304 executor tests and four stream/tool/crash integration tests passed. OpenAPI/docs regenerated and golem-client rebuilt. Focused bug-finder clean; scoped all-target Clippy with `-D warnings`, `cargo fmt --check` and `git diff --check` passed.
- Initial CI failures retained for transparency: [managed-xfs-tests](https://github.com/golemcloud/golem/actions/runs/34335507326/job/102417427577) rejected a success summary interrupted by concurrent DEBUG output; [worker-tests-group1](https://github.com/golemcloud/golem/actions/runs/34335507326/job/102417427650) had the GOL-551-tracked long-poll deletion timeout. Both tests passed on attempt 2, which instead failed the second-invocation long-poll interruption test. That test passed ten consecutive local runs using the exact CI-built HTTP fixture, and the subsequent group-1 rerun passed on attempt 3. No source or assertion changes were made to obtain green checks; the intermittent lifecycle and XFS-summary issues are not claimed fixed.
- Deletion concern reevaluated, no production fix: normal Worker deletion fences admission and retains the deleting Worker until storage cleanup finishes; ordinary concurrent recreation overlap is not established. Process loss after oplog deletion but before separate index clearing can leave stale agent-ID-keyed index rows. A real-oplog/in-memory-storage diagnostic reproduces raw stale lookup after recreation with a new fingerprint. This is not a public-Worker or process-kill reproduction, and does not establish old-session execution or data exposure: the public session lookup bypasses history when current status has none. Fingerprint-qualified namespaces remain a valid candidate awaiting approval. Diagnostic patch/log: `deletion-incarnation-diagnostic.{patch,log}` alongside the report.

The measurements below belong to dd3b838b9, before these review fixes; no new performance numbers are claimed for the follow-up revisions.

### Why performance regressed

The initial scalar regression came from delayed ACK/Nagle interactions, repeated temporary stack allocation, serialization offload for flat response types, and streaming lifecycle work on scalar invocations. This PR enables NODELAY on custom tonic listeners, reuses 32 MiB runtime stacks (retaining the guard for smaller embedding runtimes), encodes audited flat responses inline, and bypasses session work for proven scalar invocations.

Warm streaming RPC had a separate pre-existing history-scan problem: source-unavailable checks rebuilt the consumer's full journal index on each check, repeatedly reading/decompressing archived oplog blocks. Those blocks—not application chunks on the RPC wire—were the zstd workload. The bounded-index implementation then added repeated immutable attachment-evidence reads and single-batch storage-backed segment responses. Old-vs-PR producer-write spans improved even while total completion worsened, so producer writes alone did not explain the original 17.5% regression.

The final implementation uses a cached consumer head, validated immutable attachment evidence, indexed producer identity, producer-wide committed retention (32 MiB/4,096 batches), and bounded available responses (256 events/24 MiB target, singleton progress). Cold packed continuation uses bounded locator windows; deletion retains the complete exact prefix. Mutable authority is revalidated, including after long polls; replay still uses durable authority. Exact journal-lag telemetry is sampled at most every 100 ms per reader plus creation/new terminal, with failed attempts throttled. Stream-free writes skip redundant nested-handle reads. PostgreSQL singleton appends use existing autocommit instead of BEGIN/COMMIT round trips; multi-row atomicity and indeterminate-write reconciliation are unchanged.

### Real-version release measurements

Same orb, identical fixtures and runner, three iterations, sequential release builds/runs, no overlapping benchmark/test/build workloads. Actual old revision [d667feedd](https://github.com/golemcloud/golem/commit/d667feeddaff55b154ebddd75ab80f23952a17a2), measured final executor dd3b838b9. No synthetic comparator or disabled-index switch remains. Other final services are retained unchanged from the six-point build; hashes and provenance are in the report artifacts.

Median milliseconds, 512 × 4 KiB chunks:

| Case | Concurrency | Real old completion | Final completion | Real old guest first chunk | Final guest first chunk |
|---|---:|---:|---:|---:|---:|
| Tool stdout | 1 | 66.321 | 33.415 | 5.623 | 4.183 |
| Tool stdout | 10 | 1,223.169 | 25.639 | 79.177 | 5.586 |
| Ordinary generated streaming RPC | 1 | 55,662.862 | 1,136.443 | 1,427.278 | 393.887 |
| Ordinary generated streaming RPC | 10 | Earlier run failed at 600 s | 1,496.158 | Unavailable | 664.197 |

Old RPC c10 failed with repeated `InvalidAttachmentState` recovery warnings; this is not a numeric median/speedup. It was not rerun in the completing historical subset. Final RPC c1 is 49× faster than fresh real old, and 55.5× faster than the earlier PR's 63,031 ms. C10 improves from that PR checkpoint's 111,004 ms to 1,496 ms. First-chunk latency improves over old but not over every intermediate checkpoint (earlier PR: 251/416 ms). Three iterations do not establish significance for small changes.

**Durability distinction:** tool stdout is a live host-owned attachment without a per-chunk durable stream journal. Ordinary RPC commits producer records before publication and consumer journals before delivery. The remaining 34×/58× completion gap to tool stdout is not claimed solved or irreducible. Neither barrier nor the one-value ordinary record contract was weakened. Group commit or ordinary-value coalescing requires a separate design/durability review.

Selected scalar medians; every case in the complete suite beats real old:

| Case | Real old | Final |
|---|---:|---:|
| Rust echo c1 | 43.987 | 9.188 |
| Rust local RPC c10 | 248.224 | 22.565 |
| Ephemeral 5,000 host calls c10 | 51,607.174 | 180.113 |
| Persistent 5,000 host calls c10 | 59,811.014 | 511.740 |
| Persistent with commits c10 | 80,490.686 | 7,108.327 |
| Small hot invocation c100 | 800.184 | 39.463 |
| Small cold invocation c100 | 2,577.361 | 340.747 |

Not every scalar case beats every intermediate checkpoint; no exact parity claim is made. The original historical results source is https://github.com/golemcloud/benchmark-results/blob/master/results/results.json; the table here is a controlled local real-version rerun, not a comparison of different machines.

### Profile evidence

Final one-iteration diagnostic: completion 1.03 s, first chunk 312 ms. Across warm-up plus measurement, 1,024 producer writes totaled 995.921 ms (median 0.796 ms), 56 persisted producer-index lookups totaled 337.958 ms, and six available-segment reads totaled 9.131 ms. No full consumer journal-inspection spans remain. Producer-index lookup count was 998 at the six-point checkpoint, then 70 after telemetry sampling. The remaining 2,133 metadata-query spans mostly hit memory (median 0.005 ms), not storage. Nested/asynchronous durations overlap and must not be added into a wall-time breakdown.

### Cold/recovery and review fixes included

- Incremental bounded lifecycle state in AgentStatusRecord; historical lifecycle/control/producer metadata share atomic CAS coverage, with paged consumer offsets and active recovery catalogue.
- Cold status folding hydrates external >64 KiB StreamSession records; missing/malformed payloads return typed errors instead of panics.
- Raw-session KV/catch-up runs outside oplog actors, with raw and committed tip revalidation; producer dependencies load in batched frontiers.
- Indexed cold producer load, direct cursor validation, and reconciliation avoid lifetime payload reconstruction. Committed terminals retire remote recovery without deleting journals.
- Worker deletion propagates index-clear errors. Complete deletion locator prefixes are preserved.
- The one-shot index setter/Weak backlink is documented: primary needs the index while historical reads need the outer multilayer service; there is no owning Arc cycle. Extracting a separate read-only layer stack remains a larger refactor.
- Cancellation commits intent/terminal and fences the index under ownership, then releases ownership before backpressured publication. Output pumping revalidates ownership before nested mapping mutation. This fixes the MoonBit lock cycle without weakening cancellation ordering.

### Verification

```sh
export CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 CARGO_PROFILE_TEST_DEBUG=0 CARGO_PROFILE_DEV_PANIC=unwind
cargo test -p golem-worker-executor --lib -j6 -- response_encoding durable_stream durable_session::tests stream_bus::tests worker::status session_index services::worker storage::indexed::postgres indeterminate --report-time
# 294 passed
cargo test -p golem-worker-executor --test integration -j6 -- postgres_singleton_append_many postgres_append_many_rolls_back active_stream_crash_replays_pinned_activation_with_fresh_attachments concurrent_tool_attempt_identity_survives_reordered_admission_and_replay rust_generated_client_streams_live_and_handles_edges rust_tool_trap_retries_owner_without_replay_overflow --report-time
# 6 passed: two real-PG contracts/rollback tests plus four stream/tool/crash tests
cargo clippy -p golem-common -p golem-worker-executor --all-targets -j6 -- -D warnings
cargo fmt --check
git diff --check
# passed
```

Oracle approved the plan and reviewed the final implementations, including SQLx autocommit acknowledgement/error semantics. Focused telemetry and PostgreSQL bug-finder passes found no bugs. Shared-model tests passed 11/11. Final streaming/scalar suites exited 0. MoonBit direct-ABI streaming lifecycle passed 20/20 repetitions against the exact final release services. The reconstruction test also passed locally (1/1); it remains intermittent, not fixed by that pass.

For the measured dd3b838b9 revision, CI attempt 1 failed only `incomplete_custom_durability_waits_for_completed_reconstruction` (“completed reconstruction body did not settle”); the failed-job rerun passed without source or assertion changes. This remains GOL-551, not a claimed fix. That revision's applicable jobs passed; latest-head CI is linked at the top.

### Explicit follow-ups and artifacts

- GOL-551: intermittent worker-group-1 reconstruction/long-poll/live-stream test failures. A prior canceled run also had a TS bridge post-test process hang, unconfirmed locally and absent in final CI; no speculative SDK fix was made.
- GOL-552: dedicated history-size/cold/reconnect/recovery performance matrix. Those paths are implemented and have correctness/bounded-read tests; warm benchmarks do not measure their latency. Prior estimate: history+cold 2–3 days; basic full matrix 5–8 days, excluding discovered fixes.
- Overlapping retained producers can duplicate cancellation terminals; the real-oplog diagnostic remains `overlap-reproducer.rs` alongside the report.
- Once-observed multilayer initial-index race and explicitly deferred deletion/cleanup performance.

Full report: `.amp/in/artifacts/gol-514/report.md`, final section “Completed hot-path measurements”. Raw results: `hotpath-old-{streaming,scalar}.json`, `hotpath-final-{streaming,scalar}.json`; hashes: `hotpath-provenance.txt`. Obtain these through [the implementation thread](https://ampcode.com/threads/T-01a0773b-dab0-74ae-9630-f0658221e29e). Earlier checkpoints and intermediate measurements remain labeled separately in the report.


golem-worker-executor/src/services/worker.rs
```diff
@@ -819,13 +969,17 @@ impl WorkerService for DefaultWorkerService {
         result
     }
 
-    async fn remove(&self, owned_agent_id: &OwnedAgentId) {
+    async fn remove(&self, owned_agent_id: &OwnedAgentId) -> Result<(), WorkerExecutorError> {
         record_worker_call("remove");
 
         if let Some(agent_mode) = self.get_agent_mode(owned_agent_id).await {
             self.oplog_service.delete(owned_agent_id, agent_mode).await;
```
noise64, line 1086: > The oplog is deleted before its derived stream index is cleared. If the process stops between these operations,       
  > recreating the same agent can inherit the old coverage and metadata; catch_up may then consider the new oplog already 
  > indexed. Could the index be validated against the worker fingerprint/incarnation, or deletion otherwise be made crash-
  > safe?

golem-worker-executor/src/durable_host/durable_stream.rs
noise64: > On a cold indexed producer, only Global metadata is initially loaded, so finished_sessions does not yet contain       
  > persisted sessions. Could this hydrate ProducerMetadataKey::Session(session_key.clone()) through index_for before     
  > checking the set? Otherwise, after a restart we can append mapping or topology records to an already-finished session.

golem-worker-executor/src/worker/invocation.rs
```diff
@@ -1382,6 +1386,33 @@ pub fn method_uses_streams(
     contains_stream(input) || method.uses_streams(&agent_type.schema)
```
noise64: > method.uses_streams() already checks the declared input and output schemas. Is the recursive contains_stream(input)   
  > traversal needed for valid invocations? It adds another O(payload) pass to the scalar path this PR is optimizing.

cli/golem/src/main.rs
```diff
@@ -16,12 +16,14 @@
...
 use golem_cli::command_handler::CommandHandler;
-use golem_cli::main_wrapper;
 use std::process::ExitCode;
 use std::sync::Arc;
 
 fn main() -> ExitCode {
-    main_wrapper(|| {
-        CommandHandler::handle_args(std::env::args_os(), Arc::new(ServerCommandHandler {}))
-    })
+    golem_worker_executor::bootstrap::create_runtime()
```
noise64: I think we should not enable this for all CLI commands, only when we actually run the server

golem-worker-executor/src/services/worker.rs
vigoo: Reevaluated this against the complete Worker lifecycle. Normal deletion retains the Worker in the active cache in the deleting state until durable cleanup finishes, fences new invocation admission, and stops the status writers/reconcilers. Ordinary concurrent recreation bypassing that exclusion was not established.

The narrower crash-consistency concern is valid: oplog deletion and derived-index clearing are separate awaited operations. A process loss between them can leave the agent-ID-keyed projection behind. A diagnostic using the real primary oplog/index implementations with in-memory storage reproduced an old raw-index row after deleting the oplog and recreating it with a new fingerprint.

That diagnostic does not establish end-to-end stale-session execution or data exposure: it simulates the interruption boundary rather than killing/restarting a Worker, and the public session lookup bypasses historical index reads when the current status has no session history. Oracle independently confirmed this distinction.

vigoo chose to reevaluate this concern rather than authorize the incarnation-index rewrite in this PR. Therefore no production fix was made here; this is not a claim that the crash boundary is safe. Fingerprint-qualified namespaces remain a valid candidate, but authoritative fingerprints must reach all readers/writers, including readers that currently discover identity from the index itself. The report retains the diagnostic and the unresolved crash-consistency limitation separately from deferred deletion/cleanup performance.

noise64 (approved):

PR #3838, opened 2026-09-07
Support streaming agent methods in generated guest bridges

vigoo:
## Summary

Implement generated guest bridges for stream-bearing agent methods in Rust, TypeScript, Scala, and MoonBit. Components can use generated clients to invoke streaming methods through awaited Wasm-RPC without handwritten transport code.

- Generate native `AgentStream` signatures, recursive stream-aware codecs, and schema-specific producer factories/adapters. Support nested streams, dynamically shaped values, and multiple sibling streams without fixed positional slots.
- Preserve affine endpoint ownership and direct unread forwarding, with cleanup for partial conversion failures, aborted calls, and failed invocation starts. Omit triggered and scheduled variants for stream-bearing methods while preserving non-streaming and external bridge behavior.
- Fix runtime defects exposed by cross-component tests: attachment slots colliding across consumers, loopback handles incorrectly entering foreign attachment control, local nested streams stalling on an attachment they cannot acquire, queued child drains being lost after parent completion, and unread resource drops failing to notify producers. Root production remains attachment-gated; nested production is admitted after its introducing parent item is committed.
- Add shared generator fixtures, generated-client compilation and codec tests, and language-specific cross-component lifecycle E2Es. Wire the suites into the guest-bridge CI shard, document the contract, and provision the required guest toolchains in orb setup.

Resolves GOL-511

## Validation

- Rust formatting checks and `git diff --check` passed.
- `cargo clippy -p golem-cli -p golem-worker-executor --all-targets --no-deps -- -D warnings` passed.
- CLI generator unit tests: **57 passed**.
- Durable session/stream runtime unit tests: **84 passed**.
- Combined Rust, TypeScript, and MoonBit generated-client matrix: **4 passed**, including executed schema-specific codec checks; Scala generated-client compilation also passed.
- Cross-component E2Es passed in all four languages, including input/output streaming, direct forwarding and loopback, recursive/dynamic siblings, stream-of-stream values, producer failures, cancellation/readable drop, and backpressure coverage. All **5 TypeScript E2Es** passed together against the updated runtime.
- SDK validation included Rust **131 passed / 4 ignored**, TypeScript **738 passed / 20 skipped**, MoonBit **54 passed**, and Scala core **608 passed** plus targeted native stream tests. TypeScript and Scala guest runtime artifacts were rebuilt for runtime testing.
- Focused Oracle reviews informed codec ownership, abort cleanup, and runtime admission/attachment decisions. Bounded bug reviews finished with no actionable findings.

Full repository CI and multi-executor deployment testing were not run locally.


golem-worker-executor/src/worker/mod.rs
```diff
@@ -3861,10 +3869,11 @@ impl<Ctx: WorkerCtx> Worker<Ctx> {
...
-                .await
-                .map_err(WorkerExecutorError::runtime)?
+            if producer.owns_handle_identity(&mapping.handle)
+                || streams
+                    .has_journaled_consumer_terminal(mapping)
+                    .await
+                    .map_err(WorkerExecutorError::runtime)?
             {
                 continue;
             }
```
noise64, line 3895: > Could we distinguish a relayed producer-owned handle from a fresh owner-local root stream here?
>
> Minimal example:
>
> 1. Agent A starts entity E asynchronously, so A is free to accept calls.
> 2. E creates a fresh stream S and synchronously calls A.consume(S).
> 3. E's stream drain waits for A to activate an attachment before producing.
> 4. A sees that its producer owns S and skips both attachment preparation and activation.
> 5. A waits for S's first item, while S waits for A's activation.
>
> This is not a cyclic awaited RPC: A is not awaiting E. The deadlock comes from treating “producer-owned” as equivalent to “already attached.” The current loopback test only returns an already-producing handle to its producer, so it does not cover this fresh-root case. Could we explicitly admit fresh owner-local roots, or reject this otherwise-supported entity-to-owner streaming call, and add the corresponding regression test?

vigoo: "Agent A starts entity E asynchronously, so A is free to accept calls." there is no such thing. Entities are tools and tool middlewares, and they are invoked as part of an agent invocation. The tool cannot call (and await) the parent agent via RPC because the agent is in a running invocation and invocations cannot overlap.

noise64 (approved):

PR #3840, opened 2026-09-07
Configure local server system memory overrides

noise64:
- lets local servers leave memory for other applications using `--system-memory-override`, `GOLEM_LOCAL_SERVER_SYSTEM_MEMORY_OVERRIDE`, or the manifest's `localServer.systemMemoryOverride`
- matches the executor's `memory.system_memory_override` naming and accepts byte-size strings such as `2GiB`, `1mb`, and `1MB` through one `humanize-rs` parser
- applies flag > environment > manifest > automatic detection precedence; accepts whole-number quantities and rejects zero, negative, fractional, malformed, non-ASCII, and overflowing sizes
- retains the executor's 80% admission threshold; this controls agent admission and eviction, not a hard process RSS ceiling
- updates the manifest schema, documentation, and local-server skill; retains manifest/configuration unit coverage and removes the extra command wiring tests

Validation after the parser switch: CLI library check, all-targets Clippy (`--no-deps -D warnings`), scoped formatting, and two production parser tests run through an isolated Cargo harness. These cover decimal/binary units, case variants, ASCII whitespace, unitless bytes, the u64 maximum, overflow, fractions, and Unicode rejection. Regenerated skill docs and docs lint/format pass. Earlier manifest/configuration unit and property coverage included 400 JSON/YAML round trips; that broader suite was not rerun for this parser switch. Earlier rebuilt CLIs and startup checks predate this parser change. Application integration runs exposed a separate memory-pressure replay failure, now reduced to a small reproduction; those integration runs are not passing.


Cargo.toml
```diff
@@ -185,6 +185,7 @@ opentelemetry = "0.30.0"
 opentelemetry-otlp = { version = "0.30.0" }
 opentelemetry-prometheus-text-exporter = "=0.2.0"
 opentelemetry_sdk = "0.30.0"
+parse-size = "1.1.0"
```
vigoo: We already use `humansize` everywhere, do we need another one?

vigoo (approved):

PR #3841, opened 2026-09-08
Built-in and native tools

vigoo:
Resolves GOL-24


golem-common/src/base_model/deployment.rs
```diff
@@ -159,4 +162,19 @@ declare_structs! {
         pub name: ToolName,
         pub hash: Hash,
     }
+
+    pub struct DeploymentPlanAmbientToolEntry {
+        pub release_id: crate::model::tool_release::ToolReleaseId,
+        pub name: ToolName,
+        pub version: String,
```
vigoo: Define either a new transparent newtype for this like DeploymentVersion or use that, if appropriate. Same applies for `metadata_version` below.

golem-common/tests/agent_extraction.rs
```diff
@@ -66,3 +68,13 @@ async fn can_extract_agent_type_schemas_from_component_importing_p3_http() -> an
     assert!(!result?.is_empty());
     Ok(())
 }
+
+#[test]
+async fn can_extract_tool_metadata_from_component_bytes() -> anyhow::Result<()> {
+    let wasm = std::fs::read(
+        "../test-components/tool-streaming/golem-temp/agents/golem_it_tool_streaming_rust_provider_release.wasm",
```
vigoo: Should not use a temp build dir - use the wasm from `../test-components` where the test component build script places it

golem-registry-service/src/bootstrap/mod.rs
```diff
@@ -121,6 +121,7 @@ pub struct Services {
     pub environment_state_service: Arc<EnvironmentStateService>,
     pub http_api_deployment_service: Arc<HttpApiDeploymentService>,
     pub mcp_deployment_service: Arc<McpDeploymentService>,
+    pub native_tool_catalog: Arc<crate::services::native_tool_catalog::NativeToolCatalog>,
```
vigoo: Why not import the trait - don't use fully qualified paths where not necessary (not just this line)

golem-worker-executor-test-utils/src/lib.rs
```diff
@@ -1954,6 +2078,29 @@ struct TestServerBootstrap {
 
 #[async_trait]
 impl WorkerCtx for TestWorkerCtx {
+    fn native_tool_catalog() -> Arc<NativeToolCatalog<Self>> {
+        let mut registrations = vec![NativeToolRegistration {
+            definition: native_test_helper_definition(),
+            handler: Arc::new(NativeToolAdapter(
+                __GolemNativeToolInvokerNativeDurableHelperImplNativeDurableHelper::new(
```
vigoo: What is this horrible type name and (if it is macro generated) why do we need to explicitly use here? Let's avoid that

sdks/rust/golem-tool-metadata/src/extended_tool_type.rs
```diff
@@ -0,0 +1,5743 @@
...
+        other => ToolBuildError::EncodeError(other.to_string()),
+    }
+}
+
+/// Encodes a metadata-time literal (option/positional default, `value-is`
+/// literal) into its self-contained `schema-value-tree` wire form. Type
+/// conformance of these literals against their referenced type node is checked
+/// separately by [`validate_tool`], which has the per-command field context
+/// needed to resolve `value-is` references.
+#[cfg(any())]
```
vigoo: There are 12 `#[cfg(any())]` blocks in this file (none existed before the move). That is dead code that is merely compiled out — delete it instead. We do not keep old code paths around (see the backward-compatibility section in AGENTS.md).

golem-registry-service/src/services/native_tool_catalog.rs
```diff
@@ -0,0 +1,506 @@
...
+                    security_overrides: false,
+                },
+                &auth,
+            )
+            .await
+            .unwrap();
+
+        let caller_wasm = std::fs::read(concat!(
+            env!("CARGO_MANIFEST_DIR"),
+            "/../test-components/tool-streaming/golem-temp/agents/",
```
vigoo: Same problem as in `golem-common/tests/agent_extraction.rs`: this reads a `golem-temp/` build-intermediate artifact. Use the wasm from `../test-components` where the test component build script places it.

Also, this test boots the full `Services` stack from a unit test in this crate. That belongs in `integration-tests`, or should at least not depend on a built test component from the registry-service unit test suite.

golem-registry-service/src/services/builtin_tool_provisioner.rs
```diff
@@ -0,0 +1,538 @@
...
+            blob_storage: BlobStorageConfig::default_in_memory(),
+            component_compilation: ComponentCompilationConfig::Disabled(Empty {}),
+            ..Default::default()
+        };
+        let owner = config.initial_accounts["builtin_tool_owner"].id;
+        let mut join_set = JoinSet::new();
+        let services = Services::new(&config, &mut join_set).await.unwrap();
+        let wasm = std::fs::read(concat!(
+            env!("CARGO_MANIFEST_DIR"),
+            "/../test-components/tool-streaming/golem-temp/agents/",
```
vigoo: Same: do not read from the `golem-temp/` build dir, use the wasm from `../test-components`.

Cargo.toml
```diff
@@ -37,6 +37,7 @@ golem-cli = { version = "=0.0.0", path = "cli/golem-cli" }
 golem-client = { version = "=0.0.0", path = "golem-client" }
 golem-common = { version = "=0.0.0", path = "golem-common", default-features = false }
 golem-component-compilation-service = { path = "golem-component-compilation-service" }
+golem-native-tool = { version = "=0.0.0", path = "sdks/rust/golem-native-tool" }
```
vigoo: The root workspace now path-depends on crates that live under `sdks/rust` (`golem-native-tool`, and transitively `golem-tool-metadata` / `golem-rust-macro`). If these are used by the executor (or other services), they must be moved out of `sdks/rust` and become regular members of the main workspace (with their deps declared in `[workspace.dependencies]`). The SDK workspace is deliberately not part of the main build flow; this coupling is also what causes the `large_enum_variant` clippy failure in `golem-rust-macro` in CI now.

golem-worker-executor/src/workerctx/mod.rs
```diff
@@ -145,6 +157,11 @@ pub trait WorkerCtx:
     + Sized
     + 'static
 {
+    /// Native implementations installed in this executor process.
+    fn native_tool_catalog() -> Arc<crate::native_tool::NativeToolCatalog<Self>> {
```
vigoo: This is a static fn whose implementations build a fresh `Arc<NativeToolCatalog>` (validating every definition and recomputing digests) on each call, and it is called on every native dispatch from `durable_host/tool/mod.rs`. The catalog must not be rebuilt per invocation — construct it once at executor bootstrap and pass it in like the other services (via `All`/`HasAll` or the context).

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -2184,12 +2226,21 @@ impl<Ctx: WorkerCtx> DurableWorkerCtx<Ctx> {
     }
 
     pub fn component_metadata(&self) -> &Component {
```
vigoo, line 2254: This PR turned `state.component_metadata` into `Option<Component>` and `component_metadata()` now panics for native contexts. There are still ~20 call sites going through `component_metadata()`, and some are reachable from a native invocation (e.g. `get-oplog` in `v1x.rs`, `call_coordinator`).

Can we change this back to a non-optional `Component`? For a native entity the owner agent's component is the meaningful one anyway, and everything that is executable-specific should be behind `WorkerCtxExecutable`. If it really has to stay optional, make the accessor return `Option`/`Result` so the compiler forces every call site to handle it instead of panicking at runtime.

golem-worker-executor/src/worker/entity_invocation.rs
```diff
@@ -347,46 +524,117 @@ where
...
         entity_name = entity.name(),
         invocation_start_index = scope.invocation_id().start_index().as_u64(),
-        executable_component_id = %executable.component_id,
-        executable_component_revision = executable.component_revision.get(),
+        executable_component_id = tracing::field::display(executable.map(|target| target.component_id.to_string()).unwrap_or_else(|| "native".to_string())),
+        executable_component_revision = executable.map(|target| target.component_revision.get()),
         activation_fingerprint = %scope.activation().fingerprint(),
         execution_mode = ?scope.mode(),
     );
+    struct AbortRelay(tokio_util::sync::CancellationToken);
```
vigoo: Moving to a cooperative `CancellationToken` is good, but this relay is convoluted: we spawn a task per invocation that does nothing but wait forever so that its `Drop` cancels the token, and then keep an `AbortHandle` to that task around as "the abort". Just make `slot.attach_abort` (and the retained native context) accept the `CancellationToken` directly and cancel it — no extra task, no drop-order subtlety.

sdks/rust/golem-native-tool/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+            return Err(format!(
+                "native tool '{}@{}' metadata identity does not match its definition",
+                self.id, self.implementation_version
+            ));
+        }
+        Ok(())
+    }
+}
+
+fn metadata_digest(metadata_version: &str, tool: &Tool) -> Result<[u8; 32], String> {
```
vigoo: This re-implements the `golem:tool-metadata:v1` digest that already exists as `golem_common::model::tool_release::tool_metadata_digest`. If these two ever diverge, every native registration fails `validate_dispatch`. Share the implementation (another argument for moving this crate into the main workspace where it can depend on `golem-common`); if that is impossible for some reason, add a cross-crate test asserting they produce the same digest.

sdks/rust/golem-native-tool/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+        pub mod wit {
+            pub mod wire {
+                pub use golem_schema::schema::tool::*;
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum Principal {
```
vigoo: Duplicate of `golem_common::base_model::agent::Principal`. Same as the digest below: reuse the common type instead of re-declaring the contract here.

golem-common/src/base_model/tool.rs
```diff
@@ -97,6 +97,37 @@ pub enum SecretKeyScope {
...
+#[cfg_attr(
+    feature = "full",
+    derive(desert_rust::BinaryCodec, golem_schema_derive::PoemSchema)
+)]
+#[cfg_attr(feature = "full", desert(evolution()))]
+#[serde(tag = "kind", content = "keys", rename_all = "camelCase")]
+pub enum ConfigKeyScope {
+    #[default]
+    All,
+    Keys(BTreeSet<CanonicalAgentSecretPath>),
```
vigoo: Config key scopes are typed as `CanonicalAgentSecretPath`, the CLI resolves them via `ManifestSecretKeyScope`, and the manifest schema points `configKeysReadable` at `#/definitions/secretKeyScope`. It works because the canonicalization is the same, but the naming is misleading. Introduce a neutral `CanonicalAgentConfigPath` (or rename to a shared `CanonicalAgentPath`) and a config-scope schema definition instead of reusing the secret ones.

golem-registry-service/src/services/deployment/read.rs
```diff
@@ -84,18 +85,22 @@ pub struct DeploymentService {
...
     deployment_repo: Arc<dyn DeploymentRepo>,
+    native_tool_catalog: Arc<NativeToolCatalog>,
 }
 
 impl DeploymentService {
     pub fn new(
         environment_service: Arc<EnvironmentService>,
         application_service: Arc<ApplicationService>,
         deployment_repo: Arc<dyn DeploymentRepo>,
+        _component_repo: Arc<dyn crate::repo::component::ComponentRepo>,
```
vigoo: Unused parameter — remove it (and the argument at the call sites).

golem-registry-service/src/bootstrap/mod.rs
```diff
@@ -463,6 +470,29 @@ impl Services {
...
+            builtin_tool_owner_account_id,
+            &application_service,
+            &environment_service,
+            &component_service,
+            &component_write_service,
+            &deployment_service,
+            &deployment_write_service,
+            &tool_release_service,
+        )
+        .await
```
vigoo, line 480: Built-in tool provisioning failure aborts registry startup here, while built-in plugin provisioning just above only warns. Pick one behaviour deliberately for both. Hard-failing is arguably right for protected system releases, but then the plugin path should match (or the difference should be documented).

golem-registry-service/src/repo/model/tool_release.rs
```diff
@@ -120,11 +120,6 @@ impl ToolReleaseRecord {
                 "system tool release coordinate does not match its definition"
```
vigoo, line 120: The "protected system tool releases must use a host source" check was removed here so that component-implemented built-ins can also be protected system releases. Nothing now enforces the source/availability pairing though: `Ambient` ⇒ `Host` source and `Grantable` ⇒ `Component` source. Please reinstate that invariant in its new form (here or in `ToolReleaseService::provision_system_release`).

golem-worker-executor-test-utils/src/lib.rs
```diff
@@ -1655,6 +1659,126 @@ pub struct TestWorkerCtx {
     agent_id: AgentId,
 }
 
+static NATIVE_TEST_EFFECTS: AtomicUsize = AtomicUsize::new(0);
+static NATIVE_TEST_HELPER_EFFECTS: AtomicUsize = AtomicUsize::new(0);
+static NATIVE_TEST_TOOL_METADATA: OnceLock<golem_common::schema::tool::Tool> = OnceLock::new();
```
vigoo: Process-global mutable state in test utils (`OnceLock` + `install_native_test_tool_metadata`) makes the native test tool registration depend on which test ran first in the process and leaks between tests. Register test tools explicitly on the test context/catalog instead — this goes away naturally once the catalog is a bootstrap-time service passed into the context rather than a static fn.

golem-registry-service/src/services/deployment/write.rs
```diff
@@ -252,21 +257,63 @@ impl DeploymentWriteService {
...
+        let ambient_catalog = self
+            .native_tool_catalog
+            .active()
+            .into_iter()
+            .map(|tool| (tool.release.name.clone(), tool))
+            .collect::<std::collections::BTreeMap<_, _>>();
+        let ordinary_remote_tools = data
             .remote_tools
+            .iter()
+            .filter(|deployment| !ambient_catalog.contains_key(&deployment.name))
```
vigoo, line 284: The CLI rejects a top-level `tools` entry whose name collides with an ambient tool, but the API silently accepts it: the user's `release`/`provision`/`environment_binding` is dropped here and replaced with the ambient one, and only `agent_bindings` survive. The server should reject the collision too (or at least error when the requested release differs from the ambient release) instead of rewriting the request.

golem-registry-service/src/services/deployment/write.rs
```diff
@@ -252,21 +257,63 @@ impl DeploymentWriteService {
...
+                    .map(|deployment| deployment.agent_bindings.clone())
+                    .unwrap_or_default(),
+            };
+            remote_tools.push((
+                canonical,
+                Some(
+                    crate::services::environment_tool_grant::ResolvedGrantedToolRelease {
+                        owner: AccountSummary {
+                            id: release.owner_account_id,
+                            name: "Golem system".to_string(),
```
vigoo: Hardcoded display name; take it from the actual built-in tool owner account instead of a literal.

golem-worker-executor/src/durable_host/tool/mod.rs
```diff
@@ -1976,6 +1992,164 @@ async fn invoke_tool_sidecar<Ctx: WorkerCtx>(
...
+            if let Some(error) = stdout_limit_error(
+                stdout_observer
+                    .as_ref()
+                    .and_then(AttachmentObserver::terminal_snapshot)
+                    .is_some_and(|terminal| terminal.host_resource_exhausted),
+            ) {
+                encode_tool_operation_terminal(Err(error)).await
+            } else {
+                encode_tool_operation_terminal(result).await
+            }
```
noise64, line 2137: > Could we mirror the component invocation path's post-body `has_pending_live_admission_rejection` guard for native tools? A native handler can catch an attachment admission error and return success; this path then encodes an ordinary terminal, while the later accounting check does not replace that response with the pending resource-exhaustion failure. Please preserve the rejection before terminal encoding and cover a handler that catches the host error.

golem-worker-executor/src/native_tool.rs
```diff
@@ -0,0 +1,256 @@
...
+            .invoke(
+                ctx,
+                golem_native_tool::NativeToolInvocation {
+                    command_path: invocation.command_path,
+                    input: invocation.input,
+                    principal: invocation.principal,
+                    stdin: invocation.stdin.map(|input| Box::new(input) as _),
+                    stdout: invocation.stdout.map(|output| Box::new(output) as _),
+                },
+            )
```
noise64, line 157: > Could we expose and forward a safe cancellation handle through `golem_native_tool::NativeToolInvocation`? The executor invocation carries a `CancellationToken`, but `NativeToolAdapter` constructs the public invocation without it, so macro/SDK-authored native tools cannot observe cancellation even though direct executor handlers can. A cancellation test through the adapter would pin the supported authoring path.

golem-worker-executor/src/durable_host/tool/mod.rs
```diff
@@ -1976,6 +1992,164 @@ async fn invoke_tool_sidecar<Ctx: WorkerCtx>(
...
+    )
+    .await;
+    retained
+        .context_mut()
+        .durable_ctx_mut()
+        .set_invocation_principal(None);
+    if let Err(error) = retained.prepare_parent_end().await {
+        return (Err(error), Some(Box::new(retained)));
+    }
+    let result = match result {
```
noise64, line 2126: > Could we preserve the primary handler or infrastructure error when `prepare_parent_end` also fails, matching the component path? The native path currently returns the cleanup failure before examining the handler result, which can mask the failure that caused the invocation to abort. Substituting the cleanup error only when the body succeeded would retain the more useful diagnosis.

cli/golem-cli/src/command_handler/component/mod.rs
```diff
@@ -1146,6 +1155,71 @@ impl ComponentCommandHandler {
         let mut diffable_remote_tool_deployments = BTreeMap::new();
         let mut pending_remote_initial_files = Vec::new();
 
+        for ambient in ambient_tools {
+            if implementations.contains_key(&ambient.name)
+                || app.tool_declarations().contains_key(&ambient.name)
```
noise64, line 1160: What is the intended rollout policy when a future Golem release introduces an ambient tool whose name is already used by an application or selected account-owned tool? AFAIU existing deployments remain intact, but the next deployment will fail. I guess this is fine (and similar to account shared ones), just wanted to double check.

golem-common/tests/goldenfiles/diff_model_fingerprint_v5.txt
```diff
@@ -1 +1 @@
-9b8236ff15b23a2da0c9bbaaa28149cd9875f93d6a519dab7b5ed98c8dbdfd75
+8432554acdeedaabbb278c764a8a66243102ea9197309ffa6248071564c9a3c3
```
noise64: This should bump the diff model version, and create a v6 fingerprint

cli/golem-cli/src/command_handler/app/mod.rs
```diff
@@ -1073,8 +1073,19 @@ impl AppCommandHandler {
...
+        // Ambient tools are registry-owned input. Resolve them before local validation and
+        // hashing; planning remains read-only.
+        let ambient_plan = self
+            .ctx
+            .golem_clients()
+            .await?
+            .environment
+            .get_environment_deployment_plan(&environment.environment_id.0)
+            .await
+            .map_service_error()?;
```
noise64, line 1085: - `prepare_deployment` now fetches the same deployment plan twice; I think one read should be enough (also the first read calls it `ambient_plan`, which is not entirely correct)
- `ambientTools` appears to be the first field in this DTO that is not information identifying or describing the staged diff side itself, but is auxiliary registry metadata used to construct the proposed side, so i think it would be clearer to separate it explicitly in the response DTO (or have some comment on it / naming convention for possible future similar ones?)

cli/golem-cli/src/model/app_raw.rs
```diff
@@ -365,6 +365,13 @@ pub enum ManifestSecretKeyScope {
     Keys(Vec<String>),
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum ManifestConfigKeyScope {
+    All(String),
+    Keys(Vec<String>),
+}
```
noise64: Some of our checks are only visible in the JSON schema, but that is extra meta, not the main validation we use in the CLI:
> The JSON Schema constraints are useful diagnostics, but they are not currently enforced by Rust when the manifest deserializes successfully: the untagged scope enums accept any string and any string list, while path validation happens only during deployment. Please enforce the wildcard/list shape through serde and validate every path while loading/building the application in `app_raw.rs` or `app.rs`. The same issue exists for the pre-existing secret scopes, and the generic audit also found that tool-binding keys rely on the schema regex without equivalent Rust load-time validation.

cli/schema.golem.cloud/app/golem/1.6.0-dev.9/golem.schema.json
```diff
@@ -930,6 +936,22 @@
...
+          "type": "array",
+          "items": {
+            "type": "string",
+            "not": {
+              "const": "*"
+            }
+          }
+        }
+      ]
+    },
```
noise64, line 954: It seems we keep modifying these dev one, unlike how they were originally intended to be used. I guess this can be the new norm, but we might want to adjust the skills so we do not modify non dev versions. And we might want to think about automating the publishing of these, to guarantee no update on the non dev one (not in this PR, just want to mention).

vigoo: [GOL-591](https://linear.app/golem-cloud/issue/GOL-591/define-manifest-schema-publication-policy-and-enforce-immutable)

noise64 (approved):

PR #3842, opened 2026-09-08
Tool middleware registry and cli support

vigoo:
Resolves GOL-39


vigoo (commented):
Review of the registry/CLI middleware support. Overall the model looks right (merge-mode semantics, chain order, pinning, WIT copies and protobuf round-trips all check out), but there are a few things to address before this leaves WIP — see the inline comments:

- manifest JSON schema `compatibilityMode` enum doesn't match the Rust enum
- server never returns middleware state in deployment plans/summaries
- nominal compatibility mode is names-only, not the spec'd semantics
- structural error compatibility ignores `kind`/`exit_code`; strict-equality compares docs/version
- structural input projection: emit warnings for discarded inputs
- `synthesize_effective_definition`: remark on error propagation / cross-graph `Ref` comparison
- explicit reserved-name error for a tool named `middleware`
- `middlewareMergeMode` should be agent-only
- MoonBit: custom-error names lost, `expected` should be optional
- reduce the duplicated release/grant stacks to a minimum
- DB migrations: dead columns in 036, opaque blobs in 037
- dedicated API error codes for middleware
- test gaps (end-to-end chain compile, snapshot round-trip, compat variance)
- license headers on new files
- move inline `mod tests` into separate `tests.rs` files

CI is also red on 4 jobs (clippy `large_enum_variant` on `ToolMiddlewareScope::Monomorphic`, stale `golem-rust-macro` fixtures using `ToolInvokeError<TypedSchemaValue>`, the `bug_finder_guest_custom_error_wire_value_matches_declared_case_payload` expectation, and prettier on `middlewareRuntime.ts`); the integration/worker suites never ran because of that.

cli/schema.golem.cloud/app/golem/1.6.0-dev.9/golem.schema.json
```diff
@@ -1250,6 +1298,9 @@
         "compatibilityCheck": {
           "type": "boolean"
         },
+        "compatibilityMode": {
+          "enum": ["nominal", "structural"]
+        },
```
vigoo, line 1316: The JSON schema enum here (`nominal`, `structural`) does not match the Rust `ToolCompatibilityMode` enum in `golem-schema/src/schema/tool/compatibility.rs`, which serializes (kebab-case) as `strict-equality`, `structural-subtype` and `nominal`.

Consequences: the default value `structural-subtype` fails schema validation in editors, and the `structural` value accepted by the schema is rejected by serde. Please align this to `["strict-equality", "structural-subtype", "nominal"]` (ideally generated from / tested against the Rust enum so they can't drift again).

golem-registry-service/src/repo/model/deployment.rs
```diff
@@ -295,6 +301,8 @@ impl DeploymentIdentity {
                 .collect::<Result<Vec<_>, _>>()?,
             remote_tools,
             published_tools,
+            remote_tool_middlewares: Vec::new(),
+            published_tool_middlewares: Vec::new(),
```
vigoo: The server never populates middleware state for plans/summaries: these are hard-coded to empty here, and the same happens in `DeploymentIdentity::into_plan` (`remote_tool_middleware_deployments`, `published_tool_middlewares`, `universal_tool_middlewares`, `tool_compatibility_mode`, `environment_tool_middleware_bindings`, `agent_tool_middleware_bindings` all `Default::default()` a few lines below) and in `TryFrom<DeployedDeploymentIdentity> for DeploymentSummary`. The `DeploymentPlan`/`DeploymentSummary` API types also cannot carry the universal chain, compatibility mode or bindings at all.

Effect on the CLI: the quick "already up to date" check still works because it uses the server-stored hash, but the detailed current-vs-local diff, `diff_stage` and `is_stage_same_as_current` can't see middleware, so whenever anything else changes every middleware binding shows up as newly "added". `DeploymentPlan.remote_tool_middlewares` / `published_tool_middlewares` are effectively dead fields right now.

The stored snapshot (`DeploymentToolMiddlewareSnapshotRecord`) has the data; please surface it through the plan/summary conversions and extend the API types accordingly, plus a round-trip test (see the test-gaps comment).

golem-schema/src/schema/tool/compatibility.rs
```diff
@@ -0,0 +1,1549 @@
...
+    // Nominal compatibility is deliberately names-only. Do not validate or
+    // inspect unrelated command and schema details in this mode.
+    if mode == ToolCompatibilityMode::Nominal {
+        if expected.name() != inner.name() {
+            return Err(vec![err("name", "tool names differ")]);
+        }
+        return Ok(CompiledToolCompatibility {
+            mode,
+            commands: Vec::new(),
+        });
```
vigoo: Nominal mode is implemented as a root-name-only check. GOL-39 specifies nominal compatibility with its own variance rules over the command tree, type tree and error cases, and explicitly says the validator must implement the mode semantics rather than just dispatch by name. With this implementation a nominal chain accepts an inner tool with a completely different command tree and produces an empty `commands` projection.

Please implement the nominal rules from the spec (or, if names-only is the intended interpretation, update the ticket/docs so they agree — but the current comment "deliberately names-only" contradicts the ticket as written).

golem-schema/src/schema/tool/compatibility.rs
```diff
@@ -0,0 +1,1549 @@
...
+                None => (None, None),
+            };
+            ErrorProjection {
+                name: n.name.clone(),
+                inner_index: i,
+                expected_index,
+                payload,
+            }
+        })
+        .collect()
```
vigoo, line 572: Structural error compatibility only matches error cases by `name` and then compares payloads. `ErrorCase.kind` and `exit_code` are ignored entirely, so an inner tool that keeps the name but changes the kind/exit code passes as compatible. Please include those in the structural check (and in the strict-equality comparison).

Related, for strict-equality: `strictly_equal` currently also compares documentation and version strings, so a doc-comment edit on the inner tool breaks a strict chain. That seems too strict — docs at least should not participate in compatibility. Also `validate_tool_middleware` in `validation.rs` never validates `version`.

golem-schema/src/schema/tool/compatibility.rs
```diff
@@ -0,0 +1,1549 @@
...
+            });
+        } else {
+            errors.push(err(
+                format!("{}.input.{}", format_path(path), target.name),
+                "inner-only input has no representable default",
+            ));
+            return None;
+        }
+    }
+    let discard = (0..ef.len()).filter(|i| !retained.contains(i)).collect();
```
vigoo: Inputs that exist on `expected` but not on the inner tool end up in `discard` silently. This is type-correct under record width subtyping, but semantically a middleware that injects e.g. a `--token`/`retries` input towards a tool that has no such input gets it thrown away with no signal at all.

Let's emit warnings for every discarded input (path + name) so they reach the deploy diagnostics, rather than dropping them silently. The `CompiledToolCompatibility` result could carry a `warnings: Vec<...>` alongside the projections.

golem-registry-service/src/services/deployment/tool_middlewares.rs
```diff
@@ -0,0 +1,820 @@
...
+            registration.definition.scope,
+            ToolMiddlewareScope::Universal
+        )
+    {
+        return Err("middleware is installed in the wrong scope".to_string());
+    }
+    Ok(registration)
+}
+
+fn synthesize_effective_definition(
```
vigoo, line 344: Remark, not necessarily something to fix: `synthesize_effective_definition` propagates inherited error cases to *every* presented command regardless of the command path they originate from (the test name suggests this is intentional). It also compares payload `SchemaType`s across two different schema graphs, where `Ref` ids are graph-local, so equality of two `Ref`s doesn't mean the referenced types are equal. If path-insensitive propagation is intended, a short comment explaining why would help; the cross-graph `Ref` comparison I think is worth a second look.

cli/golem-cli/src/model/app_raw.rs
```diff
@@ -151,6 +156,16 @@ impl ToolDeclarations {
     pub fn into_entries(self) -> impl Iterator<Item = (String, serde_json::Value)> {
         self.0.into_iter()
     }
+
+    pub fn into_tools_and_middleware(
+        mut self,
+    ) -> (
+        IndexMap<String, serde_json::Value>,
+        Option<serde_json::Value>,
+    ) {
```
vigoo, line 165: Because `middleware` is a reserved key under `tools`, a tool that is literally named `middleware` is silently swallowed here (and the JSON schema `propertyNames` pattern still admits that name). Please turn this into an explicit error: reject a tool declaration named `middleware` with a clear "reserved name" message, and reflect the reservation in the schema (e.g. a `not: { enum: ["middleware"] }` on the tool property names or a note in the description).

golem-common/src/base_model/tool.rs
```diff
@@ -137,6 +137,17 @@ pub struct ToolBindingInput {
...
+    #[cfg_attr(feature = "full", desert(default), oai(default))]
+    pub filesystem_access: ToolFilesystemAccess,
+    /// `None` means no middleware list was authored; `Some([])` explicitly clears the
+    /// per-tool list when combined with `Replace`.
+    #[serde(default)]
+    #[cfg_attr(feature = "full", desert(default), oai(default))]
+    pub middleware: Option<Vec<crate::base_model::tool_middleware::ToolMiddlewareInstallation>>,
+    #[serde(default)]
+    #[cfg_attr(feature = "full", desert(default), oai(default))]
+    pub middleware_merge_mode: crate::base_model::tool_middleware::ToolMiddlewareMergeMode,
```
vigoo: `middleware_merge_mode` lives on the shared `ToolBindingInput`, so it is accepted and transmitted for environment-level bindings as well (and `middlewareMergeMode` is present on the shared manifest binding struct in `app_raw.rs`). Per GOL-39 the merge mode is an agent-binding concept only: it controls how the agent's list merges with the environment default and must never affect the universal chain.

Please make it agent-only — either move it to the agent binding type, or reject it at validation time when set on an environment binding — so users can't author a setting that has no effect.

sdks/moonbit/golem_sdk/tool-middleware/runtime.mbt
```diff
@@ -96,7 +98,9 @@ pub async fn UnderlyingTool::invoke(
   input : @model.TypedSchemaValue,
   stdin : MiddlewareStdin?,
 ) -> Result[OwnedInvocationResult, ToolInvokeError[@model.TypedSchemaValue]] {
-  self.invoke_with(command_path, input, stdin, value => Ok(value))
+  self.invoke_with(command_path, input, stdin, fn(_name, value) {
+    Some(Ok(value))
+  })
 }
```
vigoo, line 110: Two MoonBit issues:

1. `UnderlyingTool::invoke` passes `fn(_name, value) { Some(Ok(value)) }` and the corresponding conversion in `raw.mbt` (`ToolError::CustomError({ name: "custom", ... })`) synthesizes the literal name `"custom"`, so the declared custom-error case name coming back from the inner tool is lost. The Rust and TS runtimes preserve the name; MoonBit should too.
2. `registry.mbt` models `MonomorphicToolMiddlewareScope.expected` as a mandatory `Tool` and wraps it in `Some(...)` on conversion, while the WIT declares `expected: option<tool>`. Please align with the WIT (and with the other SDKs) so a middleware can omit `expected`.

Also the "unknown error" naming differs across SDKs (`UnknownCustomError` / `'unknown-error'` / `UnknownToolError`); worth unifying while touching this.

golem-registry-service/src/repo/tool_middleware_release.rs
```diff
@@ -0,0 +1,624 @@
+// Copyright 2024-2026 Golem Cloud
```
vigoo, line 1: This file (and `repo/environment_tool_middleware_grant.rs`, `services/tool_middleware_release.rs`, `services/environment_tool_middleware_grant.rs`, `api/tool_middleware_releases.rs`, `api/environment_tool_middleware_grants.rs`, and `command_handler/app/tool_middleware.rs` in the CLI) is a near-verbatim copy of the corresponding tool release / environment tool grant stack — roughly 2.4k lines duplicated in total.

GOL-39 asked to extend the existing model rather than build a parallel path. Please reduce the duplication to the minimum: e.g. share the repo/service/api implementations via a generic over the release "kind" (tool vs middleware), or a common trait + thin per-kind adapters, so grants/releases/listing/immutability/protection logic exists once. Related: `compile_tool_middleware_chains` re-validates the whole registry per binding (O(bindings × registry)) and emits duplicate diagnostics, and `write.rs` falls back to the name `"unknown-middleware"` — both would be easier to fix once the code lives in one place.

golem-registry-service/db/migration/postgres/036_tool_middleware_releases.sql
```diff
@@ -0,0 +1,85 @@
...
+    state_changed_at TIMESTAMP NOT NULL,
+    state_changed_by UUID NOT NULL,
+    CONSTRAINT tool_middleware_releases_pk PRIMARY KEY (tool_middleware_release_id),
+    CONSTRAINT tool_middleware_releases_owner_account_fk FOREIGN KEY (owner_account_id) REFERENCES accounts,
+    CONSTRAINT tool_middleware_releases_component_revision_fk FOREIGN KEY (component_id, component_revision) REFERENCES component_revisions,
+    CONSTRAINT tool_middleware_releases_source_kind_check CHECK (source_kind = 0),
+    CONSTRAINT tool_middleware_releases_lifecycle_check CHECK (lifecycle IN (0, 1, 2)),
+    CONSTRAINT tool_middleware_releases_origin_check CHECK (origin IN (0, 1)),
+    CONSTRAINT tool_middleware_releases_no_system_availability_check CHECK (system_availability IS NULL),
+    CONSTRAINT tool_middleware_releases_no_host_source_check CHECK (host_tool_id IS NULL AND implementation_version IS NULL)
```
vigoo: This table copies columns from `tool_releases` that middleware releases can never use (`host_tool_id`, `implementation_version`, `system_availability`; and the version column is named `tool_version`) and then adds CHECK constraints forcing them to `NULL`. That is dead schema by construction, which goes against the repo policy of not adding compat/dead paths. Please drop these columns (and the `source_kind = 0` constant column if only one kind can exist) and rename `tool_version` to `version`/`middleware_version`. Same applies to the SQLite migration.

golem-registry-service/db/migration/postgres/037_deployment_tool_middlewares.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE deployment_tool_middleware_snapshots
+(
+    environment_id UUID NOT NULL,
+    deployment_revision_id BIGINT NOT NULL,
+    registered_middlewares BYTEA NOT NULL,
+    compiled_chains BYTEA NOT NULL,
+    universal_installations BYTEA NOT NULL,
+    compatibility_mode BYTEA NOT NULL,
+    published_names BYTEA NOT NULL,
+    remote_names BYTEA NOT NULL,
```
vigoo: All middleware deployment state is stored as opaque `desert` blobs, including things that are plain scalars/lists (`compatibility_mode`, `published_names`, `remote_names`, `universal_installations`). The tool side of deployments uses relational rows for the same kind of data, so this is inconsistent and makes the middleware state unqueryable/undiffable server-side. Please model at least the scalar/list fields relationally (compat mode as a column, installations/bindings as rows), keeping blobs only for the compiled projections if needed.

golem-registry-service/src/api/error.rs
```diff
@@ -978,6 +993,63 @@ impl From<EnvironmentToolValidationError> for ApiError {
...
 }
 
+impl From<ToolMiddlewareReleaseError> for ApiError {
+    fn from(value: ToolMiddlewareReleaseError) -> Self {
+        let error = value.to_safe_string();
+        match value {
+            ToolMiddlewareReleaseError::ToolMiddlewareReleaseNotFound(_)
+            | ToolMiddlewareReleaseError::ReferencedToolMiddlewareReleaseNotFound
+            | ToolMiddlewareReleaseError::ParentAccountNotFound(_) => {
+                Self::not_found(api::error_code::TOOL_RELEASE_NOT_FOUND, error)
```
vigoo: Middleware errors reuse the tool error codes (`TOOL_RELEASE_NOT_FOUND` here, `ENVIRONMENT_TOOL_GRANT_NOT_FOUND` / `_ALREADY_EXISTS` / `_CONFLICT` in the grant mapping below), and `DeployRepoError::ToolReleaseImmutableConflict` is reused for middleware releases. Clients cannot tell a missing tool release from a missing middleware release. Please introduce dedicated codes (`TOOL_MIDDLEWARE_RELEASE_NOT_FOUND`, `ENVIRONMENT_TOOL_MIDDLEWARE_GRANT_*`, `TOOL_MIDDLEWARE_RELEASE_IMMUTABLE_CONFLICT`, …) in `golem-common/src/base_model/api.rs` and regenerate the OpenAPI/docs.

golem-registry-service/src/services/deployment/tool_middlewares.rs
```diff
@@ -0,0 +1,820 @@
...
+    ToolMiddlewareCompileDiagnostic {
+        agent_type_name: binding.agent_type_name.clone(),
+        tool_name: binding.tool_name.clone(),
+        middleware_name,
+        message: message.into(),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
vigoo: Test gaps: the unit tests here cover merge-mode and pinning in isolation, but there is no end-to-end chain compile test that exercises order (universal → per-tool → leaf), scope (monomorphic vs universal), version/account pinning and compatibility mode together, and there is no deployment snapshot round-trip test with middleware (which would have caught the empty plan/summary fields). The compatibility tests in `golem-schema` also miss input variance, command variance, options/flags, recursive types and error `kind`/`exit_code` cases. Please add these.

Please move this inline `mod tests` into a separate `tests.rs`: promote the module to a directory (`<name>/mod.rs` + `<name>/tests.rs`, or `<name>.rs` + `<name>/tests.rs`) so the implementation and its tests live in two files.

golem-common/src/base_model/tool_middleware_release.rs
```diff
@@ -0,0 +1,106 @@
+// Copyright 2024-2026 Golem Cloud
+// Licensed under the Golem Source License v1.1; see http://license.golem.cloud/LICENSE
```
vigoo: This file uses a 2-line license header; every other file in the repo uses the standard 13-line header (see e.g. `golem-common/src/base_model/tool.rs`). Please use the standard header. Also, `rustfmt` doesn't format inside `declare_structs!`, so please expand the fields one-per-line manually to match the rest of the codebase.

golem-common/src/base_model/environment_tool_middleware_grant.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright 2024-2026 Golem Cloud
+// Licensed under the Golem Source License v1.1; see http://license.golem.cloud/LICENSE
```
vigoo: This file uses a 2-line license header; every other file in the repo uses the standard 13-line header (see e.g. `golem-common/src/base_model/tool.rs`). Please use the standard header. Also, `rustfmt` doesn't format inside `declare_structs!`, so please expand the fields one-per-line manually to match the rest of the codebase.

cli/golem-cli/src/model/tool_middleware.rs
```diff
@@ -0,0 +1,239 @@
+// Copyright 2024-2026 Golem Cloud
+// Licensed under the Golem Source License v1.1; see http://license.golem.cloud/LICENSE
```
vigoo: This file uses a 2-line license header; every other file in the repo uses the standard 13-line header (see e.g. `golem-common/src/base_model/tool.rs`). Please use the standard header. Also, `rustfmt` doesn't format inside `declare_structs!`, so please expand the fields one-per-line manually to match the rest of the codebase.

golem-registry-service/src/services/deployment/tool_middlewares.rs
```diff
@@ -0,0 +1,820 @@
+// Copyright 2024-2026 Golem Cloud
+//
+// Licensed under the Golem Source License v1.1 (the "License");
+// you may not use this file except in compliance with the License.
```
vigoo, line 4: Truncated license header (4 lines) — please use the standard 13-line header used everywhere else.

cli/golem-cli/src/command_handler/app/tool_middleware.rs
```diff
@@ -0,0 +1,218 @@
...
+        GrantExecutionDecision::StopAfterPlan
+    } else if !stage && !plan && has_changes {
+        GrantExecutionDecision::ConfirmApply
+    } else {
+        GrantExecutionDecision::ContinueReadOnly
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
vigoo: Please move this inline `mod tests` into a separate `tests.rs`: promote the module to a directory (`<name>/mod.rs` + `<name>/tests.rs`, or `<name>.rs` + `<name>/tests.rs`) so the implementation and its tests live in two files.

golem-registry-service/src/services/environment_tool_middleware_grant.rs
```diff
@@ -0,0 +1,726 @@
...
+fn environment_owner(environment: &Environment) -> EnvironmentOwnerPattern {
+    EnvironmentOwnerPattern::Environment {
+        account: environment.owner_account_email.clone(),
+        application: environment.application_name.clone(),
+        environment: environment.name.clone(),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
vigoo: Please move this inline `mod tests` into a separate `tests.rs`: promote the module to a directory (`<name>/mod.rs` + `<name>/tests.rs`, or `<name>.rs` + `<name>/tests.rs`) so the implementation and its tests live in two files.

golem-schema/src/schema/tool/compatibility.rs
```diff
@@ -0,0 +1,1549 @@
...
+}
+fn err(path: impl Into<String>, message: impl Into<String>) -> ToolCompatibilityError {
+    ToolCompatibilityError {
+        path: path.into(),
+        message: message.into(),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
vigoo: Please move this inline `mod tests` into a separate `tests.rs`: promote the module to a directory (`<name>/mod.rs` + `<name>/tests.rs`, or `<name>.rs` + `<name>/tests.rs`) so the implementation and its tests live in two files.

golem-schema/src/schema/tool/validation.rs
```diff
@@ -1350,3 +1425,68 @@ fn collect_refs(constraint: &Constraint) -> Vec<&Ref> {
         Constraint::Forbids(forbids) => forbids.lhs.iter().chain(forbids.rhs.iter()).collect(),
     }
 }
+
+#[cfg(test)]
+mod tests {
```
vigoo: Please move this inline `mod tests` into a separate `tests.rs`: promote the module to a directory (`<name>/mod.rs` + `<name>/tests.rs`, or `<name>.rs` + `<name>/tests.rs`) so the implementation and its tests live in two files. (this PR introduces the inline test module here)

golem-schema/src/schema/tool/wit.rs
```diff
@@ -857,3 +901,91 @@ fn decode_error_case(dec: &GraphDecoder, e: &wire::ErrorCase) -> Result<ErrorCas
             .transpose()?,
     })
 }
+
+#[cfg(test)]
+mod tests {
```
vigoo: Please move this inline `mod tests` into a separate `tests.rs`: promote the module to a directory (`<name>/mod.rs` + `<name>/tests.rs`, or `<name>.rs` + `<name>/tests.rs`) so the implementation and its tests live in two files. (this PR introduces the inline test module here)

golem-schema/src/schema/tool/compatibility.rs
```diff
@@ -0,0 +1,1421 @@
...
+        let (t, tid) = resolve(self.tg, target);
+        if let (Some(a), Some(b)) = (sid, tid) {
+            if let Some(&node) = self.active.get(&(a.clone(), b.clone())) {
+                return Some(self.push(ProjectionNode::Recursive { node }));
+            }
+            let placeholder = self.push(ProjectionNode::Identity);
+            self.active.insert((a.clone(), b.clone()), placeholder);
+            let body = self.compile_resolved(s, t, path, errors)?;
+            self.nodes[placeholder] = ProjectionNode::Recursive { node: body };
+            self.active.remove(&(a, b));
```
noise64: > Could we memoize completed source/target type pairs and enforce a node or depth budget while compiling projection plans? `Compiler::active` detects only cycles and removes a pair after each subtree, so sibling references in a compact diamond graph repeatedly expand the same pair and can produce an exponential `ProjectionNode` graph during deployment. Please return a `ToolCompatibilityError` on the limit and add a diamond-shaped regression case.

golem-registry-service/src/services/deployment/tool_middlewares/mod.rs
```diff
@@ -0,0 +1,609 @@
...
+        .and_then(|b| b.middleware.as_ref())
+        .cloned()
+        .unwrap_or_default();
+    let Some(agent_binding) = agent else {
+        return environment;
+    };
+    let Some(agent_installations) = &agent_binding.middleware else {
+        return environment;
+    };
+    match agent_binding.middleware_merge_mode.unwrap_or_default() {
```
noise64, line 275: > Could we reject `middleware_merge_mode` for environment bindings at the server boundary or use distinct environment and agent input types? The CLI rejects it, but the shared API still accepts and persists the value while effective-chain construction reads merge mode only from the agent binding. Direct API clients can therefore create hashed state whose configured value has no execution effect.

golem-schema/src/schema/tool/compatibility.rs
```diff
@@ -0,0 +1,1421 @@
...
+        if mode == ToolCompatibilityMode::StructuralSubtype
+            && !nb.constraints.iter().all(|c| eb.constraints.contains(c))
+        {
+            errors.push(err(
+                format_path(path),
+                "inner command adds a constraint not guaranteed by the expected command",
+            ));
+            continue;
+        }
+        if eb.stdin != nb.stdin {
```
noise64: > Could we compare only semantic `StreamSpec` fields and add a documentation-only regression test? `stdin` and `stdout` are compared directly, including prose, even though the compatibility modes otherwise erase documentation. Editing only stream docs can therefore reject deployment in strict, structural, and nominal modes without changing the runtime contract.

golem-schema/src/schema/tool/compatibility/mod.rs
```diff
@@ -0,0 +1,1421 @@
...
+                    cases: a
+                        .iter()
+                        .map(|n| b.iter().position(|x| x == n).unwrap())
+                        .collect(),
+                }
+            }
+            (SchemaType::Flags { flags: a, .. }, SchemaType::Flags { flags: b, .. })
+                if a.iter().all(|n| b.contains(n)) =>
+            {
+                ProjectionNode::Flags {
```
noise64, line 878: > Could we choose and document one index direction for enum and flags projections before runtime traversal is implemented? Enum stores target positions by source index, while flags stores source positions by target index and uses `usize::MAX` for missing values. A single representation with `Option` for absence would remove the sentinel and make traversal less error-prone.

golem-worker-executor/src/durable_host/tool/mod.rs
```diff
@@ -1549,8 +1558,55 @@ where
...
+                        .payload
+                        .clone()
+                        .unwrap_or_else(|| SchemaType::tuple(Vec::new())),
+                },
+            })
+            .collect(),
+    };
     let args = command.args;
 
+    if activation_snapshot
```
noise64, line 1605: > Could we reject nonempty middleware chains during deployment, or gate the manifest capability until executor traversal is available? The current flow accepts, compiles, persists, and exposes the chain, then deterministically returns an internal “not implemented” error on the first invocation. Failing closed is correct, but deployment success makes the application look runnable and delays an actionable configuration failure until production traffic.

cli/golem-cli/src/command_handler/app/mod.rs
```diff
@@ -1258,6 +1319,69 @@ impl AppCommandHandler {
...
+                        },
+                        publications,
+                    },
+                )
+                .await
+                .map_service_error()?
+                .publication_plan;
+            log_action("Planning", "tool middleware publications");
+            for entry in &publication_plan {
+                log_preformatted(format!(
```
noise64: > Could we include `entry.reason` when rendering middleware publication conflicts and cover the resulting diagnostic? The registry already tells the caller whether to choose a new version, disable `versionCheck`, or restore a de-published release, but the CLI prints only `name@version: conflict` and then exits with a generic message. Preserving that reason would make this workflow much easier to recover from.

cli/golem-cli/src/command_handler/app/deploy_diff.rs
```diff
@@ -115,6 +118,12 @@ impl DeployDiff {
             || !self.diff.mcp_deployments.is_empty()
             || !self.diff.remote_tools.is_empty()
             || !self.diff.published_tools.is_empty()
+            || !self.diff.remote_tool_middleware_deployments.is_empty()
+            || !self.diff.published_tool_middlewares.is_empty()
+            || self.diff.universal_tool_middlewares_changed
+            || self.diff.tool_compatibility_mode_changed
```
noise64: > Could we make tool compatibility mode an environment property, synchronized with the other environment deployment options, rather than treating a mode change as deployment content? The manifest declares `compatibilityMode` under `environments.<name>.deployment` alongside `compatibilityCheck`, `versionCheck`, and `securityOverrides`, but `DeploymentOptions::to_diffable`, `EnvironmentCreation` / `EnvironmentUpdate`, and `ensure_environment_deployment_options` omit it. The CLI instead threads it into each `DeploymentCreation`, so it appears here as `tool_compatibility_mode_changed`. Please persist and diff `tool_compatibility_mode` on the environment, include it in environment creation/update and the environment-property synchronization step, and have deployment planning use that effective environment setting. While doing so, naming the manifest/API field `toolCompatibilityMode` would also make its middleware-specific scope clear.

cli/golem-cli/src/command_handler/component/staging.rs
```diff
@@ -104,8 +105,10 @@ impl ComponentDiff {
         match self {
             ComponentDiff::All => true,
             ComponentDiff::Diff { diff } => {
-                !diff.agent_type_provision_config_changes.is_empty()
+                diff.wasm_changed
```
noise64: > Could we remove `diff.wasm_changed` from `provision_config_changed`, or add a focused test and explanation for the additional invariant it serves? WASM content and agent-type updates are already staged independently through `open_wasm_if_changed` and `agent_types_if_changed`. With no provision-config diff, letting a WASM-only change through this gate just traverses the manifest-file machinery and produces the same empty `ChangedComponentFiles` that the early return would produce. Keeping this predicate limited to actual provisioning changes makes its contract clearer and avoids unnecessary work on every WASM update.

cli/golem-cli/src/model/app_raw.rs
```diff
@@ -289,6 +326,79 @@ pub struct RegistrySubjectById {
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct ToolMiddlewareDeclaration {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub component: Option<ComponentName>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub release: Option<ToolMiddlewareRegistrySubject>,
+    #[serde(default, skip_serializing_if = "LenientTokenList::is_empty")]
+    pub templates: LenientTokenList,
+    #[serde(flatten)]
```
noise64: We (I) had many UX / DX problems related to error reporting and validation using flatten, so please let's not reintroduce it.
> Could we avoid `#[serde(flatten)]` for `ToolMiddlewareDeclaration::properties` and add a Rust parsing test that rejects a misspelled middleware property? This file already keeps the analogous component and agent layer helpers outside Serde parsing because flatten weakens `deny_unknown_fields`; the middleware declaration should preserve the same invariant. The later `AppBuilder` validation correctly covers component/release exclusivity, names, references, installations, and publication rules, but it cannot recover a manifest field silently discarded during deserialization. The JSON Schema check is useful as an additional diagnostic, not a substitute for rejecting the same invalid input in Rust.

golem-schema/src/schema/tool/compatibility/mod.rs
```diff
@@ -0,0 +1,1421 @@
...
+}
+fn err(path: impl Into<String>, message: impl Into<String>) -> ToolCompatibilityError {
+    ToolCompatibilityError {
+        path: path.into(),
+        message: message.into(),
+    }
+}
+
+#[cfg(test)]
+mod tests;
```
noise64, line 1483: >Could we use the repository's directory module layout consistently for the new external test modules, moving each parent implementation into its directory as `mod.rs`? This PR introduces all six current flat-parent `mod tests;` cases: `schema/tool/{compatibility,validation,wit}`, `services/deployment/tool_middlewares`, `services/environment_tool_middleware_grant`, and `command_handler/app/tool_middleware`. The repository >overwhelmingly uses the directory layout for modules with children; applying it to all six additions avoids establishing a second layout convention in the same change.

golem-registry-service/db/migration/postgres/036_tool_middleware_releases.sql
```diff
@@ -0,0 +1,78 @@
...
+    CONSTRAINT tool_middleware_releases_origin_check CHECK (origin IN (0, 1))
+);
+
+CREATE UNIQUE INDEX tool_middleware_releases_owner_name_version_uk
+    ON tool_middleware_releases (owner_account_id, tool_middleware_name, middleware_version)
+    WHERE lifecycle != 2;
+CREATE INDEX tool_middleware_releases_component_revision_idx
+    ON tool_middleware_releases (component_id, component_revision);
+
+CREATE FUNCTION validate_tool_middleware_release_component_owner() RETURNS TRIGGER AS $$
```
noise64: > Could we avoid introducing a PostgreSQL function/trigger and a separate pair of SQLite triggers for release-owner validation, and enforce this invariant in shared transactional repository logic instead? Migration 035 added the analogous tool-release triggers only very recently, so copying that pattern here would establish a database-trigger convention with backend-specific DDL and error behavior. The service already validates middleware ownership before publication; any repository-level protection needed for other callers could use a shared check in the release creation transaction. If a concrete race requires database enforcement beyond that, please document it and introduce one deliberate, consistently tested strategy for both tool and middleware releases rather than another ad hoc trigger pair.

golem-registry-service/src/repo/environment_tool_middleware_grant.rs
```diff
@@ -0,0 +1,936 @@
...
+    fn with_ro(&self, api_name: &'static str) -> DBP::LabelledApi {
+        self.db_pool.with_ro(METRICS_SVC_NAME, api_name)
+    }
+
+    fn with_rw(&self, api_name: &'static str) -> DBP::LabelledApi {
+        self.db_pool.with_rw(METRICS_SVC_NAME, api_name)
+    }
+}
+
+impl DbEnvironmentToolMiddlewareGrantRepo<PostgresPool> {
```
noise64: > Could we share the grantability query and decision path between PostgreSQL and SQLite, isolating only the locking operation that actually differs? `grantable_release_exists` and `grant_has_available_release` are repeated in large backend-specific blocks, with the meaningful PostgreSQL distinction being `FOR SHARE`. Keeping the eligibility predicates, bindings, and sequencing duplicated makes future behavior drift likely. A shared query builder/helper plus a small backend locking primitive would keep the semantics in one place; please also retain parity tests for both repositories.

golem-schema/src/schema/tool/compatibility.rs
vigoo: Implemented in [the review-fix commit](https://github.com/golemcloud/golem/commit/3c8af14f1). Completed source/target named-type pairs now reuse the same projection node, while active pairs still close recursive references. Compilation checks a 4,096-node budget before allocation and a 128-level depth budget before descent, returning ToolCompatibilityError rather than expanding without a bound. Added compact-diamond sharing, both sides of each budget boundary, and recursive directional-projection regressions. The schema suite passed (338 tests).

golem-registry-service/src/services/deployment/tool_middlewares/mod.rs
vigoo: The server boundary already rejects middleware_merge_mode on environment bindings in validate_tool_binding_input; that validation applies to both local and remote bindings before persistence. Added focused deployment-validation tests for explicitPrepend through both paths in [the review-fix commit](https://github.com/golemcloud/golem/commit/3c8af14f1). The pure chain compiler remains responsible for compiling validated input rather than duplicating the API validation. Post-merge registry unit tests pass (171), as do the repository/integration tests (184).

golem-schema/src/schema/tool/compatibility.rs
vigoo: Fixed in [the review-fix commit](https://github.com/golemcloud/golem/commit/3c8af14f1): stream comparison now compares presence, MIME type, and requiredness, excluding documentation. Tests change stdin and stdout documentation independently in strict, structural, and nominal modes and confirm those changes remain compatible; negative tests retain rejection for MIME and requiredness changes. The schema suite passed (338 tests).

golem-schema/src/schema/tool/compatibility/mod.rs
vigoo: Unified enum and flags projections as source-index → target-index mappings in [the review-fix commit](https://github.com/golemcloud/golem/commit/3c8af14f1). The usize::MAX sentinel is gone. Absence does not need an Option here: compatibility requires every source case/flag to exist in the target, and additional target flags have no source mapping and remain unset when traversal initializes the target flags. A regression uses source [b,a] and target [c,a,b] and asserts [2,1] for both representations. The representation documents this direction; runtime traversal remains the separate GOL439 work.

golem-worker-executor/src/durable_host/tool/mod.rs
vigoo: This split is deliberate: runtime middleware-chain traversal is assigned to GOL439 and is planned next. This PR provides registry/deployment metadata, validation, and compiled plans; the executor fails closed for a nonempty chain rather than silently invoking the leaf without middleware. After discussing the tradeoff, the decision is to keep that scope and not add a temporary deployment gate that would immediately need removal. Until GOL439 lands, a successfully deployed nonempty chain is not executable; the current error makes that limitation explicit.

cli/golem-cli/src/command_handler/app/mod.rs
vigoo: Fixed in [the review-fix commit](https://github.com/golemcloud/golem/commit/3c8af14f1). Middleware publication conflicts now retain entry.reason in the CLI diagnostic alongside name@version, preserving the registry advice instead of replacing it with a bare conflict label. Added diagnostic coverage for the rendered reason. The full post-merge CLI library suite passes (635 tests).

cli/golem-cli/src/command_handler/app/deploy_diff.rs
vigoo: Implemented environment authority in [the review-fix commit](https://github.com/golemcloud/golem/commit/3c8af14f1): toolCompatibilityMode is persisted on the environment, included in creation/update and environment-property diff/synchronization, and read by deployment planning from the effective environment revision. Immutable deployment snapshots still retain the resolved mode and compiled plans; changing the setting does not silently recompile an active revision.

The manifest spelling remains deployment.compatibilityMode. [Agent Tools specification, part 3](https://linear.app/golem-cloud/document/agent-tools-imported-specification-part-3-of-3-7a6ce17c8422), §6.4.3, explicitly specifies the deployment block’s compatibilityMode field alongside versionCheck and securityOverrides; §6.4.2 supplies it to phase 4. That specifies manifest placement, not persistence ownership, so environment authority and immutable snapshots can coexist. Renaming the manifest field would be a deliberate spec change, not a correction.

Mode-only edits follow the existing environment-option workflow: environment sync-deployment-options applies them explicitly, or the next content deployment synchronizes them. They are no longer treated as deployment content just to force a deployment. Tests cover persistence, revision-sensitive planning, CLI synchronization, and immutable snapshots; post-merge registry and CLI suites pass.

cli/golem-cli/src/command_handler/component/staging.rs
vigoo: Removed diff.wasm_changed from the provision-config gate in [the review-fix commit](https://github.com/golemcloud/golem/commit/3c8af14f1). WASM staging still follows open_wasm_if_changed independently. Focused staging tests cover WASM-only and provisioning-only changes so one path cannot accidentally depend on the other; the full post-merge CLI library suite passes (635 tests).

cli/golem-cli/src/model/app_raw.rs
vigoo: Removed serde(flatten) from ToolMiddlewareDeclaration in [the review-fix commit](https://github.com/golemcloud/golem/commit/3c8af14f1). The declaration now spells out its seven provisioning properties and converts explicitly to the internal layer, following ToolPreset. Added Rust YAML coverage for valid properties, a misspelled property, and serialization round trips, alongside schema parity checks. This makes deny_unknown_fields enforcement direct; it does not rely on a later JSON Schema check to recover invalid input. Post-merge CLI parsing/schema tests pass as part of the 635-test library suite.

golem-schema/src/schema/tool/compatibility/mod.rs
vigoo: Applied the directory-module layout to all six named modules in [the review-fix commit](https://github.com/golemcloud/golem/commit/3c8af14f1): compatibility, validation, and wit under schema/tool; deployment/tool_middlewares; environment_tool_middleware_grant; and app/tool_middleware. Their implementations now live in the corresponding mod.rs files, keeping the child test modules and registrations intact. Schema, registry, and CLI tests pass after the moves and subsequent main integration.

golem-registry-service/db/migration/postgres/036_tool_middleware_releases.sql
vigoo: Replaced the owner triggers in migrations 035/036 for both PostgreSQL and SQLite with shared transaction-local repository validation in [the review-fix commit](https://github.com/golemcloud/golem/commit/3c8af14f1). Both tool and middleware release creation/restoration paths validate the component source owner within the write transaction and return an explicit SourceOwnerMismatch error; the existing relational constraints remain.

The supported repository APIs do not mutate the component → environment → application → account ownership links, so an ownership transfer cannot race this check through a supported write path. No separate backend-specific trigger strategy is needed for that contract. Tests exercise mismatched owners and restoration on both backends; the post-merge registry integration suite passes all 184 tests, including PostgreSQL, PostgreSQL TLS, and SQLite.

golem-registry-service/src/repo/environment_tool_middleware_grant.rs
vigoo: Shared the release-grant eligibility queries, bindings, and decision path using the existing release/grant lifecycle machinery in [the review-fix commit](https://github.com/golemcloud/golem/commit/3c8af14f1). Only backend locking remains specialized; PostgreSQL retains FOR SHARE inside the grant transaction so de-publication cannot pass between eligibility validation and insertion. The shared rules retain de-publication and supersession behavior.

Both-backend lifecycle tests pass. A PostgreSQL race regression runs real middleware grant creation, blocks its insertion at the environment foreign key, and verifies concurrent de-publication waits for the eligibility lock before both operations finish. The full post-merge registry integration suite passes (184 tests).

vigoo:
@noise64 The agreed fixes from your review are pushed, and all 12 inline comments have substantive replies explaining the changes or agreed scope decisions. Ready for another look. Latest head: [8b5d83e26](https://github.com/golemcloud/golem/commit/8b5d83e26e89577da0897b779743e60e1608482a).

[Docs](https://github.com/golemcloud/golem/actions/runs/35008911230) is green. [CI](https://github.com/golemcloud/golem/actions/runs/35008911235) is still running with no failures currently reported; the two previously failing shards (bridge generation and worker group 1) have passed on this head.

noise64 (approved):
I still think that we should rename compatibilityMode in the manifest too (that is what users see), otherwise looks good!

PR #3843, opened 2026-09-08
Fix RPC durability recovery at replay tails

noise64:
- resumes live RPC activation when replay reaches a settled durability tail, while preserving recorded authorization for an incomplete activation
- consumes recorded span completions during replay and appends a completion only after the cursor has genuinely exhausted and published live execution; mismatched replay entries remain strict failures
- adds deterministic coverage for recovery after RPC creation, invocation completion, and completion delivery, with control cases and exactly-once child mutation assertions
- validates focused recovery, authorization denial, cancellation, completion delivery, concurrent replay, replay mismatch, formatting, package checks, and scoped Clippy


vigoo:
I'm going to propose a more generic fix for this soon

noise64:
Closing in favour of https://github.com/golemcloud/golem/pull/3861

PR #3846, opened 2026-09-09
Port #3830 improvements to main

kmatasfp:
Not as big of an improvement here after #3820 and #3827  than it was on 1.5.x


Aditya1404Sal (approved):
looks good

PR #3848, opened 2026-09-10
Port #3720 to main: cap http benchmark retries and log failure bodies

kmatasfp:
Ports #3720 to `main`, then goes further after review: failed attempts are now part of the benchmark results, not only the logs.

The first commit is the clean cherry-pick (`invoke_and_await_http` on `main` was byte-identical to what the fix was written against).

The second commit addresses the review comment that logged failures are invisible in the daily run and retried invocations were folded into the measured numbers:

- Every failed attempt (non-success status, transport error, timeout) is recorded as a failure against its measurement key, with an exact count and up to five distinct sample messages (bodies truncated to 500 bytes). It lands in the results JSON as a new `failures` map on each run result, next to `duration_results`. Empty maps are not serialized, so old entries and the dashboard are unaffected.
- An invocation that only succeeded after a retry or timeout is recorded under `{prefix}invocation-recovered` instead of `{prefix}invocation`, so the primary series only measures first-attempt successes. The retry and timeout counts are still recorded as before.
- The text report prints a red `FAILURES` banner before and after the per-benchmark output, and each affected benchmark lists its failed attempts with their sample messages.
- The `benchmarks` binary exits with status 1 after the results have been written and printed when any failure was recorded. The daily workflow now marks that step `continue-on-error`, still commits the results, and fails in a final step, so the run goes red without losing the day's data.

A permanent 4xx or an exhausted retry budget still panics as in #3720; that already fails the run, and continuing past it would only send the remaining calls at a known-broken endpoint.

`invoke_and_await_agent` gets the same failure recording. Its unbounded retry loop is unchanged.

Once this lands the second commit needs porting back to `1.5.x`, where #3720 has the same gap.


vigoo:
I think we need to do something with the failed results, as in the automated daily benchmarks logging them is not noticeable, and otherwise the failing results are just counted to the normal measured results.

kmatasfp:
Addressed in 356c5b4. Failures are now recorded in the results themselves rather than only logged:

- each failed attempt is counted against its measurement key and lands in the results JSON as a `failures` map (count plus up to five sample messages), so it is preserved in `benchmark-results` too;
- invocations that only succeeded after a retry or timeout go to `{prefix}invocation-recovered` instead of `{prefix}invocation`, so the measured series no longer includes them;
- the report prints a red `FAILURES` banner at the top and bottom, and the binary exits 1 after writing the results. The daily workflow commits the results either way and then fails, so a run with failures shows up as a red workflow instead of a log line.

One thing worth your eye: I chose to keep recovered invocations out of the primary series rather than leave them in and only flag the run. If you would rather keep the old measurement semantics and rely on the banner and exit code alone, that is a two-line change in `InvokeResult::record`.

vigoo (approved):

PR #3867, opened 2026-09-11
Effect.ts SDK for Golem 1.6

vigoo:
Resolves GOL-338


cli/golem-cli/src/bridge_gen/typescript/effect_external.rs
```diff
@@ -0,0 +1,593 @@
...
+use crate::bridge_gen::parameter_naming::ParameterNaming;
+use crate::bridge_gen::{BridgeGenerator, bridge_client_directory_name};
+use camino::{Utf8Path, Utf8PathBuf};
+use golem_common::model::agent::{AgentConfigSource, AgentMode};
+use golem_common::schema::agent::{AgentMethodSchema, AgentTypeSchema, InputSchema, OutputSchema};
+use golem_common::schema::schema_type::SchemaType;
+use heck::ToUpperCamelCase;
+use serde_json::json;
+
+const EFFECT_VERSION: &str = "4.0.0-beta.98";
```
vigoo: Let's have the effect-specific modules in a separate `bridge_gen/effect` parent module, and also make sure we only have a single place where `EFFECT_VERSION` is specified

cli/golem-cli/templates/effect/default/component-dir/src/counter-agent.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    return {
+      value: () => Ref.get(state).pipe(Effect.map(({ count }) => count)),
+      increment: () =>
+        Ref.updateAndGet(state, ({ count }) => ({ count: count + 1 })).pipe(
+          Effect.map(({ count }) => count),
+        ),
+    };
+  }),
+  (_restoration, _id, snapshotting) =>
```
vigoo: This is very wrong, we are repeating the agent's implementation here! The `implement` DSL should somehow make it easy that you only have to write the constructor and the load-snapshot initialization path (optionally!) and never duplicate the implementation itself

vigoo:
## Why the Effect SDK migration adds a `Resumed` oplog entry

Reference explanation requested by Daniel Vigovszky (@vigoo).

**Effect itself does not require a new oplog entry. Its acceptance tests exposed a language-independent executor interrupt/resume bug, and the migration includes that executor fix.** This is not an Effect-specific execution mechanism.

### The problem: execution resumed, but durable status stayed `Interrupted`

Golem reconstructs agent status from the oplog. Interrupting an invocation recorded `Interrupted`, but resuming the same unfinished invocation did not record the opposite transition.

For example, an invocation waiting on a long timer had this simplified history:

```text
AgentInvocationStarted(key=K)
Start(clock-wait)
Interrupted
```

After a resume request, the executor could reconstruct the invocation and continue waiting. However:

- Replay does not append another `AgentInvocationStarted`: this is still invocation K.
- The original clock wait remains incomplete, so no new `End` exists yet.
- Status reconstructed from the oplog therefore still says `Interrupted`, even though the invocation has resumed.

This affects behavior, not just displayed metadata. The interrupt decision logic ignores an interrupt when status already says `Interrupted`, so **a second interrupt could do nothing**. Connecting to the agent event stream also rejects agents reported as interrupted.

### The solution: record admission to resume, not a new invocation

The executor now commits a timestamp-only `Resumed` entry when starting an unloaded durable agent with an interrupted, unfinished invocation. It commits the entry before loading/executing the instance.

```text
Before                                After
----------------------------------    ----------------------------------
AgentInvocationStarted(key=K)         AgentInvocationStarted(key=K)
Start(clock-wait)                     Start(clock-wait)
Interrupted                          Interrupted
<resume request: no entry>            Resumed

Status: Interrupted                  Status: Running
Second interrupt: ignored            Second interrupt: effective
```

`Resumed` means **the unfinished invocation was admitted to resume**, not that replay completed or the timer finished. The status reducer maps it to `Running`.

It is an oplog **hint**: durable lifecycle information that replay skips, rather than a guest host call that replay must match. The original invocation key and pending host-call `Start` remain unchanged.

### Why a dedicated entry?

| Alternative | Problem |
| --- | --- |
| Change only cached/in-memory status | Loses the transition when status is reconstructed from the oplog. |
| Append `AgentInvocationStarted` | Falsely records another invocation start for work already in progress. |
| Append `Restart` | Its existing status semantics are `Idle`, not continuation of this unfinished invocation. |
| Append `NoOp` | It is not a hint; it participates in positional replay. |
| Wait for the next ordinary host-call entry | Leaves status wrong while replay or the pending wait continues—the failing window. |

A dedicated hint expresses the missing durable transition. **Adding an enum case is a design choice, not an intrinsic Effect requirement.** Reworking an existing lifecycle entry would be another, broader contract change.

### Why this surfaced during SDK work, and comparison with TypeScript

The Effect skill-harness acceptance included interrupt/resume of a long-lived method that repeatedly updates state and calls `Effect.sleep`. That exercised the missing lifecycle transition.

The executor regression test uses a **non-Effect P3 host-API clock component** (`interrupt_while_parked_in_p3_sleep`), demonstrating that this is below the SDK layer. TypeScript and Rust agents reaching the same execution state are subject to the same underlying problem; the fix has no Effect-specific branch. It concerns Golem agent interruption/resumption, not Effect fiber cancellation or stream representation.

The regression verifies that:

- Resume reports `Running` while the original wait is still incomplete.
- A second interrupt records another interruption.
- After another resume and actual executor teardown/reconstruction, the original invocation and wait complete with exactly one invocation start/finish and one wait start/end.

### Relevant implementation and scope

- `golem-common/src/base_model/oplog/mod.rs`: timestamp-only `Resumed` hint.
- `golem-worker-executor/src/worker/mod.rs`: commit the hint when admitting an interrupted, unfinished durable invocation to resume.
- `golem-worker-executor/src/worker/status.rs`: fold `Resumed` into `Running`.
- `golem-worker-executor/src/worker/lifecycle.rs`: interrupt decisions based on durable status.
- `golem-worker-executor/tests/wasi.rs`: `interrupt_while_parked_in_p3_sleep` regression.

Adding an oplog variant required updating public/WIT representations and their consumers, including rebuilding the OTLP plugin. This should be understood as a separate executor bug fix discovered during SDK acceptance, rather than a requirement imposed by introducing the Effect SDK.

golem-skills/skills/effect/golem-read-only-method-effect/SKILL.md
```diff
@@ -0,0 +1,19 @@
...
+# Read-only Effect methods
+
+Set `readOnly` on a `method`; handlers must remain ordinary Effects and must not mutate state or perform writes.
+
+```ts
+value: method({ input: {}, success: Schema.Number, readOnly: true })
+cached: method({
+  input: {},
+  success: Schema.String,
+  readOnly: { cache: { ttlNanos: 5_000_000_000n }, usesPrincipal: false },
```
vigoo: usesPrincipal should not be user-defined, it should be set in the metadata based on whether the method is having a Principal parameter or not

golem-skills/tests/harness/src/migrate-effect-skills.ts
```diff
@@ -0,0 +1,1280 @@
+import { spawn } from "node:child_process";
+import * as fs from "node:fs/promises";
+import * as path from "node:path";
+import { isDeepStrictEqual, parseArgs } from "node:util";
+import { fileURLToPath } from "node:url";
+import * as yaml from "yaml";
+import { z } from "zod";
+
```
vigoo: What is this file for? Was it a temporary thing for the migration, or do we need to keep it for anything?

vigoo:
## Executor change 1/6: Ignite MERGE authorization

Reference explanation requested by Daniel Vigovszky (@vigoo). This is a platform fix exposed by Effect SDK acceptance, not an Effect-specific executor API.

### Problem

Ignite accepts this upsert syntax:

```sql
MERGE INTO ignite_counters (id, count) VALUES (?, 0)
```

The generic SQL parser used to extract permission targets did not accept it, so authorization rejected a valid operation before it reached the database.

### Solution

`src/durable_host/authorization/targets.rs` now tries normal parsing first, then handles a restricted Ignite MERGE shorthand by parsing an INSERT-shaped equivalent **for permission analysis only**. The SQL executed by Ignite is unchanged.

For example:

```sql
MERGE INTO destination (id, count)
SELECT id, count FROM source
```

requires `Mutate` on `destination` and `Query` on `source`. The fallback rejects multiple statements and unsupported clauses, and does not broaden PostgreSQL or MySQL parsing.

### Why it entered the migration and verification

Effect Ignite acceptance exercised an existing database operation that the executor could not classify. Four unit tests cover parameterized/quoted destinations, SELECT sources, malformed or unsupported statements, and isolation from other dialects.

All source paths above are relative to `golem-worker-executor/`.

vigoo:
## Executor change 2/6: Fire-and-forget RPC replay and span crash windows

Reference explanation requested by Daniel Vigovszky (@vigoo).

### Problem

The serialized fire-and-forget RPC path records:

```text
Start(RPC) → StartSpan → End(RPC) → FinishSpan
```

Previously, replay tried to resolve the RPC result before reconstructing `StartSpan`. It could block waiting for `End` while the unconsumed span entry prevented replay from advancing.

Partial histories also need explicit handling:

| Committed history at crash | Required recovery |
| --- | --- |
| Start only | Admit repair of the incomplete RPC, then create the missing span. |
| Start → StartSpan | Reconstruct the span, then repair the incomplete RPC. |
| Start → StartSpan → End | Reuse the recorded result and finish the span without dispatching again. |

### Solution

`src/durable_host/wasm_rpc/mod.rs` reconstructs the span before resolving the result when recorded span history exists. A Start-only tail uses the normal incomplete-call eligibility and live-admission checks. When a recorded End exhausts replay, the executor performs the checked transition to live before appending a missing FinishSpan.

**The original RPC Start index and derived idempotency key remain unchanged.** This neither tolerates mismatched replay nor permits completed RPCs to execute again.

### Why it entered the migration and verification

Fire-and-forget RPC recovery was exercised during acceptance. Fixing ordinary replay also required covering crashes between individual durable writes.

Two tests in `tests/rpc.rs` cover completed replay and crashes after Start, StartSpan, and End:

- `completed_fire_and_forget_rpc_replays_span_before_result`
- `fire_and_forget_rpc_recovers_from_committed_crash_prefixes`

They count target-side increments and verify preservation of the original RPC identity, one terminal, and one completed span. This is language-independent RPC behavior.

All source paths above are relative to `golem-worker-executor/`.

vigoo:
## Executor change 3/6: Report automatically flushed oplog entries to explicit commit callers

Reference explanation requested by Daniel Vigovszky (@vigoo).

### Problem

Appending an entry can automatically flush the oplog buffer when its threshold is reached. However, callers use the entries returned by explicit commit() to update derived worker status.

Previously:

```text
append A, B → automatic flush persists A, B
explicit commit() → returns empty
status reducer → does not receive A, B through this commit
```

Persistence succeeded, but incremental status accounting could miss those entries.

### Solution

`src/services/oplog/primary.rs` distinguishes the last persisted entry from the last entry returned by an explicit commit. Explicit commit returns previously persisted-but-unreported entries together with newly persisted entries. A subsequent commit returns nothing unless more entries appeared.

The internal flush used by `wait_for_replicas()` is separate from the reporting commit: a replication barrier makes data durable without consuming the report intended for the status reducer.

### Why it entered the migration and verification

This addresses another way lifecycle/status accounting can miss durable entries. It is distinct from the Resumed fix: recording the right transition is insufficient if the incremental reducer never receives it.

Tests in `src/services/oplog/tests.rs` force automatic commits with a low threshold, verify exact entries and append receipts, and check that reports are returned once:

- `explicit_commit_reports_threshold_commits_once_and_preserves_add_receipts`
- `wait_for_replicas_does_not_consume_explicit_commit_report`

This is shared oplog behavior, not an Effect-specific persistence mechanism. Change 4 explains the archive-boundary safeguard needed by this reporting design.

All source paths above are relative to `golem-worker-executor/`.

vigoo:
## Executor change 4/6: Preserve unreported oplog entries across archiving

Reference explanation requested by Daniel Vigovszky (@vigoo). This is a companion to change 3, explicit-commit reporting of automatically flushed entries.

### Problem

The reporting fix rereads persisted-but-unreported entries from primary storage. Archiving those entries first would remove the data it needs. The previous archive boundary tracked additions rather than the prefix returned by explicit commits.

### Solution

`src/services/oplog/multilayer.rs` transfers entries only through the last reported commit index, rather than the latest appended index:

```text
Persisted:       A B C D
Commit reported: A B
May archive:     A B
Must retain:         C D
```

The boundary advances monotonically. Later additions cannot accidentally become eligible just because an earlier outer commit finishes asynchronously. Background archiving indicates that further work remains when newer entries are not yet eligible.

### Tradeoff and verification

Some persisted entries remain in primary longer. This changes archive eligibility, not durability or oplog indices.

`archiving_auto_committed_entries_does_not_consume_explicit_commit_report` in `src/services/oplog/tests.rs` covers archiving before explicit reporting and an interleaving where later automatic commits occur before an earlier outer commit finishes.

This safeguard was necessary to keep the reporting fix correct with the existing multi-layer storage architecture; it is not a new SDK feature.

All source paths above are relative to `golem-worker-executor/`.

vigoo:
## Executor change 5/6: Settle streaming-output tail work before invocation completion

Reference explanation requested by Daniel Vigovszky (@vigoo).

### Problem

Materializing a durable streaming result can leave Store tasks with remaining durable work. Marking the invocation finished too early allows positional host-call Start/End entries to appear after AgentInvocationFinished, violating the invocation replay boundary.

### Solution

After materializing a durable streaming result, the executor runs the existing `run_guest_call_settled` mechanism before declaring success. This applies to normal execution in `src/worker/invocation_loop.rs` and recovery in `src/durable_host/mod.rs`.

Materialization finishes first; a separate ready-root settlement pass then handles remaining tail work. Settlement errors are handled as failures, not ignored.

```text
Materialize streaming result
→ settle remaining positional tail work
→ record AgentInvocationFinished
→ finish protocol/session bookkeeping
```

### Why it entered the migration and verification

Streaming acceptance exercised completion and early-drop paths, rather than only full stream consumption. This is a language-independent invocation-boundary fix, not an Effect-specific stream representation.

A TypeScript-only regression in `tests/rpc.rs`, `typescript_early_output_drop_releases_streaming_cleanup_before_next_invocation`, drops an output early and immediately invokes the target again. It verifies completion of the next call, durable session completion, and absence of positional calls after invocation completion.

`host_only_materialization_waits_before_ready_root_settlement` in `tests/concurrent_delivery_order.rs` checks the materialization/settlement sequencing with a bare Wasmtime Store. Change 6 explains the complementary session-cleanup progress fix.

All source paths above are relative to `golem-worker-executor/`.

vigoo:
## Executor change 6/6: Keep the Store event loop running during streaming-session cleanup

Reference explanation requested by Daniel Vigovszky (@vigoo).

### Problem

A pending host stream operation can retain a lock needed by streaming-session cleanup. Awaiting cleanup outside the Store event loop can deadlock:

```text
Cleanup waits for a lock
→ the lock holder needs another Store poll
→ the executor waits for cleanup instead of polling the Store
```

AgentInvocationFinished may already be recorded, so metadata can say Idle while the next invocation remains stuck.

### Solution

`complete_durable_streaming_session` now runs inside `store.run_concurrent`, keeping Store tasks progressing while cleanup waits. Both normal execution (`src/worker/invocation_loop.rs`) and replay-to-live completion (`src/durable_host/mod.rs`) use this approach.

Together with change 5, ordering remains:

```text
Positional tail work settles
→ AgentInvocationFinished
→ protocol/session-completion hints
→ next invocation
```

Protocol hints may follow AgentInvocationFinished; positional durable-call Start/End entries may not.

### Why it entered the migration and verification

Streaming acceptance exposed the early-output-drop cleanup path. The TypeScript-only regression `typescript_early_output_drop_releases_streaming_cleanup_before_next_invocation` verifies that dropping output early does not block the next call, that the streaming session is durably finished before that call starts, and that cleanup does not append positional calls after invocation completion.

This is shared executor scheduling behavior. It does not add an Effect-specific scheduler or cancellation protocol.

### Scope and CI context for these six comments

These platform fixes accompany SDK acceptance and should be reviewed as platform changes. Supporting executor changes consist of regression tests, walkthrough documentation, Resumed conversion/classification plumbing explained in the earlier comment, and removal of an unnecessary AccountId.clone() in a new test for clippy.

The completed CI run passed the worker and unit suites. The separate shard-reassignment/concurrent-oplog-writer failure observed on its first attempt was not fixed by these six changes; that job passed on an unchanged-commit rerun.

All source paths above are relative to `golem-worker-executor/`.

sdks/effect/integration-test/.agents/skills/golem-add-agent-effect/SKILL.md
```diff
@@ -0,0 +1,163 @@
+---
```
vigoo: We should not duplicate these skills in the integration-test - let's git-ignore the whole .agents/skills dir

sdks/effect/src/internal/agentStream.ts
```diff
@@ -0,0 +1,332 @@
...
+  static fromEffect<T, E, R>(
+    source: EffectStream.Stream<T, E, R>,
+  ): Effect.Effect<AgentStream<T>, never, R> {
+    return Effect.map(Effect.context<R>(), (context) =>
+      AgentStream.from(new AbortableStreamIterable(source, context)),
+    )
+  }
+
+  /** View this affine stream as an Effect Stream; early termination closes it. */
+  toEffect<E>(onError: (cause: unknown) => E): EffectStream.Stream<T, E> {
```
vigoo: Do we need this AgentStream wrapper? Can't we use EffectStream.Stream directly to represent the streams in our agents and tools?

PR #3872, opened 2026-09-12
Capture and restore agent filesystems in the lifecycle

kmatasfp:
Resolves GOL-577, delivery step T3 of Filesystem snapshotting.

Gives the agent-filesystem lifecycle the two operations that snapshotting needs: `capture`, which copies a resident filesystem at a boundary, and a baseline that can start from such a copy instead of from the initial files alone. Restoring a capture and replaying the whole oplog must leave the same tree, so the rule that installs initial files is rewritten once and used by materialisation, automatic and manual updates, and entity provisioning alike. Nothing calls `capture` in production yet; T5 wires it up.

Read-only initial files also get their 1.5.x semantics back, and a permission error from an agent's call now fails that call instead of invalidating the generation.

## Where to start

The initial-file rule is the heart of the change, and everything else reads more easily after it.

1. `lifecycle/initial_files.rs`: `rule` and `plan` are pure and short, and the doc on `plan` states the rule in full. `observe` turns the tree into the `PathState` that `plan` reads.
2. `lifecycle/baseline.rs`: `capture`, and `materialize_baseline` with and without a restore.
3. `lifecycle/mod.rs`: `execute_coordinated_open` and `coordinated_target` for the read-only checks, and `invalidates_generation` for the permission-error classification.
4. `sandbox_filesystem/adapter.rs`: `seed` and `copy_contents`.
5. `lifecycle/tests/capture_and_baseline.rs`: the property test and the reference model written from the issue text.

## Capture

- `capture` copies a resident filesystem at a boundary.
  - It holds the read guard of the sandbox for the whole copy. `delete` takes the write guard after the calls and nodes drain, so a delete that starts during a capture waits for the copy instead of failing at the exclusive-ownership check.
  - It stops new filesystem calls and waits up to a deadline for open calls; after the deadline it returns `CaptureError::Busy`.
  - It copies the tree with `copy_contents` into `.scratch/<uuid>/tree/` and writes `record.json` next to it. The record holds the paths whose bytes the tree leaves out, the hard-link groups, and the initial and provisioned declarations.
  - The comment on why an unpolled WASI stream write can hold a lease across a boundary stays at the deadline.

## Restore and the baseline

- `materialize_baseline(filesystem, prepared, Option<RestoreTree>)` replaces `materialize_initial_files`. With a restore:
  - It restores into its own `.scratch` directory, seeds `tree/` (`FromSource`, `CreateNew`) and makes the recorded hard links.
  - It seeds every left-out file from its recorded declaration (`ReadOnly`, `CreateNew`). Then it gives the root, and each directory that a link or a left-out seed changed, the modification time that `tree/` records for it. The tree then equals the captured tree, except in the times of the left-out files.
  - `copy_contents` now gives its destination the permissions and the modification time of the source root, so `tree/` records the time of the root.
  - A seed of `tree/` that fails after a part of the tree is made keeps that part. Before each retry, `seed_with_retry` removes all that is under the root, which held nothing before, so the retry runs on an empty root with `CreateNew`. A single-file seed needs nothing before a retry, because it persists a temporary file in one step. When the decision is a failure, the partial tree goes away with the sealed filesystem.
  - It runs the same `observe` and `install` as an automatic update, from the recorded declarations to the current ones, and discards the directory. So a manual update loads the old content of a left-out file that it then replaces or removes.
  - A failed discard is only logged, because startup cleans `.scratch`.

## One initial-file rule

- One rule decides every install: materialisation, automatic and manual update, and entity provisioning.
  - The rule acts only at a path whose declarations differ. The install never reads a path with an equal declaration, so a read-write file that the agent changed stays when the new revision declares it as the old one does.
  - At a path whose declarations differ, the install expects what the old declarations left there: Golem's file where they declare the path, and nothing where they do not. A path in that state gets what the new declarations give: `Seed(CreateNew)`, `Seed(Replace)` or `Unlink`. An empty path that the new declarations do not have stays empty. Anything else is a conflict: the whole install fails, names the first conflicting path, and changes nothing.
  - Read-only and read-write declarations follow the same rule. A read-write file that the agent did not change follows the new declaration. A read-write file that the agent changed or deleted is a conflict when its declaration changes or goes. Anything at a path that only the new declarations have is a conflict too.
  - The install decides each path against the tree as its own removals leave it. Golem's file that it unlinks does not block a path under it. A directory that holds only Golem's files that it unlinks does not block a new file at its path. `apply` unlinks, then removes such directories deepest first, then seeds. So one update turns a declared file into a directory of declared files, and back.
  - `initial_files::plan` is pure. `install` loads every source before its first change, so a conflict or a failed load changes nothing. A failure after that point invalidates the generation.
  - Two costs follow from the rule:
    - `observe` reads the whole content of a read-write file whose size equals the old declaration, when that declaration changes or goes. The inode shortcut covers only files without write bits.
    - `observe` lists each directory at a path that only the new declarations have, and each directory under it, and stops at the first entry that blocks.

- Golem's file at a path is a regular file whose content equals the old declaration. For a read-only declaration it also has no write bits.
  - The decision reads only the tree and the declarations. A restore therefore decides as a replay does.
  - Each install records the file's inode. `holds_golem_file` skips the blake3 hash for a file that still has that inode and no write bits.
  - The capture leaves out such a read-only file only while it has one name.
  - `SandboxAttributes` no longer carries a creation time.

- A conflicting install has its own error, `Error::InitialFileConflict`, in place of a `FilesystemStorageError`. The message names the path and what is at it: another object, a file above the path, or nothing where the old declarations left Golem's file. The install plans the whole tree before it loads a source, so a conflict changes nothing. The text a user read before was "failed to install initial files because of a conflict at filesystem &lt;path&gt;": the `Display` of `FilesystemStorageError` puts "filesystem" between the operation and the path, so no message routed through that type can read as a sentence. A conflict cannot reach a guest descriptor, so p2 and p3 trap on it with the other internal lifecycle failures.

## Read-only initial files

- They have the 1.5.x semantics again. Their contents and times cannot change, and their descriptors report no write flag. Delete, rename, hard link, and a move of a directory above one all work.
  - The check reads the file's permission bits. `authorize_writable_target` checks opens for write, create or truncate, and path time changes. `authorize_attribute_target` refuses set-size and set-times through a read descriptor.
  - A resolved target carries the permission bits from its resolution. `coordinated_target` coordinates the entry, and for a path time change or a writable open resolves it again under the coordination and authorizes and runs from that resolution. Before, a rename that put a read-only file at the path between the first resolution and the coordination could get its times changed. An open that needs no authorization resolves once.
  - A time change that follows a final symlink opens the object that the symlink names, under the coordination on the entry, and checks its permission and changes its times through that descriptor. The coordination covers only the entry, so a rename onto the symlink's referent could otherwise complete between the check and the change; the owner may set the times of a read-only file, so the kernel would not stop it. A time change without follow keeps the path form: the coordination stops a rename onto the entry. The open expects the kind that a read gave; a mismatch restarts the operation at most twice.
  - A writable open has the same window. The open itself pins the object, and the adapter already reads the metadata of the opened object to check its kind, so `SandboxOpened` now carries whether that object is a regular file without write permission. Every open that needs a mutable target refuses such an object with `NotPermitted` after the open, at no extra syscall. An executor that is not root gets EACCES from the kernel in that window, which now gives `not-permitted` too; the check matters for an executor that runs as root, which the kernel does not stop.
  - `authorize_policy_target` is deleted. It resolved every declared read-only path on each edit and broke writes after a directory move.

## Permission errors from an agent's call

- Such an error fails only that call. The guest gets `not-permitted`, and the generation stays valid.
  - Since #3728 the lifecycle classified an operation's error with `is_terminal_failure`, which counts `PermissionDenied`, so the call invalidated the generation and trapped. A hard link of a directory is one such refusal. A writable open through a symlink is another, on an executor that is not root, when a concurrent rename puts a read-only file behind the symlink after the check.
  - `classify_failure` and `classify_query_error` now ask `invalidates_generation`, which leaves out a permission error. The error is then a guest failure with no effect, and `returned_storage_error` turns it into `NotPermitted`, which the read-only checks give too, so an executor that runs as root and one that does not answer the same. The hard-link special case that read the source to find a directory is deleted, so a hard link of a directory gives `not-permitted` again, as on 1.5.x, on every platform.
  - `is_terminal_failure` does not change. The postcondition reads of a failed operation follow the agent's rule too, because they are a part of the agent's call. Only the usage reads after a storage exhaustion keep `is_terminal_failure`, because they read the storage of the executor. A permission error from the XFS usage read, for example after the executor lost `CAP_SYS_ADMIN`, still invalidates the generation.
  - A permission error stays terminal in the executor's own setup: the component cache, the initial-file cache, the managed XFS projects and the directories of an unmanaged filesystem. Those paths do not use this classification. EIO, ESTALE, ENODEV, a read-only filesystem, invalid data and a failed native task still invalidate the generation.
  - An initial-file seed that gets a permission error fails with `NotPermitted`. An install still invalidates the generation after any failed step. The seed of a restored tree runs outside an install and does not. A restored hard link that the sandbox refuses gives `Error::Sandbox`, because that step returns its error directly; both fail the start.
  - `p3_agent_write_result` takes its errors from `p3_agent_error`, as a P2 write takes them from `p2_agent_error`. Before, it had its own copy of the mapping, and that copy trapped on `NotPermitted`. A side effect: a cross-device error without an OS error number now gives `cross-device` instead of `io`, as in P2.
  - The read that a failed namespace operation makes to check its postcondition needs no check of its own. That read runs only when the error of the operation does not prove that nothing changed, so a failed read leaves the effect unknown, and an unknown effect invalidates the generation. The guard that invalidated early is gone from `execute_create_directory`, `execute_create_symlink`, `execute_hard_link`, `execute_rename` and `execute_remove`, and the reason sits on `mutation_failure_evidence`. `execute_open` keeps its guard, because its read runs for every creating open, also when the error of the open proves that nothing changed.
  - `error_proves_no_effect` counts a permission error as no effect, because the kernel checks permission before it changes anything. The exception is an LSM or a fanotify listener that refuses an open after the kernel created the file: the guest gets `not-permitted`, and the empty file stays.

## Every file an agent creates is writable by its owner

- On Unix, `AgentFilesystems::new` clears bit 0o200 of the process umask and keeps the other bits. Linux reads the mask from `/proc/thread-self/status`, which is the mask of the calling thread and, in the executor, of the process; other Unix platforms probe it with 0o022.
- Every process that runs agents starts that service: the worker executor, `golem server run` and the debugging service.
- On Windows an agent cannot set the read-only attribute, so no change is needed.
- A test runs `AgentFilesystems::new` under umask 0o227 and expects 0o027. Other tests of the binary change the process mask, so the test runs on a thread that first takes its own copy of the filesystem attributes with `unshare(CLONE_FS)`; its mask is then private to it. No test spawns a process. That is why the Linux reader uses `/proc/thread-self/status`: `/proc/self` is the thread group leader.

## Sandbox and state

- `seed` took over the install, so `sandbox_filesystem/file_update.rs` and its rollback are deleted. No half-applied tree stays in use: a failed automatic update records the failed update and stops the start (`finalize_pending_automatic_update_access`, `on_worker_update_failed`), and a failed provisioning invalidates the generation.
- `copy_contents` copies a regular file or a symlink with several names once, and returns `Box<[LinkGroup]>`.
  - The restore makes the other names with `hard_link`.
  - The sandbox `hard_link` does not follow a symlink at its source, as POSIX `linkat` without `AT_SYMLINK_FOLLOW` does. Any caller can therefore give a symlink several names, and a copy that gives back the same tree must keep them. `sandbox_filesystem` stays independent of its callers.
  - `OnExisting::Keep` is removed. `OnExisting` is renamed to `SeedPlacement`, with the variants `CreateNew` and `Replace`, so that a seed that normally succeeds does not read as a failure.
- The generation holds its initial-file state as `Mutex<Arc<InitialFileState>>`, and the state holds `Arc<Declarations>` for the initial and the provisioned declarations and `Arc<Path>` keys for the installed files. An install builds a new state that shares what it keeps. The functions of the rule take borrowed declaration views, so an install clones no declaration. `Declarations` is `HashMap<Arc<Path>, Arc<InitialAgentFile>>`, so the provisioned map that a provisioning merges shares each entry with the map before it.
- `.scratch` is made once in `AgentFilesystems::new`, which is now async, and so are `ActiveAgents::new`, `new_with_probe` and `Bootstrap::create_active_agents`. golem-cloud does not depend on golem-worker-executor.

## How it is tested

A property test runs random histories on unmanaged storage: initial files, filesystem operations, component updates and entity provisioning. It has two oracles. A replay of the whole history is the oracle for a restore at the capture point and for a manual update from that restore, and a reference model written from the issue text is the second. Right after the restored agent starts with the declarations of the record, its tree, with the modification times of the root, of every directory and symlink, and of every file that is not left out, must equal the tree of the captured agent at the capture.

The generator moves read-only files, and the directories above them, away before the capture and back after it, nests declared paths across revisions, hard links directories, and revises the first declarations so each one stays, changes content, changes permission, or goes. It runs 2048 cases with a fixed seed in about 16 s; `PROPTEST_CASES` overrides the count.

## Notes for review

- **A truncating open still truncates at the native open.** The review asked for the truncation to move onto the pinned descriptor, after the read-only check, because a concurrent rename can put a read-only file behind the symlink an open follows and the native open would empty it before the check sees which object it opened. That was built and then reverted. The window only loses data where the kernel stops enforcing the permission bits of the read-only file, which means uid 0 or `CAP_DAC_OVERRIDE`; any executor that is not privileged gets EACCES from the native open, keeps the bytes, and returns not-permitted. Deferring the truncation cost the atomicity of a truncating open, and the executor should not be running as root in the first place, because it hosts untrusted guest Wasm. The fix belongs in the deployment: golem-cloud already runs the executor as a non-root uid with every capability dropped, and GOL-623 gives the published images the same posture.
- **`capture` has no production caller.** Its items keep `#[allow(dead_code)]` until T5 (GOL-578) wires it into the invocation and update boundaries.
- **The conflict error is what T5 reports.** GOL-578 records a failed update whose error names the conflicting path, and keeps the agent running on its current revision, which is why a conflict stays a retriable error rather than a permanent one.
- **Windows.** The test code of this crate has not compiled on Windows since #3728 added `sandbox_filesystem`. #3909 gates it, stacked on this branch. Production code is unaffected: the Unix-only paths are already behind `cfg`.

## Not in this change

- Wiring `capture` into invocation and update boundaries, the snapshot service, uploads, retention and the oplog confirmation: GOL-578.
- Snapshot clean-up on delete, revert and fork: GOL-579.
- The flaky `concurrent_first_append_coordination_converges_across_hard_links` from #3728, which counts a lookup before it takes the registry lock.


golem-worker-executor-test-utils/src/lib.rs
```diff
@@ -2410,7 +2410,7 @@ impl InvocationContextManagement for TestWorkerCtx {
 
 #[async_trait]
 impl Bootstrap<TestWorkerCtx> for TestServerBootstrap {
-    fn create_active_agents(
+    async fn create_active_agents(
```
kmatasfp: why is this async function now, seems to be unrelated change?

golem-worker-executor/src/sandbox_filesystem/adapter.rs
```diff
@@ -487,51 +482,36 @@ impl SandboxResolvedNamespaceTarget {
         SandboxPath::at(self.parent.clone(), PathBuf::from(&self.name))
     }
 
-    /// Returns the resolved object identity for policy checks.
+    /// Tells whether the resolved object is a regular file without write permission.
     ///
-    /// `follow` selects the final symlink or its referent. Resolution failures are preserved as
-    /// storage errors rather than silently treating the object as absent.
-    pub(crate) fn target_identity(
```
kmatasfp: this functionality is not needed anymore?

golem-worker-executor/src/sandbox_filesystem/xfs.rs
```diff
@@ -836,31 +836,6 @@ pub(super) fn clone_file(target: &File, source: &File) -> std::io::Result<()> {
     ioctl_ficlone(target, source).map_err(errno_to_io)
 }
 
-pub(super) fn reflink_file(
```
kmatasfp: hmm why was this deleted, we still want to provide this functionality

golem-worker-executor/Cargo.toml
```diff
@@ -118,8 +118,10 @@ zstd = { workspace = true }
 [target.'cfg(windows)'.dependencies]
 windows-sys = { workspace = true, features = ["Win32_Storage_FileSystem"] }
 
-[target.'cfg(target_os = "linux")'.dependencies]
+[target.'cfg(unix)'.dependencies]
```
kmatasfp: why did we make this change?

golem-worker-executor/src/services/agent_filesystem/lifecycle/tests/capture_and_baseline.rs
```diff
@@ -0,0 +1,3579 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::*;
+use futures::StreamExt as _;
+use std::collections::{BTreeMap, BTreeSet};
+use std::convert::Infallible;
+use std::ffi::OsStr;
+use std::os::unix::fs::{MetadataExt as _, PermissionsExt as _};
```
vigoo, line 20: this will not compile on Windows

golem-worker-executor/src/services/agent_filesystem/lifecycle/mod.rs
```diff
@@ -2687,28 +2370,32 @@ async fn execute_coordinated_open<Adapter: SandboxFilesystemAdapter>(
...
-        authorize_resolved_open(&generation, &resolved, options).await?;
+        let Some(CoordinatedTarget {
+            mut coordination,
+            resolved,
+        }) = coordinated_target(&generation, &target, kind, change).await?
+        else {
+            continue;
+        };
+        let opened = execute_open(Arc::clone(&generation), resolved.target(), options).await?;
+        if change.is_some() && opened.is_read_only_file() {
```
vigoo, line 2391: Agent found an edge case here:

> **Defer truncation until after checking the opened object**

> This check is too late for `TruncateExisting` and `CreateOrTruncate`: the adapter passes `truncate(true)` to the native open (`sandbox_filesystem/adapter.rs:1119–1123`), so `execute_open` can already have destroyed the contents.

> For example, `alias` points to a writable `target` when authorization succeeds. A concurrent rename then replaces `target` with a read-only file. Coordination on `alias` does not prevent that replacement. With a root/DAC-override executor, the native open succeeds and truncates the read-only file; this guard subsequently returns `NotPermitted`, but the bytes are gone and the generation remains valid.

> Could we open without truncation, authorize the pinned descriptor, and only then truncate that same descriptor? Please also extend the scripted symlink-race test to request truncation and assert that the read-only file’s contents survive. The existing `Existing + Write` case checks rejection but cannot catch this data loss.

kmatasfp: Reverted in `64cd2b1bd` — truncation happens at the native open again.

The race is real, but it only destroys data where the kernel stops enforcing the read-only file's permission bits, which means uid 0 or `CAP_DAC_OVERRIDE`. Any unprivileged executor gets EACCES from the open, keeps the bytes, and returns `not-permitted` either way. Deferring the truncation onto the pinned descriptor bought that case at the cost of atomicity: between the open and the `set-size`, another call of the same agent can read the old contents or write into the gap.

And the filesystem isn't really the point of not running as root. Every host function — sockets, HTTP, rdbms, secrets — runs with the executor's uid and capabilities, so hardening one open path inside a root process doesn't move the needle. GOL-623 gives the published images the non-root posture our clusters already have.

The post-open check stays, so the rename race is still refused, and so does the real-filesystem test asserting a read-only initial file keeps its bytes.

golem-worker-executor/src/services/agent_filesystem/lifecycle/tests/capture_and_baseline.rs
kmatasfp: addressed plus separate followup PR https://github.com/golemcloud/golem/pull/3909

vigoo (approved):

PR #3873, opened 2026-09-13
TS and Effect reflection cleanups + tools

noise64:
- resolves GOL-45, GOL-406, and GOL-407
- defines the shared host and WIT surfaces for environment-scoped agent identities and recoverable tool RPC construction
- aligns TypeScript and Effect around caller-defined static clients, immutable discovered `Reflected*` clients, and schema-authority-free `Dynamic*` clients
- adds agent and tool discovery, caller-owned typed tool definitions, canonical command lookup, graph-aware validation, and explicit packing and decoding helpers
- preserves structured remote errors, declared defaults and constraints, lazy TypeScript JSON results, Effect causes, and owned stream and handle cleanup
- updates generated declarations, SDK documentation, reflection skills, cross-SDK fixtures, and shared schema coverage

### Client approaches

### Verification

- focused TypeScript, Effect, host, schema, lifecycle, configuration, and tool tests pass
- existing deployed fixtures cover optional tools, reflected calls, dynamic calls, configuration, and principal identity round trips; current CI passes


noise64:

vigoo:
**P2 — Reflected command input validation skips schema restrictions (TS and Effect)**

`ToolCommand.validateJson` delegates to `packJson`, which packs the value and checks command constraints, but does not run the schema restriction validation performed by `inputSchema.validateJson`.

Locations: `sdks/ts/packages/golem-ts-sdk/src/toolReflection.ts:185–196` and `sdks/effect/src/ToolReflection.ts:171–182`.

I reproduced this in both SDKs with a numeric input whose schema has minimum 10: the command-level `validateJson` reports success for 9, while `command.inputSchema.validateJson` correctly rejects it. This is incomplete explicit client-side input validation, not a host-validation bypass: the invocation path subsequently validates the value and rejects it.

Could command-level validation apply the schema restrictions as well as the command constraints, preserving the validation issues? A regression should assert that the same out-of-range input is rejected by both validation entry points.

vigoo:
**P2 — Effect reflected/dynamic trigger still uses the trapping tool constructor**

The trigger paths in `sdks/effect/src/ToolReflection.ts:422–426` and `633–638` call `host.rpc(...)`, rather than the fallible `host.createRpc(...)` used for reflected creation by `liveToolStart(..., reflected = true)`.

`host.rpc` uses the ordinary WIT constructor; `createRpc` uses the new result-returning factory. Reflected/dynamic names and runtime owner context can fail, so these trigger paths can still take the host-trap path instead of reporting the recoverable creation error. Wrapping a constructor call in `Effect.try` does not give it the new fallible host contract.

A transport spy confirmed that trigger calls `rpc` and never calls `createRpc`. Could both reflected/dynamic trigger paths use the fallible factory, with tests asserting factory selection and recoverable failure behavior? Ordinary generated clients can retain the constructor as intended.

vigoo:
**P2 — Effect collect interrupts stdout on result failure instead of waiting for both channels**

The `collect` implementations in `sdks/effect/src/ToolReflection.ts:317–324`, `354–356`, and `605–607` use:

```ts
Effect.all([result, Stream.runCollect(stdout)], { concurrency: "unbounded" })
```

This runs concurrently but fails fast: if the structured result fails first, Effect interrupts the stdout collector. I reproduced this with an immediate result error and delayed stdout; collection returned the error without draining stdout to EOF. Conversely, a stdout failure can interrupt result observation, preventing the required structured-result error priority when both channels fail.

The tool collect contract requires both channels to settle, with structured-result failure taking precedence. The existing TS regression at `sdks/ts/packages/golem-ts-sdk/tests/tool-streams.test.ts:133–165` exercises that behavior.

Could these Effect collectors capture both channel outcomes, await both, then select the error/result? Please cover a result failure before delayed stdout EOF and a stdout failure before a delayed result failure.

vigoo:
**Nonblocking suggestion — Tool reflection snapshots are only shallowly immutable**

In `sdks/ts/packages/golem-ts-sdk/src/toolReflection.ts:158–160` and `sdks/effect/src/ToolReflection.ts:144–146`, the constraints array is copied/frozen, but its nested objects and stdin/stdout descriptors retain references to the supplied metadata. Mutating an original nested constraint after constructing the reflected tool changes how that snapshot validates inputs; I reproduced this in both SDKs.

Effect additionally creates a new mutable record root at `ToolReflection.ts:161–166` and passes it to `SchemaRef.fromImmutableGraph`. That helper (`sdks/effect/src/SchemaRef.ts:38–44`) freezes the wrapper, not the supplied root. Mutating the exposed synthesized root can therefore change accepted input shapes.

This differs from the immutable-snapshot contract. **I’m not sure we need to fix this:** the demonstrated failures require caller mutation, so this can be a decision about whether deep immutability is an enforced guarantee or a documented usage contract. If it is enforced, clone/freeze nested metadata and the synthesized root, with a mutation regression; otherwise clarify the guarantee.

vigoo:
**P2 — Effect reflected creation loses structured RPC errors**

In `sdks/effect/src/Tool.ts:123–125`, a fallible `createRpc` error is wrapped in `ToolClientError`. `sdks/effect/src/BridgeTool.ts:99–106` then maps that wrapper through `protocol("tool invocation failed", error)`, flattening the original structured error into a generic string.

I reproduced an original `protocol-error("invalid tool name: bad name")` becoming:

```text
protocol-error("tool invocation failed: Tool client invoke failed: [object Object]")
```

The original diagnostic is lost, and other RPC error variants can likewise lose their identity at this boundary. This defeats the purpose of recoverable, structurally inspectable reflected creation errors.

Could this mapping preserve the structured RPC error carried by `ToolClientError.cause`, using the generic protocol fallback only for unrecognized failures? Please add a reflected-start regression asserting that the original factory error tag and payload survive the Effect adapter.

vigoo:
**P2 — Snapshot copying breaks discovery of tools with binary defaults**

The new `immutableSnapshot` recursively copies every non-array object using `Object.fromEntries(Object.entries(...))`. WIT binary values contain `Uint8Array`, so this converts a valid byte buffer such as `Uint8Array.of(1, 255)` into a plain object instead of preserving its representation.

Locations: `sdks/ts/packages/golem-ts-sdk/src/toolReflection.ts:488–494` and `sdks/effect/src/ToolReflection.ts:607–613`.

I reproduced this independently in both SDKs: construct a valid tool registration with a binary positional type and a binary default containing bytes `[1, 255]`, then call `new ToolType(registered)`. Discovery fails while decoding the copied default:

```text
binary-value.val.bytes must be a Uint8Array
```

This is a regression introduced by the snapshot fix, not the earlier nonblocking concern about callers mutating metadata: callers do not need to mutate anything to encounter it.

Please preserve byte-array representation and ownership when copying metadata, rather than treating typed arrays as plain objects. Add a regression that discovers a tool with a nonempty binary default, verifies the default bytes survive, and checks that later mutation of the source buffer does not change the snapshot.

vigoo:
**P2 — Align TS/Effect with the new single-option canonical input rule**

The latest change in `golem-schema/src/schema/tool/canonical.rs:601–624` deliberately avoids adding an outer carrier when a non-required, defaultless argument already has declared type `option<T>`. The host therefore expects a single `option<T>` for that field.

The TS and Effect reflection builders still wrap these fields unconditionally:

- TS `sdks/ts/packages/golem-ts-sdk/src/toolReflection.ts:118–133,563–576`.
- Effect `sdks/effect/src/ToolReflection.ts:134–150,742–754`.

For a defaultless, non-required positional declared as `option<string>`, both reflected clients currently construct `option<option<string>>`. The executor compares the submitted graph against the canonical graph before invocation, so this fails with `tool input schema does not match the selected command`, even for a supplied value.

I ran the new host regression `optional_surfaces_do_not_nest_an_author_supplied_option_carrier`: it passes. Independent TS and Effect probes constructing a registration with the same declared optional type both fail an assertion that the reflected field schema equals `option<string>`; both produce the extra option layer. Scalar options have the same unconditional wrapping path. The existing TS typed codec in `src/internal/tool/model.ts:705` also still wraps unconditionally.

Please align SDK schema construction and value encoding with the selected shared canonical rule, rather than changing only the graph. Add cross-boundary coverage for omitted and supplied arguments whose declared type is already optional, covering both fixed positionals and scalar options.

vigoo (approved):

PR #3874, opened 2026-09-13
Rust SDK: agent and tool reflection

noise64:
- resolves GOL-48 and GOL-408
- adds Rust caller-defined static agent clients in method-only and full forms, discovered `ReflectedAgentClient` and `ReflectedToolClient` surfaces, and fully dynamic invocation
- supports durable identity rebinding while keeping lifecycle factories off method-only and fully dynamic clients and leaving ephemeral final identities unavailable for rebinding
- adds caller-owned typed tool definitions, canonical command lookup, immutable shared schema graphs, explicit JSON and schema-value packing, and structured reflection errors
- applies the same cardinality and schema checks to awaited and pending reflected results and keeps namespace metadata separate from callable commands
- fixes generated scheduling-name collisions and updates SDK tests, UI cases, documentation, skills, and cross-SDK fixtures

### Client approaches

### Verification

- focused Rust SDK, macro, schema, UI, stream, and integration tests pass
- existing deployed fixtures cover optional tools and principal identity round trips; current CI passes


noise64:

golem-rust-macro/src/agentic/agent_client_impl.rs
```diff
@@ -0,0 +1,429 @@
...
+            &remote_type_name.value(),
+            &data_defs,
+            &data_idents,
+            &config_defs,
+            &config_idents,
+            &[],
+            args.agent_is_durable,
+        );
+        AgentConfigAttrRemover.visit_item_trait_mut(&mut item_trait);
+        return Ok(quote! { #item_trait #remote_client });
```
vigoo, line 222: ### P1 — Preserve recoverable errors for complete caller contracts

Complete `#[agent_client]` contracts delegate to the static client generator, whose methods return bare values and panic on RPC failures. This differs from partial caller contracts and the documented fallible caller-contract API. A complete contract declaring `add -> u32` makes `client.add(1).await?` fail to compile because the generated return type is `u32`, not `Result`.

The generated `bind` also uses the trapping `WasmRpc::new` despite returning `Result`.

Please generate fallible methods and factories for complete caller contracts without changing the existing static-agent client policy. Add coverage for complete-contract return types and recoverable construction/invocation failures.

sdks/rust/golem-rust/src/agentic/tool_reflection.rs
```diff
@@ -0,0 +1,882 @@
...
+                ))
+            })
+    }
+    fn decode_result(
+        &self,
+        result: InvocationResult,
+    ) -> Result<Option<SchemaValue>, ToolError<ReflectedToolCustomError>> {
+        match (self.output_schema(), result.result) {
+            (None, None) => Ok(None),
+            (Some(schema), Some(value)) if value.graph().root == *schema.root() => {
```
vigoo, line 551: ### P1 — Resolve referenced definitions when checking tool result schemas

This implements a checked-output policy, but comparing only the roots does not establish schema compatibility. Both roots can be `Ref("example.Result")` while their definitions differ.

A reproducer with discovered `{old_field: string}` and returned `{new_field: string}` definitions passes this check and value validation. The returned graph is then discarded, so subsequent JSON decoding interprets the value using the old field name. Reordered enum/variant cases have the same risk. The custom-error decoder at lines 573–583 repeats this comparison.

The runtime validates against the serving declaration; that does not establish compatibility with the caller’s earlier snapshot. Please use a graph-aware comparison that resolves references, as the TS/Effect reflection implementation does, and test differing definitions behind identical reference IDs.

Scala and MoonBit currently share this issue. Choosing to trust runtime results instead is a separate cross-SDK policy decision; root-only comparison is not a sound implementation of the current checked-output policy.

sdks/rust/golem-rust/src/agentic/reflection.rs
```diff
@@ -0,0 +1,2141 @@
...
+        input: SchemaValue,
+    ) -> Result<PendingInvocation, GolemReflectError> {
+        if self.definition.input.contains_stream() {
+            return Err(GolemReflectError::InvalidType(format!(
+                "method `{}` has streaming input; use invoke_value or invoke_json",
+                self.definition.raw.name
+            )));
+        }
+        self.definition.input.validate_value(&input)?;
+        self.transport.pending(&self.definition.raw.name, input)
```
vigoo, line 959: ### P2 — Apply the same output policy to awaited and pending reflected calls

`pending_value` returns the schema-free `PendingInvocation`, whose completion only wire-decodes the result. Unlike `invoke_value`, it does not check the reflected output contract—not even whether a declared result is missing or a unit method unexpectedly returns a value.

This makes switching between awaited and pending invocation change which results are accepted. Please apply the same selected output policy to both completion paths and add parity tests for missing, unexpected, and schema-incompatible outputs.

This is a consistency issue, not a request to introduce deep validation everywhere. Agent responses contain no result schema graph, so these checks also cannot promise detection of every stale-schema change.

sdks/rust/golem-rust/src/agentic/tool_reflection.rs
```diff
@@ -0,0 +1,882 @@
...
+                ))
+            })?;
+        let input = self
+            .definition
+            .canonical_input_model(index)
+            .map_err(|error| ToolReflectionError::InvalidMetadata(error.to_string()))?;
+        let body = self.definition.commands.nodes[index]
+            .body
+            .as_ref()
+            .expect("resolved body");
```
vigoo, line 184: ### P2 — Allow reflection of namespace-only tool commands

`command_index_by_path` is an invocation-oriented lookup that excludes nodes without command bodies. Using it for reflection makes valid namespace-only nodes inaccessible: for example, `tool.command(&[])` returns `InvalidMetadata` for a namespace root.

The reflection contract explicitly allows namespace-only nodes. Please separate metadata lookup from callability checks so callers can inspect their help, arguments, and children without being able to invoke them. Add coverage for both root and nested namespace nodes.

golem-rust-macro/src/agentic/agent_client_impl.rs
```diff
@@ -0,0 +1,429 @@
...
+            {
+                let input = #input;
+                self.inner.method::<#input_ident, #output_type>(#method_name)?.pending(&input)
+            }
+
+            pub fn #schedule_ident(&self, at: #golem_rust::ScheduledTime, #(#params),*)
+                -> Result<#golem_rust::ScheduledInvocation, #golem_rust::GolemReflectError>
+            {
+                let input = #input;
+                self.inner.method::<#input_ident, #output_type>(#method_name)?.schedule(at, &input)
```
vigoo, line 391: ### P2 — Avoid collisions with the generated scheduling parameter

The generated scheduler adds a parameter named `at` alongside the user’s parameters. A valid contract method such as `fn search(&self, at: String) -> String` consequently fails compilation with duplicate-parameter and type-mismatch errors.

Please use the existing fresh-parameter naming approach and add a compile-pass test with a method argument named `at`.

sdks/rust/golem-rust/src/agentic/tool_reflection.rs
```diff
@@ -0,0 +1,882 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub struct ToolCommand {
+    tool: ToolType,
+    index: usize,
+    path: Vec<String>,
+    input: SchemaRef,
+    wire_input: SchemaGraph,
+    arguments: Vec<ToolArgument>,
```
vigoo, line 305: ### P2 — Complete or explicitly defer the caller-owned typed tool tier

The [Rust tool reflection plan](https://linear.app/golem-cloud/document/native-tool-reflection-rust-sdk-plan-fdaa2f524f29) includes a caller-owned partial `ToolClientDefinition` retaining typed codecs. This PR exposes discovered and dynamic clients, but I could not find that tier or an equivalent implementation. Existing generated tool clients are the static tier, not a replacement.

Please implement and test the partial typed tier, or explicitly agree to defer it and narrow the PR’s completion claim. This is a planned-scope gap rather than a defect in the existing dynamic invocation path.

sdks/rust/golem-rust/src/agentic/tool_reflection.rs
```diff
@@ -0,0 +1,882 @@
...
+    }
+    pub fn doc(&self) -> &Doc {
+        &self.tool.definition.commands.nodes[self.index].doc
+    }
+    pub fn output_schema(&self) -> Option<SchemaRef> {
+        self.body().result.as_ref().map(|spec| {
+            SchemaRef::new(SchemaGraph {
+                defs: self.tool.definition.schema.defs.clone(),
+                root: spec.type_.clone(),
+            })
```
vigoo, line 332: ### Non-blocking — Share immutable schema definitions

`output_schema()` clones the definition pool on each access, and argument reflection also clones definitions per argument. The plan calls for decoding once and sharing an immutable graph across schema views. Please consider reusing graph-backed `SchemaRef` roots and the existing decoder infrastructure to avoid repeated graph allocation and decoding.

sdks/rust/golem-rust/src/agentic/tool_reflection.rs
```diff
@@ -0,0 +1,882 @@
...
+}
+
+pub fn get_tool_type(name: &str) -> Result<ToolType, ToolReflectionError> {
+    ToolType::from_registered(
+        host::get_tool(name).ok_or_else(|| ToolReflectionError::NotFound(name.to_string()))?,
+    )
+}
+
+#[derive(Clone, Debug)]
+pub struct ReflectedToolClient {
```
vigoo, line 285: ### Non-blocking — Reconcile the remaining public API differences

A few public API choices differ from the Rust plan: `ToolCommand` dispatches rather than remaining metadata-only; `ReflectedToolClient` returns that same command; and `ToolReflectionError` is introduced instead of extending `GolemReflectError`.

There is also no public command-level `validate_json`/`pack_json` that includes command constraints—`input_schema().pack_json(...)` checks only the structural schema.

Please align these surfaces with the plan or record the intentional design changes. In particular, callers need a clear way to validate and pack effective command input, including constraints, without invoking the tool.

vigoo (changes requested):
Independent review against the declared TypeScript-stack base and the current agent/tool reflection plans. Six requested changes and two non-blocking suggestions are attached.

Verification: Rust SDK library tests (80 passed, 4 ignored), schema rendering tests (4 passed), client-generation macro tests (15 passed), and the SDK agent test target check passed. Focused temporary reproducers confirmed complete-contract return types, root-only tool schema comparison, and the generated `at` parameter collision. Other findings are based on source and contract inspection; no deployed-host or full stream/cancellation lifecycle suite was run.

For output validation, the comments distinguish runtime validation against the serving declaration from SDK checks against the caller's snapshot. They do not assume a blanket requirement for deep validation at every RPC boundary.

vigoo:
The new reflected-tool path uses the updated native canonical model, but the existing generated typed-tool client path still appears to advertise the old input schema.

In `golem-tool-metadata/src/extended_tool_type.rs:1111–1154`, global/body options still use `option_collected_graph(&option.shape)` and fixed positionals still use `positional.type_.clone()`. For a non-required scalar option or fixed positional without a default, these produce `T`, whereas the shared host change in #3873 now requires `option<T>`.

This model is still used by generated clients in `golem-rust-macro/src/tool/client.rs:114` (and by the inherited-prefix path through `canonical_input_fields`). For example, an optional, defaultless string argument produces `record { label: string }` on this path, but the host expects `record { label: option<string> }`. The executor checks graph equivalence before execution and returns `tool input schema does not match the selected command`, even when the argument is supplied.

Could we update the existing extended canonical model and its encoding/forwarding consumers to the same optional-carrier contract here? Please add a regression comparing the extended model schema with the corresponding native Tool schema, plus generated typed-client coverage for both omitted and supplied optional values. Required/defaulted arguments and repeatable collections should retain their existing representations. This concerns existing typed-tool clients, not the new reflection path.

golem-tool-metadata/src/extended_tool_type.rs
```diff
@@ -1311,11 +1367,18 @@ pub fn build_canonical_input_with_prefix(
...
-            name: value.name.clone(),
-            aliases: value.aliases.clone(),
-            short: value.short,
-            schema: value.schema.clone(),
+        .map(|value| {
+            let schema = command_fields
+                .iter()
+                .find(|field| canonical_surfaces_overlap(value, field))
+                .map(|field| field.schema.clone())
+                .unwrap_or_else(|| value.schema.clone());
```
vigoo: ### P2 — Preserve the effective parent schema when building subtree input

The direct-call canonical model now handles optional carriers, but the inherited-prefix path from the [earlier optional-input comment](https://github.com/golemcloud/golem/pull/3874#issuecomment-5730084792) is still incorrect. This code selects the standalone child's schema when names overlap, and otherwise falls back to the captured Rust type. Generated capture in `golem-rust-macro/src/tool/client.rs:975–984` retains the authored type rather than the parent's effective canonical schema.

Two focused tests against the composed native canonical model fail:

* An optional, defaultless parent `verbose: u32` that is not redeclared by the child is emitted as `u32`, but the host expects `option<u32>`.
* A required parent string redeclared as an optional, defaultless child option is emitted as `option<string>`, but the host expects `string`.

Composition explicitly makes the ancestor the source of truth for requiredness and defaults. The executor's graph-equivalence check rejects both cases with `tool input schema does not match the selected command`, even when a value is supplied.

Please retain the effective parent canonical schema through subtree capture and use it when assembling the outgoing record, rather than inferring it from the child or raw Rust type. Add generated subtree-client coverage for both an inherited global absent from the child declaration and a compatible redeclaration with different requiredness/defaults.

sdks/rust/golem-rust/src/agentic/tool_reflection.rs
```diff
@@ -0,0 +1,1432 @@
...
+        output: Option<SchemaRef>,
+    ) -> Result<(), ToolReflectionError>
+    where
+        I: crate::IntoSchema,
+        P: IntoIterator<Item = S>,
+        S: Into<String>,
+    {
+        let path = path.into_iter().map(Into::into).collect::<Vec<_>>();
+        if path.is_empty() {
+            return Err(ToolReflectionError::InvalidMetadata(
```
vigoo: ### P2 — Allow the empty path for caller-owned root commands

The new typed-tool builder rejects every empty path, but the shared tool contract defines `[]` as the root command path. A tool can have a callable root body, so this prevents `ToolClientDefinition` from representing a valid command that reflected and dynamic clients can invoke.

I reproduced this with `ToolClientDefinition::unnamed().unit_command::<Input, _, _>(Vec::<String>::new())`, where `Input` is a schema-derived record: it returns `InvalidMetadata("tool client command paths cannot be empty")`.

Please allow the empty path for both value and unit commands, while continuing to reject empty path segments and duplicate paths. The assertion in `caller_owned_definition_rejects_duplicate_paths` at lines 1289–1294 currently locks in the incorrect root-path rejection; replace it with root-command acceptance coverage.

golem-rust-macro/src/agentic/agent_client_impl.rs
```diff
@@ -0,0 +1,450 @@
...
+
+            #(#client_methods)*
+        }
+    })
+}
+
+fn fresh_param_ident(params: &[(Ident, Type)], preferred_name: &str) -> Ident {
+    let occupied = params
+        .iter()
+        .map(|(ident, _)| ident.to_string())
```
vigoo: ### P2 — Normalize raw identifiers in the scheduling collision check

The [original `at` collision](https://github.com/golemcloud/golem/pull/3874#discussion_r4046700314) is fixed for ordinary identifiers, but `Ident::to_string()` preserves the `r#` prefix. Rust treats `r#at` and `at` as the same binding, while this set treats them as different names.

This valid contract still fails `cargo check` with E0415 (duplicate `at`) and E0308 (`String` supplied as `ScheduledTime`):

```rust
#[agent_client]
trait Probe {
    fn reschedule(&self, r#at: String);
}
```

Please normalize identifiers, for example with `IdentExt::unraw`, before testing occupancy. Add compile-pass cases for `r#at` and for a raw suffixed candidate such as `at` alongside `r#at1`.

sdks/rust/golem-rust/src/agentic/tool_reflection.rs
```diff
@@ -0,0 +1,1432 @@
...
+            })?;
+        let input = self
+            .definition
+            .canonical_input_model(index)
+            .map_err(|error| ToolReflectionError::InvalidMetadata(error.to_string()))?;
+        let body = self.definition.commands.nodes[index]
+            .body
+            .as_ref()
+            .expect("resolved body");
+        let input_graph = Arc::new(input.record_schema.clone());
```
vigoo, line 227: ### Non-blocking — Graph sharing is improved but still incomplete

The output-schema and per-argument clones from the [earlier suggestion](https://github.com/golemcloud/golem/pull/3874#discussion_r4046700319) are improved. However, each command lookup still constructs a canonical input graph and clones its definition pool into a new `Arc` here, rather than sharing one decoded pool across the tool tree.

The agent path also still clones and decodes the entire wire graph for every selected root in `reflection.rs:1425–1433` (`decode_root`), even though the resulting views share an `Arc`.

Please consider sharing the existing graph pool for schema views and reusing a decoder across agent roots. A self-contained graph can still be projected where the wire format requires one; it need not be duplicated for every metadata view.

sdks/rust/golem-rust/src/agentic/reflection.rs
```diff
@@ -0,0 +1,2289 @@
...
+impl Future for ReflectedPendingInvocation {
+    type Output = PendingResult;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.get_mut();
+        match Pin::new(&mut this.inner).poll(cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Err(error)) => Poll::Ready(Err(error)),
+            Poll::Ready(Ok(invocation)) => {
+                if let Err(error) = validate_declared_output(
```
vigoo, line 1174: ### Non-blocking — Retain regression tests for the output-policy fixes

The pending path now applies the awaited output policy, and the reference-resolving tool comparison also fixes the earlier correctness issue. The checked-in tests do not yet directly guard those failure cases.

Please add:

* A reflected pending/awaited parity test for a missing declared output, an unexpected unit-method output, and an incompatible value. Testing `validate_declared_output` alone would still pass if either invocation path stopped calling it.
* A reflected tool-result regression where both roots use the same reference ID but the definition changes from `{old_field: string}` to `{new_field: string}`. The newly added positive test for equivalent differently named references exercises the opposite direction. Include the analogous custom-error payload case.

My temporary result-graph regression passes on this revision, including acceptance of equivalent differently named definitions; checking these cases in would preserve that guarantee.

vigoo (changes requested):
Follow-up review of the latest changes. The original complete-client error handling, graph-aware output comparison, reflected pending output policy, and namespace discovery issues are fixed in implementation. Caller-owned typed tool definitions and constraint-aware packing are now present, and the intentional metadata/error API choices are documented.

The attached comments cover three remaining correctness issues and two non-blocking follow-ups. Local verification passed: 108 SDK library tests (4 ignored), 15 macro tests, 62 metadata tests, 50 schema tests, 14 focused tool tests, and the trybuild UI suite. A full caller-client `.await?` compile probe and the stale-result-graph regression probe also pass. Separate focused probes reproduce the root-path rejection, both subtree schema mismatches, and the raw-identifier scheduling collision. The subtree probes verify the graph mismatch against the host's canonical model; they are not deployed-host tests.

vigoo (approved):

PR #3875, opened 2026-09-13
Scala SDK: agent and tool reflection

noise64:
- resolves GOL-52 and GOL-409
- adds Scala caller-defined static agent clients in method-only and full forms, discovered `ReflectedAgentClient` and `ReflectedToolClient` surfaces, and fully dynamic invocation
- adds caller-owned typed tool definitions, canonical namespace and command lookup, graph-aware input and result validation, and explicit JSON and schema-value packing
- makes reflection construction fallible, preserves structured RPC and agent errors, and keeps declared tool-error precedence while settling streamed output and results
- retains representable text, numeric, and binary restrictions in JSON Schema projection and applies schema-value restrictions alongside command constraints
- updates code generation, tests, test agents, documentation, the Scala reflection skill, and generated runtime declarations

### Client approaches

### Verification

- focused Scala model, core, code generation, reflection, configuration, tool, and integration tests pass
- existing deployed fixtures cover optional tools and principal identity round trips; current CI passes


noise64:

vigoo (commented):
**Optional suggestion — decode schema graphs once and share the decoded roots.**

`ToolType.schemaAt` decodes the full descriptor graph for every selected argument, result, and error, and each `command` lookup rebuilds these views. Agent reflection similarly repeats `schemaGraphFromWit`/`graphFromJs` per input field, output, and config declaration (`Reflection.scala:275,287,298`). This adds repeated decoding and allocation for large descriptors instead of the planned shared immutable snapshot.

Consider decoding the graph and its selectable roots once per discovered type, then constructing `SchemaRef` views over those roots and shared definitions. The TypeScript `schemaGraphRootsFromWit` approach is a useful reference. This is a non-blocking performance/maintainability suggestion.

sdks/scala/core/js/src/main/scala/golem/reflection/Reflection.scala
```diff
@@ -0,0 +1,580 @@
...
+    config: List[golem.config.ConfigOverride]
+  ): Either[GolemReflectError, Transport] =
+    encode(constructor).flatMap { payload =>
+      try {
+        val phantomArg = phantom.fold[js.UndefOr[JsSchemaUuid]](js.undefined)(uuid =>
+          JsSchemaUuid(js.BigInt(uuid.highBits.toString), js.BigInt(uuid.lowBits.toString))
+        )
+        Right(
+          new Transport(
+            WasmRpcApi.newClient(typeName, payload, phantomArg, golem.config.ConfigOverrideEncoder.encode(config))
```
vigoo, line 539: **P1 — Use the fallible RPC factory for reflection construction.**

`WasmRpcApi.newClient` calls `new RawWasmRpc`, not `WasmRpc.create`. The host constructor delegates to `create` and converts its ordinary RPC error into an `anyhow::Error` (`golem-worker-executor/src/durable_host/wasm_rpc/mod.rs:236-259`). For example, an unknown agent type or invalid constructor input traps the guest instead of returning the advertised `Left`; the Scala `try/catch` here cannot recover an imported host trap.

Please bind and use the fallible `WasmRpc.create` API for reflection/caller-codec/dynamic construction, preserving its RPC error. Add deployed tests for an unknown type and invalid constructor arguments. This is the recoverable-construction requirement from GOL-44, and the TS reflection transport already uses the fallible factory.

sdks/scala/core/js/src/main/scala/golem/reflection/ToolReflection.scala
```diff
@@ -0,0 +1,511 @@
...
+          ToolError.MalformedRemoteOutput(s"tool error '$name' has a malformed payload")
+        case _ => ToolError.Tool(NamedToolError(name, payload))
+      }
+    case ToolRpcFailure.RemoteToolError(error) => ToolError.RemoteTool(error)
+  }
+
+  private[reflection] def decodeResult(value: ToolInvokeResult): Either[ToolError[Nothing], Option[SchemaValue]] =
+    (result, value.result) match {
+      case (None, None)                                                       => Right(None)
+      case (Some(schema), Some(payload)) if payload.graph.root == schema.root =>
```
vigoo, line 315: **P1 — Compare resolved schema graphs, not only root nodes.**

Root equality ignores the definitions behind `RefType`. I reproduced an expected `Ref(Result) -> record{name: string}` accepting a returned `Ref(Result) -> record{password: string}` with `RecordValue(List(StringValue("hello")))`. Validation uses the old graph, so the value can be silently interpreted under the wrong field name. Conversely, two equivalent record schemas using different reference IDs are rejected.

The custom-error check at line 305 has the same problem. Please compare structure after resolving references against each graph, as the existing typed-client `ToolGraphs.schemaShapesMatch` and TS reflection do. Add tests for both same-ID/different-definition and different-ID/equivalent-definition results and error payloads.

sdks/scala/core/js/src/main/scala/golem/reflection/Reflection.scala
```diff
@@ -0,0 +1,580 @@
...
+  def triggerValue(input: SchemaValue): Either[GolemReflectError, InvocationMetadata]                 = transport.trigger(name, input)
+  def scheduleValue(at: Datetime, input: SchemaValue): Either[GolemReflectError, ScheduledInvocation] =
+    transport.schedule(at, name, input)
+}
+
+private[reflection] final class Transport private (raw: WasmRpcApi.WasmRpcClient) {
+  def invokeAndAwait(method: String, input: SchemaValue): Future[Either[GolemReflectError, Invocation[SchemaValue]]] =
+    encodeAsync(input).flatMap { payload =>
+      raw.asyncInvokeAndAwaitWithMetadata(method, payload) match {
+        case Left(error)                => Future.successful(Left(GolemReflectError.Remote(error.toString)))
```
vigoo, line 485: **P2 — Preserve structured agent RPC errors and align the common SDK contract.**

This converts the RPC error to `GolemReflectError.Remote(String)`; pending failures use `getMessage`, and trigger/schedule stringify errors too. Callers lose the ability to distinguish authorization, protocol, not-found, and remote-agent error variants or recover custom typed payloads without parsing text, even though `WasmRpcApi.RpcError` already retains `kind` and `agentError`.

This is the same cross-SDK inconsistency discussed in https://github.com/golemcloud/golem/pull/3876#discussion_r4046520541. Please settle the common behavior there and align Scala with it, preferably retaining structured RPC/agent variants and custom payloads through construction, immediate failures, and pending-result failures. Tests should distinguish two different custom payloads and the authorization/protocol/remote-agent variants.

sdks/scala/core/js/src/main/scala/golem/reflection/ToolReflection.scala
```diff
@@ -0,0 +1,511 @@
...
+  private implicit val ec: ExecutionContext = ToolInvokerRuntime.executionContext
+
+  val result: Option[SchemaRef]                 = body.result.map(spec => tool.resultSchema(spec.tpe))
+  val errors: List[(String, Option[SchemaRef])] =
+    body.errors.map(error => error.name -> error.payload.map(tool.resultSchema))
+
+  def packJson(input: Json): Either[ToolError[Nothing], SchemaValue] =
+    try
+      inputSchema.packJson(input).left.map(issue => ToolError.InvalidInput(issue.message)).flatMap { value =>
+        validateConstraints(value).map(_ => value)
```
vigoo, line 221: **P2 — Include schema-value restrictions in combined JSON packing.**

`inputSchema.packJson` checks the JSON shape but does not enforce numeric/text/etc. restrictions, and `validateConstraints` only checks command-level relationships. I reproduced a required `s32` positional with maximum 3 returning `Right` from `command.packJson({"count":4})`. `AgentType.packConfigJson` (`Reflection.scala:181-188`) similarly accepts an out-of-range config value.

Please run schema value validation in the combined packing helpers, in addition to command constraints, and add restricted-value boundary tests. The later `startValue`/factory paths validate again, so this is not a host-validation bypass; it makes the public combined pack/validation result incorrect for callers using it independently.

sdks/scala/core/js/src/main/scala/golem/reflection/SchemaRef.scala
```diff
@@ -0,0 +1,615 @@
...
+          "type"                 -> Json.String("object"),
+          "properties"           -> Json.Object(name -> payload.fold[Json](typed("null"))(schemaJson(graph, _))),
+          "required"             -> Json.Array(Json.String(name)),
+          "additionalProperties" -> Json.Boolean(false)
+        )
+        Json.Object("oneOf" -> Json.Array(side("ok", ok), side("err", err)))
+      case TextType(_) =>
+        typed(
+          "object",
+          "properties"           -> Json.Object("text" -> typed("string"), "language" -> typed("string")),
```
vigoo, line 165: **P2 — Retain refinements in the exported JSON Schema.**

The wildcard drops `TextRestrictions`; numeric branches above and binary branches below similarly discard declared restrictions. I reproduced a text schema with `regex = Some("^https://")` and `minLength = Some(12)` exporting `text: {"type":"string"}` without either `pattern` or `minLength`. Schema-driven callers can therefore generate values that `validateValue` rejects.

Please project representable numeric/text/binary restrictions into the JSON Schema, following the canonical renderer contract. TS already retains the text pattern/length keywords. Add restricted leaves and restricted union-branch tests; the current union test uses unrestricted string bodies and does not verify the claimed retention of branch restrictions.

sdks/scala/core/js/src/main/scala/golem/reflection/ToolReflection.scala
```diff
@@ -0,0 +1,511 @@
...
+    val nodes    = definition.commands.nodes
+    val resolved = path.foldLeft(Option((0, List.empty[Int]))) { (current, segment) =>
+      current.flatMap { case (index, indices) =>
+        nodes(index).subcommands.find { child =>
+          nodes.lift(child).exists(node => node.name == segment || node.aliases.contains(segment))
+        }.map(child => (child, indices :+ child))
+      }
+    }
+    resolved match {
+      case Some((index, indices)) if nodes(index).body.nonEmpty =>
```
vigoo, line 128: **P2 — Complete the planned tool metadata and caller-owned client surfaces before closing GOL-409.**

This rejects valid namespace-only nodes. I reproduced `command(Nil)` returning a discovery error for a tool whose root is a namespace with a callable child. `ToolCommand` requires a body and exposes no child traversal, so a caller must inspect raw `WitTool` to discover such metadata. GOL-406 explicitly permits namespace nodes and specifies metadata-only `ToolCommand`; resolution should succeed for those nodes, while invocation can reject a non-callable command.

There is also no Scala `ToolClientDefinition` implementation for the partial caller-owned contract required by GOL-409 (the TS counterpart is in `toolClient.ts:35-58`); `ReflectedToolClient` only forwards command lookup. Please implement these planned surfaces, or explicitly keep that scope/ticket incomplete rather than claiming GOL-409 is resolved. Add namespace-root and nested-namespace metadata tests.

sdks/scala/core/js/src/main/scala/golem/reflection/ToolReflection.scala
```diff
@@ -0,0 +1,511 @@
...
+    ec: ExecutionContext
+  ): Future[Either[ToolError[NamedToolError], (Option[SchemaValue], Array[Byte])]] = {
+    def drain(stream: ToolInputStream, chunks: Vector[Array[Byte]]): Future[Array[Byte]] =
+      stream.read().flatMap {
+        case Right(Some(bytes)) => drain(stream, chunks :+ bytes)
+        case Right(None)        => Future.successful(chunks.flatten.toArray)
+        case Left(failure)      => Future.failed(new ToolStreamException(failure))
+      }
+    val bytes = stdout.fold(Future.successful(Array.emptyByteArray))(drain(_, Vector.empty))
+    ToolReflectionFailures.recover(result.zip(bytes).map { case (terminal, output) => terminal.map(_ -> output) })
```
vigoo, line 417: **P2 — Preserve declared tool errors when collecting stdout.**

If stdout fails, `drain` produces a failed future, so `result.zip(bytes)` fails and is recovered as `Rpc(Protocol(...))`. This masks a declared `ToolError.Tool(...)` from the invocation result and can return before that result settles.

The existing non-reflection Scala `ToolInvocation.collect` in `ToolStreams.scala:104–119` settles both channels and explicitly gives declared tool errors precedence over stdout failure. `ToolClientSpec` tests that behavior. TypeScript, Rust, and MoonBit also wait for both channels and prioritize terminal result errors; Effect currently uses fail-fast collection.

Please preserve the existing Scala behavior in the reflected, JSON, and dynamic collectors: settle both channels and retain a declared tool error when stdout also fails. Add tests for a declared error plus stdout failure, including stdout failing before the invocation result arrives.

This does not require automatically cancelling the invocation on stdout failure; cancellation remains a separate capability.

vigoo:
The new `ToolReflection.scala` path wraps optional input fields correctly, but the existing generated typed-tool client path still appears to advertise the old input schema.

In `sdks/scala/model/src/main/scala/golem/tool/ExtendedToolType.scala:116–131`, global/body options still use `ToolGraphs.optionCollectedGraph(o.shape)` and fixed positionals still use `p.tpe`. For a non-required scalar option or fixed positional without a default, these produce `T`, whereas the shared host change in #3873 now requires `option<T>`.

This model is still consumed by `ToolClient.staticInputModel` in `sdks/scala/model/src/main/scala/golem/tool/ToolClient.scala:271`. For example, an optional, defaultless string argument produces `record { label: string }` on this path, but the host expects `record { label: option<string> }`. The executor checks graph equivalence before execution and returns `tool input schema does not match the selected command`, even when the argument is supplied. The correct schema in the new reflection implementation does not fix this separate typed-client path.

Could we align the existing extended canonical model and its encoding/forwarding consumers with the reflected/host representation here? Please add regression coverage asserting the optional field schema and exercising generated typed-client calls with both omitted and supplied values. Required/defaulted arguments and repeatable collections should retain their existing representations.

sdks/scala/core/js/src/main/scala/golem/reflection/ToolReflection.scala
```diff
@@ -0,0 +1,574 @@
...
+    }
+
+  def client(targetName: String): Either[GolemReflectError, Client] =
+    create(name.getOrElse(targetName))
+
+  private def create(lookupName: String): Either[GolemReflectError, Client] =
+    if (lookupName.trim.isEmpty)
+      Left(GolemReflectError.Discovery("tool client target name cannot be empty"))
+    else
+      try Right(createClient(lookupName))
```
vigoo, line 257: **P2 — Use recoverable construction throughout the caller-owned static tool path.**

The new wrapper catches Scala exceptions, but the generated client factory it is intended to wrap still eagerly calls `ToolRpcClient.transport(lookupName)` (`ToolRpcCodegen.scala:486–489`). That function uses `new RawToolRpc`, not `RawToolRpc.create` (`core/js/src/main/scala/golem/runtime/tool/client/ToolRpcClient.scala:40–48`).

For example, `ToolClientDefinition.unnamed(name => SomeGeneratedToolClient(name)).client("NOT-valid")` passes the nonempty-name check, then the host constructor rejects the invalid tool name with an `anyhow::Error` (`golem-worker-executor/src/durable_host/tool/mod.rs:5508–5527`). That is a guest trap, not a recoverable exception for this catch block, so the advertised `Either` does not protect the caller. The fallible host factory returns this validation failure as a structured RPC error instead.

Please route generated caller-owned clients through the fallible factory and retain the error. Add a deployed invalid-target construction test using an actual generated client; the current `identity[String]` test cannot exercise this path. This finding is source-traced, not deployed-tested locally.

sdks/scala/core/js/src/main/scala/golem/reflection/Reflection.scala
```diff
@@ -0,0 +1,639 @@
...
+          .flatMap(value => validate(declaration.schema, value).map(_ => value))
+          .map(value => ConfigOverride(entry.path, TypedSchemaValue(declaration.schema.graph, value)))
+      }
+    })
+
+  def validateConfig(overrides: List[ConfigOverride]): Either[GolemReflectError, List[ConfigOverride]] =
+    ReflectionInternals.sequence(overrides.map { entry =>
+      configDeclaration(entry.path).flatMap { declaration =>
+        validate(declaration.schema, entry.value.value).map { _ =>
+          ConfigOverride(entry.path, TypedSchemaValue(declaration.schema.graph, entry.value.value))
```
vigoo: **P2 — Check the caller's config schema before replacing it.**

`validateConfig` validates only `entry.value.value`, then discards `entry.value.graph` and substitutes the declaration graph. Because records and variants carry positional values, this can silently change the meaning of a well-formed typed config value.

I reproduced a declaration `record { name: string }` accepting a `TypedSchemaValue` whose graph is `record { password: string }` and value is `RecordValue(List(StringValue("secret")))`. The method returns `Right` with the graph changed to `{ name: string }`, so the host receives `name = "secret"` rather than rejecting the incompatible override. Reordered fields of the same primitive types have the same problem.

Please compare the supplied and declared schema graphs structurally before substituting the graph, while still allowing equivalent definitions with different reference IDs. Add incompatible-field/reordered-field and equivalent-graph tests. This is an additional existing issue in the PR, not a regression from the latest fix; the current TypeScript config path also does this and should be aligned.

sdks/scala/core/js/src/main/scala/golem/reflection/SchemaRef.scala
```diff
@@ -0,0 +1,646 @@
...
+          case None         => typed("string")
+        }
+        typed(
+          "object",
+          "properties"           -> Json.Object("text" -> typed("string", textRestrictions: _*), "language" -> language),
+          "required"             -> Json.Array(Json.String("text")),
+          "additionalProperties" -> Json.Boolean(false)
+        )
+      case BinaryType(restrictions) =>
+        def base64UrlLength(bytes: Int): Int = (bytes * 4 + 2) / 3
```
vigoo: **P2 — Avoid integer overflow when projecting binary length restrictions.**

`bytes * 4 + 2` is evaluated as a 32-bit `Int`. I reproduced `BinaryRestrictions(maxBytes = Some(1073741824))` exporting `maxLength: 0`, rather than `1431655766`. Other large positive limits overflow to negative lengths, yielding invalid JSON Schema. This only requires rendering the schema; no large binary allocation is needed to reproduce it.

Please promote to `Long`/`BigInt` before multiplying and retain a sufficiently wide result. Add boundary coverage around the first overflow as well as the 1 GiB example, for both minimum and maximum restrictions.

sdks/scala/core/js/src/main/scala/golem/reflection/ToolReflection.scala
```diff
@@ -0,0 +1,574 @@
...
+        case _ => ToolError.Tool(NamedToolError(name, payload))
+      }
+    case ToolRpcFailure.RemoteToolError(error) => ToolError.RemoteTool(error)
+  }
+
+  private[reflection] def decodeResult(value: ToolInvokeResult): Either[ToolError[Nothing], Option[SchemaValue]] =
+    (result, value.result) match {
+      case (None, None) => Right(None)
+      case (Some(schema), Some(payload))
+          if ToolGraphs.schemaShapesMatch(payload.graph, SchemaGraph(schema.graph.defs, schema.root)) =>
```
vigoo, line 399: **P2 — Do not reject equivalent result schemas at the comparator's depth limit.**

Using a graph-aware comparison fixes the original reference-ID issue, but `ToolGraphs.schemaShapesMatch` has a hard `ShapeMatchMaxDepth = 32` budget and returns false when it is exhausted (`model/src/main/scala/golem/tool/ToolGraphs.scala:327,367`). That conservative fallback is now being used to reject remote results, even when both schemas are identical.

I reproduced a valid string schema wrapped in 32 `ListType` nodes, with the matching nested `ListValue`, returning `MalformedRemoteOutput("missing, unexpected, or mismatched structured result")` from `decodeResult` when the payload carries exactly the expected graph. The custom-error comparison at line 384 has the same limit.

Please use a cycle-safe structural comparison that does not reject supported finite nesting solely because of this helper's budget. Add tests on both sides of the depth boundary and keep coverage for equivalent graphs with different reference IDs and genuinely incompatible graphs.

vigoo (commented):
**P1 — The generated typed-tool optional input schema is still incompatible with the host.**

Follow-up to https://github.com/golemcloud/golem/pull/3875#issuecomment-5730085342: this remains reproducible on the latest head. `sdks/scala/model/src/main/scala/golem/tool/ExtendedToolType.scala:116–130` still uses `optionCollectedGraph(o.shape)` for options and `p.tpe` for fixed positionals without adding the canonical presence wrapper for optional, defaultless fields.

An independent test using a valid descriptor with an optional, defaultless string positional produced `record { label: string }` from `canonicalInputRecordSchema`, versus `record { label: option<string> }` from the reflected view of the same descriptor. Generated clients consume the static model through `ToolClientRuntime.staticInputModel`, so their input graph does not match the host contract. The reflected optional round-trip fixture does not exercise this path.

Please align the extended canonical model and its generated encoding/decoding consumers with the host representation, and test generated typed-client calls with both omitted and supplied values. Required/defaulted arguments and repeatable collections should keep their existing representation. Posting this unanchored because the owning `ExtendedToolType.scala` lines are not in this PR's diff.

vigoo (approved):

PR #3876, opened 2026-09-13
MoonBit SDK: agent and tool reflection

noise64:
- resolves GOL-46 and GOL-410
- adds MoonBit caller-defined static agent clients in method-only and full forms, discovered `ReflectedAgentClient` and `ReflectedToolClient` surfaces, and fully dynamic invocation
- adds caller-owned typed tool clients, canonical namespace and command lookup, reachable input and result schema graphs, and explicit JSON and schema-value packing
- validates native configuration and reflected values against recursive schemas while retaining JSON-specific integer limits and declared constraints
- preserves structured RPC, agent, and tool errors and releases futures, streams, schema values, and other owned handles on success, failure, cancellation, and rejected output paths
- updates generated interfaces, examples, tests, documentation, and the MoonBit reflection skill

### Client approaches

### Verification

- focused MoonBit reflection, schema, tool, generator, stream, configuration, and example checks pass
- existing deployed fixtures cover repeated convenience calls, optional tools, and principal identity round trips; current CI passes


noise64:

sdks/moonbit/golem_sdk/reflection/transport.mbt
```diff
@@ -0,0 +1,121 @@
...
+  }
+}
+
+///|
+async fn Transport::invoke_value(
+  self : Transport,
+  method_name : String,
+  input : @model.SchemaValue,
+) -> Invocation[@model.SchemaValue?] {
+  let wire = @model.schema_value_to_wit(input) catch {
```
vigoo, line 50: **P1 — Fix before merging: awaited agent calls cannot send locally produced streams.**

`AgentStream::produce` / `new` creates native endpoints, but the synchronous `schema_value_to_wit` used here rejects these with `native schema value stream was not prepared for transfer`. I reproduced this with a record containing an `AgentStream[String]`, exercising the exact encoder used here (not a deployed RPC test). Because this transport is shared, full, method-only, reflected-native, and dynamic agent clients are affected.

Use the asynchronous transfer encoder (`schema_value_to_wit_async`) with transfer-failure cleanup. Please add a regression test for a successful awaited call with a locally produced stream input; stream detection and non-awaited-call rejection do not cover this path.

sdks/moonbit/golem_sdk/reflection/tool_reflection.mbt
```diff
@@ -0,0 +1,804 @@
...
+    None => None
+  }
+  let input_root = @model.schema_type(Record(input_fields))
+  {
+    tool: self,
+    path: canonical,
+    body,
+    arguments,
+    input_schema: SchemaRef::new({ defs: decoded.defs, root: input_root, }),
+    wire_input: { defs: decoded.defs, root: input_root, },
```
vigoo, line 306: **P1 — Fix before merging: reflected tool inputs carry unrelated definitions and fail provider validation.**

`wire_input` copies every definition from the tool descriptor, including definitions used only by results, errors, or other commands. For a command with a single `String` input and a named `Output` result, this produces one definition in the input graph, whereas the MoonBit provider's canonical input has zero. `tool-core/canonical_input.mbt:382–385` compares the complete graphs and rejects the call with `input schema does not match the canonical input schema`. My producer-to-consumer probe reproduced this with identical roots and differing definition lists.

Construct the wire graph using the canonical input's reachable definitions and matching ordering. Please cover named result/error types and multi-command descriptors in a producer-to-provider regression test; primitive-only descriptors miss this.

sdks/moonbit/golem_sdk/reflection/reflected_client.mbt
```diff
@@ -0,0 +1,708 @@
...
+
+///|
+fn validate_invocation_output(
+  entry : AgentMethod,
+  invocation : Invocation[@model.SchemaValue?],
+) -> Invocation[@model.SchemaValue?] raise ReflectError {
+  match (entry.output, invocation.value) {
+    (None, None) => invocation
+    (Some(schema), Some(value)) => {
+      schema.validate_value(value)
```
vigoo, line 697: **P2 — Fix before merging: rejected agent outputs leak owned resources.**

If validation raises here, the decoded `SchemaValue` is never returned to the caller and its resources are not released. I reproduced a stream remaining present after returning it where the reflected snapshot expects `String`. The `(None, Some(_))` branch similarly discards an unexpected value without cleanup. Dropping the completed RPC future does not dispose of resources already extracted from it.

Please release unreturned outputs on validation/decoding failure, without releasing values successfully transferred to the caller. Audit the JSON-output and caller-codec paths too: `caller_codec.mbt:288–312` has the same gap for failed typed decoding and unexpected unit-method output. Add resource-ownership tests for rejected outputs, including stale-schema cases.

sdks/moonbit/golem_sdk/reflection/types.mbt
```diff
@@ -0,0 +1,88 @@
...
+}
+
+///|
+fn agent_error_message(error : @common.AgentError) -> String {
+  match error {
+    InvalidInput(message) => "invalid input: " + message
+    InvalidMethod(message) => "invalid method: " + message
+    InvalidType(message) => "invalid type: " + message
+    InvalidAgentId(message) => "invalid agent id: " + message
+    CustomError(_) => "custom remote agent error"
```
vigoo, line 75: **P2 — Resolve before merging: agent RPC error handling appears inconsistent across SDKs; we should probably agree on one common behavior.**

Here every RPC error becomes `Remote(String)`, and every `CustomError` becomes the same fixed message. I reproduced two errors with different typed string payloads becoming equal. This prevents callers from recovering custom payloads or reliably branching on RPC/agent error variants.

I checked the other reflection PRs: TypeScript (#3873) preserves structured variants and a decoded `TypedSchemaValue` in `RemoteCallError.cause`; Effect (also #3873) preserves the original WIT `RpcError` and payload; Rust (#3874) preserves variants and `RemoteAgentError::Custom(Box<TypedSchemaValue>)`. Scala (#3875) also collapses errors to `GolemReflectError.Remote(String)`, although it attempts to render custom payloads as text rather than unconditionally discarding them. So this is not MoonBit-only, and the SDKs currently expose different behavior.

Can we settle a common error contract across the SDKs, preferably retaining distinguishable RPC/agent variants and custom payloads, then align MoonBit and Scala with it? Add a test distinguishing two custom payloads as well as authorization, protocol, and remote-agent failures.

sdks/moonbit/golem_sdk/reflection/tool_reflection.mbt
```diff
@@ -0,0 +1,804 @@
...
+    index = match next {
+      Some(next) => next
+      None => raise Discovery("tool command '" + segment + "' was not found")
+    }
+    indices.push(index)
+    canonical.push(nodes[index].name)
+  }
+  let body = match nodes[index].body {
+    Some(body) => body
+    None => raise Discovery("tool command has no body")
```
vigoo, line 194: **P2 — Fix before merging: namespace-only tool commands cannot be reflected.**

A command node without a body is valid namespace metadata, not a missing command. This unconditional rejection means even the PR's own `sample_tool().command([])` fails: its root is a valid namespace containing `run`. It prevents inspecting root/intermediate namespace nodes through the reflected command API.

[GOL-406](https://linear.app/golem-cloud/issue/GOL-406/reflection-finalize-generic-native-tool-api-shape) explicitly allows namespace-only nodes, and the TypeScript `ToolType.command` implementation returns their metadata. Please allow lookup/reflection without a body and require a body only when creating an executable client or invoking the command. Add coverage for `[]` and intermediate namespace lookup, alongside rejection of namespace invocation.

vigoo:
**Nonblocking conformance concern (P3): decode-once / immutable graph ownership.**

[GOL-406](https://linear.app/golem-cloud/issue/GOL-406/reflection-finalize-generic-native-tool-api-shape) and [GOL-410](https://linear.app/golem-cloud/issue/GOL-410/tool-reflection-moonbit-sdk) describe decoding discovery metadata once and sharing an immutable graph across reflected schema roots. In `reflection/tool_reflection.mbt:79–95`, discovery decodes the graph but discards the result; each selected argument/result subsequently decodes the descriptor again, and command lookup decodes it again at lines 292–294. Agent schema-root extraction also repeats whole-graph decoding. The public structs expose mutable arrays, so callers can mutate metadata/schema snapshots.

Could we retain one decoded graph per discovered type and expose immutable/read-only views (or otherwise prevent mutation through public aliases), following the planned ownership model and the TypeScript implementation? This would avoid repeated whole-descriptor work and keep the advertised snapshot stable. I am flagging this as a maintainability/performance and plan-conformance follow-up, separate from the concrete invocation failures in the inline comments.

vigoo:
**Nonblocking scope clarification: partial tool caller-codec definitions and JSON Schema projection.**

I could not find a public partial `ToolClientDefinition` surface or JSON Schema projection in this PR. [GOL-410](https://linear.app/golem-cloud/issue/GOL-410/tool-reflection-moonbit-sdk) explicitly includes partial caller-codec tool definitions, and I found no explicit deferral in the context researched for this review. Other SDK reflection implementations also expose JSON Schema projection, but its requirement for the current MoonBit scope is less clear than in the historical reflection plan.

Please make the scope decision explicit: are partial tool definitions and JSON Schema projection intended here, or deferred to separately tracked work? These are distinct questions: the partial caller-codec surface has a direct current-plan reference; JSON Schema export needs clarification rather than treating the older plan as authoritative. This is not a claim of a proven runtime defect or an additional merge blocker.

sdks/moonbit/golem_sdk_tools/lib/tools_emit.mbt
```diff
@@ -2593,6 +2593,10 @@ pub fn generate_tool_client_code(
       escape_string_lit(tool.name) +
       "\") } }\n\n///|\npub fn " +
       client +
+      "::new_for(name : String) -> " +
```
vigoo, line 2596: **P2 — New regression: `new_for` collides with accepted tool command names.**

The generator now emits `Client::new_for`, but `tools.mbt:1475` still reserves only `new` and `drop`. This valid source is accepted by `parse_tools`:

```moonbit
#derive.tool
struct Probe {}
pub fn Probe::new_for(value : String) -> String { value }
```

`generate_tool_client_code` then emits both the new constructor and the command method as `ProbeClient::new_for`, producing duplicate method definitions. My independent generator probe counted two declarations where there must be one (`2 != 1`).

Please reserve or disambiguate `new_for` alongside the existing lifecycle names and add a collision regression test. The current test only asserts that the constructor exists, so it misses this case.

sdks/moonbit/golem_sdk/reflection/tool_reflection.mbt
vigoo: **Still not fully fixed (P1)** at [the reviewed head](https://github.com/golemcloud/golem/commit/1023b7d1bf469df080dedc8f38773756d5467827). Removing unreachable definitions fixes the output-only-definition case, but definition ordering still causes valid calls to fail.

`tool_reflection.mbt:363–366` calls `schema_graph_closure`, which preserves the whole descriptor's definition order (`schema_model/fingerprint.mbt:125–140`). For a valid two-command MoonBit tool where an earlier command uses named type `B` and the selected command takes `A` then `B`, reflection sends definitions `[B, A]`, while the selected command's provider-side `canonical_input_model` expects `[A, B]`.

I reproduced this through the real `ToolDef::to_tool` → `ToolType::command` → `checked_input` → `decode_canonical_input` path. The provider rejects the reflected input with `InvalidInput("target: input schema does not match the canonical input schema")`. A control using the same values with the provider's graph succeeds.

Please make definition ordering consistent between producer and consumer, not just prune unreachable definitions. Add a producer-to-provider regression test with multiple commands and differently ordered named types; the new test only checks that an output-only definition is removed.

sdks/moonbit/golem_sdk/reflection/types.mbt
vigoo: **MoonBit's data-loss bug is fixed; the cross-SDK contract discussion remains open.** The latest implementation retains RPC/agent variants in `AgentRpcError`, carries custom `TypedSchemaValue` payloads, and tests that two different custom payloads remain distinguishable.

I have not found a recorded agreement on the common cross-SDK behavior requested above. Please document that agreement or explicitly track the remaining SDK alignment separately. This is not a claim that MoonBit still stringifies RPC errors, nor a fresh assessment of the latest heads of the other SDK PRs.

vigoo:
**Still not fixed — nonblocking decode-once / immutable graph concern.** Follow-up to [the original comment](https://github.com/golemcloud/golem/pull/3876#issuecomment-5729678621), checked at [the latest reviewed head](https://github.com/golemcloud/golem/commit/1023b7d1bf469df080dedc8f38773756d5467827).

`reflection/tool_reflection.mbt:129–145` still decodes and discards the discovery graph, then decodes again for each selected schema root; command lookup also decodes it again at lines 351–353. Public metadata/schema arrays remain mutable. The new reachable-definition projection does not address either of these concerns.

Please retain a decoded graph and prevent mutation through public aliases, or explicitly record this as a deferred conformance follow-up. The current PR description calls these immutable snapshots, which does not match the exposed ownership model. This remains nonblocking, as in the original comment.

vigoo:
**Partially addressed — JSON Schema scope is still unresolved.** Follow-up to [the original scope comment](https://github.com/golemcloud/golem/pull/3876#issuecomment-5729680630), checked at [the latest reviewed head](https://github.com/golemcloud/golem/commit/1023b7d1bf469df080dedc8f38773756d5467827).

The new named/unnamed `ToolClientDefinition` factories and generated `new_for` entry point address the previously missing caller-owned client-definition surface. JSON Schema projection is still absent, and I have not found an explicit decision that it is either in scope or deferred.

Please record that scope decision. As before, this is a nonblocking clarification, not a claim that the historical plan alone makes JSON Schema export a required runtime fix. The separate `new_for` name-collision finding concerns the implementation of the newly added factory support.

vigoo (approved):

PR #3878, opened 2026-09-14
Add failed invocation recovery guidance

vigoo:
## Summary

- show contextual recovery guidance for exact previous-invocation failures from agent invoke, stream/connect, and resume
- print a shell-safe revert command that preserves profile and environment selection
- require confirmation before destructive agent reverts
- preserve typed previous-invocation failure details across executor/service boundaries
- cover the recovery flow with focused unit tests and one CLI integration workflow

Resolves GOL-520


cli/golem-cli/src/command_handler/agent/mod.rs
```diff
@@ -3113,6 +3147,50 @@ fn all_modes_filter() -> AgentFilter {
...
+        shlex::try_quote(value)
+            .expect("CLI arguments cannot contain NUL bytes")
+            .into_owned()
+    }
+
+    let mut args = vec![
+        quote(executable),
+        "--profile".to_string(),
+        quote(profile_name),
+    ];
```
noise64: While there is always a selected profile, most of the time it is implicilty selected, and in general we do not recommend using profiles in "manifest mode", only in "non-manifest" dev-ops cases, so unconditionally adding them here can be confusing compared to what we suggest in other docs. I recommend enriching what load_profile calculates, and remember there if there were explicit profiles in use, then also add that info to profile_name() on context (or have effective_profile_name and explicit_profile_name).

And ftr: we could say that environment can be also implicit on the original command, but I think that is actually nice to unconditionnaly include in the recommended command.

golem-service-base/src/error/worker_executor.rs
```diff
@@ -506,6 +507,12 @@ impl From<std::io::Error> for WorkerExecutorError {
 
 impl From<WorkerExecutorError> for Status {
     fn from(value: WorkerExecutorError) -> Self {
+        if matches!(value, WorkerExecutorError::PreviousInvocationFailed { .. }) {
+            let message = value.to_string();
+            let details = golem::worker::v1::WorkerExecutionError::from(value).encode_to_vec();
+            return Self::with_details(tonic::Code::FailedPrecondition, message, details.into());
+        }
```
noise64: Is this extraction from the match is reuired to be able consuming value?

noise64 (commented):
Two inline comments on the current diff.

golem-service-base/src/error/worker_executor.rs
vigoo: Fixed in 4dc441f. The typed previous-invocation case now lives directly in the exhaustive match using an `error @ ...` binding, which allows rendering the message before consuming the value into protobuf details.

cli/golem-cli/src/command_handler/agent/mod.rs
vigoo: Fixed in 4dc441f. Context now retains the explicitly supplied profile separately from the effective profile selected for execution. The recovery command emits `--profile` only for an explicit `--profile`/`GOLEM_PROFILE`; implicit/default profiles are omitted. Environment selectors remain preserved. Added coverage for implicit-profile omission and explicit-profile quoting.

noise64 (approved):

PR #3880, opened 2026-09-14
Custom HTTP handlers

vigoo:
## Purpose

Implements the custom HTTP handler platform (GOL-554–563) and the MoonBit, Rust, TypeScript, and Scala SDK APIs (GOL-564, GOL-565, GOL-567, GOL-568).

| Task | Implemented scope |
| --- | --- |
| [GOL-554](https://linear.app/golem-cloud/issue/GOL-554) | Shared language-neutral conformance corpus, JSON Schema, and TypeScript/Bun integrity checker. |
| [GOL-555](https://linear.app/golem-cloud/issue/GOL-555) | Router/file/provider metadata, validation, codecs, extraction, persistence, and low-level SDK support. |
| [GOL-556](https://linear.app/golem-cloud/issue/GOL-556) | Mounted dispatch descriptors, immutable read-only indexes, strict path decoding, routing precedence, security barriers, and requested-method CORS preflight. |
| [GOL-557](https://linear.app/golem-cloud/issue/GOL-557) | Exact-absolute-path, metadata-first bounded executor filesystem reads, shared with existing CLI/REST inspection. |
| [GOL-558](https://linear.app/golem-cloud/issue/GOL-558) | Request-private ephemeral invocation sessions, bounded input/output, same-execution reattachment, completion tracking, and disconnect cleanup. |
| [GOL-559](https://linear.app/golem-cloud/issue/GOL-559) | Direct immutable-file serving from shared blob storage, deployment-index caching, ranges, validators, and GET/HEAD without executor calls. |
| [GOL-560](https://linear.app/golem-cloud/issue/GOL-560) | Offline OpenAPI 3.1 validation, mount-relative rebasing, deterministic merging with typed operations, and reference/security/conflict checks. |
| [GOL-561](https://linear.app/golem-cloud/issue/GOL-561) | Live durable-agent filesystem serving over existing executor reads, including initialization/restoration, ranges, bounded streaming, and cancellation. |
| [GOL-562](https://linear.app/golem-cloud/issue/GOL-562) | Canonical streaming HTTP request/response envelopes and mounted handler delivery through invocation sessions. |
| [GOL-563](https://linear.app/golem-cloud/issue/GOL-563) | Lazy deployment-pinned OpenAPI provider invocation, bounded single-flight caching, invalidation fencing, and JSON/YAML serving. |
| [GOL-564](https://linear.app/golem-cloud/issue/GOL-564) | MoonBit router declarations, parser/emitter/runtime support, streaming HTTP helpers, live-file exposure, OpenAPI providers, and examples. |
| [GOL-565](https://linear.app/golem-cloud/issue/GOL-565) | Dedicated Rust router macros/APIs, canonical streaming types, file mappings, OpenAPI providers, compile-time validation, and deployed fixtures. |
| [GOL-567](https://linear.app/golem-cloud/issue/GOL-567) | TypeScript `defineHttpRouter`, Web and raw streaming adapters, host-free contract helpers, `exposeFiles`, OpenAPI serialization, and CLI integration coverage. |
| [GOL-568](https://linear.app/golem-cloud/issue/GOL-568) | Scala router annotations/macros/codegen, lightweight HTTP/AgentStream types, live-file exposure, OpenAPI providers, and deployed examples. |

The normative `HTTP_HANDLERS.md` specification is attached to GOL-554, not tracked in Git. The shared corpus lives in `golem-service-base/tests/fixtures/http-handlers/`.

## Routing, authorization, and file serving

- Complete concrete-method endpoints precede eligible mounted fallbacks; reserved routes retain precedence. Selection/security failures do not fall through to parent mounts. Existing atomic deployment selection is reused.
- Paths decode once per segment. HTTP mappings resolve to exact absolute filesystem paths before executor calls; the executor performs no URI decoding or mount/root/suffix resolution. Existing CLI/REST inspection uses the same bounded read protocol.
- Mount authentication, session policy, and CORS remain host-owned and precede file lookup or handler execution. Trusted ingress is explicit and defaults to no trust; routing, redirects, and guest envelopes share validated public-origin information.
- Immutable indexes reuse upload-computed BLAKE3 hashes and sizes: deployment does not redownload or rehash contents. Worker-service reads the selected deployment's read-only files directly from shared blob storage, caching indexes rather than file contents.
- Immutable GET/HEAD supports MIME inference, nosniff, strong BLAKE3 ETags, conditional requests, and single byte ranges. HEAD ignores ranges; malformed/multiple ranges return the full representation. Indexed missing blobs and size mismatches are server errors, never handler fallback; late stream errors abort delivery. Filesystem, S3, memory, and SQLite reads use bounded chunks.
- Live-file HTTP access initializes absent durable agents and restores non-resident ones without invoking an exported guest method. Reads use the concrete constructor identity and current normal lifecycle revision, while the selected target stays fixed. Ordered mappings advance only on absence, not directories, symlinks, permission errors, or initialization/read failures.
- Live reads support GET/HEAD and single ranges without immutable validators. Metadata and bytes come from one admitted read; bounded read-ahead and the namespace/serialization guard prevent mixing file generations. HTTP deadlines belong to worker-service, not the executor. EOF, failure, cancellation, and disconnect release the read.

## Streaming handlers

- A handler receives the canonical seven-field request and returns a three-field response with `stream<list<u8>>` bodies. Raw adapters preserve valid repeated headers and opaque values, including repeated Set-Cookie; cross-name ordering is not promised.
- Each HTTP exchange owns one private ephemeral invocation. Input is not polled before acceptance supplies stream mappings. Transport recovery reattaches the same execution and reconciles ACKs/cursors; it does not create a replacement invocation after ambiguous acceptance or process loss.
- An early result is not invocation completion. Unknown-length responses withhold successful EOF until session success; positive Content-Length responses retain their final byte. HEAD/204/205/304/zero-length responses dispose bodies unpolled and await completion. Post-head failure aborts transport instead of fabricating successful EOF.
- Executor execution stays Running through post-result stream production and settlement, including replay. Interruption retains its typed outcome rather than becoming a runtime error/retry. Disconnect cleanup cancels streams and uses existing pending cancellation/interruption; interruption is pause, not rollback. Serving-process death cannot guarantee remote cleanup without a lease, and no lease is added.
- Official Hyper, hyper-util, and Poem remain the HTTP stack; no vendored parser or second framing parser. Equal duplicate Content-Length may normalize; HTTP/1 TE+CL follows chunked framing and closes the connection. Differing/comma Content-Length is rejected. Strict rejection of distinctions already normalized by the transport is unsupported by design. Guest response heads are validated before encoding.

## OpenAPI providers

- `openapi_provider_method` names an ordinary parameterless, non-streaming method. Providers run only for OpenAPI requests, under a fixed system context and at the selected deployment's component revision; ordinary routing, handlers, and static/live files do not execute provider code.
- OpenAPI 3.1 documents are validated offline. Paths and supported local references are rebased under mounts and merged deterministically with typed operations. Equivalent-path/method collisions, incompatible components/path metadata, duplicate operation IDs, and unsupported references are rejected. External references are never fetched.
- Golem owns the public document's top-level identity/origin; effective operation security is preserved. Provider timeout, size, validation, or merge failures affect only OpenAPI responses and produce safe diagnostics, not provider payloads or secrets.
- Bounded single-flight generations cache JSON/YAML from the same merged model. Deployment, domain, security, and secret changes invalidate the cache; generation fencing prevents invalidated work from publishing stale results.

## SDKs and remaining scope

MoonBit, Rust, TypeScript, and Scala now expose named ephemeral routers with independently optional handlers, immutable mappings, and OpenAPI providers, plus ordered live-file mappings on eligible ordinary durable agents. They reuse normal configuration, registration, and invocation machinery. Tests cover declaration validation, metadata, generated clients, streaming ownership, and deployed examples. TypeScript provides both normalized Web APIs and a raw escape hatch; Scala does not require an unrelated HTTP/OpenAPI framework.

**GOL-566 (Effect HttpRouter/OpenAPI integration) remains follow-up work and is not claimed as implemented here.** Low-level Effect metadata support and shared TypeScript contract helpers are present. External deployment/access changes and backward-compatibility layers are not included.

## Verification

- The last fully green combined platform run before the four SDK additions was [CI 35517318356](https://github.com/golemcloud/golem/actions/runs/35517318356): 54 passed, 7 skipped, with [Docs passing](https://github.com/golemcloud/golem/actions/runs/35517318304). These links describe that exact earlier head, not the newest merge.
- Platform verification includes executor unit/lifecycle/replay tests; SQLite/PostgreSQL HTTP, file, and OpenAPI integrations; real HTTP/1.1 and HTTP/2 streaming/disconnect tests; storage range/drop tests; provider/cache race tests; and shared corpus consumers.
- The live-file backpressure fixture remains 32 MiB but uses deterministic poorly compressible bytes: a protobuf+gzip regression proves it exceeds the inner receive window. Both PostgreSQL and SQLite held-response mutation tests passed without weakening blocking assertions or increasing timeouts.
- SDK additions include Rust macro/runtime tests, MoonBit parser/emitter/runtime tests, Scala macro/codegen/stream tests, TypeScript contract/type/runtime tests, and CLI build/deployment/client-generation tests. Corpus integrity checks alone are not runtime conformance certification.
- Latest-main merge follows main's removal of tracked Effect WASMs; all three remain generated, ignored build artifacts. Local `npm run check:contracts`, `npm run check:artifacts`, and merge diff checks passed.
- Full CI for the newly merged head is pending; this section will be updated with its result.

This PR remains unmerged; merging the PR or deploying is not part of this update.


golem-common/src/schema/agent/http.rs
```diff
@@ -0,0 +1,247 @@
...
+use crate::base_model::agent::http_files::valid_decoded_segment;
+use crate::base_model::agent::{
+    AgentMode, FileMapping, HttpMethod, HttpMountDetails, PathSegment, Snapshotting,
+};
+use crate::schema::schema_type::NamedFieldType;
+use crate::schema::validation::is_equivalent_cross_graph;
+use crate::schema::{SchemaGraph, SchemaType};
+use std::collections::HashSet;
+
+fn record(fields: impl IntoIterator<Item = (&'static str, SchemaType)>) -> SchemaType {
```
vigoo: This helper belongs to SchemaType itself so we can use it everywhere

golem-common/src/schema/agent/http.rs
```diff
@@ -0,0 +1,247 @@
...
+                }
+            }
+            Ok(())
+        }
+    }
+}
+
+fn validate_router(agent: &AgentTypeSchema) -> Result<(), String> {
+    if agent.mode != AgentMode::Ephemeral {
+        return Err("router-mode".into());
```
vigoo: We should introduce a proper error enum for these cases

golem-common/src/base_model/agent.rs
```diff
@@ -484,6 +486,48 @@ pub struct HttpMountDetails {
     pub phantom_agent: bool,
     pub cors_options: CorsOptions,
     pub webhook_suffix: Vec<PathSegment>,
+    pub static_bindings: Vec<FileMapping>,
+    pub filesystem_bindings: Vec<FileMapping>,
+    pub openapi_provider: Option<String>,
```
vigoo: It is hard to understand what this field is (it is holding a method name). This should be made explicit, at least in the field name, maybe with some newtype?

golem-registry-service/src/services/deployment/deployment_context.rs
```diff
@@ -1316,13 +1512,718 @@ mod tests {
...
+            input_schema: InputSchema::Parameters(vec![]),
+            output_schema: golem_common::schema::OutputSchema::Single(Box::new(
+                SchemaType::string(),
+            )),
+            http_endpoint: vec![],
+            read_only: None,
+        }
+    }
+
+    #[test]
```
vigoo: Let's move these tests to their own submodule (in a separate file). Follow the existing module style

golem-registry-service/src/services/deployment/router_file_index.rs
```diff
@@ -0,0 +1,331 @@
...
+            .iter()
+            .any(|route| matches!(route.behaviour, RouteBehaviour::HttpRouter(_)))
+        {
+            return Ok(());
+        }
+        let _permit = self
+            .builds
+            .try_acquire()
+            .map_err(|_| DeploymentWriteError::RouterFileIndexBusy)?;
+        tokio::time::timeout(self.config.timeout, async {
```
vigoo: What do we expect here to timeout and why?

golem-registry-service/src/services/deployment/router_file_index.rs
```diff
@@ -0,0 +1,331 @@
...
+use golem_common::model::component::{AgentFilePermissions, InitialAgentFile};
+use golem_common::model::environment::EnvironmentId;
+use golem_service_base::custom_api::{RouteBehaviour, RouterFileIndexEntry};
+use golem_service_base::service::initial_agent_files::InitialAgentFilesService;
+use sha2::{Digest, Sha256};
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
+use tokio::sync::Semaphore;
+
+pub(super) struct RouterFileIndexBuilder {
```
vigoo: Explain the router file index's purpose (also add as a doc comment)

golem-worker-executor/src/services/active_agents/mod.rs
```diff
@@ -781,13 +783,21 @@ impl<Ctx: WorkerCtx> ActiveAgents<Ctx> {
         let cache_key = owned_agent_id.clone();
         let deps = deps.clone();
         let invocation_context_stack = invocation_context_stack.clone();
+        let card_interest_index = self.card_interest_index.clone();
+        let origin = TraceOrigin::capture_current();
         let active_agent = self
             .agents
-            .get_or_insert_simple(&cache_key, || {
-                Box::pin(async move {
+            .get_or_insert_simple_spawned(&cache_key, move || {
```
vigoo: Explain this change

golem-worker-executor/src/services/agent_filesystem/lifecycle/inspection_stream/tests.rs
```diff
@@ -0,0 +1,289 @@
+use super::inspection::native_resident;
+use super::*;
+use futures::StreamExt;
+use golem_common::model::filesystem::{
+    FileByteSelection, FileReadError, FileReadExtent, FileReadHead, FileReadTarget,
+};
+use golem_service_base::model::FileReadResponse;
+use test_r::{test, timeout};
+use tokio::time::Instant as Deadline;
+
```
vigoo, line 11: This file should follow existing practices - inspection_stream made to a module directory, with mod.rs and tests.rs in it

golem-worker-executor/src/services/agent_filesystem/lifecycle/inspection/tests.rs
```diff
@@ -0,0 +1,330 @@
+use super::*;
+use golem_common::model::filesystem::{
+    FileByteSelection, FileReadError, FileReadExtent, FileReadHead, FileReadTarget,
+};
+use test_r::{test, timeout};
+
```
vigoo, line 11: This file should follow existing practices - inspection made to a module directory, with mod.rs and tests.rs in it

golem-worker-executor/src/services/agent_filesystem/lifecycle/mod.rs
```diff
@@ -48,6 +48,11 @@ use std::task::{Context, Poll};
 
 const WRITE_PRESSURE_RECOVERY_TIMEOUT: std::time::Duration = std::time::Duration::from_millis(250);
 
+mod inspection;
```
vigoo, line 52: Explain what inspection is and why do we need it for our custom http router plan

golem-worker-executor/src/worker/invocation_loop.rs
```diff
@@ -2680,40 +2747,54 @@ impl<Ctx: WorkerCtx> Invocation<'_, Ctx> {
...
-                // This will delay processing of the next invocation and is quite unfortunate.
-                // A possible improvement would be to check whether we are on a copy-on-write filesystem
-                // if yes, we can make a cheap copy of the file here and serve the read from that copy.
-
-                let (latch, latch_receiver) = oneshot::channel();
-                let drop_stream = DropStream::new(stream, || latch.send(()).unwrap());
-                let _ = sender.send(Ok(ReadFileResult::Ok(Box::pin(drop_stream))));
-                latch_receiver.await.unwrap();
+        let deadline = _permit.deadline();
+        let access = tokio::select! {
```
vigoo, line 2946: Why do we have timeouts on reading files from the filesystem?

golem-common/src/schema/agent/http.rs
vigoo: Moved the helper to SchemaType::record_from_fields and switched the HTTP envelope schemas to use it.

Addressed in https://github.com/golemcloud/golem/commit/1d25f5656216c132258cddbcc3b36d71e4e182bd.

golem-common/src/schema/agent/http.rs
vigoo: Added HttpAgentValidationError with typed variants, Display and Error implementations. Validation consumers and tests now use the typed error rather than string sentinel codes.

Addressed in https://github.com/golemcloud/golem/commit/1d25f5656216c132258cddbcc3b36d71e4e182bd.

golem-common/src/base_model/agent.rs
vigoo: Renamed the field to openapi_provider_method throughout the shared model, protocol, WIT, SDKs and generated artifacts. It explicitly identifies the ordinary method providing the OpenAPI document.

Addressed in https://github.com/golemcloud/golem/commit/1d25f5656216c132258cddbcc3b36d71e4e182bd.

golem-registry-service/src/services/deployment/deployment_context.rs
vigoo: Moved deployment_context into a directory module with mod.rs and a separate tests.rs.

Addressed in https://github.com/golemcloud/golem/commit/1d25f5656216c132258cddbcc3b36d71e4e182bd.

golem-registry-service/src/services/deployment/router_file_index.rs
vigoo: Removed the timeout and concurrency settings. Index construction no longer downloads blobs: it copies the existing upload-computed BLAKE3 digest and size from selected provisioning metadata. The same digest is the blob identity and immutable ETag; no additional SHA-256 is needed. F01/F03 and the corpus are aligned.

Addressed in https://github.com/golemcloud/golem/commit/1d25f5656216c132258cddbcc3b36d71e4e182bd.

golem-registry-service/src/services/deployment/router_file_index.rs
vigoo: Added a doc comment explaining that this is a selected-revision index of read-only initial files, not the live filesystem. It lets static lookup and serving use immutable metadata/blob storage without activating an executor. Hashes and sizes come from upload metadata, with no blob reread at deployment.

Addressed in https://github.com/golemcloud/golem/commit/1d25f5656216c132258cddbcc3b36d71e4e182bd.

golem-worker-executor/src/services/active_agents/mod.rs
vigoo: Initialization now uses the existing spawned cache-creation API so one cancelled first-access waiter cannot abandon the shared creation entry and strand other callers. Normal Worker creation and initialization remain the source of truth; only ownership of that shared creation future changes. Added the rationale inline. Initializer-enqueue storage errors are propagated instead of panicking in the detached creator. Cache cancellation and runtime concurrent-first-access tests cover this.

Addressed in https://github.com/golemcloud/golem/commit/1d25f5656216c132258cddbcc3b36d71e4e182bd.

golem-worker-executor/src/services/agent_filesystem/lifecycle/inspection_stream/tests.rs
vigoo: Moved inspection_stream into a directory with mod.rs and tests.rs, using the existing module convention.

Addressed in https://github.com/golemcloud/golem/commit/1d25f5656216c132258cddbcc3b36d71e4e182bd.

golem-worker-executor/src/services/agent_filesystem/lifecycle/inspection/tests.rs
vigoo: Moved inspection into a directory with mod.rs and tests.rs, using the existing module convention.

Addressed in https://github.com/golemcloud/golem/commit/1d25f5656216c132258cddbcc3b36d71e4e182bd.

golem-worker-executor/src/services/agent_filesystem/lifecycle/mod.rs
vigoo: This replaces the implementation behind the existing get_file_contents operation used by CLI/REST; it is not a new parallel filesystem API. The old read buffered the whole file. The new metadata-first bounded stream retains filesystem/lifecycle ownership until consumer completion or drop, so HTTP live-file bindings can read coherently through initialization, replay and concurrent updates. It still accepts an exact absolute path; HTTP mapping stays outside the executor. Static router files use blob storage instead. Added this explanation to the durable-execution reference and walkthrough.

Addressed in https://github.com/golemcloud/golem/commit/1d25f5656216c132258cddbcc3b36d71e4e182bd.

golem-worker-executor/src/worker/invocation_loop.rs
vigoo: Removed executor-owned file-read deadlines from admission, queueing, streaming, configuration and protocol errors. A connected read now waits through normal initialization/invocation scheduling and completes or returns an actual error. Bounded admission and consumer-drop cleanup remain. An HTTP exchange deadline belongs to the serving layer, which can drop its read consumer. Runtime tests cover a read waiting behind guest work and completing after normal resume.

Addressed in https://github.com/golemcloud/golem/commit/1d25f5656216c132258cddbcc3b36d71e4e182bd.

test-components/agent-rpc/golem-it-agent-rpc-rust/src/http_router.rs
```diff
@@ -0,0 +1,234 @@
...
+        // Disposal of a suppressed body must not suppress the invocation failure.
+        let _ = input.next().await;
+        panic!("intentional HTTP response producer failure");
+    });
+    output
+}
+
+#[agent_implementation]
+impl RawHttpRouter for RawHttpRouterImpl {
+    fn __register_agent_type() {
```
vigoo, line 100: This should be macro-generated, not hand-written here

golem-worker-service/src/custom_api/http_envelope.rs
```diff
@@ -0,0 +1,924 @@
...
+        "upgrade",
+        "proxy-authenticate",
+        "proxy-authorization",
+    ] {
+        map.remove(name);
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
vigoo: Split to separate mod.rs/tests.rs

golem-worker-service/src/custom_api/http_completion.rs
```diff
@@ -0,0 +1,382 @@
...
+        self.aborted = true;
+        GateOutput {
+            bytes: None,
+            terminal: Some(GateTerminal::Abort(error)),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
vigoo: Split to separate mod.rs/tests.rs

golem-worker-service/src/custom_api/http_session/mod.rs
```diff
@@ -0,0 +1,1218 @@
...
+    end_sequence: u64,
+    _permit: Option<InputCredits>,
+}
+
+struct AcceptedIdentity {
+    accepted: InvocationAccepted,
+    input_mapping: DurableStreamMapping,
+}
+
+async fn run_session(
```
vigoo, line 700: This code is too complicated - need to be refactored. Extract methods and helper types, etc

golem-worker-service/src/custom_api/raw_handler.rs
```diff
@@ -0,0 +1,780 @@
...
+
+fn map_session_error(error: super::http_session::HttpSessionError) -> RequestHandlerError {
+    match error {
+        super::http_session::HttpSessionError::Deadline => RequestHandlerError::RawDeadline,
+        _ => RequestHandlerError::RawBadGateway,
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
vigoo: Split to mod.rs/tests.rs

golem-worker-service/src/custom_api/raw_handler.rs
```diff
@@ -0,0 +1,780 @@
...
+
+impl RawHandler {
+    pub(super) fn new(worker_service: Arc<WorkerService>, limits: HttpSessionLimits) -> Self {
+        Self {
+            worker_service,
+            limits,
+        }
+    }
+
+    async fn invoke(
```
vigoo: Chunk this up using helper methods/types to make it more readable

golem-worker-service/src/custom_api/raw_handler.rs
```diff
@@ -0,0 +1,780 @@
...
+                .iter()
+                .map(|byte| ProtoValue {
+                    value: Some(schema_value::Value::U8Value(*byte as u32)),
+                })
+                .collect(),
+        })),
+    }
+}
+
+fn request_value(
```
vigoo: Can we simplify this and the next function by using existing `IntoSchema` and `FromSchema` conversions and then just converting the schemavalue to its protobuf representation? If this is much faster we can keep it but should be measured

test-components/agent-rpc/golem-it-agent-rpc-rust/src/http_router.rs
vigoo: Fixed in https://github.com/golemcloud/golem/commit/77fd8cac484fdf2401c93c88ad60f69702dc863e. The fixture now uses agent_definition(kind = "http-router", ephemeral, ...) and endpoint(any = "/") rather than handwritten registration. Verified the generated kind, endpoint, auth/CORS and snapshot metadata with 33 SDK agent tests, compile-fail tests, a WASM rebuild and platform schema extraction.

golem-worker-service/src/custom_api/http_envelope.rs
vigoo: Fixed in https://github.com/golemcloud/golem/commit/77fd8cac484fdf2401c93c88ad60f69702dc863e. Split into mod.rs and tests.rs. The affected custom API/gateway suite passes all 227 tests.

golem-worker-service/src/custom_api/http_completion.rs
vigoo: Fixed in https://github.com/golemcloud/golem/commit/77fd8cac484fdf2401c93c88ad60f69702dc863e. Split into mod.rs and tests.rs. The affected custom API/gateway suite passes all 227 tests.

golem-worker-service/src/custom_api/http_session/mod.rs
vigoo: Fixed in https://github.com/golemcloud/golem/commit/77fd8cac484fdf2401c93c88ad60f69702dc863e. Extracted InputProgress, OutputProgress and RecoveryProgress, command preparation, and cancellation/drain handling. Retained input accounting, ACK/high-water handling, cursor tracking and cleanup now have named owners. Existing custom API/gateway tests pass (227), including retries, cancellation and backpressure.

golem-worker-service/src/custom_api/raw_handler.rs
vigoo: Fixed in https://github.com/golemcloud/golem/commit/77fd8cac484fdf2401c93c88ad60f69702dc863e. Split into mod.rs and tests.rs. The affected custom API/gateway suite passes all 227 tests.

golem-worker-service/src/custom_api/raw_handler.rs
vigoo: Fixed in https://github.com/golemcloud/golem/commit/77fd8cac484fdf2401c93c88ad60f69702dc863e. Split invoke into session setup, response-head reception, bodyless completion and unread-request handling, with streaming output state kept separate. The custom API/gateway suite passes all 227 tests.

golem-worker-service/src/custom_api/raw_handler.rs
vigoo: Fixed in https://github.com/golemcloud/golem/commit/77fd8cac484fdf2401c93c88ad60f69702dc863e. Replaced manual request/header codecs with canonical structs deriving IntoSchema/FromSchema and the existing protobuf conversions. Kept only transport stream-reference handling explicit. Response schema validation preserves strict record arity; regression tests cover extra header fields and integer overflow.

golem-worker-service/src/custom_api/immutable_files.rs
```diff
@@ -0,0 +1,666 @@
...
+                    }
+                }
+            }
+        }
+    })
+    .boxed()
+}
+
+#[cfg(test)]
+mod tests {
```
vigoo: Split to mod.rs/tests.rs

golem-worker-service/src/custom_api/route_resolver.rs
```diff
@@ -181,12 +274,28 @@ impl RouteResolver {
...
         Ok(DomainHttpApi {
-            environment_id,
-            router,
+            domain_exists: true,
+            reserved: build_router(reserved)?,
+            typed: build_router(typed)?,
+            mounts: Arc::new(mounts),
             openapi_spec,
+            created_at: Instant::now(),
         })
```
vigoo, line 190: That is this "magic" 0..3 loop?

golem-worker-service/src/custom_api/route_resolver.rs
```diff
@@ -60,9 +70,14 @@ impl SafeDisplay for RouteResolverError {
     }
 }
 
+type DomainApiCache = Cache<(Domain, u64), (), Arc<DomainHttpApi>, ()>;
```
vigoo: Let's make a struct with named fields for (Domain, u64) for better readability

test-components/agent-rpc/golem-it-agent-rpc-rust/src/http_router.rs
```diff
@@ -0,0 +1,275 @@
...
+    fn new() -> Self;
+    #[endpoint(any = "/")]
+    async fn route(&self, request: HttpRequest) -> HttpResponse;
+}
+
+struct StaticHttpRouterImpl;
+
+#[agent_implementation]
+impl StaticHttpRouter for StaticHttpRouterImpl {
+    fn __register_agent_type() {
```
vigoo, line 100: This is supposed to be generated by the macro, why is it inlined here??? This is not the first time - make sure you never make this mistake again

vigoo: Or is this a temporary hack because GOL-565 is not done yet? If that's the case it can remain

golem-worker-service/src/custom_api/immutable_files.rs
vigoo: Split immutable_files into mod.rs and tests.rs, preserving the production code and tests unchanged apart from formatting. Verified all 231 custom API tests, scoped Clippy with warnings denied, and formatting. Oracle and bug-finder reviews found no regressions.

Addressed in https://github.com/golemcloud/golem/commit/92bfc8e91a13ad27dca507c19af16e7631aed60e.

golem-worker-service/src/custom_api/route_resolver.rs
vigoo: This bounds snapshot admission attempts, not registry/network retries. An attempt is discarded if a registry invalidation changes the generation while its lookup is in flight, or if the cached snapshot has exceeded its maximum age. It permits one initial attempt plus two retries; persistent churn returns an error rather than admitting a known-obsolete snapshot or retrying forever. Three is a bounded retry policy, not a correctness requirement.

Named it MAX_SNAPSHOT_ADMISSION_ATTEMPTS and documented that rationale. The existing deterministic invalidation-storm test pins the three-attempt behavior; all 231 custom API tests pass.

Addressed in https://github.com/golemcloud/golem/commit/92bfc8e91a13ad27dca507c19af16e7631aed60e.

golem-worker-service/src/custom_api/route_resolver.rs
vigoo: Replaced the tuple with DomainApiCacheKey { domain, generation }, including named-field access during invalidation. Equality/hash semantics and the cache behavior are unchanged. All 231 custom API tests and scoped Clippy pass.

Addressed in https://github.com/golemcloud/golem/commit/92bfc8e91a13ad27dca507c19af16e7631aed60e.

test-components/agent-rpc/golem-it-agent-rpc-rust/src/http_router.rs
vigoo: Yes: this is specifically the temporary static-mapping fixture pending GOL-565, so I have left it in place under your allowance. I checked the ticket and current macro: agent_definition accepts kind/mode/snapshotting/mount/auth/CORS/webhook metadata, but not static mappings; get_http_mount_details initializes static_bindings to an empty list. GOL-565 explicitly owns ordered immutable static mappings in the Rust macro/builder API.

RawHttpRouter remains macro-generated. Only the separate StaticHttpRouter fixture overrides registration so GOL-559 can exercise real deployment and immutable-file delivery now. Once GOL-565 exposes those declarations, this override should be replaced by the macro. I should have made this dependency explicit when adding the fixture; handwritten registration is not the intended pattern for SDK-supported declarations.

golem-common/src/schema/agent/http/mod.rs
```diff
@@ -0,0 +1,336 @@
+// Copyright 2024-2026 Golem Cloud
```
vigoo, line 1: Follow the existing file layout practice and make this `http/mod.rs`

golem-rust-macro/src/agentic/agent_definition_attributes.rs
```diff
@@ -124,30 +199,109 @@ pub fn parse_agent_definition_attributes(
...
+    }
+    if (has_filesystem || !http.static_files.is_empty() || http.openapi_provider_method.is_some())
+        && http.mount.is_none()
+    {
+        return Err(Error::new(
+            proc_macro2::Span::call_site(),
+            "file mappings and OpenAPI providers require a mount",
+        ));
+    }
+    if kind == "HttpRouter" {
```
vigoo, line 216: The user should not be able to set `kind` in `#[agent_definition]` - we introduced the special `#[http_router]` for that

golem-schema/src/schema/schema_type.rs
```diff
@@ -660,6 +676,24 @@ pub struct NamedFieldType {
     pub metadata: MetadataEnvelope,
 }
 
+#[cfg(test)]
+mod record_constructor_tests {
```
vigoo: why does this function have its own separate test module? clean this up

golem-service-base/src/custom_api/mod.rs
```diff
@@ -318,7 +357,7 @@ pub struct CompiledRoutes {
 #[derive(Debug)]
 pub struct CompiledRoute {
     pub route_id: RouteId,
-    pub method: HttpMethod,
+    pub route_match: RouteMatch,
     pub path: Vec<PathSegment>,
     // TODO: move this into the individual route behaviours
```
vigoo: Does this old todo comment still make sense? Let's see what it would take to satisfy it

golem-service-base/src/custom_api/mod.rs
```diff
@@ -241,6 +243,29 @@ pub struct CompiledInputSchema {
     pub input_schema: InputSchema,
 }
 
+impl CompiledInputSchema {
```
vigoo, line 246: It's not strictly related to the new changes in this PR - but what's the reason for these types and methods to live in `golem-service-base` and not in the worker service, that implements the http routes?

golem-worker-executor/src/durable_host/mod.rs
```diff
@@ -6007,76 +6008,16 @@ impl<Ctx: WorkerCtx> ExternalOperations<Ctx> for DurableWorkerCtx<Ctx> {
...
+                        } else {
+                            invoke_result
+                        };
                         match invoke_result {
                             Ok(InvokeResult::Succeeded {
-                                result: mut invocation_result,
+                                result: invocation_result,
                                 consumed_fuel,
                             }) => {
-                                if uses_streams
```
vigoo, line 6015: What's the reason for this change?

golem-worker-executor/src/services/agent_filesystem/lifecycle/inspection/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+    Rejected(FileReadHead),
+}
+
+/// Opens a canonical absolute path without following any intermediate or final symlink.
+///
+/// The caller must hold the resident Store and exclusive owner lane from before this call until
+/// the returned file has finished streaming. This is observation, not a durable guest invocation.
+/// Metadata is captured from the opened descriptor; path metadata only rejects unsafe opens.
+/// Directories never select an index file. No returned error contains a host path.
+pub(crate) async fn open_file_for_inspection<Adapter: SandboxFilesystemAdapter>(
```
vigoo, line 38: What do we need the new "inspection" operation for, regarding the custom http routing / file serving spec?

golem-worker-executor/src/services/file_read_admission.rs
```diff
@@ -0,0 +1,323 @@
...
+use crate::services::golem_config::FileReadConfig;
+use golem_common::model::OwnedAgentId;
+use golem_common::model::filesystem::FileReadError;
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+use tokio::sync::{OwnedSemaphorePermit, Semaphore};
+
+/// Executor-local admission, shared by all requests, including requests for absent agents.
+#[derive(Debug)]
+pub struct FileReadAdmission {
```
vigoo: Do we need a separate mechanism to limit file reads? We already have various limits coming from the user's plan, enforced by the executor.

golem-worker-executor/src/worker/invocation.rs
```diff
@@ -276,13 +279,110 @@ async fn invoke_observed<Ctx: WorkerCtx>(
...
 
-    let call_result = apply_invocation_deadline(&mut store, deadline, call_result).await;
+    apply_invocation_deadline(&mut store, deadline, call_result).await
+}
 
-    if manages_owner_execution_status {
-        store.data().set_suspended();
+/// Publishes the early stream-bearing result and drives its producers to completion.
+/// Live execution and replay use the same trap classification as the guest export.
+pub(crate) async fn materialize_streaming_result<Ctx: WorkerCtx>(
```
vigoo, line 287: Explain what needed to be changed about streaming invocation, why, and how it affects other use cases

golem-worker-executor/src/worker/invocation_loop.rs
```diff
@@ -13,22 +13,24 @@
 // limitations under the License.
```
vigoo, line 13: I did not expecct other changes to the invocation loop than supporting the new file-system access commands, explain

golem-worker-executor/src/worker/mod.rs
```diff
@@ -618,7 +621,7 @@ pub struct ResolvedWorkerData<Ctx: WorkerCtx> {
     pub(crate) tasks: tasks::WorkerTasks,
     worker_event_service: Arc<dyn WorkerEventService + Send + Sync>,
 
-    queue: Arc<RwLock<VecDeque<QueuedWorkerInvocation>>>,
+    queue: Arc<StdMutex<VecDeque<QueuedWorkerInvocation>>>,
```
vigoo: Explain the change

golem-worker-service/src/custom_api/http_session/mod.rs
```diff
@@ -0,0 +1,1389 @@
+// Copyright 2024-2026 Golem Cloud
+//
+// Licensed under the Golem Source License v1.1 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://license.golem.cloud/LICENSE
+
+//! Retry-owning invocation sessions used by the custom HTTP adapter.
```
vigoo: This whole module deserves better structuring and more inline (and doc comment on extracted methods etc) documentation

golem-worker-service/src/custom_api/openapi/cache.rs
```diff
@@ -0,0 +1,231 @@
+// Copyright 2024-2026 Golem Cloud
```
vigoo: Follow the existing practice of file layout - if there is a `cache/tests.rs` then the cache module should be in `cache/mod.rs` not in `cache.rs`. Apply this rule to the whole changeset

golem-worker-service/src/custom_api/openapi/budget.rs
```diff
@@ -0,0 +1,148 @@
...
+use std::io::{self, Write};
+use std::time::Duration;
+use tokio::time::Instant;
+use tokio_util::sync::CancellationToken;
+
+pub(crate) const GENERATION_TIMEOUT: Duration = Duration::from_secs(30);
+pub(super) const DOCUMENT_BYTE_LIMIT: usize = 8 * 1024 * 1024;
+
+#[derive(Clone)]
+pub(super) struct Budget {
```
vigoo: Why do we need this?

golem-worker-service/src/custom_api/openapi/cache.rs
```diff
@@ -0,0 +1,231 @@
...
+struct Completed {
+    generation: Arc<Generation>,
+    expires_at: Instant,
+    result: Result,
+}
+
+// Pending entries cannot be evicted by completed-entry LRU pressure. Admission
+// remains with the generation's CPU/cleanup tasks, even after invalidation.
+#[derive(Default)]
+pub(super) struct CacheState {
```
vigoo: I want you to use our `Cache` type from `golem_common` for all cache implementations.

golem-worker-service/src/custom_api/openapi/service/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+use tokio::sync::{OwnedSemaphorePermit, Semaphore};
+use tokio::time::{Instant, timeout, timeout_at};
+
+const PROVIDER_TIMEOUT: Duration = Duration::from_secs(5);
+const CLEANUP_TIMEOUT: Duration = Duration::from_secs(2);
+const PROVIDER_CONCURRENCY: usize = 8;
+const GENERATION_CONCURRENCY: usize = 8;
+
+pub struct OpenApiDocument {
+    pub json: Bytes,
```
vigoo, line 35: Did we support JSON before? If not and only yaml, then let's keep it that way

golem-worker-service/src/custom_api/openapi/mod.rs
```diff
@@ -10,13 +10,70 @@
...
+
+use crate::custom_api::RichCompiledRoute;
+use golem_common::model::deployment::DeploymentRevision;
+use golem_common::model::domain_registration::Domain;
+use golem_common::model::environment::EnvironmentId;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+#[derive(Clone)]
+pub(crate) struct Freshness {
```
vigoo: What's this for? I think the generation and caching of OpenAPI specs got a bit over complicated in this PR. What I think we need is simply a cached openapi document per deployment, lazily generated on first request, and instead of just generating it from the http route metadata as before, it needs to optionally call the component's custom openapi fragment genreator and merge the two. How it calls it is the interesting part and can be completely orthogonal to the cache implementation

golem-worker-service/src/custom_api/raw_handler/mod.rs
```diff
@@ -0,0 +1,626 @@
+// Copyright 2024-2026 Golem Cloud
+//
+// Licensed under the Golem Source License v1.1 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://license.golem.cloud/LICENSE
```
vigoo, line 7: How is this module related to the `http_session` module? It's confusing to me

golem-common/src/schema/agent/http/mod.rs
vigoo: Done: moved the HTTP schema module to http/mod.rs, following the directory-module convention.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-rust-macro/src/agentic/agent_definition_attributes.rs
vigoo: Done: removed the public kind argument from #[agent_definition]. #[http_router] now passes the router kind through private typed macro expansion; ordinary agent definitions cannot select it. Updated fixtures and compiler-diagnostic coverage accordingly.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-schema/src/schema/schema_type.rs
vigoo: Done: moved the record_from_fields tests into schema_type's existing test module and removed the separate helper-specific module. The schema tests pass.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-service-base/src/custom_api/mod.rs
vigoo: Satisfied the TODO: body schema now belongs to CallAgentBehaviour rather than the enclosing compiled route. Updated compilation, persistence/protobuf conversions, request parsing, CORS and OpenAPI consumers together. Non-agent behaviours no longer carry irrelevant body-schema state. Added/retained regression coverage for webhook binary bodies.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-service-base/src/custom_api/mod.rs
vigoo: These are shared compiled route representations, not the HTTP server implementation. The registry service compiles and serializes them; the worker service loads them and executes the resulting routes. Keeping that contract and its conversions in golem-service-base avoids making the registry depend on the worker-service implementation. As agreed, no relocation is needed here; HTTP request handling remains in worker-service.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-executor/src/durable_host/mod.rs
vigoo: This makes replay use the same streaming-result materialization as live execution. A guest function can return while its returned streams are still being produced; the invocation result is not complete at the function-return boundary. Replay must drain/materialize those results and handle stream failures consistently with live execution. If replay reaches live mode while finishing an interrupted invocation, it must also complete the durable streaming session so resumed clients receive completion. This applies to streaming invocations generally, not only custom HTTP handlers. We retained the shared path as agreed.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-executor/src/services/agent_filesystem/lifecycle/inspection/mod.rs
vigoo: The file operation serves the existing agent filesystem API as well as HTTP live-file delivery; it is not a separate HTTP-only executor API. The old implementation buffered the complete file. Streaming needs safe opening, metadata/byte selection and bounded production while retaining filesystem ownership against concurrent mutation.

Following our discussion, removed the separate inspection scheduling/lifecycle policy. Reads and listings now enter common worker scheduling and use normal activation, initialization, reconstruction, updates and failure handling. Native requests/bytes are not journaled. Existing filesystem ownership is held while the producer reads and released after successful production, rather than waiting for consumer-observed EOF. Static immutable files still use blob storage and need no executor. Runtime tests cover initialization, recovery, Suspend/resume, update and backpressure.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-executor/src/services/file_read_admission.rs
vigoo: Agreed; removed FileReadAdmission, its registry/reservations/semaphore, configuration and admission-only errors. No replacement per-account quota was introduced. Bounded buffering remains per response, but there is no new aggregate outstanding-file-read limit. Slow-consumer backpressure may keep the agent occupied just as ordinary sequential work can; that is the accepted behaviour.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-executor/src/worker/invocation.rs
vigoo: The old suspension point could mark an invocation suspended when the function returned, although returned streams were still running. That left the post-return part of the invocation with incorrect interrupt handling.

We keep execution running through stream materialization/settlement and suspend afterward. This is more than moving one call: interrupt delivery, typed trap/error classification, replay failure handling and session settlement must remain consistent over that extended interval. It affects streaming RPC/entity/custom-HTTP invocations; non-streaming invocations have no additional stream-production interval. Oracle reviewed the broader changes and confirmed they are related correctness fixes, which we agreed to retain with regression coverage.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-executor/src/worker/invocation_loop.rs
vigoo: The revision removes inspection-specific scheduling and lifecycle machinery. Native reads/listings use the common loop and normal worker activation/reconstruction/update/failure handling; a suspended unfinished invocation must resume before boundary-only native work runs.

One ordering change remains: main keeps persisted invocations in oplog-derived status and transient commands in an in-memory deque, processing the latter first. To avoid a read overtaking an already accepted invocation, transient filesystem requests capture their position relative to persisted work. select_next_work applies that common policy, giving A -> read -> B when accepted in that order, without journaling the read. invocation_queue contains this ordering/pruning policy, not another execution loop or admission system. The separate streaming-interrupt correctness fix is explained in the adjacent invocation.rs thread. Runtime tests include ordinary Suspend and the concurrent-agent permit-wait path.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-executor/src/worker/mod.rs
vigoo: Restored Tokio's async RwLock. The synchronous lock had been introduced to remove cancelled requests directly from Drop; that mechanism is now removed. Abandoned transient requests are pruned at enqueue/selection and lifecycle boundaries instead. Accepted durable invocations are never pruned just because their caller disconnected. A cancelled transient record may remain until the next pruning boundary, which is the agreed tradeoff.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-service/src/custom_api/http_session/mod.rs
vigoo: Done: split http_session into driver, progress and transport responsibilities and added method/module documentation for protocol state transitions, input/output accounting and recovery ownership. HTTP envelope/body adaptation remains outside the session protocol. The separate recovery correctness changes have regression tests and are described in the RawHandler thread below.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-service/src/custom_api/openapi/cache.rs
vigoo: Done: applied the directory-module convention throughout the affected modules, including cache/mod.rs alongside cache/tests.rs and the HTTP schema module. Removed the superseded flat module files.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-service/src/custom_api/openapi/budget.rs
vigoo: Removed the OpenAPI-specific Budget, provider/generation/request timeouts, cancellation checks and timeout-driven cleanup. The provider is an ordinary invocation on a fresh ephemeral http-router instance; normal invocation admission/execution owns overload handling. It does not need a separate OpenAPI lifecycle.

Retained document-size limits and a spawn_blocking boundary for synchronous document processing. spawn_blocking avoids running CPU-bound parsing/validation/merge/serialization on an async runtime worker; it is not a timeout or provider-execution mechanism. The remaining bounded_json helpers enforce data-size bounds, not time budgets.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-service/src/custom_api/openapi/cache.rs
vigoo: Done: replaced the bespoke OpenAPI cache with golem_common::Cache, using its spawned singleflight creation path for lazy document generation. Concurrent requests for the same snapshot share generation; cancellation of an individual waiter does not abandon that shared fill. Capacity remains bounded at 256 entries. Removed custom freshness counters, stale-delivery fencing/retry loops and negative caching.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-service/src/custom_api/openapi/service/mod.rs
vigoo: Confirmed that both JSON and YAML OpenAPI endpoints existed before this PR. Kept both; this is not an added format requirement.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-service/src/custom_api/openapi/mod.rs
vigoo: Implemented the simpler model: lazily generate a document from a captured immutable route snapshot, optionally invoke providers through the ordinary ephemeral invocation path, merge their fragments, and cache the result with golem_common::Cache. Provider invocation is independent of the cache implementation.

The cache key is the immutable route-snapshot UUID, so changed route/origin/security inputs use a distinct document identity. Removed OpenAPI-specific freshness generations, stale-delivery fences, retry loops and negative caching. An already-running fill for an older captured snapshot may finish, as agreed; new snapshots use their own keys. Tests cover singleflight generation and snapshot identity/invalidation behaviour.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-service/src/custom_api/raw_handler/mod.rs
vigoo: Separated the responsibilities: MountedDispatch owns mounted-backend selection/dependencies; RawHandler adapts the HTTP request/response envelope and bodies; http_session owns the private streaming-session protocol, transport and recovery. Restructured the session into driver/progress/transport modules with documentation rather than introducing another universal driver.

Also fixed the recovery inconsistencies we discussed, with regression tests: an ambiguous transport failure now retains the exact pending resume attempt ID so an accepted attempt can be replayed. Resume versus Takeover follows attachment state and changes only after an authoritative rejection, rather than unconditionally requesting Takeover or creating a fresh attempt on every retry.

Implemented/reviewed in https://github.com/golemcloud/golem/commit/71e9f29fad01db6c8520a7c0a05c0fd8aa84eb3e. Oracle review approved the final fixes. After the latest main merge, CI is green: https://github.com/golemcloud/golem/actions/runs/35634342722 (54 passed, 7 skipped).

golem-worker-executor/src/worker/invocation_loop.rs
```diff
@@ -1627,8 +1688,48 @@ impl<Ctx: WorkerCtx> InnerInvocationLoop<'_, Ctx> {
...
+                pending.has_idempotency_key(&self.parent.initialization_idempotency_key())
+            });
+        if !needs_initialization && let Some(invocation) = queue.pop_front() {
+            return SelectedWork::Resident(invocation);
+        }
+        if !status.pending_updates.is_empty() {
+            return SelectedWork::ApplyPendingUpdate;
+        }
+        if let Some(pending) = status.pending_invocations.first() {
+            return SelectedWork::Durable(pending.clone());
```
noise64: > `select_next_work` drains every resident request before checking pending updates or durable invocations. Because filesystem requests enter an unbounded resident deque and this PR intentionally removes aggregate file-read admission, continuous live-file traffic can keep the resident queue non-empty forever, so an already accepted agent method or pending update never runs. This is stronger than merely observing the current state at an invocation boundary. Could we add a bounded fairness rule, plus a test that keeps enqueueing reads while proving a pending invocation and update still make progress?

golem-worker-service/src/custom_api/route_resolver.rs
```diff
@@ -173,20 +278,41 @@ impl RouteResolver {
...
-            Err(e) => {
-                tracing::warn!("Failed to build openapi spec for http api: {e}");
+        let finalized_routes: Vec<_> = finalized_routes.into_iter().map(Arc::new).collect();
+        let openapi_inputs = finalized_routes.iter().find_map(|route| {
+            if let RichRouteBehaviour::OpenApiSpec(behavior) = &route.behavior {
+                Some(Arc::new(OpenApiInputs {
+                    key: OpenApiKey::fresh(),
+                    public_origin: behavior.scheme.origin(domain),
+                    routes: finalized_routes.clone(),
+                }))
```
noise64, line 289: > The review thread describes one lazily generated OpenAPI document per deployment, but `fetch_and_build_domain_api` assigns a fresh random key whenever the domain router is rebuilt. With the default ten-minute router TTL, an unchanged deployment re-runs every provider periodically and leaves the previous document in the 256-entry LRU until eviction. Is that periodic refresh intentional? If not, could the key derive from stable deployment/route identity (including the origin and security inputs) and have a test for a router-cache refill with unchanged routes? If it is intentional, please document the refresh semantics because they differ from the thread resolution.

cli/golem-cli/src/model/app_raw/mod.rs
```diff
@@ -1161,6 +1161,8 @@ pub struct HttpApiDeployment {
     pub domain: Option<DeploymentDomain>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub subdomain: Option<DeploymentSubdomain>,
+    #[serde(default)]
+    pub scheme: golem_common::model::http_api_deployment::HttpApiDeploymentScheme,
```
noise64: > `scheme` is needed for a deterministic host-owned OpenAPI `servers` origin, but defaulting an omitted value to `Https` here happens before the deployment environment is resolved. Stock local manifests use plain HTTP (`*.localhost:9006`) and omit `scheme`, so their generated OpenAPI advertises an unreachable `https://...` server; the human manifest reference does not document the field either. Could the raw value remain optional and be resolved from the built-in target (HTTP for local, HTTPS for cloud), while retaining an explicit override for custom servers? Please also cover an omitted local value and document the field.

PR #3886, opened 2026-09-15
Port #3802 to main: archive ephemeral oplogs from a sweep

kmatasfp:
Ports #3802 to `main`.

`main` moved `schedule_oplog_archive_if_needed` from `Worker` onto the state actor's `StatusState`, so the ephemeral guard is in `worker/state_actor.rs`. The condition and its comment are unchanged. Doc comments that named `Worker::schedule_oplog_archive_if_needed` or `update_state_lock` now point at the new location.

`main`'s #3797 removed `SchedulerWorkerAccess::open_oplog`, so the sweep archives through `Worker::archive_oplog`, like the scheduled action. A new `ArchiveWait` argument sets when that call returns. The scheduler passes `Queued` and behaves as before. The sweep passes `Finished` and holds the agent's oplog lifecycle guard until the transfer ends, so `max_concurrency` bounds the transfers in flight. `max_tick_duration` only stops further agents from starting. When the worker declines because the agent is being deleted, its oplog was retired, its index moved, or it no longer exists, the sweep records a new `declined` outcome. It asks again on a later pass, or, if the attempt left a worker in `ActiveAgents`, once that worker expires.

Waiting for the transfer exposed a hang in `main`'s code. With every movable layer already empty, `EphemeralOplog::archive` and `MultiLayerOplog::archive` awaited a transfer that was never enqueued, so `try_archive_blocking` never returned. Both now return `false` there. `Worker::archive_oplog` with `Finished` is the only production caller, and the `read_initial_from_archive` tests now call it on a fully archived oplog.

Other differences from the 1.5.x patch:

- `owns` calls `ShardAssignment::contains`, since `shard_ids` is not a field on `main`.
- The sweep also requires a live shard lease, mirroring `check_admission`.
- `Outcome::Moving` / label `moving` is now `Outcome::Waiting` / `waiting`.
- `SWEPT_MODES` is gone; the sweep hardcodes ephemeral mode and the layer-count hazard is documented on the module.
- A declined archive logs at debug level and counts as `declined`. In #3802 an oplog that no archive step recognised logged a warning and counted as `archive_failed`.
- The executor's `shutdown` service tracks the sweeper instead of the join set. On termination it gets `SHUTDOWN_GRACE`, 10 seconds, to finish the agents under way. #3802 waited until they had drained all their layers.
- The debugging service defaults `oplog.sweep.enabled` to `false`. It never sets the worker activator that archiving goes through, so every sweep archive there would fail. With the sweep off it registers the scheduled archive for ephemeral agents, as `main` does today.
- The sweep's test doubles follow `main`'s trait signatures: shard epochs and lease revisions, `open_fresh`/`read_source`/`verify_persisted` on the archive doubles, and the 8-parameter `enqueue_invocation`. `main`'s `assign_shards` rejects an executor that never registered, so the zero-shard test reaches that state through `register(0, ...)`.
- `ReadCountingIndexedStorage` in `oplog/tests.rs` delegates the new `scan_stable` and `last_id`.
- Config files are regenerated and only add `[oplog.sweep]`.
- `OplogConfig::sweep` has no `#[serde(default)]`, but the config loader merges the defaults first, so a config file without `[oplog.sweep]` still loads, with the sweep enabled.

Tested locally: the 59 `oplog_sweep` unit tests, the oplog, scheduler and indexed-storage unit tests, `scan_stable`/`last_id` integration tests on all five backends, the `oplog_sweep` and `oplog_archive_schedule` integration tests, and clippy with `-D warnings` on the worker executor and the debugging service.


golem-worker-executor/src/services/oplog_sweep.rs
```diff
@@ -0,0 +1,4256 @@
...
+
+    /// Whether this executor is running the agent right now.
+    ///
+    /// In memory, so it costs nothing to ask. `ActiveWorkers` keys on the agent id alone, so the
+    /// environment on the id is not part of the question.
+    async fn is_resident(&self, owned_agent_id: &OwnedAgentId) -> bool {
+        self.worker_access
+            .active_worker_fingerprint(owned_agent_id)
+            .await
+            .is_some()
```
vigoo, line 1001: > **[P2] Avoid refreshing cache residency from the sweep**
>
> On `main`, `active_worker_fingerprint` calls `ActiveAgents::try_get` → `Cache::get`, which refreshes the entry’s TTL. If an invocation-result lookup reconstructs a suspended ephemeral worker after a crash, every sweep subsequently returns `Resident` and refreshes its eight-hour TTL. With the default sweep interval, that worker never expires and its stranded oplog never reaches archival.
>
> #3802 already acknowledged suspended-worker residency, but `main` now has unloaded-worker TTL eviction that this probe defeats. Please use a non-touching cache-presence check, or distinguish executing workers from observation-only shells. A regression test should cover a lookup-created suspended worker becoming eligible for archival.

golem-worker-executor/src/lib.rs
```diff
@@ -997,6 +1001,25 @@ pub async fn create_worker_executor_impl<
...
+        component_service.clone(),
+        Arc::new(lazy_worker_activator.clone() as Arc<dyn WorkerActivator<Ctx>>),
+    );
+    join_set.spawn({
+        let shutdown_token = shutdown_token.clone();
+        async move {
+            oplog_sweeper.run(shutdown_token).await;
+            Ok(())
+        }
+    });
```
vigoo, line 1021: > **[P2] Register the sweeper with the shutdown tracker**
>
> Unlike the `1.5.x` server loop, `main`’s termination path cancels `Shutdown`, awaits `wait_for_tracked`, then calls `join_set.shutdown()`, which aborts remaining tasks. Spawning the sweep here therefore does not provide the drain-on-shutdown behavior described above: an in-flight archive can be cut off as soon as the tracked tasks finish, even if it would finish within the shutdown grace period.
>
> Please register the sweep through `Shutdown::spawn` so shutdown waits for its cooperative cancellation boundary. This matters particularly for indexed-to-indexed transfers, where interruption after append but before source deletion leaves a non-idempotent retry.

golem-worker-executor/src/services/oplog_sweep.rs
kmatasfp: addressed

golem-worker-executor/src/lib.rs
kmatasfp: addressed

vigoo (approved):

PR #3891, opened 2026-09-15
Add blob listing with sizes, strict ranged reads and chunked S3 deletes

kmatasfp:
- resolves GOL-598
- targets `filesystem-snapshot`, not `main`, like the rest of the epic. It has no other pull request under it: its commits sit directly on that branch, with one merge that brings the branch up to date.

Three changes to `BlobStorage`. The filesystem snapshotting work needs all three before it can use blob storage as a backing store.

### A listing that carries sizes

`list_blobs_below` gives every blob below a path, at all depths, with its size, as `Box<[ListedBlob]>`. The snapshot bridge needs a size for each blob, and the only way to get one was a `get_metadata` call for each path.

It is a required trait method, not a defaulted one. A default must walk `list_dir` and call `get_metadata` on each entry, which is the cost this method exists to remove, and on S3 it is slow in a way that no caller can see. Required means that each backend answers with its own cheapest walk: S3 uses the paged `ListObjectsV2` walk it already runs, so a size costs no more requests; SQLite is one query with a case-sensitive range on the primary key, because `LIKE` ignores ASCII case; fs walks the tree on a blocking thread through one `try_fold`; the in-memory backend reads its map one time.

Two S3 decisions need a word, and both come from one fact: S3 has no directories.

S3 holds a flat set of keys. To record a directory that holds no blob, `create_dir` writes a zero byte object at `<dir>/__dir_marker`. Other S3 tools record a directory a second way, with a key that ends in `/`. The listing leaves out both, because a snapshot that carried them would restore two files that the guest never made. The filter reads the name of the key, not who wrote it, so it also hides an object that a guest names `__dir_marker`. That collision is older than this change and #3910 fixes it, by reserving the name. `exists` and `list_dir` still read the marker, so it keeps its meaning. Only the blob listing hides it.

S3 gives the size of each key in the listing, but the field is optional in the SDK type. When the size is absent the listing gives an error, and the message names the key. The alternative was a zero. A zero is the worse error: the blob enters the snapshot as an empty file, the snapshot looks correct, and the loss shows only when someone restores it.

### Ranged reads are inclusive, and give an error instead of a guess

`get_raw_slice(start, end)` gives `end - start + 1` bytes on every backend. A range with a byte that the blob does not have gives `BlobRangeError`.

Inclusive is not a new convention. `wit/deps/blobstore/container.wit` says "Start and end offsets are inclusive" directly above `get-data`. The implementation was exclusive on some backends and inclusive on others, so the host was off by one against the specification it ships, and it was off by one inconsistently. Every caller that this pull request touches is that off-by-one put right, in `host-api-tests`, the Scala test agent, and the TypeScript and Effect SDKs.

The two SDKs show why this belonged at the host. Each carried the same written caveat, that ranged reads "are not portable", and named which backends read `end` as exclusive. Each had the same workaround in its full-object read: ask for `size - 1`, and when the answer came back short, ask again for `size`. Both are gone. Each SDK reads the size and then makes one ranged read, where a short answer used to make it read a second time.

The Effect SDK needed one thing more. Its two test doubles, the client fake and the WASI container mock, both read `end` as exclusive, so its tests agreed with a host that no longer exists. Both now hold to the same contract as the backends.

The error type matters more than it looks. The old default was `data[start..end]`, which panics, so the fs, SQLite and in-memory backends could stop the executor on a range that a guest chose. The WASI host gives `BlobRangeError` to the guest as `InvalidInput`, which is permanent, so the guest gets the error on its first attempt and the executor keeps running. A clamp would have hidden the guest's error and given the snapshot fewer bytes than it asked for.

One function, `blob_range`, holds the rule. The default `get_raw_slice` calls it, and so does the S3 backend. "Every backend uses the same rule" is therefore a property of the call graph, not a sentence in a comment that someone must keep true.

### What S3 needs that the other backends do not

S3 is lenient where the other backends are strict, so the backend reads the answer rather than trusting it.

A `start` after `end` is an invalid range, and RFC 9110 lets a server ignore or reject it, so the backend sends no request for it.

A 416 tells the backend that the range has no byte in the object. It is permanent, so the backend does not send the request again, does not write the error to the error log, and does not count it as a failure. A missing key gets the same treatment, and one predicate now holds both, because they are one policy.

A 206 must carry a `Content-Range`. The backend reads it: the range must match the range of the request, and the body must hold `end - start + 1` bytes.

The case that needed the most work is a 200. RFC 9110 lets a server ignore the `Range` header and answer with the full object, and it says directly that a client cannot depend on a 416. A server that does this is correct, so the backend must read such an answer rather than reject it.

Before, a 200 gave an error that the guest saw as transient, and the executor sent the read again. The bytes it asked for were in the answer each time, and no number of attempts could change the result. Now the backend uses the body of a 200 as the full object and applies `blob_range` to it, which gives the bytes for a range in the object and a `BlobRangeError` for a range that is not.

Two things make that decision safe.

The `Content-Length` of a 200 tells the backend the size of the object, so a range that is not in the object gives an error before the body is read. A large object that a server sends for a bad range costs one header, not a buffer.

The status separates a 200 from a 206 that lost its `Content-Range`. A 206 holds a part of the object, so its body must not be read as the full object. `GetObjectOutput` does not carry the status, and the SDK gives it only on the error path, so the backend reads it with an SDK interceptor that records the status of each response before the SDK builds the output. Each read makes its own recorder, so two reads cannot see each other's status.

### S3 bulk deletes are chunked

`delete_many` and `delete_dir` share one private function. It sends at most 1,000 keys in each `DeleteObjects` request, one request at a time, in quiet mode, and no request for an empty list. 1,000 is the S3 limit, and past it MinIO answers `MalformedXML`. The chunks go one after another and the loop stops at the first chunk that gives an error, which decides what a failure leaves behind. A response that carries an error for one key makes the attempt an error; the behaviour section says what that costs.

### Behaviour changes to know

**A read that ends at or past the end of the blob.** This is the one change that a working guest can see.

The offsets are inclusive, so a blob of 50 bytes answers `get-data(0, 49)`. Two guests reach past that end: one that wants the full blob and does not know the size, so it sends a large end, and one that sends `(0, 50)` because it read the end as exclusive.

Before, the deployed backend decided what happened. S3 sent `Range: bytes=0-50`, got a 206 with `Content-Range: bytes 0-49/50` and 50 bytes, because HTTP cuts a range to the end of the object. The old code gave those bytes and checked nothing, so the guest asked for 51 bytes, got 50, and nothing told it. The fs, SQLite and in-memory backends used `data[start..end]`, which panics when `end` is past the length, on a range that the guest chose.

So one guest call had two failure modes and the configuration chose between them. Now all four backends give `BlobRangeError`.

**An out-of-range read on S3 is now permanent, not transient.** This follows from the 200 handling above. The guest gets `InvalidInput` on the first attempt, as it does on the other three backends, rather than a retried error that can never succeed.

**The oplog archive still panics on a delete error.** `drop_prefix` and the archive `delete` panicked before this change too. An error for one key that survives the retries now reaches that panic, where before it was ignored. The panic stays as it is, because a change there belongs in its own pull request. This is the one place where the stricter delete changes a failure mode rather than a return value.

**An error for one key sends the chunk again.** A response that carries an error for one key makes the attempt an error, so the retry sends that chunk again, all 1,000 keys of it.

Sending again is safe on its own. `DeleteObjects` is idempotent, so a key that is already gone deletes again without an error. The keys that succeeded cost one more request, not a wrong result.

The case to watch is a key that can never be deleted, such as one that an object lock holds. It makes its chunk an error on every attempt, uses the retry budget, and makes the call an error. The loop stops at the first chunk that gives an error, so the chunks behind it never go out and a prune can stop half done. The message names the count and the first key that did not delete.

There is no claim here about a size at which the repeat is free, because this code does not set the rate. The snapshot pruning will call `delete_many`, and what it deletes is not known yet. Two things hold whatever the caller sends: a chunk is at most 1,000 keys, and the retry budget bounds the attempts on it.

### Notes for review

- `s3.rs` reads as a bigger change than it is. Its inline `mod tests` moved to `s3/tests.rs`, which accounts for most of the deleted lines.
- `S3BlobStorage::with_sdk_config` is new so that those tests can run against a scripted HTTP transport with SDK retries off: no environment read, no socket opened. `new` builds its config as before. `aws-smithy-runtime-api` is a new dev-dependency, and aws-sdk-s3 already resolves that version. `tracing-subscriber` is a new dev-dependency, for the test that holds the error-log policy.
- `ListedBlob.path` is `Box<Path>` and the result is `Box<[ListedBlob]>` rather than `Vec`, because the value is built one time and never grown, so the spare capacity that a `Vec` holds is dead for the life of the result.
- #3910 follows this one and targets its branch. It holds the rest of the blob storage work: one form for a path, one contract for a name, and the agreement between the in-memory backend and S3. The backend differences that neither of the two removes are written up in GOL-622.


vigoo (approved):

kmatasfp:
Re-parented onto `filesystem-snapshot`, so this no longer waits for #3872.

The four commits were replayed onto `filesystem-snapshot` with `git cherry-pick` and applied with no conflict. The two branches touch a disjoint set of files, 17 here and 28 there, with no overlap, so the stacking was organisational rather than technical.

**The content is byte-identical.** `git diff` between the old head `979633798` and the new head `5a61549d7`, over every file this pull request touches, is empty. The diff against the new base is the same 17 files, 1518 insertions and 116 deletions as before. Only the parent commits changed.

This did rewrite the branch, which normally does not happen once a pull request is approved. It was agreed first, and there are no inline review comments on this pull request, so no review anchors were lost. The old head is `9796337984a4a6d63583ae443b34cd98eb8fc03b` if anything needs checking against it.

Checked on the new base before pushing: `cargo clippy -p golem-service-base --all-targets --no-deps -- -D warnings` exit 0, and `cargo test -p golem-service-base --test integration -- blob_storage` 233 passed, 0 failed. Full CI is running now.

The stack was removed, so #3872 is independent and unaffected. It can take `filesystem-snapshot` whenever convenient; the disjoint file sets mean no conflict.

PR #3893, opened 2026-09-16
Delete blob directories by key range instead of a case-insensitive LIKE

kmatasfp:
- The SQLite blob storage deleted a directory with `parent LIKE '<dir>/%' ESCAPE '!'`. SQLite `LIKE` ignores ASCII case, so `delete_dir("foo")` also deleted the rows whose parent was `FOO/...`.
- Through the WASI blob store, a guest that cleared or deleted the container `foo` therefore lost the objects of the container `FOO` whose names contain a `/`. An object directly in `FOO` stayed, because the exact match of the parent is case-sensitive.
- The delete now uses a range on the primary key: `parent = ?` for the direct children, and `parent >= '<dir>/' AND parent < '<dir>0'` for the deeper ones, because `0` is the character after `/`. Text comparisons use the BINARY collation, so the match is case-sensitive.
- `escape_like` had this one caller, so it is gone.

### Query plan

The change is about correctness, not speed. On the bundled SQLite 3.46.0, `EXPLAIN QUERY PLAN` gives `SEARCH blob_storage USING INDEX sqlite_autoindex_blob_storage_1 (namespace=?)` for the new statement and for the old one alike: the `OR` keeps SQLite from a search on `parent`. The range does become `(namespace=? AND parent>? AND parent<?)` when it stands alone, and `LIKE` never does, so the range is the shape that can use the primary key later. An earlier draft of this description claimed the speedup; that claim is gone.

### Verification

- `cargo fmt -p golem-service-base -- --check`: exit 0
- `cargo clippy -p golem-service-base --all-targets --no-deps -- -D warnings`: exit 0
- New test `delete_dir_keeps_siblings_that_differ_only_in_case`, over each storage and each namespace. It writes `foo/a` and, only where the store tells `FOO` from `foo`, the sibling `FOO/x/c`. After the delete of `foo`, the sibling is still there and `foo/a` is gone.
- New test `delete_dir_deletes_nested_descendants`, on SQLite, which keeps the descendant range honest. It passed before this change too, so it is a guard, not a proof.
- Red proof: with the fix backed out and the new tests in place, the three SQLite cases of the sibling test fail, because `FOO/x/c` is gone. The other cases and the descendant guard pass.
- The existing `delete_dir` tests pass, `delete_dir_escapes_like_wildcards` included: `dirXname` sorts after `dir%name0`, so the range leaves it alone.
- 23 test invocations in total, each one with tests that ran and passed. The MinIO cases ran in CI only.

### Not in this change

When the in-memory blob storage deletes a directory, it removes the directory entry and the files that the entry lists, and it does not go below that. A blob at `foo/x/b` therefore stays after `delete_dir("foo")`. That is a second defect, in a second backend, and it keeps the new descendant test on SQLite.


vigoo (approved):
Agent says:

> **Nonblocking caveat:** the PR’s index-performance claim needs qualification. On local SQLite 3.40.1, `EXPLAIN QUERY PLAN` for the [full DELETE](file:///home/user/workspace/repo/golem-service-base/src/storage/blob/sqlite.rs#L330-L339) uses the index only for `namespace=?`, not the parent bounds. This is not a regression, but verify against bundled SQLite before claiming a range-search speedup.

kmatasfp:
The caveat is right, and I measured it on the bundled SQLite rather than a local one. `EXPLAIN QUERY PLAN` on SQLite 3.46.0, the version `libsqlite3-sys 0.30.1` bundles:

```
new DELETE   SEARCH blob_storage USING INDEX sqlite_autoindex_blob_storage_1 (namespace=?)
old DELETE   SEARCH blob_storage USING INDEX sqlite_autoindex_blob_storage_1 (namespace=?)
range alone  SEARCH blob_storage USING INDEX sqlite_autoindex_blob_storage_1 (namespace=? AND parent>? AND parent<?)
LIKE alone   SEARCH blob_storage USING INDEX sqlite_autoindex_blob_storage_1 (namespace=?)
```

So the `OR` keeps SQLite from a search on `parent`, the plan is the same before and after, and there is no speedup to claim. The range does become a search on `parent` when it stands alone, which `LIKE` never does, so the shape is the one that can use the primary key later.

The description no longer claims the speedup and now carries these four plans. The change stands on the case sensitivity alone. The misleading comment in `sqlite.rs` is corrected in #3896, which touches the same statement, to keep this pull request from another CI round.

PR #3896, opened 2026-09-16
Make delete_dir behave the same on every blob storage backend

kmatasfp:
- `delete_dir` answered differently on each blob storage backend. This change gives the four backends one contract, and writes that contract into the trait doc, which said only "Returns true if the directory was deleted; false if it did not exist".
- A root path now changes nothing and gives false on every backend. Before, only SQLite refused it. The filesystem backend called `remove_dir_all` on the root of the namespace, the S3 backend listed and deleted every key of the namespace, and the in-memory backend removed the root-level entries.
- `.`, `./` and `././` are at the root too, and no backend refused them, SQLite included. This came out of the review of an earlier version of this pull request, which guarded the empty path alone. The filesystem backend gave these paths to `remove_dir_all`, which removed every blob of the namespace and then failed with `Invalid argument (os error 22)`. A guest reaches that through `delete-container(".")`, because `delete_container` gives the container name to `delete_dir` as it arrives. The loss stays inside the custom storage of the environment of that guest. SQLite failed with `Path must have a file name`, and S3 with an invalid key.
- A directory that only holds blobs is now deleted on SQLite too, and the result is true. Before, SQLite looked for a directory row that `put_raw` never writes, found nothing, and returned false without deleting the blobs below the path. The filesystem, in-memory and S3 backends already deleted them.

### The changes

- `blob_path_is_root` tells whether a path has a name in it. Each backend asks it before any other work, so a root path costs no S3 request and no directory read.
- The guard cannot move into `validate_relative_blob_path`. That function accepts a root path, because such a path is neither absolute nor a parent traversal, and the other operations need it to keep accepting one.
- SQLite loses its `exists` pre-query, which was the only part that consulted a directory row. The `DELETE` statement already covered a directory that only holds blobs, through its `parent = ?` clause, and it keeps the case-sensitive key range from the pull request below this one. `rows_affected() > 0` now gives the result.
- A file whose path equals the directory path is still left alone and still gives false, because the directory-row clause requires `is_directory = TRUE` and nothing sits below that path.
- The comment about the plan of that `DELETE` said the range can use the primary key. `EXPLAIN QUERY PLAN` on the bundled SQLite 3.46.0 gives `SEARCH blob_storage USING INDEX sqlite_autoindex_blob_storage_1 (namespace=?)`, because the `OR` keeps SQLite from a search on `parent`. The comment says that now.

### Verification

- `cargo fmt -p golem-service-base -- --check`: exit 0
- `cargo clippy -p golem-service-base --all-targets --no-deps -- -D warnings`: exit 0
- The whole blob storage suite: 260 passed, 0 failed. Every `delete_dir` test on all five storages: 74 passed. The MinIO cases are included.
- `delete_dir_root_path_is_safe_noop` is a matrix test again, instead of running on SQLite alone. It covers ``, `.`, `./` and `././`, and it keeps a nested blob as well as a root-level one, so a backend that descends is caught.
- New matrix test `delete_dir_deletes_a_directory_that_only_holds_blobs`.
- Red proof of the root path: with all four guards removed, all 15 cases fail.
- Red proof of the aliases: with the guards back at the empty-path test, 12 of the 15 cases fail and only the 3 in-memory cases pass. The filesystem cases fail with `Invalid argument (os error 22)`, the SQLite cases with `Path must have a file name: "."`, and the S3 cases with `XMinioInvalidResourceName`.
- Proof of the loss: a throwaway test that keeps the error instead of unwrapping it reported `delete_dir(".") -> Err(Invalid argument (os error 22))` on the filesystem backend, and both blobs gone afterwards. On the other three backends both blobs were still there.
- Red proof of the blobs-only directory: with the SQLite change backed out, exactly the 3 SQLite cases fail and the other 12 pass.

### Note

This is the third of three pull requests about `delete_dir`. #3895 has merged. This one is stacked on #3893, and its base becomes `main` as #3893 merges.


vigoo (approved):
One pre-existing thing you may want to fix too (but not necessarily in this one):

> **[P1] Root-path guard misses `"."` aliases** — `golem-service-base/src/storage/blob/fs.rs`, PR lines 350–352. The validator accepts `"."`, `"./"`, and `"././"`, but the new guard checks only an empty string. These aliases resolve to the namespace root and reach `remove_dir_all`, deleting root-level and nested blobs before returning an error on Linux. Recognize root-equivalent paths before backend work and extend the matrix test to cover these aliases.

> This unsafe case already existed; the finding is that it remains despite the PR’s new root-safety contract.

kmatasfp:
Fixed here in d69248446, because this pull request is the one that claims a root path is safe on every backend, and that claim was false while `.` went through.

The finding holds, and it is worse than an error return. A throwaway test that keeps the error instead of unwrapping it gives, on the filesystem backend:

```
delete_dir(".") -> Err(Invalid argument (os error 22))
survived "keep-me"     -> Ok(false)
survived "keep/me/too" -> Ok(false)
```

`remove_dir_all` removes the contents first and only then fails on the root itself, so the blobs are gone. A guest reaches it through `delete-container(".")`, since `delete_container` gives the container name to `delete_dir` as it arrives. The loss stays inside the custom storage of that environment. SQLite and S3 only failed, with `Path must have a file name` and `XMinioInvalidResourceName`.

The fix is `blob_path_is_root`, which asks whether the path has a name in it, so `""`, `"."`, `"./"` and `"././"` are all at the root. Each backend asks it in place of the empty-path test. `delete_dir_root_path_is_safe_noop` now covers the four spellings and keeps a nested blob as well as a root-level one.

Red proofs: with all four guards removed, all 15 cases fail. With the guards back at the empty-path test, 12 of 15 fail and only the in-memory cases pass. Green: 260 in the blob storage suite, MinIO included.

This commit also corrects the query-plan comment on the SQLite `DELETE`, which your note on #3893 called out. Measured plans are in that thread.

PR #3897, opened 2026-09-16
Native external-tool invocation on existing and ephemeral agents

vigoo:
## Summary
- Make external-tool invocation a native Golem feature through REST, resumable invocation sessions, and `golem tool invoke`, rather than an MCP-only capability.
- Queue tool calls on existing agents using regular invocation ordering, authorization, idempotency, scheduling, durable execution, and replay semantics.
- Support fresh component-backed ephemeral owners that instantiate the deployed component, including core initializers, with ordinary module and memory accounting. No agent type is selected and no agent constructor is called. Accepted execution is fail-stop, including interruption during core initialization.
- Inherit component-level tool bindings and configuration into real agents, and use the component baseline to provision ephemeral tool owners and permissions.
- Export selected native commands through MCP with compiled schemas, authenticated deployment selection, revision fencing, bounded byte streams, and cancellation.
- Include protocol, SDK, generated API, documentation, and regression-test updates. The implementation is organized into 12 focused commits, followed by integration with current main.

## Integration with main
- Preserve upstream native built-in tools, middleware metadata, and lifecycle recovery semantics.
- Persist owner-keyed activation snapshots, including middleware chains, and preserve named custom errors across protocol boundaries.
- Use ordinary component instantiation and resource admission for external-tool owners, with regression coverage for core initialization, constructor avoidance, and fail-stop recovery.
- Renumber the new migrations after upstream migrations and introduce deployment-diff model version 10.

## Validation
- Latest pushed implementation: [`7917d6de9`](https://github.com/golemcloud/golem/commit/7917d6de94564a86741b488f077447993d950010). [CI run 35266393701, attempt 3](https://github.com/golemcloud/golem/actions/runs/35266393701) passed on that exact commit; all PR checks passed. Final Oracle review approved the exact pushed code and bug-finder reported no bugs.
- Current common/API/executor/service library suites: 3,891 passed, 10 ignored. Nine targeted native-tool, Rust-provider, and lifecycle integration tests passed; scoped formatting and strict Clippy passed.
- One earlier integration-test timeout has no proven root cause; 40 related repeated test executions and an unchanged CI retry passed. No assertions or timeouts were weakened.

Earlier implementation-stage verification:
- Focused common-model tests: 38 passed, including the new diff fingerprint checked without golden-file updates.
- Capability-redaction regression test: passed.
- Worker-service build and OpenAPI/docs generation: passed; registry spec generated against a fresh disposable database.
- TypeScript SDK: typecheck, 38 tool-client tests, 32 bridge tests, and all three runtime-template builds passed.
- Effect SDK: declaration drift, contracts, typecheck, and 9 oplog tests passed.
- MoonBit SDK CI passed.
- CLI schema/property and deployment-domain tests: 55 passed after integration fixes.
- Common strict Clippy and focused JSON/protobuf/BinaryCodec round trips passed; scoped Rust formatting passed.
- Generator regression test passed. Oracle-reviewed CI fixes and bug-finder completed with no remaining findings.
- Full [CI run 35160937081](https://github.com/golemcloud/golem/actions/runs/35160937081) passed on [`b8830a322`](https://github.com/golemcloud/golem/commit/b8830a3221699a1ec8d8ac7193e0d39fa3c2f52e): builds, Clippy, unit checks, all SDK and CLI suites, all worker groups, and all service integration groups. Docs checks and deployment also passed.
- Rebuilt the embedded OTLP exporter for the current owner-kind oplog ABI: WASM validation and 3 plugin tests passed; registry cargo check passed. Oracle and bug-finder review gates passed.

Resolves GOL-25


golem-common/src/base_model/durable_stream.rs
```diff
@@ -541,6 +556,8 @@ pub struct StreamSessionPreparedRecordV1 {
     pub format_version: u8,
     pub attempt: StartAttemptDescriptorV1,
     pub stream_mappings: Vec<StreamSessionMappingRecordV1>,
+    pub tool_stdin: Option<u64>,
```
vigoo: Could tool stdin/out be somehow implemented the same way as input/output stream parameters of regular invocations, so we don't need special cases here (and many other places) for them?

golem-common/wit/deps/golem-1.x/golem-oplog.wit
```diff
@@ -373,12 +390,45 @@ interface oplog {
...
+    record tool-invocation-result {
+        %result: option<typed-schema-value>,
+    }
+
+    record custom-tool-error {
+        name: string,
+        payload: typed-schema-value,
+    }
+
+    variant tool-error {
```
vigoo: Shouldn't these be imported from the common tool wit package?

golem-service-base/src/mcp/native_tool.rs
```diff
@@ -0,0 +1,1009 @@
...
+
+fn flag_default(shape: &FlagShape) -> SchemaValue {
+    match shape {
+        FlagShape::BoolFlag(shape) => SchemaValue::Bool(shape.default),
+        FlagShape::CountFlag(_) => SchemaValue::U32(0),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
vigoo: Let's split here to separate mod.rs/tests.rs files

golem-worker-executor/src/worker/mod.rs
```diff
@@ -8060,7 +8418,7 @@ struct RunningAgent<Runtime, Adapter: SandboxFilesystemAdapter = SandboxFilesyst
 }
 
 struct RunningAgentRuntime<Ctx: WorkerCtx> {
-    instance: Instance,
+    instance: Option<Instance>,
```
vigoo: This is a nice optimisation that we don't need a component instance for an ephemeral tool call - but I'm not sure it worths the complexity it brings in. Calling external ephemeral tools is expected to be a rarely used feature, while the primary hot path is running agents.

vigoo (commented):
An initial set of comments

cli/golem-cli/src/command_handler/tool.rs
```diff
@@ -290,6 +311,220 @@ impl ToolCommandHandler {
...
+                .as_ref()
+                .is_none_or(|key| key.value == "-")
+        {
+            bail!("--lookup requires an explicit --idempotency-key other than '-'");
+        }
+        let key = match args.idempotency_key {
+            Some(key) if key.value != "-" => key,
+            _ => IdempotencyKey::fresh(),
+        };
+        log_action(
```
noise64: **Keep raw stdout byte-clean**

When I run `golem tool invoke --stdout ... >output.bin` with the default text format, `log_action("Using", ...)` writes the idempotency key to stdout before the raw tool stream. The resulting file starts with CLI text, so it is not the tool output. Could we route/suppress progress logging for this mode (as `agent invoke` does for raw output), including any app-context logs, and add a test that asserts stdout contains only tool bytes?

cli/golem-cli/src/command.rs
```diff
@@ -1323,6 +1329,45 @@ pub mod tool {
...
+        #[arg(long)]
+        pub agent: Option<RawAgentId>,
+        /// Component used to create a fresh ephemeral invocation owner without constructing an agent
+        #[arg(long)]
+        pub component: Option<ComponentName>,
+        /// Deployed tool name
+        pub tool_name: ToolName,
+        /// Native tool command path
+        pub command_path: Vec<String>,
+        /// Typed input JSON (`schema` and `value`), or @PATH. Defaults to an empty record.
```
noise64: **Make the new CLI input shape discoverable**

`--input` requires a full schema-typed `{schema, value}` document, whereas MCP advertises ordinary JSON, and I could not find a runnable `golem tool invoke` example in CLI help or the guide. Could we accept the command's ordinary JSON input (deriving the typed value from its deployed schema), or at least add a complete valid `@request.json` example and explain that omitting input sends an empty record? That would make the new CLI route usable without first reverse-engineering the wire format.

docs/src/content/next/invoke/mcp.mdx
```diff
@@ -23,6 +23,46 @@ mcp:
...
+```yaml
+mcp:
+  deployments:
+    local:
+      - subdomain: mcp-demo
+        agents:
+          CounterAgent: {}
+        tools:
+          file-tools:
+            ownerComponent: my-app:tool-owner
```
noise64, line 56: **Show the component binding in the MCP setup example**

Could this guide show a complete **tools-only MCP** example: the top-level tool declaration, `components.<owner>.tools.<name>` binding, and `mcp.deployments.*.tools.<name>.ownerComponent`, with no agent required? It would clarify three distinct roles: the top-level declaration provisions the implementation; the component binding authorizes a fresh owner **and** becomes the default for real agents in that component; the MCP entry exports selected commands. An agent-only binding does not satisfy the export, and the implementation and owner components need not be the same. For a remote tool, the example could also mention the grant prerequisite.

noise64 (commented):
### Document component bindings in the manifest reference

The [manifest reference](https://github.com/golemcloud/golem/blob/7917d6de94564a86741b488f077447993d950010/docs/src/content/next/app-manifest.mdx#L116) currently moves from top-level tool declarations straight to agent bindings. Could we add `components.<name>.tools` and `toolsMergeMode` to the field reference, and explain that a component binding both defaults into its agents and authorizes component-targeted fresh owners? A short note that an agentless owner component is the clean pattern for tools that should **not** become agent defaults would prevent a permission/DX surprise.

### Teach the new binding and MCP export pattern in the manifest skill

The [manifest-editing skill](https://github.com/golemcloud/golem/blob/7917d6de94564a86741b488f077447993d950010/golem-skills/skills/common/golem-edit-manifest/SKILL.md#L96) lists component fields without `tools`/`toolsMergeMode`, and its [MCP example](https://github.com/golemcloud/golem/blob/7917d6de94564a86741b488f077447993d950010/golem-skills/skills/common/golem-edit-manifest/SKILL.md#L404) only exports agents. Could we update it with the component-binding inheritance/opt-out rule and a tools-only MCP deployment using `ownerComponent`? Otherwise generated manifests can follow the skill and still miss the binding needed for a native tool export.

### Reconcile the fresh-owner lifecycle description

The PR description says a fresh owner runs "without instantiating the owner component guest", but the [MCP guide](https://github.com/golemcloud/golem/blob/7917d6de94564a86741b488f077447993d950010/docs/src/content/next/invoke/mcp.mdx#L48) says it instantiates the component (including core initializers), and the [executor test](https://github.com/golemcloud/golem/blob/7917d6de94564a86741b488f077447993d950010/golem-worker-executor/tests/tool_streaming.rs#L6620) asserts the initializer runs. Could we correct the PR description to say that no **agent type/constructor** is selected, while component initialization still happens? That distinction matters for startup cost and side effects.

noise64 (approved):

vigoo:
@noise64 — corrected the fresh-owner lifecycle description from your review: https://github.com/golemcloud/golem/pull/3897#pullrequestreview-5247813852.

A component-targeted call instantiates the actual deployed component and runs its core initializers with ordinary module/memory accounting. It does not select an agent type or call an agent constructor. The owner remains `EphemeralExternalTool`, and accepted execution is fail-stop, including interruption during core initialization.

The previous “host-only / without instantiating the owner component guest” wording described the optimization removed in the earlier review follow-up. I removed those stale claims from the PR description and updated the validation summary. The MCP guide and initializer/constructor regression test describe the current behavior correctly. This correction changes the description, not the runtime.

PR #3898, opened 2026-09-16
CLI cleanups: builtin profiles, usage errors, best-effort agent bulk ops, local server port warning, config file locking

noise64:
- resolves GOL-413
- resolves GOL-411
- resolves GOL-395
- resolves GOL-334
- resolves GOL-246
- makes the builtin `local` / `cloud` profiles authoritative for their server: stored `customUrl` / `customWorkerUrl` / `allowInsecure` are ignored consistently, `local` no longer pins `customUrl`, and `profile get` / `profile list` show the effective URLs; no new server-target flags, profiles + manifest `server:` + `GOLEM_*` env vars already cover non-manifest usage
- reports non-interactive usage errors of `new` and `profile new` like clap does: error + command help on stderr, exit code 2; `profile new` refuses existing profile names, `profile delete` distinguishes missing and active profiles
- makes `redeploy-agents` and `deploy --reset` agent deletion best-effort like `update-agents`: per-agent failures are collected into a new `errors` field of `agent.redeploy` / `agent.delete-all` (redeploy errors tell delete and recreate failures apart), `redeployed` / `deleted` are only true without errors, exit code is non-zero on any error
- strips terminal styling from error messages in structured agent outputs (`agent.update` / `agent.redeploy` / `agent.delete-all`), which leaked ANSI codes into `-F json` on a tty
- `golem server run` warns when the bound custom-request / MCP port differs from the manifest's subdomain expansion (flag override or port `0`), also for manifests without a `localServer` section; docs and skills now state that `routerAddr` does not affect subdomains, the listeners bind all interfaces, and `.localhost` resolves only on the server machine
- writes the CLI config file atomically (temp file + rename) and serializes read-modify-writes across processes with an advisory lock on `config-v4.json.lock`; `save_auth`, `profile new`, `profile delete` and `profile config set-format` are single locked updates, no-op updates do not touch the file
- re-enables the unstructured / multimodal invoke integration tests with the current argument syntax, pinned by parser unit tests
- regenerates the skill-derived how-to guides

Notes:
- config file permissions are tightened to owner-only (`0600`) on the next write, as the file holds auth tokens; the `.lock` file stays in the config directory
- `.localhost` subdomains: the GOL-395 premise turned out wrong (the HTTP API / MCP listeners always bind `0.0.0.0`), so instead of changing the host the real trap, port mismatch, got the warning


vigoo (approved):
The agent found two minor issues that I personally don't find very important - I'm fine with not solving them but here they are for reference:

> 1. **P2 — Bulk error aggregation loses failures across components.**
   [`component/mod.rs:449`](https://github.com/golemcloud/golem/pull/3898/files#diff-a8dffc6ce0b97a1c6c24f9c7f5d9a436d5c73a508e3f5e02a15aa312fba2bc77R449) merges errors using only the raw agent ID. Although agent types are unique within a deployment, moving a type from component A to B leaves A’s older agents intact. Both components can therefore contain `counter("x")`; if both redeployments fail, the second error overwrites the first, potentially hiding a failed recreation. Delete aggregation has the same identity problem. Key errors by component plus agent ID.

>  2. **P3 — Local-port warnings also fire for cloud environments.**
   [`command_handler.rs:65–66`](https://github.com/golemcloud/golem/pull/3898/files#diff-b7054c4c6986f42ca03b192cafd4a203cc0be0f14dc4654c0365f38dc672be96R65) checks only whether a manifest environment exists. With `server: cloud`, running `golem server run --custom-request-port 0` incorrectly warns that deployment subdomains require the manifest’s local port. Cloud subdomains do not use that port. Restrict the warning to local environments.

PR #3899, opened 2026-09-16
Preserve filesystem ownership across failed deletion

kmatasfp:
## Summary

- make `delete_and_verify` consume the filesystem adapter and return ownership on failure
- restore failed adapters to the same filesystem generation for safe retries
- remove unnecessary provisioning `Arc`s and centralize creation rollback

Follow-up to #3797.

## Validation

- `cargo test -p golem-worker-executor --lib` (1993 passed, 8 ignored)
- `cargo clippy -p golem-worker-executor --lib --tests -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`


vigoo:
Thanks, this is better indeed

vigoo (approved):

PR #3902, opened 2026-09-17
fix(executor): replay the component's bootstrap prefix under automatic snapshot recovery

noise64:
Fixes GOL-611

## Problem

Automatic-snapshot recovery skipped the oplog from right after `Create` up to the snapshot index. That range does not only hold the agent's history — which the snapshot substitutes for — but also the host calls the component runtime makes while it is instantiated, before the first agent invocation: clocks, random seeds, environment reads. Because those entries were skipped, the recovering instance had nothing to replay them against, and since #3856 it re-ran the initializer with durability suppressed, i.e. against live host responses.

That is only sound if the initializer's results are throwaway. They are not for every guest: a language runtime seeds its PRNG from `random_get` during initialization, and later scheduling decisions — some of which reach the host as durable calls — follow from that seed. A fresh seed on recovery makes the tail replay diverge from the recording. (This is the executor half of the Go snapshot flake in GOL-485.)

## Change

- The skip region now starts at the first `AgentInvocationStarted` (`first_agent_invocation_index`), so the bootstrap prefix of the same component revision replays positionally, exactly as on a plain restart. Automatic snapshots are revision-scoped, so the recorded prefix belongs to the component being instantiated.
- The initializer is no longer run under `begin_call_snapshotting_function()` for automatic snapshots. Manual-update snapshots keep the suppression: their prefix was recorded by a different revision and has no counterpart.

## Witness

- `test-components/agent-counters/src/snapshot_test.rs`: `_initialize` hashes a `RandomState` (keyed from `insecure-seed`) into a static and exposes it as `initializer_seed_hash`.
- `golem-worker-executor/tests/durability.rs::automatic_snapshot_every_2nd_invocation`: reads it before the first snapshot (in the slot that used to read `get == 0`, keeping the snapshot cadence) and asserts the same value after snapshot recovery.
- Fails 3/3 on main, passes 5/5 with this change. CI rebuilds `agent-counters`.

Draft until CI confirms the other snapshot / hot-update tests are unaffected.


vigoo:
The part where we reuse the persisted component initialization entries from the beginning of the oplog when applying snapshot based recovery makes sense. 

You are _not_ doing this for the case when snapshot is used to perform a manual update, which also makes sense partially - we don't want the new revision to replay the old revision's initialization code (although this probably would only break if the component's compiler / base runtime is updated or changed). 

Anyway, the problem is the combination of the two. After a manual update the new component again may do some side-effects during its component initialization, and that is now neither matched against the recorded one, nor recorded as a fresh one. So this is already a divergence point. And the above fix is even making it worse by re-enabling (or maybe not, just ignoring as well) the first revision's initialization entries when applying a snapshot-based recovery _after_ a snapshot-based update.

I don't have a simple good design in my mind - it has to be something like persisting new initialization blocks after updates, and somehow pointing the oplog cursor to those while initializing a follow-up recovery.

So I recommend to not merge this and instead create a post-1.6 follow-up ticket to support this properly (if this only affects the go runtime, as I think so).

noise64:
I think, with the our forked go, we do not need this even for, but I will recheck that, and depending on it, will create an issue.

PR #3903, opened 2026-09-17
GOL-172 Add local semantic retries to MoonBit SDK

vigoo:
## Summary

- add an idiomatic async `@api.retry_local` API for interpreting Golem semantic retry policies around arbitrary fallible user code
- support every policy and predicate node with canonical composition state, bounds, duration rounding, jitter, glob matching, malformed AST handling, and final-error preservation
- add generated WASI monotonic-clock and insecure-random bindings plus MoonBit documentation and focused tests

Local attempts remain ordinary user-space executions: they do not produce executor `RetryAttempt` oplog entries or form one host-managed durable retry sequence.

## Verification

- `moon info api`
- `moon fmt --check api/retry_local.mbt api/retry_local_wbtest.mbt api/moon.pkg`
- `moon check --target wasm api`
- `./scripts/run-sdk-tests.sh api/retry_local_wbtest.mbt api/retry_builder_test.mbt` — 17/17 passed
- `./scripts/run-sdk-tests.sh` — 319/319 passed
- `git diff --check`


sdks/moonbit/golem_sdk/api/retry_local.mbt
```diff
@@ -0,0 +1,1111 @@
...
+fn parse_policy(raw : RetryPolicy) -> LocalPolicy raise LocalRetryError {
+  if raw.nodes.is_empty() {
+    raise invalid_policy("Retry policy has no root node")
+  }
+  let visiting = Array::make(raw.nodes.length(), false)
+  parse_policy_node(raw.nodes, 0, visiting)
+}
+
+///|
+fn parse_policy_node(
```
noise64: > Could we give raw policy compilation explicit node, expanded-node/payload, and recursion-depth limits? This parser detects cycles, but it recursively rebuilds every reference. A deep acyclic chain can exhaust the stack, and a small shared DAG can be expanded repeatedly into much more work than its raw node count.
>
> The Rust sibling in #3904 already caps raw/compiled nodes, payload bytes, and depth. Matching that bounded contract here, with deep-chain and shared-DAG regression tests, would make `retry_local` robust for hand-built or host-sourced raw policies rather than relying on the builder to keep the graph small.

noise64 (approved):

PR #3904, opened 2026-09-17
GOL-172: Add user-space retry policies to Rust SDK

vigoo:
## Summary

- add an idiomatic async `RetrySchedule` API for applying Golem semantic retry policies to arbitrary Rust user code
- interpret every retry policy and predicate node locally, including composition state, bounds, jitter, filtering, and malformed raw-policy validation
- preserve the user's final error and document the distinction from host-managed retry sequences
- document resolving and locally executing named policies

## Verification

- `cargo fmt -p golem-rust --manifest-path sdks/rust/Cargo.toml -- --check`
- `cargo clippy -p golem-rust --manifest-path sdks/rust/Cargo.toml --all-targets --no-deps -- -D warnings`
- `cargo test -p golem-rust --manifest-path sdks/rust/Cargo.toml`
- `cargo build -p golem-rust --manifest-path sdks/rust/Cargo.toml --target wasm32-wasip2`

Oracle final review found no remaining fix-worthy findings. Two bounded bug-finder runs found no bugs.

GOL-172


sdks/rust/golem-rust/src/retry/local/mod.rs
```diff
@@ -0,0 +1,785 @@
...
+        if self.policy_visiting[position] {
+            return Err(RetryPolicyError::CyclicPolicyNode(index));
+        }
+        self.policy_visiting[position] = true;
+        let next_depth = depth + 1;
+        let result = match &nodes[position] {
+            retry_api::PolicyNode::Periodic(delay) => {
+                CompiledPolicy::Periodic(Duration::from_nanos(*delay))
+            }
+            retry_api::PolicyNode::Exponential(config) => {
```
noise64, line 507: > Should `RetrySchedule::try_from` follow the canonical raw-policy evaluator for exponential factors, or are the other implementations meant to become strict too? The host evaluator's `scale_duration` maps non-finite or non-positive factors to zero duration, and the MoonBit/TypeScript PRs intentionally accept a raw factor of `0`; this implementation rejects it (Scala does too).
>
> The public builders can still reject invalid inputs. The concern is the hand-built/host-sourced raw `RetryPolicy`: today the same graph has different behavior depending on SDK. Could we settle that shared contract and add one cross-SDK fixture for zero, negative, NaN, and infinity?

vigoo: We settled on a strict shared raw-policy contract rather than preserving the evaluator’s malformed-factor arithmetic: exponential must be finite and > 0; jitter must be finite and >= 0. Rust already enforced this for raw policies. Commit 35be28d adds the explicit builder/raw edge matrix (0, negative, NaN, and both infinities, with zero jitter accepted). The TypeScript, MoonBit, Scala, and host changes are being coordinated in their respective PRs/workstream; host ingress validation will also eliminate the current +Infinity jitter panic path.

noise64 (approved):

PR #3905, opened 2026-09-17
GOL-172: Add user-space retry execution to TypeScript SDK

vigoo:
## Summary

- add `retry(policy, operation, options)` for applying Golem retry policies to synchronous or Promise-returning user code
- interpret all retry policy and predicate nodes locally, with independent composition state, elapsed/count bounds, jitter, property projection, and cancellation
- validate malformed and cyclic raw policy ASTs before invoking user code and preserve the final thrown value by identity
- document that local retries are not executor-managed retry attempts and do not form a durable retry sequence
- follow the Effect SDK's minimal local `prop-matches` approximation (`*` / `?`) and direct users to host-driven retries for authoritative platform glob semantics

## Verification

- `npx --yes pnpm@10.17.1 --filter @golemcloud/golem-ts-sdk run test`
  - 41 test files passed, 1 skipped
  - 851 tests passed, 20 skipped
- `npx --yes pnpm@10.17.1 --filter @golemcloud/golem-ts-sdk run lint`
  - 0 errors; 6 pre-existing warnings
- Prettier check on changed files
- `git diff --check`
- `npx --yes pnpm@10.17.1 run build-agent-template`
  - all three runtime templates built successfully

## Review

- Oracle final focused review: no implementation blockers; documentation precision finding accepted and fixed
- Bug-finder completed two clean terminal runs with no findings


sdks/ts/packages/golem-ts-sdk/src/host/retryExecutor.ts
```diff
@@ -0,0 +1,832 @@
...
+        throw error;
+      }
+
+      retryCount += 1;
+      await sleep(step.verdict.delay, options.signal);
+    }
+  }
+}
+
+function compilePolicy(policy: RetryPolicy): CompiledPolicy {
```
noise64, line 162: > Can we add explicit raw-node, expanded-node/payload, and recursion-depth limits before compiling this graph? Active-path cycle detection prevents a loop, but a deep acyclic chain can still overflow the JS stack, and a shared DAG is recursively recompiled on every reference and can expand exponentially.
>
> The Rust sibling already memoizes shared nodes and caps raw nodes, compiled nodes, payload bytes, and depth. Matching that contract here, with deep-chain and shared-DAG tests, would keep host-sourced or hand-built raw policies bounded.

sdks/ts/packages/golem-ts-sdk/tests/retry-executor.test.ts
```diff
@@ -0,0 +1,450 @@
...
+          throw attempt;
+        }),
+      ).resolves.toBe('ok');
+    });
+
+    expect(exponentialTimes).toEqual([0, 10, 30, 70]);
+    expect(fibonacciTimes).toEqual([0, 5, 15, 30]);
+  });
+
+  it('matches Rust floating-duration rounding and accepts a raw zero exponential factor', async () => {
```
noise64: > This test name says accepting raw factor `0` “matches Rust”, but #3904 rejects every non-positive raw exponential factor. MoonBit and this implementation match the canonical host evaluator's zero-delay behavior; Rust and Scala are strict.
>
> Could we settle the raw-policy contract across the four SDKs and update this assertion/name accordingly? A shared fixture covering zero, negative, NaN, and infinity would prevent the implementations from drifting while still allowing each public builder to be stricter.

sdks/ts/packages/golem-ts-sdk/src/host/retryExecutor.ts
vigoo: Addressed in 7592b9c. Raw compilation now caps total raw nodes (4096), expanded nodes (4096), UTF-8 predicate payload (1 MiB), and combined policy/predicate depth (256). Policy and predicate nodes are memoized with active-path cycle detection; expanded complexity is still charged per occurrence so runtime state remains independent. Predicate values are snapshotted, and a compiler-wide payload counter rejects before excessive allocation. Added exact boundary, deep-chain, shared-DAG, reused-subtree, payload, membership, mutation, and early-allocation tests.

sdks/ts/packages/golem-ts-sdk/tests/retry-executor.test.ts
vigoo: Resolved by the strict cross-SDK decision in 7592b9c. Raw/WIT exponential compilation now rejects 0, negative, NaN, and ±Infinity (finite and >0 required), while raw jitter accepts 0 and rejects negative/NaN/±Infinity (finite and >=0 required). Builders remain strict. The misleading test name is now `matches canonical floating-duration rounding`, and explicit raw edge-matrix tests verify rejection before user code runs plus zero jitter as a no-op.

noise64 (approved):

PR #3906, opened 2026-09-17
GOL-172: Add user-space retry execution to Scala SDK

vigoo:
## Summary

- add Future-native `Retry.retry` overloads for semantic and raw retry policies
- interpret every retry policy node locally with independent composition state, bounds, jitter, and failure property projection
- preserve the final failed value and reject malformed/cyclic raw trees before invoking user code
- document the distinction between local attempts and host-managed durable retry behavior
- align local `prop-matches` with the Effect/TypeScript no-flags JavaScript regex approximation

## Validation

- `sbt "++3.8.2; core/scalafmtCheckAll; core/compile; core/test"` — 651 tests passed
- `sbt "++3.8.2; core/testOnly golem.host.LocalRetrySpec"` — 14 tests passed
- Oracle review approved with no blockers
- bug-finder completed clean with no findings

`testAgents/fastLinkJS` was also attempted, but this checkout lacks the three embedded guest runtime WASMs required by that task. Core compile and compile-level public API coverage pass.


sdks/scala/core/js/src/main/scala/golem/host/LocalRetry.scala
```diff
@@ -0,0 +1,377 @@
...
+  private case object GiveUp                                 extends Verdict
+  private case object EvaluationError                        extends Verdict
+
+  private[host] trait Runtime {
+    def nowNanos(): Long
+    def sleep(delay: FiniteDuration): Future[Unit]
+    def random(): Double
+  }
+
+  private object DefaultRuntime extends Runtime {
```
noise64: > Can we use a monotonic time source for `nowNanos`? `js.Date.now()` is wall clock, so an NTP/manual clock adjustment can make `elapsed` negative or jump forward. A `TimeBox` may then retry beyond its budget or give up prematurely.
>
> `performance.now()` is the natural Scala.js/browser/QuickJS-style elapsed clock if it is available in the runtime; otherwise the fallback and its limitations should be explicit. Please also add a production-clock-focused test or abstraction test where wall time moves backward/forward while monotonic time advances normally.

noise64 (approved):

PR #3908, opened 2026-09-17
Roll out stable indexed scans

vigoo:
## Summary

- replace the public scan cursor structure with an opaque, versioned token carrying oplog layer, agent mode, and backend resume state
- migrate primary and archived oplog enumeration to deletion-safe stable scans and remove the OFFSET-based indexed-storage scan API
- bound PostgreSQL and SQLite prefix scans with literal UTF-8 key intervals, using C/BINARY ordering and migration-time encoding checks
- validate cursor, resume, layer, mode, and count inputs before storage retry or panic paths, including running-status listing dispatch
- keep invocation-session protocol defects as internal server errors while mapping caller validation failures to HTTP 400
- update durable get-agents payloads, protobuf/OpenAPI contracts, CLI behavior, generated docs, embedded skills, and integration coverage

## Verification

- `cargo clippy -p golem-common -p golem-worker-executor -p golem-worker-service --all-targets --no-deps -- -D warnings`
- all-target checks for `golem-common`, `golem-worker-executor`, and `golem-cli`
- 32 indexed-storage scan integration tests across in-memory, Redis, SQLite, multi-SQLite, and PostgreSQL
- literal-prefix, Unicode-boundary, encoding, collation, query-plan, deletion-safe resume, and property tests
- opaque-cursor restart/replay integration test
- targeted cursor, oplog scan, resume validation, invocation protocol-error, and worker-service fast-path tests
- `cargo make diff-openapi`

Resolves GOL-522


kmatasfp:
> Bad cursors and count=0 return HTTP 500 instead of 400.
> - ScanCursor::from_str (golem-common/src/base_model/mod.rs:812) now accepts any string. The check in golem-worker-service/src/api/worker.rs:417 can no longer fail, so the INVALID_SCAN_CURSOR error code is never returned.
> - A malformed token now reaches the executor, which rejects it as InvalidRequest. Worker-service turns that into an API error in golem-worker-service/src/api/common.rs:284. That function has no case for InvalidRequest, so it falls through to _ => internal(INTERNAL_UNKNOWN). A GET with cursor=garbage used to get a 400 and now gets a 500.
> - The new count check in client.rs:1097 has the same problem. Before this PR, count=0 returned an empty page with a non-null cursor, which was its own bug, so rejecting it is right. Only the status code is wrong.
> - Fix: map InvalidRequest to bad_request there, or validate the cursor and count in worker-service. The first option also changes other endpoints, which I think is correct, but it's a wider change.

kmatasfp:
> The replay test doesn't check that listing resumes from the replayed cursor. get_workers_opaque_cursor_replays_after_restart calls get_agents_next_result (test-components/host-api-tests/src/golem_host_api.rs:645). That function creates the listing handle, fetches one page and drops the handle.
> - After the restart, replay only decodes the recorded page. Nothing fetches the next page, so the test would still pass if replay restored the wrong cursor.
> - It does prove the new persisted format decodes, which is worth keeping.
> - To test resume, the guest needs to fetch a page, wait on a promise that the test completes after restarting the executor, then fetch again. With 51 agents, it should get pages of 50 and 1.

gol-522-stable-indexed-scan-rollout-plan.md
```diff
@@ -0,0 +1,427 @@
+# GOL-522: roll stable indexed scans out to agent enumeration
```
kmatasfp: probably did not mean to commit this.

Also:
```
The committed plan lists tests the PR doesn't contain. Missing:
- an out-of-range layer in a multi-layer cursor
- a primary-layer scan given a non-zero layer
- a blob archive scan given a resume value
- count of 0 and above i64::MAX
- an invalid token over gRPC coming back as Failure(InvalidRequest)
- serde, binary and protobuf round trips for ScanCursor
- a test that pins the gsc1_ JSON shape, which the plan says is part of the token format
```

golem-worker-executor/src/services/oplog/multilayer.rs
```diff
@@ -773,76 +774,48 @@ impl OplogService for MultiLayerOplogService {
...
+                    Ok((ScanCursor::default(), ids))
                 } else {
-                    // Still scanning the current layer
                     Ok((new_cursor, ids))
                 }
             }
-            layer => Err(WorkerExecutorError::unknown(format!(
-                "Invalid oplog layer in scan cursor: {layer}"
-            ))),
+            _ => unreachable!(),
```
kmatasfp, line 818: Looking up `self.lower.get(layer - 1)` and returning an error instead removes the panic path

kmatasfp:
> Listing one component still reads every row in the table section for that agent mode. This was already true before the PR and isn't a regression. But performance is the PR's stated motivation, so the gain is smaller than the description suggests.
> - I ran the new Postgres query in a Postgres 17 container (default en_US.utf8 collation) with 200 components, 20 agents each and 100 oplog entries per agent.
> - With that collation, key LIKE $3 can't narrow the index lookup, so Postgres reads rows and discards the ones that don't match. Only key > $2 narrows it.
> - Listing a 20-agent component in the middle of the key order read 400,000 index rows and discarded 398,000.
> - Starting the first lookup at the component prefix cuts that to 200,000. The last page still reads to the end of the table section, because nothing gives the index an upper bound.
> - SQLite's plan has the same shape: it seeks on key>? only.
> - The OFFSET re-reads are gone, so a full walk is linear rather than quadratic. But that is linear in all agents' oplog rows across every component, not just this component's.
> - A real fix needs an upper bound the index can use, such as a C-collated key column.

kmatasfp:
In `golem-worker-service/src/api/common.rs:287`, worker-service now gives HTTP 400 for `InvalidRequest`. This is correct for the scan cursor and for the count. But two paths send a server fault through the same error.

- `decode_invocation_failure` (`golem-worker-service/src/service/worker/client.rs:168`) makes an `invalid_request` error from `InvocationFailureKind::Protocol`.
- `golem-worker-service/src/service/worker/client.rs:1871` makes an `invalid_request` error from `OneShotInvocationSessionResult::ProtocolFailure`.

These two errors show a protocol defect in the invocation session of the worker executor. For example, the session finishes before it sends a result. The test at `client.rs:2500` shows this condition.

Some errors from the gRPC layer of the worker executor have the same problem. For example, "auth_ctx not found" at `golem-worker-executor/src/grpc/mod.rs:302` happens when worker-service sends an incomplete request. This is a defect in golem. The request of the user is correct.

Before this pull request, these faults gave HTTP 500. Now they give HTTP 400 with the code VALIDATION_ERROR. This has two results:

- The caller reads that its request is invalid, but the request is correct. The caller does not send the request again.
- The fault is not in the 5xx rate. The alerts do not show the fault.

Keep the new behavior in `common.rs:287`. But do not use `invalid_request` for the two protocol paths. Give an internal error for these paths. This is a small change in `client.rs:168` and `client.rs:1871`.

kmatasfp:
`golem-worker-service/src/api/worker.rs:417` still gives HTTP 400 with the code INVALID_SCAN_CURSOR when `ScanCursor::from_str` fails. But `from_str` cannot fail. The implementation at `golem-common/src/base_model/mod.rs:812` has the error type `String`, but it always returns `Ok`. This branch is dead code.

The constant `INVALID_SCAN_CURSOR` at `golem-common/src/base_model/api.rs:209` has only this one use. No other code uses it.

The worker executor now gives the correct status for a bad scan cursor. You can remove the branch and the constant. If you keep the `FromStr` implementation, change the error type to `Infallible`.

kmatasfp (approved):

PR #3912, opened 2026-09-18
Compact raw oplog entries

vigoo:
## Summary

- reduce the in-memory size of raw oplog entries from 400 bytes to 160 bytes by boxing large cold-path payloads while keeping the frequent `Start` entry allocation-free
- remove unreleased legacy optionality and obsolete Desert evolution steps across raw/public models, protobuf, WIT, replay, rendering, CLI output, and tests
- preserve required permission-card metadata and wallet state, including lossless raw snapshot conversion
- regenerate synchronized OpenAPI, CLI schema, WIT, TypeScript, Effect, Scala, and MoonBit artifacts

## Verification

- 78 `golem-common` oplog tests
- raw snapshot WIT wallet-state roundtrip regression test
- worker-executor integration test compilation
- 638 Scala core tests
- TypeScript SDK typecheck and template builds
- Effect declaration, contract, and artifact checks
- MoonBit isolated `moon info` and `moon check`
- Rust formatting and diff checks

Resolves GOL-588


golem-common/src/model/oplog/tests.rs
```diff
@@ -66,6 +66,12 @@ use std::collections::{BTreeMap, BTreeSet, HashMap};
 use test_r::test;
 use uuid::Uuid;
 
+#[cfg(target_pointer_width = "64")]
+#[test]
+fn raw_oplog_entry_fits_in_176_bytes() {
+    assert!(std::mem::size_of::<OplogEntry>() <= 176);
```
noise64, line 72: > Why 176 and not 160 as in the description?

noise64 (approved):

golem-common/src/model/oplog/tests.rs
vigoo: There were some additional optimisations after this, i guess

PR #3854, opened 2026-09-10
Durable Streams: initial export side (DS-1–DS-6)

vigoo:
Implements the initial Durable Streams export side in separately reviewed commits.

Step completion (checked only after implementation, Oracle review, bug-finder, and CI):
- [x] GOL-540 / DS-1: stream-aware HTTP route registration and validation
- [x] GOL-101 / DS-2: executor protocol foundation
- [x] GOL-103 / DS-3: HTTP creation and reads
- [x] GOL-106 / DS-4: cancellation and tombstones
- [x] GOL-105 / DS-5: external writes and idempotent producers
- [x] GOL-541 / DS-6: reference-client compatibility harness

DS-6 progress before the current review revisions:
- [x] Unmodified @durable-streams/client@0.2.7 compatibility harness implemented
- [x] Targeted local compatibility tests, formatting, and CLI Clippy passed
- [x] Oracle approved and bug-finder clean for the compatibility harness
- [x] Separate DS-6 commit pushed; latest main merged before that push
- [x] Compatibility CLI CI shard and Docs passed on the DS-6 implementation commit
- [x] Memory-billing CI root cause fixed, regression-tested, Oracle-approved, and bug-finder clean
- [x] CI fix committed and pushed after latest-main sync
- [x] Docs passed on the cancellation-harness revision linked below
- [x] CI passed on the cancellation-harness revision linked below

Earlier status-fold CI fix: [1a096e099](https://github.com/golemcloud/golem/commit/1a096e099e94ff043465f4cea529b51f84028243). Threshold auto-commits could hide committed entries from the resident status fold, leaving metadata Running after invocation success. The fix catches up from committed storage in bounded chunks; it changes neither replay nor the DS protocol.

CI-fix verification: deterministic threshold-zero regression failed before the fix and passed afterward; existing idle-to-archive test passed; 99 status/state-actor unit tests passed; original memory-billing test passed locally with SQLite (Docker unavailable, original test restored); scoped formatting and Clippy passed, with existing warnings outside changed code. Oracle approved and bug-finder found no bugs. Walkthrough changes were rendered and inspected.

[...]

PR #3916, opened 2026-09-18
Bring filesystem-snapshot up to date with main

kmatasfp:
`filesystem-snapshot` last took `main` on 10 September and was 43 commits behind. This is that catch-up, nothing else. No feature work, and the aim throughout was that both sides' behaviour survives.

## Two decisions that want a human eye

**1. A protobuf tag collision, resolved by moving the epic's number.** Both sides claimed the same tags. `main` took 65 and 66 for `Resumed` and `RecoverySucceeded`; this branch had `SnapshotConfirmed` at 65 in `public_oplog.proto` and 64 in `raw_oplog.proto`.

`SnapshotConfirmed` moved to **67** and **66**. The reasoning is that `main` is merged and this branch is unreleased, so the unreleased side is the one that should move. Nothing in the tree pins the old numbers: no Rust code names them, and no golden or fixture file encodes them.

**The one thing that would make this wrong is a persisted oplog written from this branch**, since those bytes carry the old tag. If anyone has run this branch against storage they intend to keep, say so and it gets handled differently.

**2. `worker/status.rs` was deleted on main, and this branch had modified it.** That deletion was a refactor in #3854, not a removal: the logic moved to `worker/status/mod.rs` with tests in `worker/status/tests.rs`. So the deletion is accepted and this branch's 220 lines are ported into the new module, tests included. All six `snapshot_confirmed_*` tests pass there.

## Everything else

- `golem-common/src/model/oplog/mod.rs`: main's match arm plus `| OplogEntry::SnapshotConfirmed { .. }`. Main's earlier arm already covers every `Card*`, `Stream*`, `SetRetryPolicy` and `RemoveRetryPolicy` variant that this branch also listed, so only `SnapshotConfirmed` was genuinely new.
- `sandbox_filesystem/adapter.rs`: this branch's `copy_contents` and main's `delete_and_verify(self)` signature both kept, across the trait, the implementation and the scripted fake. 15 call sites follow main's by-value signature.
- Six import conflicts: unions.
- `SnapshotConfirmed` added to two match arms that did not conflict, `entity_parent_start_index` and `owner_start_index`, both functions main added in #3854 that this branch never saw. The compiler flagged them as non-exhaustive. It is a hint entry with no parent index, so it sits with `Snapshot` in the `=> None` arm.
- `filesystem_snapshot: None` at 7 sites main added, forced by this branch's new required field, matching every other site.

## Generated files were regenerated, not merged

The moonbit binding alone had 1327 conflict hunks behind a `DO NOT EDIT` header, so hand-merging was never an option.

- moonbit: `bash scripts/regen-bindings.sh`, using the repo-pinned `golemcloud/wit-bindgen` fork at `36866deb`, installed into a scratch root so the global toolchain was left alone.
- openapi and the REST API docs: `cargo make generate-openapi`.
- `sdks/effect/wit/deps`: `cargo make wit`.
- `plugins/otlp-exporter.wasm`: main's blob, since this branch changed no plugin source.

## Verification

- No conflict marker survives: `git grep -n '^<<<<<<<\|^>>>>>>>' | wc -l` gives 0.
- `cargo fmt --all -- --check`: exit 0
- `cargo clippy -p golem-common -p golem-worker-executor -p golem-service-base --all-targets --no-deps -- -D warnings`: exit 0
- `cargo test -p golem-common --lib`: 1274 passed, 0 failed, 1 ignored. Oplog subset: 161 passed.
- `cargo test -p golem-worker-executor --lib -- status`: 132 passed, the six ported `snapshot_confirmed_*` tests among them. `-- oplog`: 228 passed.
- `cargo build -p golem-worker-executor --all-targets`: exit 0

## Two things found along the way, neither caused by this merge

`scripts/regen-bindings.sh` does an `rm -rf` of the generated roots that preserves only `moon.pkg` and `stub.mbt`, so it deletes hand-maintained files that live beside the generated ones. It removed `gen/interface/golem/api/load-snapshot/snapshot_envelope_wbtest.mbt`, which is restored here. That looks like a latent bug in the script.

The `pkg.generated.mbti` files could not be regenerated: the installed `moonc` is v0.10.1 and cannot parse this SDK's current sources, so `moon info` fails. Only one `.mbti` differs from main, the oplog one, and it was checked by hand: main's file plus this branch's 13 `SnapshotConfirmed` lines, every declaration of which is present in the regenerated `top.mbt`. CI's `build-golem-moonbit` job runs `moon check --target wasm` over the SDK, so it is covered there rather than locally.


kmatasfp:
The protobuf question is settled: there is no meaningful oplog from main yet, so no persisted bytes carry the old tags and nothing has to be preserved.

That also matches the repository policy in `AGENTS.md`, which says to make breaking changes directly and not to add compatibility shims, fallback parsing or migration paths until that section is revised.

So `SnapshotConfirmed` keeps its new numbers, 67 in `public_oplog.proto` and 66 in `raw_oplog.proto`, and main keeps 65 and 66 for `Resumed` and `RecoverySucceeded`. No further action on this one; it is recorded here so it does not get raised again in review.

kmatasfp:
The unverified item in the description is now closed: the moonbit interface files are generated, not merged.

The host had a stale `~/.moon` from 18 August carrying `moonc v0.10.1` (built 22 June), which is not on `PATH` and not in the nix environment, and cannot parse this SDK's current sources, so `moon info` failed during the merge. Installing the version this repository pins, `MOONBIT_INSTALL_VERSION` = `0.10.11+6ff76a5f9`, into a scratch `MOON_HOME` lets it run: 121 tasks, 0 errors, across all 57 generated packages. The stale toolchain at `~/.moon` was left exactly as it was.

Exactly one file changed, `interface/golem/api/oplog/pkg.generated.mbti`, and only in ordering: the generator writes `SnapshotData` after `SnapshotConfirmedParameters`, while the merge kept main's earlier position for it. No declaration differs, so the merged content was right and only its position was wrong. Committed as a09d387e4.

Every generated file in this pull request now comes from its generator rather than from a three-way merge.

PR #3917, opened 2026-09-18
Scope worker caches by agent fingerprint

vigoo:
## What problem does this solve?

Several worker-executor data structures are persistent caches derived from an agent's oplog. Until now, most of them were keyed only by the logical agent ID.

That is unsafe when an agent is deleted and later recreated with the same ID:

1. incarnation A writes derived cache data;
2. incarnation A is deleted;
3. incarnation B is created with the same logical agent ID;
4. a delayed write, incomplete cleanup, or stale cache entry from A can be read as if it belonged to B.

The oplog already distinguishes these incarnations. Every `Create` entry contains a unique `instance_id`, represented in the executor as an `AgentFingerprint`. This PR uses that existing identity consistently for derived persistent data.

In short:

```text
Before: cache key = logical agent ID
After:  cache key = logical agent ID + fingerprint from the current Create entry
```

Data written by incarnation A therefore remains isolated from incarnation B even though both use the same logical agent ID.

### Fingerprint-scoped derived caches

The following data is now stored and read under the authoritative agent fingerprint:

- agent status and clean status checkpoints;
- invocation-result indexes;
- durable-stream/session indexes;
- the `RunningWorkers` recovery index.

Readers first resolve the current oplog `Create` entry, then access only the cache namespace for that entry's fingerprint. Resident workers already carry this fingerprint, so their writes do not need another identity lookup.

Stale entries may remain temporarily, but they cannot be mistaken for data belonging to a replacement incarnation. Renewable TTLs eventually remove them.

### The cached mode optimization is preserved

Agent mode is intentionally different. Its flat cached value is retained as a **probe-order hint**, not treated as proof that an agent exists or that a particular incarnation is current.

For example, if the hint says `Ephemeral`, the executor probes the ephemeral oplog first. It still validates the actual `Create` entry and fingerprint before accepting derived data. A stale hint can therefore cause one extra oplog probe, but cannot select stale state.

This preserves the important fast path for ephemeral agents: a correct mode hint still avoids probing durable storage first.

### Safe concurrent updates and cleanup

The storage abstraction now provides the atomic operations needed by these caches:

- set/update a value together with its expiry;
- refresh expiry safely;
- compare-and-delete or compare-and-mutate so delayed cleanup cannot remove a newer value.

These operations are implemented consistently for Redis, PostgreSQL, SQLite, multi-SQLite, namespace-routed, fault-injecting, retrying, and in-memory storage backends.

`RunningWorkers` entries include the fingerprint as well. Recovery admits only the exact recorded incarnation, and stale cleanup removes only the matching stale member instead of potentially deleting the replacement's entry.

## Failure behavior

The oplog remains authoritative. All fingerprint-scoped caches are derived accelerators:

- missing or expired data is reconstructed from the oplog;
- corrupt or mismatched entries are ignored or rebuilt rather than trusted;
- stale data from deleted incarnations is isolated and eventually expires;
- concurrent publishers use atomic metadata updates to avoid partially published cache state.

Low-cardinality metrics were added for relevant fallback and repair paths.

## Documentation

The durable-execution skill and worker-executor walkthrough now describe:

- authoritative oplog identity versus derived cache state;
- fingerprint-scoped cache namespaces;
- mode hints and the ephemeral fast path;
- exact-incarnation recovery through `RunningWorkers`.

The two evaluated implementation plans are included as `plan1.md` and `plan2.md`.

## Verification

- full PR CI is green, including all worker-executor, integration, CLI integration, unit, formatting, and Clippy jobs;
- `cargo test -p golem-worker-executor --lib services::worker:: -- --report-time` — 63 passed;
- `cargo test -p golem-worker-executor --test integration -- compare_and_delete_flat_key_contract --report-time` — 6 passed;
- `cargo test -p golem-worker-executor --lib status_flusher -- --report-time` — 16 passed;
- `cargo test -p golem-worker-executor --lib status_checkpointer -- --report-time` — 10 passed;
- targeted stale-incarnation, cache-loss, concurrent-publication, recovery, and durable-stream tests passed;
- `cargo check -p golem-worker-executor --all-targets`;
- `cargo clippy -p golem-worker-executor --lib --no-deps`;
- `cargo fmt --all -- --check`.

Resolves GOL-476


plan1.md
```diff
@@ -0,0 +1,146 @@
+# GOL-476 Plan 1: Owner-Persisted Derived State
```
vigoo: Remove this from the commit

plan2.md
```diff
@@ -0,0 +1,572 @@
+# GOL-476 Plan 2: Fingerprint-Scoped Derived Cache Namespaces
```
vigoo: Remove this from the commit

golem-worker-executor/src/services/oplog/mod.rs
```diff
@@ -192,6 +192,19 @@ pub trait OplogService: Debug + Send + Sync {
         self.read_exact(owned_agent_id, agent_mode, idx, n).await
     }
 
+    /// Reads the initial entry without treating an absent oplog or a storage failure as a
+    /// fail-stop condition. Identity resolution uses this narrow fallible path.
+    async fn read_initial_entry(
```
vigoo: Is this new method absolutely necessary? Would be nice to avoid any changes in the oplog modules

plan1.md
vigoo: Removed `plan1.md` from the PR.

golem-worker-executor/src/services/oplog/mod.rs
vigoo: No—the dedicated method was avoidable, so I removed it and all of its forwarding/archive implementations. Identity resolution now uses the existing `OplogService::read_source(INITIAL, 1)` API. The generic oplog modules (`mod`, blob, compressed, plugin, and rate-limited) are now identical to `main`. The remaining oplog changes only thread the already-known fingerprint into raw-session cache lookups, plus a `MultiLayerOplogService::read_source` override so a logical read can find an archived `Create` entry without the default exact-read path opening a nonexistent mode namespace.

plan2.md
vigoo: Removed `plan2.md` from the PR.

noise64 (approved):

PR #3922, opened 2026-09-21
Fix host-managed TimeBox recovery semantics

vigoo:
## Summary

- pass real elapsed time to host-managed `RetryPolicy::step` calls instead of `Duration::ZERO`
- persist one retry sequence's first-decision timestamp and monotonic elapsed high-water mark in `RetryPolicyState::TimeBox`
- preserve the evaluated TimeBox state when inline retry falls back to trap/replay, avoiding a second policy step that could reset the origin
- keep policies without a nested `TimeBox` on their existing state and fallback paths
- cover exact inclusive expiration, pre-boundary progress, nested composition, backward clock movement, and simulated-crash reconstruction

## Root cause and durable invariant

Host-managed semantic retry always evaluated policies with zero elapsed time, so any positive `TimeBox` was effectively infinite. The durable invariant is now: a selected host-managed retry sequence starts at its first policy decision, and every later decision—including after reconstruction—uses `max(previous_elapsed, now - started_at)`. Thus elapsed cannot regress when the wall clock moves backward, and restart cannot reset the budget. Canonical `RetryPolicy::step` semantics remain inclusive: `elapsed >= limit` gives up.

The new internal retry-state envelope is persisted in the existing raw oplog `Error.retry_policy_state`. Public oplog projections continue to expose it as the existing wrapper shape; they are not the reconstruction source. No SDK or public WIT contract changes are included.

## Recovery coverage

`time_box_elapsed_budget_survives_reconstruction` starts a failing HTTP invocation with `TimeBox(500ms, Periodic(2s))`, waits until retrying, lets 750ms elapse, then performs a simulated crash. The invocation must fail within 1.5s after reconstruction—strictly before a reset budget could authorize and execute another 2s-delayed retry.

## Verification

- `cargo test -p golem-common --lib -- time_box_exhausts_after_elapsed_limit --report-time`
- `cargo test -p golem-worker-executor --lib -- host_managed_time_box_uses_inclusive_monotonic_elapsed_time host_managed_policy_without_time_box_keeps_existing_state_shape --report-time`
- `cargo test -p golem-worker-executor --lib -- host_managed_time_box_uses_inclusive_monotonic_elapsed_time host_managed_policy_without_time_box_keeps_existing_state_shape time_box_fallback_carries_the_evaluated_state_to_trap_recovery fallback_to_trap_when_delay_exceeds_threshold --report-time`
- `cargo test -p golem-worker-executor --test integration -- retry_lifecycle::time_box_elapsed_budget_survives_reconstruction --exact --report-time`
- `cargo check -p golem-worker-executor --lib`
- `git diff --check`

Oracle's focused durable-state review found an inline-to-trap handoff issue and a recovery-test validation gap; both were fixed and its follow-up found no remaining blocker. The bounded bug-finder protocol completed two clean runs.


golem-worker-executor/src/durable_host/http/inline_retry.rs
```diff
@@ -1234,7 +1251,7 @@ pub async fn try_output_stream_inline_retry<Ctx: crate::workerctx::WorkerCtx>(
                 }
             }
         }
-        AsyncRetryDecision::Exhausted | AsyncRetryDecision::FallBackToTrap => {
+        AsyncRetryDecision::Exhausted | AsyncRetryDecision::FallBackToTrap(_) => {
             return Ok(false);
```
noise64: > Could we preserve this `FallBackToTrap` payload through the streaming helpers instead of collapsing it into `Ok(false)`? For p2 body reads/writes, the caller then invokes `DurableCallSession::try_trigger_retry` with an empty property bag, so a method/URI-specific `TimeBox` policy can be replaced by the catch-all policy and the evaluated timestamp/state is never persisted. The analogous p3 response-body and transport-error arms also discard the payload and step the policy again. Please thread the override into the trap path (without a second evaluation) and add long-delay `TimeBox` tests for these streaming fallbacks.

vigoo: Fixed in 6c9de8b. Inline HTTP retry now propagates the exact `SemanticTrapRetryOverride` through P2 stream operations, P3 response-body resume, P3 send, and the nested P2 response resend helper. The trap path consumes that evaluated payload directly instead of resolving the policy again with missing or altered properties, preserving `retry_from`, policy name, verdict, and the complete TimeBox state across recovery. Added focused complete-payload unit coverage plus response-body and transport reconstruction tests. Oracle follow-up review approved the fix with no blockers.

noise64 (approved):

PR #3923, opened 2026-09-21
Simplify SDKs to a single agent world

vigoo:
## Summary

- consolidate the Rust, TypeScript, Effect, Scala, and MoonBit SDKs on one default agent world that exports agents, tools, and tool middleware
- remove the obsolete middleware-only/mixed role selection, feature flags, templates, generated bindings, and diagnostics while keeping standalone middleware packages supported
- preserve middleware chaining and underlying-tool dispatch, including parameters, streams, resource lifetime, deployment, permission enforcement, and durable runtime behavior
- synchronize canonical WIT copies, regenerated SDK artifacts, CLI fixtures/tests, and documentation

## Validation

- `cargo make wit`
- Rust middleware unit and real cross-crate component/compile-fail tests (28 unit tests and built WASM contracts)
- Rust SDK library clippy with `export_golem_agentic`
- TypeScript middleware tests and generated component contract checks
- Effect typecheck, 888 tests, lint, formatting, declaration/artifact/package checks
- Scala generated agent guest export/role contract scripts
- MoonBit binding regeneration, checks, and 278 tests
- CLI `GuestLanguage` unit test
- docs formatting/version checks
- `cargo fmt --all -- --check`
- `git diff --check`

Resolves GOL-497


sdks/ts/packages/golem-ts-sdk/scripts/template-matrix.mjs
```diff
@@ -1,6 +1,6 @@
 export const templateMatrix = [
   {
-    role: 'agent',
+    role: 'default',
     world: 'agent-guest',
     wrapperDirectory: 'agent-template',
     sdkModuleName: '@golemcloud/golem-ts-sdk',
```
noise64, line 6: > The new wrapper embeds only `@golemcloud/golem-ts-sdk`, but the CLI still externalizes `@golemcloud/golem-ts-sdk/middleware`, and the updated docs tell standalone middleware authors to import that subpath. The user bundle will retain an import the wrapper does not provide, so this looks like a runtime module-resolution failure despite the WIT export test passing. Could you make the subpath resolve to the same embedded SDK instance and add a component test that evaluates middleware registered through the documented import?

sdks/effect/scripts/template-matrix.mjs
```diff
@@ -1,6 +1,6 @@
 export const templateMatrix = [
   {
-    role: "agent",
+    role: "default",
     world: "agent-guest",
     wrapperDirectory: "agent-template",
     sdkModuleName: "@golemcloud/effect-golem",
```
noise64, line 6: > The single Effect wrapper embeds only the root SDK, while the CLI does not externalize `@golemcloud/effect-golem/middleware`. Rollup therefore bundles the documented middleware entry into user code, giving its module-local `entries` map a different instance from the root SDK’s `toolMiddlewareGuest` in the wrapper. Middleware registered via that entry appears invisible to discovery/invocation. Could you route both imports through one registry and add a runtime test using the README’s subpath import?

sdks/ts/packages/golem-ts-sdk/scripts/template-matrix.mjs
vigoo: Fixed in 9f96631 and c88ccbf. The wrapper now embeds the `/middleware` subpath, and that subpath is built as a thin re-export of the embedded root SDK rather than as a second bundled SDK instance. The standalone TypeScript middleware fixture now registers through `@golemcloud/golem-ts-sdk/middleware` and verifies discovery through the root `toolMiddlewareGuest`; `test_ts_tool_middleware_component_roles` passes with the rebuilt template.

sdks/effect/scripts/template-matrix.mjs
vigoo: Fixed in 9f96631. The Effect wrapper now embeds `/middleware`, the CLI template externalizes that import, and the subpath bundle is a thin re-export from the root `Middleware` namespace so both entry points share one registry. The standalone Effect integration component registers via `@golemcloud/effect-golem/middleware` and asserts discovery through the root `toolMiddlewareGuest`; its real component build and metadata extraction pass.

PR #3925, opened 2026-09-21
Add the filesystem snapshot store with an in-memory adapter and conditional blob writes

kmatasfp:
- resolves GOL-602

This pull request is at the bottom of the filesystem snapshotting stack, directly on `filesystem-snapshot`. It gives the worker executor the interface of the filesystem snapshot store, an in-memory store, and one contract suite that each store must pass. It also gives blob storage the two things that the rustic store needs: a conditional write, and a namespace for filesystem snapshots. Nothing in production calls the store yet, so the module has `#[allow(dead_code)]`.

## Where to start

1. `golem-worker-executor/src/filesystem_snapshot/mod.rs`: the `FilesystemSnapshotStore` trait and its types. The trait docs are the contract.
2. `filesystem_snapshot/contract.rs`: the cases of the contract, which see a store only through the trait.
3. `filesystem_snapshot/memory.rs` and `memory/tree.rs`: the in-memory store.
4. `golem-service-base/src/storage/blob/mod.rs`: `put_raw_if_absent`, `PutIfAbsent`, the `FilesystemSnapshots` namespace and `agent_path_segment`. Then the four backends.

## The store interface

- `FilesystemSnapshotStore` has the seven methods of section 3.1 of the project: `save`, `restore`, `stat`, `list`, `delete`, `delete_scope` and `copy_scope`. They share one error type, `SnapshotStoreError`, with the six variants of section 3.1.
- A `SnapshotScope` holds a `BlobStorageNamespace`, and only `SnapshotScope::agent` knows agents. A `SnapshotName` has 1 to 64 ASCII letters, digits, `-` or `_`, so it can be a path segment and a label. `SnapshotName::new` can fail.
- The docs fix five rules that section 3.1 left open. The rustic store must keep them too:
  - A snapshot keeps the path, the kind, the content, the symlink target, the permission bits and the modification time of each entry below the tree. It does not keep the owner, the access time, extended attributes, or the metadata of the root. Each name of a hard-linked file comes back as a separate file.
  - A save gives a `created_at` later than that of each snapshot that the scope held when the save started. Sequential saves therefore list newest first, also when the clocks of two executors differ, and retention by count keeps the right snapshots.
  - `restore` refuses an `into` that is missing, that is not a directory, or that is not empty, and writes nothing.
  - `list` leaves out a snapshot whose metadata fails an integrity check. `stat` and `restore` of that name give `Corrupt`.
  - A `copy_scope` into a scope that is not empty is outside the contract.

## The in-memory store

- `InMemorySnapshotStore` keeps the snapshots of each scope as one immutable `Arc<[..]>`. Each change makes a new slice with a pure function and puts it in place under a lock that no `await` holds.
- A clone of the store is another store over the same snapshots, as another executor has. The contract uses this to check that a save through one store resolves through another.
- A restore keeps its own `Arc` to the tree, so a delete at the same time cannot change what it writes.
- The reads and writes of the tree run on the blocking pool.
- It is ordinary code, not test code, so that mutation testing reaches it through the contract.

## The contract suite

- `contract::register` adds each case as a test of a `#[test_gen]` function. It takes a function that opens stores over one new storage for each case, and it names no store. The rustic store registers the same suite in the same way.
- The suite writes its own trees and reads them back with its own code. So a defect of a store cannot hide in the code that checks it.
- `listing`, `listed_names`, `fixture` and `one_file` return a `Vec`, not a boxed slice, because the tests compare their values with `vec![...]` literals. A boxed form would make each such assertion convert its literal.
- `no_method_blocks_the_runtime` counts a local task in a `LocalSet` while `save` and `restore` run. test-r runs a test on a runtime with more than one thread, and a task outside the `LocalSet` could count on another thread while the call blocks.

## `put_raw_if_absent`

- The method writes the blob only when the path has no blob. It gives `PutIfAbsent::Written` or `PutIfAbsent::AlreadyExists`. An existing blob is an answer, not a fault, so it is a value and not an error to downcast.
- The method is required on all seven implementors. A default would check the path and then write in two steps, and two writers could both write between the steps.
- S3 sends `If-None-Match: *`. A 412 gives `AlreadyExists`, stops the retries, and stays out of the error log and the failure counter. A 409 means that a request on the same key ran at the same time, and the S3 documentation says that the write can go again, so it is retried. The trait doc says what follows from the retries: when the response of an attempt that wrote the blob is lost, the next attempt finds that blob, and the call gives `AlreadyExists`.
- The in-memory backend inserts only into a free key. SQLite uses `ON CONFLICT DO NOTHING` and reads the number of changed rows.
- The filesystem backend does not use `create_new`, although the project text names it. With `create_new` and then a write, a reader or a crash can see a partly written blob. The backend writes a temporary file in `.staging` at the storage root, outside every namespace, and moves it into place without replacing a file that is there.

## The namespace

- `BlobStorageNamespace::FilesystemSnapshots { environment_id, agent_id }` has its own location on each backend. S3 gets a new bucket field, `filesystem_snapshots_bucket`, with the default `filesystem-snapshots`, and the generated configuration files have it. A deployment needs the bucket before the feature writes to it.
- On S3, SQLite and the filesystem backend, the location of an agent is one path segment from `agent_path_segment`, not the agent id. An agent id can hold `/`, `\` and `.` segments, and the blob name rules refuse a `.` or `..` segment on S3. An agent id can also be longer than a file name can be on the filesystem backend.
- `agent_path_segment` is the function that the filesystem backend already used for oplog payloads, `FileSystemBlobStorage::filesystem_safe_oplog_payload_agent_key`, moved to `blob/mod.rs` so that three backends share it. Its body did not change, so its `for` loop is existing code. A new test fixes the form of the segment.
- `OplogPayload` on S3 still puts the raw agent id in its key. That is why the isolation test compares the new namespace with the oplog payloads of an agent whose id has no `..` segment.


golem-worker-executor/src/filesystem_snapshot/contract/tree.rs
```diff
@@ -0,0 +1,390 @@
...
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use tempfile::TempDir;
+
+/// A temporary directory that goes away with all that is in it, also when a directory or a file
+/// below it is read-only.
+pub(super) struct Scratch(TempDir);
+
+impl Scratch {
+    pub(super) fn new() -> Self {
+        Self(tempfile::tempdir().unwrap())
```
kmatasfp: can this fail?

golem-worker-executor/src/filesystem_snapshot/contract/tree.rs
```diff
@@ -0,0 +1,390 @@
...
+}
+
+/// Lists each entry below `root`, in the order of the paths. Symlinks are not followed.
+pub(super) fn listing(root: &Path) -> Vec<Listed> {
+    let mut listed = add_listed(root, Path::new(""), Vec::new());
+    listed.sort_by(|left, right| left.path.cmp(&right.path));
+    listed
+}
+
+fn add_listed(directory: &Path, relative: &Path, listed: Vec<Listed>) -> Vec<Listed> {
```
kmatasfp: is this immutalbe, if yes can use Box<[Listed]> or Arc<[Listed]> if we need to clone it later

golem-worker-executor/src/filesystem_snapshot/contract_tests/mod.rs
```diff
@@ -0,0 +1,1017 @@
+// Copyright 2024-2026 Golem Cloud
```
kmatasfp, line 1: to be consistent this should be test(s).rs under contract dir. Looks like it only contains some tests? Slightly unclear how it is supposed to be used

golem-worker-executor/src/filesystem_snapshot/contract_tests/fixture.rs
```diff
@@ -0,0 +1,390 @@
+// Copyright 2024-2026 Golem Cloud
```
kmatasfp, line 1: same as other contract comment, not obvious that this is for tests only, I think we need to change this to make it obvious it is not a production code I think, unless I am misunderstanding something. Even if I am others might have similar thoughts/be confuse so we need to improve this regardless

golem-worker-executor/src/filesystem_snapshot/contract_tests/mod.rs
kmatasfp: Moved to `filesystem_snapshot/contract_tests/mod.rs` in c4efa1ffb. The module doc now starts with "This module is test code, and it compiles only for tests." Then it says what the suite is: the behaviour that each adapter of `FilesystemSnapshotStore` must have. It also says how an adapter uses it, with the `#[test_gen]` registration from `memory/tests.rs` as an example. The rustic store of GOL-604 registers the suite in the same way.

golem-worker-executor/src/filesystem_snapshot/contract_tests/fixture.rs
kmatasfp: Moved to `filesystem_snapshot/contract_tests/fixture.rs` in c4efa1ffb. The path now says that it is test code, and the name no longer looks like the production `memory/tree.rs`. The module doc starts with the same sentence: "This module is test code, and it compiles only for tests."

